PR #1, opened 2015-11-30
Separated raw UART interface

rafalcieslak:
Since UART is probably going to be very useful for debugging, I moved raw access implementation to a separate file, cleaned it up a little bit, and added few convenience procedures.

`uart_raw.h` provides a simple interface for basic UART operations.

The `Makefile` is modified so that it supports compiling multiple translation units.


cahirwpz:
LGTM

PR #6, opened 2015-12-05
Reorganized startup procedure as specified in #2.

### LINKED ISSUE #2: Move boot code aka startup routine to separate file.
Move `_start` to `startup.s` file. Make sure `.data` section is copied to RAM before code jumps to `main` function. Loop forever when `main` exits. Fix `Makefile` and linker script respectively.

rafalcieslak:
Moved `_start` to `startup.s`. The start procedure now copies `.data` to RAM, and clears `.bss`. Renamed `main()` to `kernel_main()`. 

Fixes #2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,21 +3,29 @@
 P	= mips-mti-elf-
 QEMU	= qemu-system-mipsel -machine pic32mz-wifire -sdl -serial stdio -s -S
 CC	= $(P)gcc -mips32r2 -EL -g -nostdlib
+AS	= $(P)as -mips32r2 -EL -g
 GDB	= $(P)gdb
 OBJCOPY	= $(P)objcopy
 OBJDUMP	= $(P)objdump
 CFLAGS	= -O -Wall -Werror -DPIC32MZ
-LDFLAGS	= -T pic32mz.ld -e _start -Wl,-Map=pic32mz.map
+LDSCRIPT= pic32mz.ld
+LDFLAGS	= -T $(LDSCRIPT) -e _start -Wl,-Map=pic32mz.map
 
 PROGNAME = main
 SOURCES = main.c uart_raw.c
+SOURCES_ASM = startup.s
 OBJECTS := $(SOURCES:.c=.o)
+OBJECTS_ASM := $(SOURCES_ASM:.s=.o)
+OBJECTS += $(OBJECTS_ASM)
 
 all: $(PROGNAME).srec
 
-$(PROGNAME).srec: $(OBJECTS)
+$(PROGNAME).srec: $(OBJECTS) $(LDSCRIPT)
 	$(CC) $(LDFLAGS) $(OBJECTS) $(LIBS) -o $(PROGNAME).elf
 	$(OBJCOPY) -O srec $(PROGNAME).elf $(PROGNAME).srec
+
+%.S: %.c
+	$(AS) $(CFLAGS) -c $<
 
 %.o: %.c
 	$(CC) $(CFLAGS) -c $<
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,9 @@ LDFLAGS	= -T $(LDSCRIPT) -Wl,-Map=pic32mz.map
 
 PROGNAME = main
 SOURCES = main.c uart_raw.c
-SOURCES_ASM = startup.s
+SOURCES_ASM = startup.S
 OBJECTS := $(SOURCES:.c=.o)
-OBJECTS_ASM := $(SOURCES_ASM:.s=.o)
+OBJECTS_ASM := $(SOURCES_ASM:.S=.o)
 OBJECTS += $(OBJECTS_ASM)
 
 all: $(PROGNAME).srec
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ OBJCOPY	= $(P)objcopy
 OBJDUMP	= $(P)objdump
 CFLAGS	= -O -Wall -Werror -DPIC32MZ
 LDSCRIPT= pic32mz.ld
-LDFLAGS	= -T $(LDSCRIPT) -e _start -Wl,-Map=pic32mz.map
+LDFLAGS	= -T $(LDSCRIPT) -Wl,-Map=pic32mz.map
 
 PROGNAME = main
 SOURCES = main.c uart_raw.c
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -3,22 +3,30 @@
...
 QEMU	= qemu-system-mipsel -machine pic32mz-wifire -sdl -serial stdio -s -S
 CC	= $(P)gcc -mips32r2 -EL -g -nostdlib
+AS	= $(P)as -mips32r2 -EL -g
 GDB	= $(P)gdb
 OBJCOPY	= $(P)objcopy
 OBJDUMP	= $(P)objdump
 CFLAGS	= -O -Wall -Werror -DPIC32MZ
-LDFLAGS	= -T pic32mz.ld -e _start -Wl,-Map=pic32mz.map
+LDSCRIPT= pic32mz.ld
+LDFLAGS	= -T $(LDSCRIPT) -e _start -Wl,-Map=pic32mz.map
```
cahirwpz: Please move -e _start to the linker script.

main.c
```diff
@@ -84,6 +73,16 @@ int main()
     /* Initialize UART. */
     uart_init();
     uart_putstr("Hello, UART!\n");
+
+    /* Demonstrate access to .data */
+    uart_putstr(str);
+
+    /* Test whether .bss appears to have been cleared. */
+    char* p = empty;
+    while(p < empty + sizeof(empty))
```
cahirwpz: I can see code formatting inconsistencies, including whitespace mix of tabs and spaces. That isn't worrying at the moment but it'll become a problem in long run. Perhaps we could use GNU indent to solve the problem?

rafalcieslak: GNU Indent or astyle should be fine, we just have to pick a style.

cahirwpz:
Have a look at sources downloadable from http://training.mips.com/cps_mips/Examples/Boot_MIPS_1.03.zip

Startup routine for interAptiv family of processors is in `interAptiv/start.S` file. Note that reset vector is setup at EBASE + 0x180. After that every 0x80 bytes is an entry point for another exception handling routine, up to EBASE + 0x500.

You can find more about exception processing in chapter 5 of [MIPS32® microAptiv™ UP Processor Core Family Software User’s Manual](https://imagination-technologies-cloudfront-assets.s3.amazonaws.com/documentation/MD00942-2B-microAptivUP-SUM-01.02.pdf)

startup.s
```diff
@@ -0,0 +1,43 @@
...
+	bne	$t1, $t2, copy_next_word
+	  nop			# jump if dest != dest_end
+
+	/*     Clean .bss. It starts at _bss and ends at _ebss. */
+	la	$t1, _bss	# dest = _bss
+	la	$t2, _ebss	# dest_end = _ebss
+clear_next_word:
+	/*     Everybody loves MIPS $zero register. */
+	sw	$zero, 0($t1)	# *dest = 0
+	addiu	$t1, 4		# dest++
```
cahirwpz: You may put this instruction into a delay slot of `bne` instead of `nop`.

startup.s
```diff
@@ -0,0 +1,43 @@
...
+	bne	$t1, $t2, clear_next_word
+	  nop			# jump if dest != dest_end
+
+	/* Jump to kernel_main(). */
+	jal	kernel_main
```
cahirwpz: AFAIK every branch instruction in MIPS needs to have a delay slot filled in, including `jal`.

startup.s
```diff
@@ -0,0 +1,43 @@
...
+	la	$gp, _gp	# Prepare global pointer.
+
+	/* Copy .data from ROM to RAM.
+	 * .data is located in ROM just after .text, so it starts at _etext.
+	 * It should land at the beggining of RAM, which is pointed at by _data.
+	 * Copy words one by one, until _edata is reached. */
+	la	$t1, _data	# dest = _data
+	la	$t2, _edata	# dest_end = _edata
+	la	$t3, _etext	# src = _etext
+copy_next_word:
```
cahirwpz: I would organize this code in a different way w.r.t. symbol names and their use.

https://sourceware.org/binutils/docs/as/Symbol-Names.html

I propose to give a symbol `cope_rom_to_ram` to the beginning of the loop  and then use local label `.Lnext_word` or similar. Local labels are very handy tool in assembly.

startup.s
rafalcieslak: Apparently I should not have to insert instructions into the delay slot at all. The assembler is aware of them, and automatically placed `addiu` right after `bne`:

    Dump of assembler code for function copy_next_word:
    0xbfc00028 <+0>:	lw	t0,0(t3)
    0xbfc0002c <+4>:	sw	t0,0(t1)
    0xbfc00030 <+8>:	addiu	t1,t1,4
    0xbfc00034 <+12>:	bne	t1,t2,0xbfc00028 <copy_next_word>
    0xbfc00038 <+16>:	addiu	t3,t3,4
    0xbfc0003c <+20>:	nop

Similarly, for `jal`:

    0xbfc00060 <+16>:	jal	0xbfc000a8 <kernel_main>
    0xbfc00064 <+20>:	nop

It does not swap instructions blindly if it could modify behaviour, and places `nop`s when needed:

    0xbfc00054 <+4>:	addiu	t1,t1,4
    0xbfc00058 <+8>:	bne	t1,t2,0xbfc00050 <clear_next_word>
    0xbfc0005c <+12>:	nop

Therefore we can simply ignore all delay slots. We can also choose to manage them on our own by disabling this feature with `.set noreorder`.

cahirwpz: It's an interesting feature of GNU assembler I was not aware of. I don't expect to have large codebase in assembly language nor everyone to read it. However if in any case anyone is forced to do so I'd prefer to have everything written in very explicit style. Perhaps that's why you can find `.noreorder` and `.noat` in ImgTec sources.

pic32mz.ld
```diff
@@ -19,6 +19,7 @@ SECTIONS
 
   .text : AT(ORIGIN(flash))
   {
+    _text = ABSOLUTE(.);
```
cahirwpz: Please note that C compiler converts `name` symbol to `_name` linker symbol. Thus by a mere mistake you may end up using `text` in C but refer to the beginning of text section instead of just a variable.

If you have a Debian based distro look at linker script in `/usr/lib/ldscripts/elf_x86_64.x` file. Most of the time they use double underscore for special linker symbols.

Issue #3, opened 2015-12-01
Integrate C standard library with the kernel


cahirwpz:
It's desired to use standard C function within kernel (like `strings.h`, `stdlib.h` and so on) - there's no real benefit in implementing them by ourselves.

In the future It'd be even better if we could devise kprintf function for debugging purposes. Obviously I expect this function to reuse some code from newlib.

rafalcieslak:
Is the goal to properly integrate newlib by modifying its autotools scripts so that it can automatically building into an object file that we would then simply link with the kernel, or would it be fine to just copy the appropriate source files from newlib into our source tree to keep it buildable with a simple makefile?

I believe the second approach is probably easier to implement, it significantly simplifies the build process, and should keep our sources self-contained and therefore easier to comprehend as a whole. On the other hand, the first option makes it easier to replace newlib with a newer version or a different implementation (do we need that?), and keeps our source independent from stdlib implementation.

cahirwpz:
It's even more complicated that you may think. In fact we need two C standard libraries: one for kernel (may be incomplete) and one for user space. In first case it's acceptable we partially merge sources with our kernel - thus the license of donor library should be permissive. In second case - I guess that we'd like to have a reasonably complete solution to enable running existing software on top of our operating system. It's quite obvious we'll have to provide our method to invoke system calls, hence we should fork newlib and patch it to meet our requirements.

I'm quite convinced we should choose newlib for our user space C standard library. However I have no clear conclusions regarding kernel space one. Perhaps PDCLib?

Please read following article: http://wiki.osdev.org/C_Library

cahirwpz:
Perhaps we could just steal [_PDCLIB_print.c](https://github.com/blanham/PDCLib/blob/master/functions/stdio/_PDCLIB_print.c) file to have our own printf implementation? I believe most interesting functions from C standard library could be delivered by [libgcc](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#Other-Builtins)

rafalcieslak:
I have given PDCLib a closer look. Apparently the glue layer is very small, and almost all functions can be stubbed, therefore it should not require much effort to port the entire PDCLib, and then gradually enable the "advanced" features once implemented.

The build scripts for PDClib are not universal and it might be tricky to adapt them to our needs, but since we are fine with merging the library's sources into the kernel (PD license), we can ignore these scripts altogether.

[...]

PR #11, opened 2015-12-16
Integrating smallclib with the kernel

rafalcieslak:
This branch introduces a smallclib port, as discussed in #3. It builds the standard library using sources provided by the toolchain. A simple stdlib demonstration procedure is called by `kernel_main`.

This branch also introduces custom `kprintf`, `kputs` and `kperror` functions which use UART to output strings.


uart_raw.h
```diff
@@ -15,6 +15,8 @@ void uart_init();
 void uart_putch(unsigned char c);
 /* Transmits a character string via UART1. */
 void uart_putstr(const char* str);
+/* Transmits n characters from a string via UART1. */
+void uart_putnstr(unsigned int n, const char* str);
```
cahirwpz: Please be consistent with C standard library function naming and argument passing order, if possible.

int uart_putchar(int c);
int uart_puts(const char *s);
int uart_write(const void *s, size_t n);

uart_raw.c
```diff
@@ -39,6 +39,10 @@ void uart_putstr(const char* str){
   while(*str) uart_putch(*str++);
 }
 
+void uart_putnstr(unsigned int n, const char* str){
```
cahirwpz: Was it your intention to possibly move past '\0' ?

smallclib/extra/stdio.h
```diff
@@ -0,0 +1,21 @@
+#ifndef __custom_STDIO_H__
```
cahirwpz: Shouldn't we rename this file to kdebug.h, or will it be misleading?

smallclib/extra/stdio.h
```diff
@@ -0,0 +1,21 @@
...
+#define __custom_STDIO_H__
+
+/* This is a wrapper header for stdio.h.  It appends serveral
+ * kernel-based string output functions to the standard stdio.h
+ * header.
+ */
+
+#include_next <stdio.h>
```
cahirwpz: Nice. I did not know about it :)

smallclib/extra/stdio.h
```diff
@@ -0,0 +1,21 @@
...
+
+/* Write a formatted string to UART.
+ * Equivalent to standard printf. */
+int kprintf(const char *fmt, ...);
+/* Write a character string and a trailing newline to UART.
+ * Equivalent to standard putss */
+int kputs (const char *s);
+/* Print an error message to UART.
+ * Equivalent to standard perror. */
+void kperror (const char *s);
```
cahirwpz: Is `errno` meaningful inside the kernel at all? Is errno defined as `__thread int errno;` ?

smallclib/extra/kprintf.c
```diff
@@ -0,0 +1,23 @@
...
+
+/* Print a formatted string to UART. */
+int kprintf(const char *fmt, ...)
+{
+    int ret;
+    ReadWriteInfo rw;
+    va_list ap;
+    va_start(ap, fmt);
+    rw.m_fnptr = __low_kprintf;
+    ret = _format_parser(&rw, fmt, &ap, 0);
```
cahirwpz: Paweł Dziepak suggested not to use floating point numbers inside the kernel. If so we can use the alternative to `_format_parser` which is AFAIR `_format_parser_int`.

smallclib/Makefile
```diff
@@ -0,0 +1,140 @@
...
+LIB_PATH = $(realpath $(join $(STD_HEADER_PATH), ../../../../src/smallclib/))
+
+$(info Using $(LIBNAME) sources located at: $(LIB_PATH))
+
+INCLUDES = -I$(LIB_PATH)/include
+INCLUDES += -isystemextra
+INCLUDES += -I../ # For uart_raw.h
+
+# Explicitly picking files to compile.
+LIB_SOURCES  = ctype/ctype_.c
```
cahirwpz: Don't pull in files we don't need yet.

smallclib/Makefile
```diff
@@ -0,0 +1,140 @@
...
+	@echo "Creating $@"
+	$(AR) rvs $@ $(EXTRA_OBJECTS) $(LIB_OBJECTS) > /dev/null
+
+# Generate rules for each library file separately. A simple pattern
+# rule will not suffice, because we need to correct the file directory
+# for rule and dependency.
+define emit_lib_rule =
+CFILE := $(1)
+OFILE := $(patsubst %.S,%.o,$(patsubst %.c,%.o,$(notdir $(1))))
+$(LIB_BUILD_DIR)/$(OFILE): $(LIB_PATH)/$(CFILE) | $(LIB_BUILD_DIR)
```
cahirwpz: Nice GNU Make magic :)

Please add some reference to the documentation:
* http://www.gnu.org/software/make/manual/make.html#Prerequisite-Types
* http://www.gnu.org/software/make/manual/make.html#Eval-Function

smallclib/Makefile
```diff
@@ -0,0 +1,140 @@
+P	= mips-mti-elf-
+CC	= $(P)gcc -mips32r2 -EL -nostdlib
+AR  = $(P)ar
+
+# -Werror disabled for building a third-party external lib
+CFLAGS =  -Wall -DPIC32MZ
+# No optimisations, emit debugging info.
+CFLAGS += -O0 -g
```
cahirwpz: Why not just `-Og` i.e. optimizations that do not interfere with generating debugging information ?

cahirwpz:
I feel pretty confident we're close to merging this branch with main repo. Good job! :+1:

smallclib/Makefile
pdziepak: On 18 December 2015 at 13:43, Krystian Bacławski <notifications@github.com>
wrote:

> In smallclib/Makefile
> <https://github.com/cahirwpz/wifire-os/pull/11#discussion_r48021339>:
>
> > @@ -0,0 +1,140 @@
> > +P	= mips-mti-elf-
> > +CC	= $(P)gcc -mips32r2 -EL -nostdlib
> > +AR  = $(P)ar
> > +
> > +# -Werror disabled for building a third-party external lib
> > +CFLAGS =  -Wall -DPIC32MZ
> > +# No optimisations, emit debugging info.
> > +CFLAGS += -O0 -g
>
> Why not just -Og i.e. optimizations that do not interfere with generating
> debugging information ?
>
I'm afraid that gcc uses quite unusual definition of "not interfering with
debugging" and even with -Og you can still sometimes see that annoying
"<optimized out>" in gdb.

However, it may be a good idea to have both debug (-O0) and release (-O3
-flto) builds since optimizations may help finding some bugs in the code
(mostly undefined behaviours, sadly there is no easy way to get ubsan
working for a kernel) and lto has problems of it own, in most cases fixed
by adding some [[gnu::used]] attributes (on the other hand lto can quite
significantly reduce the size of the binary which may be quite important
for this project).

rafalcieslak:
I was not aware of `_format_parser_int`. It makes perfect sense to use it instead of the generic `_format_parser`. 

However, the `sprinf` implementation still uses `_format_parser`. We might ignore that fact, but if we either:
 - decide to never use `sprintf`
 - implement a custom `sprintf` using `_format_parser_int` (easy)

then we would no longer need to pull math functions like `frexp` and `fpclassify` as well as float operations `__umodi` and `__udivi` (which removes the need to link with libgcc), because it is only the format parser that requires them. I have no clear opinion in this matter.

I have refactored UART functions to `uart_putc`, `uart_puts` and `uart_write`, and modified their behaviour to match their stdlib equivalents. It was indeed my intention to ignore null characters in `uart_putnstr` (now `uart_write`), as I believe it may be useful to have a function that transmits bytes regardless of what they are (binary transfer?). I believe this is also how the standard `write` behaves.

I have stripped the list of smallclib files that get compiled. I hoped it was a reasonable idea to compile a larger subset of them, because the linker will only pull the ones we actually use, so adding many standard library functions has no effect on the resulting kernel image, and almost no impact on compilation times. Also whoever wishes to use a function from the standard library for the first time will have to tinker a bit with the standard library's Makefile, which seemed to me like an unnecessary effort.

Anyway, the current subset of files picked for compilation is the absolute minimum that is required to use `kprintf`, `sprintf` and `strlen`.

Whether we wish to place `kprintf` etc. in `stdio.h` or `kdebug.h` depends on how we would like to think of these function. At first I had an impression that we wish to expand the standard library with custom variants of several functions, and in that case it would only make sense to append them to `stdio.h` with a header wrapper. However, if we perceive them as functions completely unrelated to the standard library (which I don't), then a custom header like `kdebug.h` makes more sense.

The `errno` is declared as:

```
#define errno (*__errno())
```

with  `__errno()` being:
```
static int error = 0;
int * __errno ()
{
  return &error;
}
```
I believe it may be useful within the kernel. Some standard library functions may set it to meaningful values (currently none of enabled functions does that, but others do). It may also be handy to have it set by some of our custom functions in case of a soft failure, especially if we choose to stick to plain C.

cahirwpz:
> implement a custom sprintf using _format_parser_int (easy)

Firstly, I'd like not to use `sprintf` inside kernel, but rather `snprintf` to avoid buffer overflow errors. Let's reimplement it - as you mentioned it should be easy.

> then we would no longer need to pull math functions like `frexp` and `fpclassify` as well as float operations __umodi and __udivi (which removes the need to link with libgcc), because it is only the format parser that requires them.

**libgcc** normally contains implementation of some integer and floating point routines that are not available as machine instructions. It's safe to assume that kernel code will pull in some routines from the library in long run.

It's perferable to get rid of all floating point functions in the kernel. This will make context switches easier as CP1 registers AFAIK are not shadowed.

> I hoped it was a reasonable idea to compile a larger subset of them, because the linker will only pull the ones we actually use, so adding many standard library functions has no effect on the resulting kernel image, and almost no impact on compilation times. 

Let's be conservative in that manner. Having no necessary functions in kernel C library will make developers think if they really need given feature. 

> Anyway, the current subset of files picked for compilation is the absolute minimum that is required to use `kprintf`, `sprintf` and `strlen`.

Let's also include `memcpy`.

> However, if we perceive them as functions completely unrelated to the standard library (which I don't), then a custom header like kdebug.h makes more sense.

I think of C library in the kernel as a necessary evil. We won't use all of it for sure. We would not like to make developers think they can use it in an unrestricted way.

FreeBSD `libkern` is a stripped down C library - please have a look at a set of functions they use [here](http://fxr.watson.org/fxr/source/libkern/). Please note, that they don't use separate header files for `strings.h`, `stdlib.h` and `stdio.h`, but rather they put everything into [`libkern.h`](http://fxr.watson.org/fxr/source/sys/libkern.h). I'm in favor of such solution.

> I believe it may be useful within the kernel. Some standard library functions may set it to meaningful values (currently none of enabled functions does that, but others do).

But then we have to provide `errno` for each thread separately and we have no TLS implemented yet. Let's try to avoid it for now. Eventually it may be rendered useless.

rafalcieslak:
There is an integer only variant of `snprintf` available, namely `sniprintf`, but instead of using `_format_parser_int` it passes an argument to `_format_parser` asking it to use integers only, so it still pulls floating point routines.

Issue #8, opened 2015-12-15
Setup MIPS CPU internal timer to generate interrupt each milisecond


cahirwpz:
MIPS32® microAptivTM UP Processor Core Family Software User’s Manual, Chapter 5 and 7.

Use **Count Register** and **Compare Register**.

Interrupt handler should just atomically increment a global variable.

rafalcieslak:
A Release 2 MIPS can be configured to use one of following three interrupt modes:

- *Single-vector* mode (AKA legacy mode), in which all interrupts are processed by a single handler procedure
- *Vectored Interrupt* mode, where interrupts are vectored to multiple handlers, can be prioritized and can be assigned a separate shadow register set
- *Extrernal Interrupt Controller* mode, in which an external chip manages interrupt priorities, jump addresses, generally providing more features.

Given that PIC32MZ has a quite powerful external interrupt controller (http://ww1.microchip.com/downloads/en/DeviceDoc/60001108H.pdf), so I hoped I could configure it to react on Core Timer interrupts, which are triggered when `Count == Compare`.

Once again, the results I observed were inconsistent with specification, so I had to run `qemu` within a debugger and investigate sources. Indeed it turns out that the support for external interrupt controllers with MIPS is not fully implemented. Generally, it works exactly as it should, except for the part where the exception handler offset is calculated. It ignores the fact that OFF registers (which specify the offset for each interrupt, and can be set to a custom value) are present in the device (see chapter 8.6.1 in the document linked above), and assumes the offset is encoded in `Cause<15:8>`. This may be the case for other interrupt controllers, but for PIC32MZ it is not, as these bits store the current interrupt's requested priority. This obviously results in incorrectly calculated interrupt handler address, which results in a jump to a completely wrong address.

The comments nearby the incorrect calculation in qemu's sources [target-mips/helper.c:587](https://github.com/sergev/qemu/blob/master/target-mips/helper.c#L584) suggest that developers are aware of the issue, as they specifically mention PIC32MZ as an example where their assumption is incorrect.

Possible workarounds:
- Fixing qemu. I expect this particular issue would be fairly easy to fix, but preparing the fix and getting it upstream are two different stories, and we probably wouldn't like to fork qemu :-)
- Exploiting the bug. We might use interrupt priorities to encode vector offsets. This is silly, because it renders them useless, and will only work on qemu and not on the real device.
[...]

PR #14, opened 2015-12-21
Initial interrupt support, core timer interrupts 

rafalcieslak:
This branch introduces initial support for interrupt configuration and handling.

It uses a single interrupt handler (vector spacing = 0), as discussed per #8. The core timer is configured to emit a high-priority interrupt every millisecond. The handler stores register status on the stack, and then calls a C procedure, which recognizes interrupt type and processes it. In case of a core timer interrupt, it increments a millisecond counter. That counter is used to implement `mdelay` which blocks execution until that counter reaches desired value. `kernel_main` uses that function in the main loop, and reports the elapsed time (in ms) since it started.

In this implementation nested interrupts are unavailable. Should I enable them already?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,8 @@ CPPFLAGS += -isystemsmallclib/include
 LDLIBS   = smallclib/smallclib.a
 
 PROGNAME = main
-SOURCES_C = main.c uart_raw.c
-SOURCES_ASM = startup.S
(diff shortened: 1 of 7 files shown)


rafalcieslak:
A recent `git rebase` in this branch completely confused commit history. Since there was only a single commit in this branch, I will cherry-pick it into a new branch, and will open a new, identical PR.

PR #15, opened 2015-12-26
Initial interrupt support, core timer interrupts

rafalcieslak:
This branch introduces initial support for interrupt configuration and handling.

It uses a single interrupt handler (vector spacing = 0), as discussed per #8. The core timer is configured to emit a high-priority interrupt every millisecond. The handler stores register status on the stack, and then calls a C procedure, which recognizes interrupt type and processes it. In case of a core timer interrupt, it increments a millisecond counter. That counter is used to implement mdelay which blocks execution until that counter reaches desired value. kernel_main uses that function in the main loop, and reports the elapsed time (in ms) since it started.

In this implementation nested interrupts are unavailable. Should I enable them already?

[Note: This branch is a clone of `core_timer` with fixed commit history.]

### DIFF
--- a/interrupts.c
+++ b/interrupts.c
@@ -1,11 +1,9 @@
 #include "interrupts.h"
+#include "clock.h"
 #include <libkern.h>
 
 /* Provided by linker script. */
 extern const char __ebase[];
-
-/* This counter is incremented every millisecond. */
-static volatile unsigned int timer_ms_count;
 
 void intr_init(){
 
@@ -63,70 +61,13 @@ void intr_init(){
     mtc0(C0_STATUS, 0, status);
 }
 
-
-void intr_timer_init(){
-    /* Disable interrupts while we are configuring them. */
-    unsigned s = di();
-
-    /* Clear instruction counter. */
-    mtc0 (C0_COUNT, 0, 0);
-    /* Set compare register. */
-    mtc0 (C0_COMPARE, 0, TICKS_PER_MS);
-
-    /* Clear ms counter. */
-    timer_ms_count = 0;
-
-    /* Clear core timer interrupt status. */
-    IFSSET(0) = 1 << 0;
-    /* Set core timer interrupts priority to 6 (highest) */
-    unsigned p = IPC(0);
-    p &= ~PIC32_IPC_IP0(7); // Clear priority 0 bits
-    p |=  PIC32_IPC_IP0(6); // Set them to 6
-    IPC(0) = p;
-    /* Enable core timer interrupts. */
-    IECSET(0) = 1 << 0;
-
-    /* It is safe now to re-enable interrupts. */
-    ei(s);
-}
-
-unsigned intr_timer_get_ms(){
-    return timer_ms_count;
-}
-
-void intr_handler(){
+void intr_dispatcher(){
     unsigned irq_n = PIC32_INTSTAT_VEC(INTSTAT);
-    unsigned compare,count;
     /* Recognize interrupt type. */
     switch(irq_n){
     case PIC32_IRQ_CT:
         /* Core timer interrupt. */
-
-        /* Mark the interrupt as handled. */
-        //IFSCLR(0) = 1 << PIC32_IRQ_CT;
-
-        compare = mfc0(C0_COMPARE, 0);
-        count   = mfc0(C0_COUNT, 0);
-        signed int diff = compare - count;
-        if (diff > 0){
-            /* Should not happen. Potentially spurious interrupt. */
-            return;
-        }
-
-        /* This loop is necessary, because sometimes we may miss some ticks. */
-        while (diff < TICKS_PER_MS){
-
-            compare += TICKS_PER_MS;
-
-            /* Increment the ms counter. This increment is atomic, because
-               entire _interrupt_handler disables nested interrupts. */
-            timer_ms_count += 1;
-
-            diff = compare - count;
-        }
-        /* Set compare register. */
-        mtc0(C0_COMPARE, 0, compare);
-
+        hardclock();
         break;
     default:
         kprintf("Received unrecognized interrupt: %d!\n", irq_n);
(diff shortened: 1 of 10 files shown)


startup.S
```diff
@@ -4,6 +4,7 @@
 /* Exception vector. */
 .section .exception
 .globl __reset_vector
+.org 0x0
```
cahirwpz: I don't understand what [.org](https://sourceware.org/binutils/docs/as/Org.html) does here. Can you explain?

intr.S
```diff
@@ -0,0 +1,91 @@
...
+    sw $t2,  8 ($sp)
+    sw $t3, 12 ($sp)
+    sw $t4, 16 ($sp)
+    sw $t5, 20 ($sp)
+    sw $t6, 24 ($sp)
+    sw $t7, 28 ($sp)
+    sw $t8, 32 ($sp)
+    sw $t9, 36 ($sp)
+
+    /* Asking early, as these instructions require a few cycles. */
```
cahirwpz: > All multiply operations (except the MUL instruction) write to the HI/LO register pair. All integer operations write to the general purpose registers (GPR). Because MDU operations write to different registers than integer operations, integer instructions that follow can execute before the MDU operation has completed. The MFLO and MFHI instruc- tions are used to move data from the HI/LO register pair to the GPR file. **If an MFLO or MFHI instruction is issued before the MDU operation completes, it will stall to wait for the data.**

Isn't that comment somewhat misleading? Or perhaps I don't understand sth?

cahirwpz:
Please fix your code to adhere to [coding conventions](https://github.com/cahirwpz/wifire-os/wiki/Coding-conventions). Choose `intr` or `irq` as abrreviation for the module.

interrupts.c
```diff
@@ -0,0 +1,113 @@
...
+
+    /* Changing EBase must be done with Status:BEV set to 1. */
+    unsigned status = mfc0(C0_STATUS, 0);
+    mtc0(C0_STATUS, 0, status | ST_BEV);
+
+    /* Set EBase. */
+    mtc0(C0_EBASE, 1, __ebase);
+
+    /* Restore Status, set BEV to 0. */
+    status &= ~ST_BEV;
```
cahirwpz: If we had a header file with bit operations you could use BITCLR here.

interrupts.h
```diff
@@ -0,0 +1,35 @@
...
+/* Initializes and enables interrupts. */
+void init_interrupts();
+
+/* This is the single interrupt handler procedure. */
+void interrupt_handler();
+
+
+/* -- CORE TIMER -- */
+
+#define TICKS_PER_MS (1000 * MHZ / 2)
```
cahirwpz: Please, move this line to `global_config.h`.

startup.S
rafalcieslak: It explicitly forces the routine to be placed exactly at the beginning of the `.exception` section. This was unimportant before changes introduced by this branch, because there was only one chunk of code to be placed in that section, so it was okay to assume it is going to be placed there. Not using this directive should yield the same result, but I prefer to be explicit about placement within `.exception`, as offsets there have a special meaning and routines shall not be placed arbitrarily. So this directive simply underlines that this routine has to be at `0xbfc00000`.

In particular, this may save us from problems (by emitting warnings/errors) in case `intr.S` was processed before `startup.S` (without this directive `__reset_vector_` would just get appended after interrupt procedures, which is clearly wrong), and in the very unlikely case that the `__reset_vector_` does not fit before other routines placed at further offsets.

intr.S
rafalcieslak: Apparently I have confused MIPS-II with MIPS32-II, there is indeed no restriction on the instructions surrounding `mflo`/`mfhi`.

startup.S
cahirwpz: Please, add a short comment, so nobody ever be tempted to remove this line.

intr.S
```diff
@@ -0,0 +1,91 @@
+    .set noreorder // Disable automatic instruction reordering
+
+    /* Interrupts vector */
+    .section .exception
+    .globl _interrupt_handler
+    .org 0x200
+_interrupt_handler:
```
cahirwpz: Have you compared your CPU context save / restore routines with [exception.S](http://nxr.netbsd.org/xref/src-freebsd/sys/mips/mips/exception.S#268)?

interrupts.h
```diff
@@ -0,0 +1,35 @@
+#ifndef __INTERRUPTS_H__
+#define __INTERRUPTS_H__
+
+#include "pic32mz.h"
+#include "global_config.h"
+
+/* Disables interrupts. Returns the previous value of Status
+   register. */
+#define di() ({                             \
```
cahirwpz: FYI: FreeBSD declares two helper functions to disable preemption:

    void critical_enter();
    void critical_exit();

cahirwpz:
In near future we'll have to think about interrupt chaining. The idea is that drivers or other kernel-side clients will register their interrupt handler that will get called according to assigned priorities.

interrupts.c
```diff
@@ -0,0 +1,113 @@
...
+
+/* Provided by linker script. */
+extern const char __ebase[];
+
+/* This counter is incremented every millisecond. */
+static volatile unsigned int timer_ms_count;
+
+void init_interrupts(){
+
+    /* Changing EBase must be done with Status:BEV set to 1. */
```
cahirwpz: Please, add short comment on EBase and BEV. What are they?

intr.S
rafalcieslak: My implementation was mildly inspired by [Sergie's LiteBSD](https://github.com/sergev/LiteBSD/tree/master/sys/mips/pic32), and I made sure the information set that gets stored is correct. The routines that you have linked to store much more information than it is actually needed, and I wanted to avoid redundancy.

cahirwpz:
I find this patchset to be almost ready. Just a few small changes are needed:
* factor out *Core timer interrupt* to `hardclock()` routine,
* rename `intr_timer` to `clock` and move to `clock.{c,h}`,
* rename `intr_handler` to `intr_dispatcher` as this will, eventually, call chains of device specific interrupt handlers.

Issue #10, opened 2015-12-15
Print out basic information about CPU capabilities


cahirwpz:
Read CP0 registers to find out:
* cache size and organization,
* FTLB or/and VTLB sizes,
* minimum and maximum page size,
* ... and other interesting facts.

cahirwpz:
Clearly most of this information isn't really useful. However at some point we may choose to expose it through a special filesystem (similar to `/proc`).

Please remove all information that is completely irrelevant (zeros?). Leave only what's mentioned in task description and comment out the rest.

Can you find something similar within Linux or FreeBSD sources and link references to the code?

cahirwpz:
http://nxr.netbsd.org/xref/src-freebsd/sys/mips/include/cpuinfo.h

PR #16, opened 2015-12-27
Cp0 config

Bozydarek:
This branch adds function which prints the configuration registers values, together with their description. (Issue #10)

In the case of bit value - 0 or 1 means respectively false (disabled, not present, etc) or true (enabled, present, etc).

In other cases, I put a description of the meaning of the respective values ( in square brackets ).


Bozydarek:
Now I'm stuck because:
* cache size is 0 and i don't know that is my fault or is it true ?
* I can't find in documentation a clear answer to the question about min and max size pages
* I don't even know what else is interesting in these configure registers

Moreover I think that I should add 1 to mips_cpuinfo.ic_nways and mips_cpuinfo.dc_nways

bitmap.h
```diff
@@ -9,6 +9,7 @@
 #define BIT_CLR(n, b) ((n) &= ~(1 << (b)) )
 #define BIT_INV(n, b) ((n) ^=  (1 << (b)) )
 #define BIT_GET(n, b) (((n) & (1<<(b))) != 0)
+#define BITS_GET(n, l, b) ( ((n) & (( (1<<(l)) - 1) << (b) )) >> b)
```
cahirwpz: Please reimplement it as a [assembly inline](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html) using `EXT` instruction. Implement `BITS_PUT` with `INS` instruction as well.

cahirwpz: Oops... actually those are defined in `mips-mti-elf/include/mips/mips32.h` from toolchain directory, please use them.

main.c
```diff
@@ -58,7 +73,66 @@ void udelay (unsigned usec)
...
     }
 }
-
+/* 
+ * Read configuration registers value, its interpretation 
+ * and save it into the struct mips_cpuinfo
+ */
+void read_config()
+{
+    int config = mfc0(16, 0);
```
cahirwpz: Please use symbolic names for CP0 registers. Look at `mips-mti-elf/include/mips/m32c0.h` in the toolchain directory.

cahirwpz:
Please `git rebase` your branch on top of current master.

main.c
```diff
@@ -70,6 +86,75 @@ void mdelay (unsigned msec)
     while(final > clock_get_ms());
 }
 
+/* 
+ * Read configuration registers value, its interpretation 
+ * and save it into the struct mips_cpuinfo
+ */
+void read_config()
+{
+    int config = mfc0(16, 0); // CP0 Register 16, Select 0
```
cahirwpz: `mips/m32c0.h` contains following definition:

```
#define mips32_getconfig0() mips32_get_c0(C0_CONFIG0)
```

PR #17, opened 2015-12-28
A very basic physical memory manager

rafalcieslak:
Since we are going to need dynamic memory allocation pretty soon, I've started by implementing a simple physical memory manager. It uses a bitmap to track which frames are in use and which are free, and uses first-fit for finding free regions.

The manager is aware that the kernel stores some data in the memory (`.data`, `.bss`, etc.) and the frames containing it are marked as used during initialisation.

The algorithms used in this branch are not efficient. I believe substituting them with optimized variants should be considered a separate task.

The exposed interface boils down to:

```
void init_phys_mem();
void* alloc_frames(size_t n);
void free_frames(const void* p, size_t n);
```

Upon this basis an actual dynamic memory allocator could be implemented. We might also use some existing, fancy `malloc` implementation (smallclib provides one), as usually they only need the interface for getting and freeing n frames.

Toggling `DEBUG_PHYS_MEM` in `phys_mem.h` will cause a simple demonstration to be run during initialization.

### DIFF
--- a/bitops.h
+++ b/bitops.h
@@ -1,0 +1,8 @@
+#ifndef __BITOPS_H__
+#define __BITOPS_H__
+
+#define   SET_BIT(n, b) (n |=  (1 << b) )
+#define CLEAR_BIT(n, b) (n &= ~(1 << b) )
+#define   GET_BIT(n, b) ((n & (1<<b)) != 0)
+
+#endif // __BITOPS_H__
--- a/bitops.h
+++ b/bitops.h
@@ -1,8 +1,10 @@
 #ifndef __BITOPS_H__
 #define __BITOPS_H__
 
-#define   SET_BIT(n, b) (n |=  (1 << b) )
-#define CLEAR_BIT(n, b) (n &= ~(1 << b) )
-#define   GET_BIT(n, b) ((n & (1<<b)) != 0)
+#define   SET_BIT(n, b) ((n) |=  (1 << (b)) )
+#define CLEAR_BIT(n, b) ((n) &= ~(1 << (b)) )
+#define   INV_BIT(n, b) ((n) ^=  (1 << (b)) )
+
+#define   GET_BIT(n, b) (((n) & (1<<(b))) != 0)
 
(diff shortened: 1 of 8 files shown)


bitops.h
```diff
@@ -0,0 +1,8 @@
+#ifndef __BITOPS_H__
+#define __BITOPS_H__
+
+#define   SET_BIT(n, b) (n |=  (1 << b) )
+#define CLEAR_BIT(n, b) (n &= ~(1 << b) )
```
cahirwpz: For consistency and to avoid macro expansion related errors:

#define CLR_BIT(n, b) ((n) &= ~(1 << (b)))

Also, please add INV_BIT.

cahirwpz:
I gave early memory allocation some thought and I reckon it's too early for doing physical memory allocator. Please, have a look at [pmap_bootstrap_alloc](https://github.com/sergev/LiteBSD/blob/master/sys/mips/pic32/pmap.c#L181).

We clearly need to allocate memory before we start virtual memory subsystem. I suspect kernel startup will allocate structures that'll have to be available throughout system lifetime - so no free operation is required. Such memory can be obtained through moving end of `bss` segment.

If in any case such early allocated memory needs to be freed, we at least need to allocate it on page basis. In other words - after vm subsystem starts with memory within `[bss_end, ram_end]` range, we can add some pages from `bss` segment that we consider to be free.

For physical memory management I'm leaning towards simplified freelist method (McKusick, Figure 6.7). 

Could you have a look how early memory allocation is solved in LiteBSD and FreeBSD. After that share your findings with us and we'll decide which path to choose. We should identify real requirements for early memory allocator so we better understand what we really need to implement.

cahirwpz:
I'd like you to extract bit operations on words and arrays to separate file. The interface should be similar to [bitset.h](http://nxr.netbsd.org/xref/src-freebsd/sys/sys/bitset.h) and be generic. I'm sure this will prove to be extremely useful.

cahirwpz:
I chose completely different way to handle this problem. I won't import your PR.

PR #23, opened 2016-01-04
The first proposal of the circular linked list interface.

Killavus:
I tried to follow the naming principles from the project's wiki. This API is
    highly influenced by the model of FreeBSD TAILQ queues. The only difference is
    that this list is circular.
    
    I'm not happy with this whole 'head type' approach. It makes it convoluted and
    adds an additional argument to insertion methods. What I'd like to propose is
    to ditch the head list concept at all. So we have only list elements, we do not
    distinguish head. We do not provide the IS_HEAD, but IS_ELEM generalization. We
    also get rid of APPEND, UNSHIFT and FIRST.
    
    But this is ugly, because there is no natural representation of an empty list.
    There can be so-called hacks, like making a "guard" element, or invalidating
    next and prev pointers to NULL since in a 'normal' at least one-sized circular
    list there is no situation where those pointers are NULL.
    
    But I don't know whether we'd like to do that or not. Let me know what you think
    about it. Maybe I'm missing something?

### DIFF
--- a/klist.h
+++ b/klist.h
@@ -1,0 +1,50 @@
+#ifndef __KLIST_H__
+#define __KLIST_H__
+
+/* First proposal - we distinguish the head of the list.
+ * It makes operations like empty possible, but makes inserting slightly more
+ * complicated. It makes all operations like INSERT_AFTER dependent on the head
+ * structure too. */
+
+// Make a list element structure from a given type.
+#define KL_CLIST_ELEMENT(type) 0
+
+// Make a list head type for a given type.
+#define KL_CLIST_HEAD(head_name, type) 0 
+
+// Initialize the list with an element passed.
+#define KL_CLIST_INIT(head, elem) 0
+
+// Makes a new head.
+#define KL_CLIST_UNSHIFT(head, new_head) 0
+
+// Adds an element to the end (esentially, being the predecessor of a head) 
+// of a list.
+#define KL_CLIST_APPEND(head, elem) 0
+
+// Inserts a new element after the given element.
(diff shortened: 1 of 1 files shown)


cahirwpz:
I don't like reinventing a wheel. Why can't we just take the implementation from FreeBSD? After all, it's well tested and was good enough to be the choice for BSD kernels. Please resist [NIH](https://en.wikipedia.org/wiki/Not_invented_here). Are arguments for not using BSD lists solid and reasonable?

Killavus:
Current proposal is a subset of FBSD implementation with names changed to follow GLib rules. How about it?

cahirwpz:
I gave it a second thought and the conclusion is - please steal `queue.h` and `tree.h` from FreeBSD and integrate with our sources without changing those files.

Killavus:
I added lists from FreeBSD to master already, as well as examples for them. I need to port `tree.h`, but I understood that linked lists are top priority for now :).

cahirwpz:
Please **DO NOT** push to _master_ without my permission! I gave you push access, but I cannot limit it to branches only. Please always create pull request that I'll have an opportunity to review / modify before merging with trunk.

cahirwpz:
I've just reverted your change. Please prepare a proper _pull request_.

cahirwpz:
I took `queue.h` from FreeBSD 4. I'm still waiting for import of `tree.h`.

Killavus:
Ok, so I see from the commit from an hour that both files are properly stolen now?

I'll find another task then. Sorry for inconveniences caused.

Issue #20, opened 2015-12-30
Single header files directory


cahirwpz:
We need one directory to hold all header files.

This is simple build tree restructuring task.

PR #24, opened 2016-01-04
Moved header files to a dedicated directory (#20)

rafalcieslak:


cahirwpz:
Could you move `libkern.h` to `include` as well?

rafalcieslak:
Wouldn't it makes sense now to move `smallclib/extra/*` to `smallclib`?

cahirwpz:
Well... quite a lot, actually :)

Issue #25, opened 2016-01-09
There are two conflicting definitions for _estack


rafalcieslak:
I noticed this issue while implementing a bootstrap memory allocator, but it can certainly haunt us in any context.

Our intention is to place the kernel stack at the end of the RAM (as it grows downwards). This is expressed in `pic32mz.ld` with:

    _estack = ORIGIN(ram) + LENGTH(ram);

This value apparently makes sense, `pic32mz.map` confirms:

                    0x00000000a0080000                _estack = (ORIGIN (ram) + 0x80000)

which is, indeed, 512kB after RAM's beggining. This address value is correct.

To initialize stack pointer, we use the `la` pseudo-instruction: (`startup.S`)
<pre>
_start:
    <b>la  $sp, _estack</b>    // Set stack pointer.
    la  $gp, _gp        // Prepare global pointer.
</pre>

I would expect that the linker will set the value in this instruction so that the effect of this instruction will be setting `sp` to `0xa0080000`. However, it is not:

<pre>
(gdb) disas _start
Dump of assembler code for function _start:
   0xbfc00500 <+0>:	 lui	sp,<b>0xa000</b>
   0xbfc00504 <+4>:  addiu	sp,sp,<b>4236</b>
   0xbfc00508 <+8>:	 lui	gp,0xa000
   0xbfc0050c <+12>: addiu	gp,gp,0
   [...]
</pre>

so in fact (in this case) the stack pointer will be set instead to `0xa000108c`. And it indeed is:

<pre>
(gdb) info registers
[...]
            t8       t9       k0       k1       gp       sp       s8       ra
 R24  00000000 00000000 00000000 00000000 00000000 <b>a000108c</b> 00000000 00000000 
[...]
</pre>

Obviously, this is incorrect. It will cause all sorts of issues, starting with the fact that the stack ends up having just 1kiB before it crashes onto `.bss`. I originally found this issue when I tried to allocate more than 1kiB with the bootstrap memory allocator, and cleared the allocated region, which happened to override stack frames.

I have no idea what may be causing this. The only interesting thing I noticed while investigating this is that apparently that incorrect address is always `_end + 1kiB`, so increasing `.data` or `.bss` by N bytes pushes the stack these N bytes further.

rafalcieslak:
[...]

PR #26, opened 2016-01-09
Renamed _estack to _custom_estack.

rafalcieslak:
This is to work around linker preferring some another definition of `_estack` and ignoring ours (see #25).


rafalcieslak:
This is an invalid fix, see [this comment](https://github.com/cahirwpz/wifire-os/issues/25#issuecomment-170269862).

PR #28, opened 2016-01-10
Implemented early dynamic memory allocation

rafalcieslak:
Implemented a bootstrap memory allocator as suggested in [this comment](https://github.com/cahirwpz/wifire-os/pull/17#issuecomment-168301900). The implementation is similar to [LiteBSD's](https://github.com/sergev/LiteBSD/blob/master/sys/mips/pic32/pmap.c#L181).

I was unsure to which namespace the allocator shall belong, so I ended up using `vm`.

Currently, allocating more than ~4000B will crash the kernel due to #25. However, the currently proposed fix for #25 solves the issue.

### DIFF
--- a/kmem.c
+++ b/kmem.c
@@ -1,0 +1,19 @@
+#include "kmem.h"
+#include "libkern.h"
+#include "common.h"
+
+/* The end of the kernel's .bss section. Provided by the linker. */
+extern char __ebss[];
+
+/* This is the pointer to the extended end of kernel's image.
+ * It shall be always word-aligned. See km_early_alloc for details. */
+static uintptr_t km_kernel_image_end = (uintptr_t)__ebss;
+
+void* km_early_alloc(size_t size){
+    // TODO: Explicitly crash if this function is called after
+    // physical memory manager has been initialized.
+    void* ptr = (void*)km_kernel_image_end;
+    bzero(ptr, size);
+    km_kernel_image_end = ALIGN(km_kernel_image_end + size, sizeof(word_t));
+    return ptr;
+}
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ CPPFLAGS += -Iinclude
 LDLIBS   = smallclib/smallclib.a
 
 PROGNAME = main
-SOURCES_C = main.c uart_raw.c interrupts.c clock.c bitmap.c
+SOURCES_C = main.c uart_raw.c interrupts.c clock.c bitmap.c kmem.c
 SOURCES_ASM = startup.S intr.S init_gpr.S init_cp0.S init_caches.S init_tlb.S
 SOURCES = $(SOURCES_C) $(SOURCES_ASM)
 OBJECTS = $(SOURCES_C:.c=.o) $(SOURCES_ASM:.S=.o)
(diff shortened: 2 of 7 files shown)


vm.c
```diff
@@ -0,0 +1,22 @@
+#include "vm.h"
+#include "libkern.h"
+
+/* This is the pointer to the extended end of kernel's image. See
+   vm_early_alloc for details. */
+char* vm_kernel_image_end;
```
cahirwpz: Should it be defined as static not to leak outside of the compilation unit?

vm.c
```diff
@@ -0,0 +1,22 @@
...
+
+void vm_init(){
+    /* The end of the kernel's .bss section. Provided by the linker. */
+    extern char __ebss[];
+
+    vm_kernel_image_end = __ebss;
```
cahirwpz: Hmm... can't you just use `__ebss instead` of `vm_kernel_image_end` ?

cahirwpz:
Note that bootstrap memory allocator doesn't belong to virtual memory subsystem, but rather kernel memory allocator. Thus the module should be named 'kmem` and the function... well... maybe `km_early_alloc` or `kmalloc_early`.

vm.c
rafalcieslak: I was slightly hesitant about it. I suspected that the memory manager which will deal with the entire RAM pool `[bss_end, stack_end]` will need to know how much memory was allocated this way, and it might be implemented in a different module.

However, instead of peeking at `vm_kernel_image_end` it could as well allocate a 0-size block. So probably there is no need to expose this variable to different TU's.

vm.c
rafalcieslak: I believe I can't! `__ebss` is a linker symbol. It does not exist in the memory, there is no address under which it's value would be stored. It makes no sense assigning values to it on runtime, as it is not available there. It makes sense, however, to dereference it, since it points to some address, but that's not a good place to store values, as it's the end of the `.bss` section indeed, and it will be used for other purposes. Copying the value of `__ebss` to a variable makes sense too, as the linker places a constant under `__ebss`:

<pre>
(gdb) disas km_init 
Dump of assembler code for function km_init:
   0xbfc00c98 <+0>:	lui	v0,0xa000
   0xbfc00c9c <+4>:	addiu	v0,v0,144
   0xbfc00ca0 <+8>:	jr	ra
   0xbfc00ca4 <+12>:	sw	v0,40(gp)
End of assembler dump.
</pre>

kmem.c
```diff
@@ -0,0 +1,22 @@
...
+
+    km_kernel_image_end = __ebss;
+}
+
+void* km_early_alloc(size_t size){
+    // TODO: Explicitly crash if this function is called after
+    // physical memory manager has been initialized.
+    void* ptr = km_kernel_image_end;
+    bzero(ptr, size);
+    km_kernel_image_end += size;
```
cahirwpz: MIPS CPU will trap on a word access to unaligned address. Please make sure that address returned by `km_early_alloc` is divisible by `sizeof(word_t)`.

rafalcieslak:
While fixing address alignment, I've also added `ALIGN_WORD_UP` and `ALIGN_WORD_DOWN` macros to `common.h`, they may come in handy.

PR #30, opened 2016-01-11
Uniform formatting for source files

rafalcieslak:
In this branch, I propose a set of options for astyle that will unify code formatting style. The options are listed in `astyle.options`, each with brief explanation. I have chosen the options so that the current formatting of source files matches the options most closely. Turns out `k&r` style is mostly correct.

The command for running astyle with these options is

    astyle --options=astyle.options --recursive "*.h" "*.c"

and it is also listed, for convenience, inside `astyle.options`. We may consider hooking it so that it always runs before `git commit`, but it must be kept in mind that as it is not a standard tool, it may be not installed on the particular machine.

The second commit applies the proposed formatting to all source files. Clearly, the `pic32mz.h` file was an utter tabs/spaces mess, so the fix for that takes up 90% of the entire diff.

### DIFF
--- a/astyle.options
+++ b/astyle.options
@@ -1,0 +1,30 @@
+# This file defines the options for astyle.
+# To run astyle with these options on all files, use:
+#     astyle --options=astyle.options --recursive "*.h" "*.c"
+# It will format all files in-place, saving a backup in .orig file.
+#
+# The documentation listing available options can be viewed at:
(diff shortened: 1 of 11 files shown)


cahirwpz:
I'll perform reformatting in one step in single commit. Thanks for astyle template, I pushed it to master.

PR #43, opened 2016-02-14
Added physical page allocator

coodie:
I tried to make code readable. I hope i won't have to explain anything.

However there are some points to make:

- I added special macro `FAIL`, which prints message and loops forever. This was used by me to debug the code, but i decided to leave these macros, since they show how the function should be used and we don't have assertions so maybe in the future they can be improved or changed to something else.
- Addres of space which should remain unallocated (end of .bss section) is automatically calculated from linker, so if someone adds some code, it automatically recalculates manageable segment. There are some constants like `PAGE_SIZE`, `NUM_OF_PAGES`, `STACK_SPACE`. Initialisation is based on all of these constans.

### DIFF
--- a/include/vm_phys.h
+++ b/include/vm_phys.h
@@ -1,0 +1,43 @@
+#ifndef _VM_PHYS_H_
+#define _VM_PHYS_H_
+
+#include <queue.h>
+#include <common.h>
+#include <global_config.h>
+
+#define VM_NFREEORDER 9
+#define VM_RESERVED 1
+#define VM_FREE 2
+
+typedef uintptr_t vm_addr_t;
+typedef uintptr_t vm_paddr_t;
+
+typedef struct vm_page {
+  TAILQ_ENTRY(vm_page) listq;
+  vm_addr_t virt_addr;
+  size_t size;
+
+  TAILQ_ENTRY(vm_page) freeq;
+  vm_paddr_t phys_addr;
+  size_t order;
+  uint32_t flags;
+} vm_page_t;
+
+TAILQ_HEAD(vm_freelist, vm_page);
+
+typedef struct vm_phys_seg {
+  TAILQ_ENTRY(vm_phys_seg) segq;
+  vm_paddr_t start;
+  vm_paddr_t end;
+  struct vm_freelist free_queues[VM_NFREEORDER];
+  struct vm_page *page_array;
+} vm_phys_seg_t;
+
+void vm_phys_add_seg(vm_paddr_t start, vm_paddr_t end);
+void vm_phys_init();
+vm_page_t *vm_phys_alloc(size_t order);
+void vm_phys_free(struct vm_page *page);
+void vm_phys_print_free_pages();
+void vm_phys_reserve(vm_paddr_t start, vm_paddr_t end);
+
+#endif /* _VM_PHYS_H */
--- a/vm_phys.c
+++ b/vm_phys.c
@@ -1,0 +1,241 @@
+#include <vm_phys.h>
+#include <queue.h>
+#include <malloc.h>
+#include <libkern.h>
+
+#define POW2(x) (1 << (x))
+#define PG_START(pg) ((pg)->phys_addr)
+#define PG_END(pg) ((pg)->phys_addr + (1 << (pg->order)) * PAGESIZE)
+#define PG_FREEQ(seg, i) (&(seg)->free_queues[(i)])
+
+TAILQ_HEAD(vm_seglist, vm_phys_seg);
+
+static struct vm_seglist seglist;
+
+void vm_phys_print_free_pages() {
+  vm_phys_seg_t *seg_it;
+  log("free pages:");
+  TAILQ_FOREACH(seg_it, &seglist, segq) {
+    for (int i = 0; i < VM_NFREEORDER; i++) {
+      if (!TAILQ_EMPTY(PG_FREEQ(seg_it, i))) {
+        kprintf("size %dKiB:", (PAGESIZE / 1024) << i);
+        vm_page_t *pg_it;
+        TAILQ_FOREACH(pg_it, PG_FREEQ(seg_it, i), freeq)
+          kprintf(" (%p,%p)", PG_START(pg_it), PG_END(pg_it));
+        kprintf("\n");
+      }
+    }
+  }
+  kprintf("\n");
+}
+
+void vm_phys_add_seg(vm_paddr_t start, vm_paddr_t end) {
+  struct vm_phys_seg *seg = kernel_sbrk(sizeof(struct vm_phys_seg));
+  int page_array_size = (end - start) / PAGESIZE;
+
+  seg->start = start;
+  seg->end = end;
+  seg->page_array = kernel_sbrk(page_array_size);
+  TAILQ_INSERT_TAIL(&seglist, seg, segq);
+
+  if ((vm_paddr_t)ALIGN(seg->start, PAGESIZE) != (vm_paddr_t) seg->start)
+    panic("start not page aligned");
+  if ((vm_paddr_t)ALIGN(seg->end, PAGESIZE) != (vm_paddr_t) seg->end)
+    panic("end not page aligned");
+
+  for (int i = 0; i < page_array_size; i++) {
+    seg->page_array[i].phys_addr = seg->start + PAGESIZE * i;
+    seg->page_array[i].order = 0;
+    seg->page_array[i].flags = 0;
+  }
+
+  for (int i = 0; i < VM_NFREEORDER; i++)
+    TAILQ_INIT(PG_FREEQ(seg, i));
+  int last_page = 0;
+  int seg_size = page_array_size;
+  for (int ord = VM_NFREEORDER - 1; ord >= 0; ord--)
+    while (POW2(ord) <= seg_size) {
+      vm_page_t *page = &seg->page_array[last_page];
+      page->flags |= VM_FREE;
+      page->order = ord;
+      TAILQ_INSERT_HEAD(PG_FREEQ(seg, page->order), page, freeq);
+      seg_size -= POW2(ord);
+      last_page += POW2(ord);
+    }
+}
+
+void vm_phys_init() {
+  TAILQ_INIT(&seglist);
+}
+
+/* Takes two pages which are buddies, and merges them */
+static vm_page_t *merge(vm_page_t *page1, vm_page_t *page2) {
+  if (page1->order != page2->order)
+    panic("trying to connect pages of different orders");
+  if (page1 > page2) {
+    vm_page_t *tmp = page1;
+    page1 = page2;
+    page2 = tmp;
+  }
+  if (page1 + POW2(page1->order) != page2)
+    panic("pages are not buddies");
+  page1->order++;
+  return page1;
+}
+
+static vm_page_t *find_buddy(vm_phys_seg_t *seg, vm_page_t *page) {
+  vm_page_t *buddy = NULL;
+
+  /* Page is left buddy if it's addres is divisible
+   * by 2^(page->order+1), otherwise it's right buddy */
+
+  if ((page - seg->page_array) % POW2(page->order + 1) == 0)
+    buddy = page + POW2(page->order);
+  else
+    buddy = page - POW2(page->order);
+
+  if (buddy->order != page->order)
+    return NULL;
+
+  /* Check if buddy is on freelist */
+  vm_page_t *it;
+
+  TAILQ_FOREACH(it, PG_FREEQ(seg, page->order), freeq)
+    if (it == buddy)
+      break;
+
+  return buddy;
+}
+
+static void split_page(vm_phys_seg_t *seg, vm_page_t *page) {
+  int order = page->order;
+
+  if (TAILQ_EMPTY(PG_FREEQ(seg, order)))
+    panic("cannot split segment");
+
+  /* Works because page is some indice of page_array */
+  vm_page_t *buddy = &page[1 << (order - 1)];
+
+  TAILQ_REMOVE(PG_FREEQ(seg, page->order), page, freeq);
+
+  page->order = order - 1;
+  buddy->order = order - 1;
+
+  TAILQ_INSERT_HEAD(PG_FREEQ(seg, page->order), page, freeq);
+  TAILQ_INSERT_HEAD(PG_FREEQ(seg, buddy->order), buddy, freeq);
+  buddy->flags |= VM_FREE;
+}
+
+static void vm_phys_reserve_from_seg(vm_phys_seg_t *seg, vm_paddr_t start,
+                                     vm_paddr_t end) {
+  for (int i = VM_NFREEORDER - 1; i >= 0; i--) {
+    vm_page_t *pg_ptr;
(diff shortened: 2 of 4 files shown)


include/vm_phys.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <queue.h>
+#include <common.h>
+
+#define VM_NFREEORDER 9
+#define VM_RESERVED 1
+
+typedef uintptr_t vm_addr_t;
+typedef uintptr_t vm_paddr_t;
+typedef struct vm_page* vm_page_t;
```
cahirwpz: Please do not. You're forcing a reader to remember that `vm_page_t` is not simply derived from `struct vm_page` but it's actually a pointer to `struct vm_page`. It's evil.

vm_phys.c
```diff
@@ -0,0 +1,211 @@
...
+/* Note that generally size of page isn't constant
+ * This constant is minimal granularity of page */
+
+#define PAGE_SIZE 4096
+#define NUM_OF_PAGES 128
+#define STACK_SPACE 4 /* stack space in number of pages */
+
+#define ORDER_SIZE(x) (1 << (x))
+
+#define FAIL(MSG...) do { kprintf(MSG); while(1); } while (0);
```
cahirwpz: Please define a panic function in separate module. Have a look at: [panic.c](http://fxr.watson.org/fxr/source/boot/common/panic.c)

vm_phys.c
```diff
@@ -0,0 +1,211 @@
...
+    vm_page_t page_buddy = &page[1 << (order-1)];
+
+    TAILQ_REMOVE(&basic_seg.free_queues[order], page, freeq);
+
+    page->order = order-1;
+    page_buddy->order = order-1;
+
+    TAILQ_INSERT_HEAD(&basic_seg.free_queues[order-1], page, freeq);
+    TAILQ_INSERT_HEAD(&basic_seg.free_queues[order-1], page_buddy, freeq);
+    return 1;
```
cahirwpz: If you're not planning to return anything else than 1 then don't return anything.

vm_phys.c
```diff
@@ -0,0 +1,211 @@
...
+            seg_size -= ORDER_SIZE(ord);
+            last_page += ORDER_SIZE(ord);
+        }
+
+}
+
+/* Takes two pages which are buddies, and coalesce them */
+static vm_page_t coalesce(vm_page_t page1, vm_page_t page2) {
+    if(page1->order != page2->order)
+        FAIL("coalesce: trying to connect pages of different orders\n");
```
cahirwpz: You're the only user of `coalesce` function and you managed to put more defensive code into this function that actual functionality. I'd get rid of it.

vm_phys.c
```diff
@@ -0,0 +1,211 @@
+#include <vm_phys.h>
+#include <queue.h>
+#include <kmem.h>
+#include <libkern.h>
+
+/* Note that generally size of page isn't constant
+ * This constant is minimal granularity of page */
+
+#define PAGE_SIZE 4096
```
cahirwpz: Rename to VM_PAGESIZE and move to header file.

vm_phys.c
```diff
@@ -0,0 +1,211 @@
...
+#define NUM_OF_PAGES 128
```
cahirwpz: Please redesign your code so you don't need to use these constants.

vm_phys.c
```diff
@@ -0,0 +1,211 @@
...
+#define ORDER_SIZE(x) (1 << (x))
+
+#define FAIL(MSG...) do { kprintf(MSG); while(1); } while (0);
+
+static struct vm_phys_seg basic_seg;
+static struct vm_page vm_page_array[NUM_OF_PAGES];
+
+/* Next three constants are provided by linker */
+
+/* Start of data section is start of pageable memory. */
```
cahirwpz: Please redesign your code so you don't need to use these symbols.

coodie: What's wrong with using these symbols? And what do you mean by not using them.

cahirwpz: They're exposing some nasty details about the platform. Please re-read my post - I edited it.

cahirwpz: I'd like to keep this file as platform independent as possible. Usable ranges of RAM can be delivered in many ways. The most basic one is to calculate it from linker symbols. Another, more sophisticated, is to fetch it from firmware and boot loader.

include/vm_phys.h
coodie: I understand your point.

There is one curiosity however. In FreeBSD's code vm_page_t is defined as **pointer** to page.

cahirwpz: As `vm_page` is at the very core of kernel they could make a mistake early enough so that changing it now would involve modification of too many lines of code. However, if you find a sensible reason behind it I'll reconsider.

coodie:
I updated pull request. I hope the code is ok now. This took me a while because writing reserving pages was quite nightmare, but i finally made it work.

cahirwpz:
Could you invoke `git rebase master` - otherwise I'll see my changes in your PR.

cahirwpz:
Please, run `astyle` as well if you haven't done it yet.

coodie:
I did git rebase master, and it automatically merged me into your branch, that's why this pull request contains your changes, seems like there must be other way to rebase remote branches.

cahirwpz:
I just want to be 100% sure I don't pull in any unnecessary changes. You may try to `git rebase -i master`, then squash everything to single commit and invoke `git push --force`.

Nevertheless I want PR to be clean.

vm_phys.c
```diff
@@ -0,0 +1,251 @@
+#include <vm_phys.h>
+#include <queue.h>
+#include <kmem.h>
+#include <libkern.h>
+
+
+#define ORDER_SIZE(x) (1 << (x))
+
+/* this is a stub, should be moved to other file */
+#define panic(MSG...) do { kprintf(MSG); while(1); } while (0);
```
cahirwpz: Please do it now, as a separate and small PR. I'll import it into master and you'll rebase this branch.

vm_phys.c
```diff
@@ -0,0 +1,251 @@
...
+    int page_array_size = (seg->end - seg->start) / PAGE_SIZE;
+    seg->page_array = km_early_alloc(page_array_size);
+    TAILQ_INSERT_TAIL(&seglist, seg, segq);
+    if(ALIGN(seg->start, PAGE_SIZE) != (uintptr_t) seg->start)
+        panic("vm_phys_add_seg: start not page aligned ")
+    if(ALIGN(seg->end, PAGE_SIZE) != (uintptr_t) seg->end)
+        panic("vm_phys_add_seg: end not page aligned ")
+
+    int i;
+    for(i = 0; i < page_array_size; i++) {
```
cahirwpz: Can't you write `for(int i = 0; ...` in C99?

coodie:
I rebased the branch already (i should just use git push --force, otherwise i'll have another merge commit which i don't want to have).

 I try to use astyle but for some reason i can't use options:

style=google
remove-brackets

include/vm_phys.h
```diff
@@ -0,0 +1,42 @@
...
+struct vm_page {
+    TAILQ_ENTRY(vm_page) listq;
+    vm_addr_t virt_addr;
+    size_t size;
+
+    TAILQ_ENTRY(vm_page) freeq;
+    vm_paddr_t phys_addr;
+    size_t order;
+    uint32_t flags;
+};
```
cahirwpz: The convention inside kernel is that every `struct X` has a `typedef struct X X_t`.

vm_phys.c
coodie: I have been using `make qemu` command. And here is error i was getting:

`error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode`

include/vm_phys.h
```diff
@@ -0,0 +1,42 @@
+#ifndef _VM_PHYS_H_
+#define	_VM_PHYS_H_
+
+#include <queue.h>
+#include <common.h>
+
+#define VM_NFREEORDER 9
+#define VM_RESERVED 1
+#define PAGE_SIZE 4096
```
cahirwpz: Move it to `global_config.h`, please.

cahirwpz:
Please use `astyle.option` configuration file from main repo directory.

vm_phys.c
cahirwpz: Interesting... I pushed a small change that force C11 standard. Can you merge and check again ?

Run `make qemu VERBOSE=1` to check what command actually perfoms compilation.

coodie: The command: `mips-mti-elf-gcc -mips32r2 -EL -g -nostdlib -T pic32mz.ld -Wl,-Map=pic32mz.map main.o uart_raw.o interrupts.o clock.o kmem.o context.o vm_phys.o startup.o intr.o init_gpr.o init_cp0.o init_caches.o init_tlb.o context-mips.o smallclib/smallclib.a -o main.elf`

The flags: 
`mips-mti-elf-gcc -mips32r2 -EL -g -nostdlib -T pic32mz.ld -Wl,-Map=pic32mz.map`

vm_phys.c
```diff
@@ -0,0 +1,251 @@
...
+struct vm_page * vm_phys_alloc(size_t order) {
+
+    struct vm_phys_seg *seg_it;
+    TAILQ_FOREACH(seg_it, &seglist, segq) {
+        struct vm_page *res;
+        res = vm_phys_alloc_from_seg(seg_it, order);
+        if(res != NULL) return res;
+    }
+
+    panic("vm_phys_alloc: cannot allocate page of given order");
```
cahirwpz: Ekhm... if this function returns `NULL` then a swapper (that we don't have yet) should be launched.

vm_phys.c
```diff
@@ -0,0 +1,251 @@
...
+    struct vm_phys_seg *seg_it;
+    TAILQ_FOREACH(seg_it, &seglist, segq) {
+        if(page->phys_addr >= seg_it->start && 
+            page->phys_addr + page->size <= seg_it->end  )
+        {
+            vm_phys_free_from_seg(seg_it, page);
+            return;
+        }
+    }
+    panic("vm_phys_free: page out of range\n");
```
cahirwpz: Good! But it'd be even nicer if you dumped segment list `(start, end)` and page address under question :)

vm_phys.c
coodie: So the solution would be to make this simply return NULL in term of fail, right?

vm_phys.c
```diff
@@ -0,0 +1,251 @@
...
+}
+
+static void vm_phys_free_from_seg(struct vm_phys_seg *seg, struct vm_page *page) {
+
+    /* check if page is reserved */
+    if((page->flags & VM_RESERVED) != 0)
+        panic("vm_phys_free: trying to free reserved page: %p\n", page->phys_addr);
+
+    /* check if page was free'd before */
+    struct vm_page *it;
```
cahirwpz: Isn't it better to just add VM_FREE flag and check it?

vm_phys.c
```diff
@@ -0,0 +1,251 @@
...
+        if(seg_it->start <= start && seg_it->end >= end) {
+            vm_phys_reserve_from_seg(seg_it, start, end);
+            return;
+        }
+    }
+
+}
+
+static struct vm_page * vm_phys_alloc_from_seg(struct vm_phys_seg *seg, size_t order) {
+    struct vm_page *res = NULL;
```
cahirwpz: `s/res/page/g`

vm_phys.c
cahirwpz: I'm confused - how on earth linker complains about for loop?

vm_phys.c
cahirwpz: Yes.

cahirwpz:
Ok. I think we're close to state where PR is ready to be merged with master. Most of my remarks are now focused primaryly on coding style. Please fix as much as possible - I'll take care of rest.

vm_phys.c
coodie: Does it complain on your computer?

cahirwpz: If I don't merge e07f0d097eed568e6b3d06a46eead99a51ab12f4 then it does complain. Please rebase your branch.

coodie:
I corrected everything you said, i'm looking forward to finally close this issue, however if you find any bugs please inform me.

vm_phys.c
```diff
@@ -0,0 +1,249 @@
...
+
+static void vm_phys_free_from_seg(vm_phys_seg_t *seg, vm_page_t *page) {
+
+  page->flags |= VM_FREE;
+  /* check if page is reserved */
+  if ((page->flags & VM_RESERVED) != 0)
+    panic("vm_phys_free: trying to free reserved page: %p\n", page->phys_addr);
+
+  /* check if page was free'd before */
+  if ((page->flags & VM_FREE) != 0)
```
cahirwpz: Hmm... this will always panic ;-) Look at L212.

coodie: You're good one, too bad i'm quite tired now and i just want to have it done, that's why i don't see such simple bugs.

include/vm_phys.h
```diff
@@ -0,0 +1,43 @@
...
+#define VM_RESERVED 1
+#define VM_FREE 2
+
+typedef uintptr_t vm_addr_t;
+typedef uintptr_t vm_paddr_t;
+
+typedef struct vm_page {
+  TAILQ_ENTRY(vm_page) listq;
+  vm_addr_t virt_addr;
+  size_t size;
```
cahirwpz: One major fix is needed in your code. Fields in lines 16-18 must not be used by code in `vm_phys.c`. They're here for, yet to come, virtual page allocator.

vm_phys.c
```diff
@@ -0,0 +1,249 @@
...
+static struct vm_seglist seglist;
+
+void vm_phys_print_free_blocks() {
+  vm_phys_seg_t *seg_it;
+  TAILQ_FOREACH(seg_it, &seglist, segq) {
+    for (int i = 0; i < VM_NFREEORDER; i++) {
+      kprintf("order %d:\n", i);
+      vm_page_t *pg_it;
+      TAILQ_FOREACH(pg_it, &seg_it->free_queues[i], freeq)
+      kprintf("(%p,%p) ", pg_it->phys_addr,
```
cahirwpz: I suggest creating helper macros:
* `pg_start(pg) ((pg)->phys_addr))`
* `pg_end(pg) ((pg)->phys_addr + (1 << (pg)->order))`

... and use them throughout your whole code. This will increase readability.

coodie: True. I was thinking about this.

vm_phys.c
cahirwpz: I'm kinda tired as well. Let's finish the work tomorrow in the morning.

vm_phys.c
cahirwpz: Yeah. If you see some frequenty used code it's probably a good candidate to abstract out as a macro. For instance there's plenty of `seg->free_queues[order]` in your code.

vm_phys.c
coodie: Good idea, i'll finally get something solid to eat :)

coodie:
I rebased and updated issue, please review it.

cahirwpz:
Regardless of current state I'll take it. I'll rework the PR and ask you to review it. In the meantime please move to another task.

PR #47, opened 2016-02-18
Add physical memory allocator

cahirwpz:
Ready for merge with master.

### DIFF
--- a/vm_phys.c
+++ b/vm_phys.c
@@ -1,0 +1,241 @@
+#include <vm_phys.h>
+#include <queue.h>
+#include <malloc.h>
+#include <libkern.h>
+
+#define POW2(x) (1 << (x))
+#define PG_START(pg) ((pg)->phys_addr)
+#define PG_END(pg) ((pg)->phys_addr + (1 << (pg->order)) * PAGESIZE)
+#define PG_FREEQ(seg, i) (&(seg)->free_queues[(i)])
+
+TAILQ_HEAD(vm_seglist, vm_phys_seg);
+
+static struct vm_seglist seglist;
+
+void vm_phys_print_free_pages() {
+  vm_phys_seg_t *seg_it;
+  log("free pages:");
+  TAILQ_FOREACH(seg_it, &seglist, segq) {
+    for (int i = 0; i < VM_NFREEORDER; i++) {
+      if (!TAILQ_EMPTY(PG_FREEQ(seg_it, i))) {
+        kprintf("size %dKiB:", (PAGESIZE / 1024) << i);
+        vm_page_t *pg_it;
+        TAILQ_FOREACH(pg_it, PG_FREEQ(seg_it, i), freeq)
+          kprintf(" (%p,%p)", PG_START(pg_it), PG_END(pg_it));
+        kprintf("\n");
+      }
+    }
+  }
+  kprintf("\n");
+}
+
+void vm_phys_add_seg(vm_paddr_t start, vm_paddr_t end) {
+  struct vm_phys_seg *seg = kernel_sbrk(sizeof(struct vm_phys_seg));
+  int page_array_size = (end - start) / PAGESIZE;
+
+  seg->start = start;
+  seg->end = end;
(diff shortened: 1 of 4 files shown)


coodie:
Am i supposed to review the code, and accept merge pull request?

cahirwpz:
You're supposed to review my changes. I'll merge PR myself.

vm_phys.c
```diff
@@ -0,0 +1,241 @@
...
+    vm_page_t *pg_it = TAILQ_FIRST(PG_FREEQ(seg, i));
+    while (pg_it) {
+      if (PG_START(pg_it) >= start && PG_END(pg_it) <= end) {
+        /* if segment is containted within (start, end) remove it from free
+         * queue */
+        pg_ptr = pg_it;
+        while (pg_ptr++ < pg_it + POW2(pg_it->order))
+          pg_ptr->flags |= VM_RESERVED;
+        TAILQ_REMOVE(PG_FREEQ(seg, pg_it->order), pg_it, freeq);
+        /* Go to start of list because it's corrupted */
```
cahirwpz: Could you explain why is it corrupted? This comment is worrisome.

coodie: You should not worry.

Usualy when you iterate over some structure, and modify it while traversing it, the iterator isn't reliable anymore. So the safest way to keep on traversing is to simply start doing it from the very start. And this is what i am doing here. So actually this comment is quite wrong, because list is totally ok, it's more like iterator is corrupted, or more like 'the way of iterating'.  In my opinion i should be able to remove iterator while iterating, because removing out of list doesn't change pointers of iterator, but pointers of elements adjacent to iterator, but i didn't want to rely on this, and (i don't know why) it didn't work.

In my opinion, the best idea would be to create set of elements to be deleted, but this means i'd need to use some kind of dynamic allocation to keep necessary information. 

So basically, i chose the safest solution.

vm_phys.c
```diff
@@ -0,0 +1,241 @@
...
+      } else if ((PG_START(pg_it) < start && PG_END(pg_it) > start) ||
+                 (PG_START(pg_it) < end && PG_END(pg_it) > end)) {
+        /* if segments intersects with (start, end) split it in half */
+        split_page(seg, pg_it);
+        /* Go to start of list because it's corrupted */
+        pg_it = TAILQ_FIRST(PG_FREEQ(seg, i));
+      } else {
+        pg_it = TAILQ_NEXT(pg_it, freeq);
+        /* if neither of two above cases is satisfied, leave in free queue */
+      }
```
coodie: Some grammar mistakes in comments in this function, also i'd change:
`/* Go to start of list because it's corrupted */`
to
`/* Go to start of list because iterator isn't reliable after removal */`

vm_phys.c
```diff
@@ -0,0 +1,241 @@
+#include <vm_phys.h>
+#include <queue.h>
+#include <malloc.h>
+#include <libkern.h>
+
+#define POW2(x) (1 << (x))
+#define PG_START(pg) ((pg)->phys_addr)
+#define PG_END(pg) ((pg)->phys_addr + (1 << (pg->order)) * PAGESIZE)
+#define PG_FREEQ(seg, i) (&(seg)->free_queues[(i)])
```
coodie: Just a comment. I preffered older macro. I mean if you look at line:
`TAILQ_INSERT_HEAD(PG_FREEQ(seg, page->order), page, freeq);`
It's not immediately clear, what is happening here. And there is small opportunity of making a mistake, when choosing order of queue to insert page to. I'd rather prepare page, set up all necessary fields, and simply use:
`INSERT_FREE_PAGE(seg, page); `
Sure, we lose opportunity to insert at start or end, but algorithm isn't based on position of page in list, but on it's existence in list.

vm_phys.c
cahirwpz: Ok. I'll include this change.

vm_phys.c
cahirwpz: I'll reconsider it before merging with master.

PR #49, opened 2016-03-29
Basic interrupt handler registration

coodie:
Not really pull request. Just a place to discuss this topic before final implementation, and show what I've done already.

Couple of questions:

1. For what types of interrupts are we going to use this handler? This is important since this will determine the implementation. Can you provide simple example?
2. With this implementation and my knowledge about interrupts I need to keep list of all possible intr_events, and then pick proper interrupt event based on irq which comes. Is it a problem or is there different (possibly better) solution? On FreeBSD intr_event handlers are called individually by specific interrupts. In order to apply this solution we'll need to chose set of events we'd like to handle.

We can't really implement this well until we have threads. Other solution which would solve problem with time-consuming interrupts (without using threads) would be to put them on special list and keep on running them in main thread, until nothing is done. But obviously this isn't good.

As long as I don't have threads I can't implement this properly.


intr_handler.c
```diff
@@ -0,0 +1,45 @@
...
+  events[irq] = ie;
+}
+
+void intr_event_remove_handler(intr_handler_t *ih) {
+  TAILQ_REMOVE(&ih->ih_event->ie_handlers, ih, ih_next);
+}
+
+void intr_event_execute_handlers(intr_event_t *ie) {
+  intr_handler_t *it;
+  TAILQ_FOREACH(it, &ie->ie_handlers, ih_next) {
```
cahirwpz: What will happen if all handler report FILTER_STRAY? Will a spurious interrupt request be handled correctly, so that it will not be triggered again just after interrupt handler routine finishes?

coodie: This routine doesn't check what handlers return (note that there is no filter routine yet), they are executed every time interrupt comes. Can you provide example situation when stray_interrupt occurs?

cahirwpz: Imagine an example where you have two devices attached to the same interrupt line. One of those devices is correctly configured and handled by interrupt handler. But the other was not reset in a proper way, which has an effect of sending interrupt requests on apparently random occasions. Thus if you don't clear interrupt request flag for specific line a processor will reenter the interrupt handler as soon as you return from it.

coodie:
I updated this code.

When it comes to unwanted changes in main.c, they are result of running astyle on this file. Therefore I recommend to run astyle by you in main repo (I pushed the modified script), and when trying to merge this with master branch simply resolve conflicts in main.c . Unless you don't care and want some of lines in main.c to be overwritten by me running astyle (but you know, we want to keep real history on blame probably).

Issue #37, opened 2016-01-18
Generic kernel memory allocator


cahirwpz:
For reference look at [sys/malloc.h](http://fxr.watson.org/fxr/source/sys/malloc.h) and [kern/kern_malloc.c](http://fxr.watson.org/fxr/source/kern/kern_malloc.c). Please also read [malloc(9)](https://www.freebsd.org/cgi/man.cgi?query=malloc&sektion=9) manual.

**Depends on #42**

```
#define MB_MAGIC 0xC0DECAFE

typedef struct mem_block {
    uint32_t mb_magic;               /* if overwritten report a memory corruption error */
    uint32_t mb_size;                /* size < 0 => free, size > 0 => alloc'd */
    TAILQ_HEAD(, mem_block) mb_list;
    uint8_t mb_data[0];
} mem_block_t;

#define MB_UNIT sizeof(mem_block_t)

#define MA_SINGLE 0 /* Single large block allocated within arena. */
#define MA_BLOCKS 1 /* Arena provides space for small blocks. */

/* Allocation larger than MA_THRESHOLD will be handled by allocation of single arena. */
#define MA_THRESHOLD (2 * VM_PAGESIZE)

typedef struct mem_arena {
    TAILQ_ENTRY(mem_arena) ma_list;
    uint16_t ma_pages; /* Size in pages. */
    uint16_t ma_flags;
    TAILQ_HEAD(, mem_block) ma_freeblks;
    TAILQ_HEAD(, mem_block) ma_usedblks;
} __attribute__((aligned(MB_UNIT))) mem_arena_t;

/* Flags to malloc */
#define M_WAITOK    0x0000 /* ignore for now */
#define M_NOWAIT    0x0001 /* ignore for now */
#define M_ZERO      0x0002 /* clear allocated block */

typedef struct malloc_pool {
    SLIST_ENTRY(malloc_pool) mp_next;  /* Next in global chain. */
    uint32_t mp_magic;                 /* Detect programmer error. */
    const char *mp_desc;               /* Printable type name. */
    TAILQ_HEAD(, mem_arena) *mp_arena; /* First managed arena. */
} malloc_pool_t;

/* Defines a local pool of memory for use by a subsystem. */
#define MALLOC_DEFINE(pool, desc)     \
    malloc_pool_t pool[1] = {         \
        { NULL, M_MAGIC, desc, NULL } \
    };

#define MALLOC_DECLARE(pool) \
    extern malloc_pool_t pool[1]

[...]

PR #51, opened 2016-05-03
Add a generic memory allocator.

Darge:


cahirwpz:
You should have push access to my repo. Could you create the branch in my repo so I can apply some changes.

cahirwpz:
Have you had a closer look at #37 ?

include/memory_allocator.h
```diff
@@ -0,0 +1,61 @@
+#ifndef __MEMORY_ALLOCATOR_H__
+#define __MEMORY_ALLOCATOR_H__
+
+#define MAGIC 0xC0DECAFE
+#define ALIGNMENT 8
+
+TAILQ_HEAD(mb_list, mem_block);
+TAILQ_HEAD(mp_arena, mem_arena);
+
+typedef struct mem_block {
```
cahirwpz: Most likely you don't need to expose it to malloc's user. If so please put it into `memory_allocator.c`.

main.c
```diff
@@ -10,17 +10,18 @@
 #include <uart_cbus.h>
 #include <rtc.h>
 #include <pci.h>
+#include "memory_allocator.h"
 
 typedef struct cpuinfo {
```
cahirwpz: This change does not relate to what you're implementing. If so please remove irrelevant modification and postpone it to another patch.

memory_allocator.c
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  return NULL;
+}
+
+void deallocate(void *addr, malloc_pool_t *mp) {
+  mem_block_t *mb = (mem_block_t *)(((char *)addr) - sizeof(mem_block_t));
+
+  if (mb->mb_magic != MAGIC || mp->mp_magic != MAGIC)
+    kprintf("Memory corruption detected!\n");
```
cahirwpz: Panic here! Check for consistency also when you're allocating memory. Don't be too defensive (e.g. don't scan whole list every time) but still try to catch a memory misuse at every reasonable opportunity.

memory_allocator.c
```diff
@@ -0,0 +1,203 @@
...
+    if ((char *)addr >= start && (char *)addr < start + current->ma_size) {
+      remove_used_memory_block(current, mb);
+      add_free_memory_block(current, mb, mb->mb_size + sizeof(mem_block_t));
+    }
+  }
+}
+
+void print_free_blocks(malloc_pool_t *mp) {
+  mem_arena_t *current_arena = NULL;
+  TAILQ_FOREACH(current_arena, &mp->mp_arena, ma_list) {
```
cahirwpz: From my experience is much easier to debug memory allocator if you dump whole list at once without making difference between allocated or free block.

include/memory_allocator.h
Darge: I moved all the struct definitions to `memory_allocator.c`. Is that what you meant, or I did too much?

main.c
Darge: Do you mean using the linter?

memory_allocator.c
Darge: Done.

main.c
cahirwpz: astyle

memory_allocator.c
Darge: I haven't implemented the "size is less than zero, so it's a free block" feature yet, because I am not convinced that it's better than simply another field boolean field in the struct. In the current implementation, the size is always > 0.
Should I implement it?

main.c
Darge: That's a valid point. Done.

memory_allocator.c
```diff
@@ -0,0 +1,255 @@
...
+#define MAGIC 0xC0DECAFE
+#define ALIGNMENT 8
+
+TAILQ_HEAD(mb_list, mem_block);
+TAILQ_HEAD(mp_arena, mem_arena);
+
+typedef struct mem_block {
+  uint32_t mb_magic; /* if overwritten report a memory corruption error */
+  uint32_t mb_size;
+  uint16_t mb_flags;
```
cahirwpz: If you drop this a block header will take only 16 bytes instead of 24.

Darge: Done.

memory_allocator.c
```diff
@@ -0,0 +1,258 @@
...
+  - use the mp_next field of malloc_pool
+*/
+
+#define MAGIC 0xC0DECAFE
+#define ALIGNMENT 8
+
+TAILQ_HEAD(mb_list, mem_block);
+TAILQ_HEAD(mp_arena, mem_arena);
+
+int32_t abs(int32_t x)
```
cahirwpz: static inline ;-)

memory_allocator.c
```diff
@@ -0,0 +1,258 @@
...
+{
+  if (x < 0)
+    return -x;
+  return x;
+}
+
+typedef struct mem_block {
+  uint32_t mb_magic; /* if overwritten report a memory corruption error */
+  int32_t mb_size; /* size < 0 => free, size > 0 => alloc'd */
+  TAILQ_ENTRY(mem_block) mb_list;
```
cahirwpz: Hmm... is there any good reason to keep the list of used blocks ? Perhaps just a list of free blocks would do, and we could save some space by:
```
union {
  TAILQ_ENTRY(mem_block) mb_list;
  uint64_t mb_data[0];
};
```
What do you think ?

Darge: The list of used blocks might be useful for debugging, because we keep track of every block. Of course it might be more of a convenience thing, because we can recreate the list of free blocks: everything that isn't a free block is a used block.

That's a clever idea, but from my understanding, that union won't save us any space, because uint64_t mb_data[0] doesn't use any memory: it just might add padding to the end, so that the structure's size is a multiple of 8.
Without saving any space, I'd personally leave the used blocks be, because they use only about 16 bytes in every `mem_arena_t`.

include/memory_allocator.h
```diff
@@ -0,0 +1,27 @@
...
+
+#define MALLOC_DECLARE(pool) \
+    extern malloc_pool_t pool[1]
+
+/* Flags to malloc */
+#define M_WAITOK    0x0000 /* ignore for now */
+#define M_NOWAIT    0x0001 /* ignore for now */
+#define M_ZERO      0x0002 /* clear allocated block */
+
+void malloc_add_arena(malloc_pool_t *mp, void *start, size_t size);
```
cahirwpz: It'd be nice to have consistent naming scheme for these, and later integrate those with `malloc.h`.

memory_allocator.c
cahirwpz: For debugging purposes you can scan all blocks in arena without using pointers anyway - just start at the beginning and move forward by `abs(blk->mb_size)`.

You're right union won't save any space, but the smallest usable block would be 16 bytes with header then (for up to 8 byte allocations, of course). I reckon it's worth considering since the implementation of that is trivial.

memory_allocator.c
```diff
@@ -16,9 +16,16 @@
...
+{
+  if (x < 0)
+    return -x;
+  return x;
+}
+
 typedef struct mem_block {
   uint32_t mb_magic; /* if overwritten report a memory corruption error */
-  uint32_t mb_size;
+  int32_t mb_size; /* size < 0 => free, size > 0 => alloc'd */
```
cahirwpz: Wouldn't you be better off with the opposite encoding - i.e. `/* size > 0 => free, size < 0 => alloc'd */`?

You would have to consider `size < 0` case only for allocated blocks then, instead of using `abs` all the time.

memory_allocator.c
Darge: With union:
> 
> (gdb) print mb
> $11 = (mem_block_t *) 0x80001770
> 
> (gdb) print &mb->mb_list
> $10 = (struct {...} *) 0x80001778
> 
> (gdb) print mb->mb_data 
> $7 = 0x80001778
> 
> 
> 
> 
> 
<br>

Without union:
> 
> (gdb) print mb
> $9 = (mem_block_t *) 0x80001770
> 
> (gdb) print &mb->mb_list 
> $8 = (struct {...} *) 0x80001778
> 
> (gdb) print mb->mb_data
> $7 = 0x80001780

It does save us space!

However: [https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html](url)
> A structure containing a flexible array member, or a union containing such a structure (possibly recursively), may not be a member of a structure or an element of an array. (However, these uses are permitted by GCC as extensions.)

Do you still want to do this? I think that this feature is really shady. For example, this doesn't compile with my gcc version 4.8.4, even though that it, in my opinion, should:

> struct widget
> {
>   int a;
>   union {
>     struct {int x; int y;};
>     int b[];
>   };
> };
> 
> int main()
> {
>   return 0;
> }

> gcc -std=gnu99 file.c
> gcc -std=gnu11 file.c
> both give:
> untitled.c:6:9: error: flexible array member in union

It does compile with g++ though.

memory_allocator.c
Darge: You are right. With the opposite, I have to use `abs` only once, and the calculations are simpler too.

include/memory_allocator.h
Darge: Please give me some directions on how I can improve the naming scheme.

Do you mean to use the name `malloc` instead of `allocate`? I could do that, but I would have to use the `-fno-builtin` flag, because right now gcc complains:
> include/memory_allocator.h:21:7: error: conflicting types for built-in function ‘malloc’ [-Werror]

memory_allocator.c
cahirwpz: Note, that there's a difference between `int b[]` and `int b[0]`. With the latter you won't get an error message and it will be happily accepted even by older compilers (i.e. like 10-15 years old if you really care about portability).

include/memory_allocator.h
cahirwpz: Well... it's a shame we haven't used `-fno-builtin` so far. As we don't use external standard C library we really should have turned that option on long time ago. Does adding it to `Makefile.common` break anything right now?

cahirwpz: I personally like following interface [malloc.h](http://fxr.watson.org/fxr/source/sys/malloc.h) from FreeBSD. We could draw from it.

memory_allocator.c
```diff
@@ -0,0 +1,260 @@
...
+  kprintf("\n");
+}
+
+void test_memory_allocator() {
+  char text[] = "this is a pool for testing purposes";
+  MALLOC_DEFINE(test_pool, text);
+  TAILQ_INIT(&test_pool->mp_arena);
+
+
+  char array[2100];
```
cahirwpz: Please allocate memory from buddy systems. Kernel's stack is usually limited in size to one or two pages, thus deep recursion and array allocations are highly disadvised.

PR #54, opened 2016-05-13
ge_table is now array of addresses

coodie:
Look at title for more description.
@cahirwpz Please review this code. I'm bad assembly programmer, so I'd rather make sure everything works, before this goes to repo.

Besides I think irq_handler asm code can be simplified and we can directly jump to proper functions. I guess it is enough to mark proper functions with ((interrupt)) attribute. For example we could remove entire 'irq_handler' label in intr.S and then mark hardclock with ((attribute)), and other functions aswell. Note that there are more awesome properties of attributes, for more information check here: 

https://gcc.gnu.org/onlinedocs/gcc/MIPS-Function-Attributes.html

### DIFF
--- a/interrupts.c
+++ b/interrupts.c
@@ -4,6 +4,28 @@
 
 extern const char _ebase[];
 
+void intr_init() {
+  /*
+   * Enable Vectored Interrupt Mode as described in „MIPS32® 24KETM Processor
+   * Core Family Software User’s Manual”, chapter 6.3.1.2.
+   */
+
+  /* The location of exception vectors is set to EBase. */
+  mips32_set_c0(C0_EBASE, _ebase);
+  mips32_bc_c0(C0_STATUS, SR_BEV);
+  /* Use the special interrupt vector at EBase + 0x200. */
+  mips32_bs_c0(C0_CAUSE, CR_IV);
+  /* Set vector spacing for 0x20. */
+  mips32_set_c0(C0_INTCTL, INTCTL_VS_32);
+
+  /*
+   * Mask out software and hardware interrupts. 
+   * You should enable them one by one in driver initialization code.
+   */
+  mips32_set_c0(C0_STATUS, mips32_get_c0(C0_STATUS) & ~SR_IPL_MASK);
+
+  intr_enable();
+}
 
 static intr_event_t *events[8];
 
@@ -58,7 +80,6 @@ void intr_event_execute_handlers(intr_event_t *ie) {
 #endif
 }
 
-
 static const char *exceptions[32] = {
   [EXC_INTR] = "Interrupt",
   [EXC_MOD]  = "TLB modification exception",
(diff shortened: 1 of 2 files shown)


interrupts.c
```diff
@@ -118,4 +96,56 @@ void kernel_oops() {
...
+
+  /*
+   * Mask out software and hardware interrupts. 
+   * You should enable them one by one in driver initialization code.
+   */
+  mips32_set_c0(C0_STATUS, mips32_get_c0(C0_STATUS) & ~SR_IPL_MASK);
+
+
+  /* Prepare general exception table with proper exception handler addresses. */
+  general_exception_table[EXC_INTR] =   (uint32_t)kernel_oops;
```
cahirwpz: Why can't you initialize the table the same way I did with `execeptions`? It should work if you define it as `void *general_exception_table[32]`.

intr.S
```diff
@@ -37,10 +37,11 @@ general_exception:
...
         nop
 
-        sll     $k0, $k0, 1            /* Multiply cause by two */
-        la      $k1, ge_table          /* Compute address of proper handler in ge table */
-        add     $k1, $k1, $k0
-        jr      $k1                    /* Jump to handler */ 
+        la      $k1, general_exception_table
+        sll     $k0, $k0, 2                           /* Multiply exc_code by 4 */  
+        add     $k1, $k1, $k0                         /* Add exc_code */  
+        lw      $k1, 0($k1)                           /* Compute address of proper handler in ge table */
```
cahirwpz: If an entry is `NULL` choose `kernel_oops` to avoid filling in `general_exception_table` in entries that are not handled.

interrupts.c
```diff
@@ -4,28 +4,6 @@
 
 extern const char _ebase[];
 
-void intr_init() {
```
cahirwpz: Please try to avoid moving code around the file. This distracts a reader! I personally learned that it's best to keep PR as small as possible.

interrupts.c
coodie: It's impossible because addresses of these functions aren't constants, and are defined at link time.

interrupts.c
coodie: I don't like reading code like this aswell so I totally understand your pain. But this has to be moved because tlb_refill_handler isn't visible yet, so intr_init has to be at the end of file.

cahirwpz: Ok then.

interrupts.c
cahirwpz: I'll investigate that. It should work by means of data-to-code relocations.

cahirwpz:
Please verify if my change work (I don't have ovpsim installed locally) and merge.

intr.S
```diff
@@ -20,8 +20,8 @@
 
         .org 0x0
 tlb_refill:
-        j tlb_exception_handler
-        nop
+        sw $k0, 0($zero)
```
cahirwpz: Was that intended or is it a testing artifact ?

coodie:
Everything works fine. I'll merge this. (Yayz, first merged by me pull request!)

Issue #39, opened 2016-02-12
Run queues


cahirwpz:
This is a basic structure to maintain a list of threads ready to be executed. On top of it a simple scheduler should be implemented. For reference look at: [runq.h](http://fxr.watson.org/fxr/source/sys/runq.h) and [kern_switch.c](http://fxr.watson.org/fxr/source/kern/kern_switch.c). The only dependency for this task is to have a kernel thread structure defined.

cahirwpz:
I mused on *wait channels*, *turnstiles* and scheduler for a while. This structure seems to be essential for dealing with threads. Please implement it as soon as possible. This task should be relatively easy to complete.

cahirwpz:
Use [include/runq.h](https://github.com/cahirwpz/mimiker/blob/master/include/runq.h) as a starting point please.

cahirwpz:
While the task seems to be implemented, it lacks proper integration. I'll close it when runq gets its first use by a thread short term scheduler.

cahirwpz:
As of 3a989a7aaef2a850042f98b0d3cfe20281960729 run queues are in use and work well. Good job!

PR #55, opened 2016-05-25
Implement runq

Darge:
Implements #39


cahirwpz:
Please reorganize the test in such a way that it's runnable outside of kernel code. Look for `_KERNEL` identifier in [subr_unit.c](http://fxr.watson.org/fxr/source/kern/subr_unit.c) to understand what I mean. Modify `Makefile` to compile `runq-test` from `runq.c` as well.

Issue #40, opened 2016-02-12
Sleep queues


cahirwpz:
Develop a mechanism that will have similar responsibility to FreeBSD's [sleepqueue](https://www.freebsd.org/cgi/man.cgi?query=sleepqueue&sektion=9). For reference look at: [sleepqueue.h](http://fxr.watson.org/fxr/source/sys/sleepqueue.h) and [subr_sleepqueue.h](http://fxr.watson.org/fxr/source/kern/subr_sleepqueue.c).

> Sleep queues provide a mechanism for suspending execution of a thread until some condition is met.  Each queue is associated with a specific wait channel when it is active, and only one queue may be associated with a wait channel at any given point in time.

cahirwpz:
Please use [include/sleepq.h](https://github.com/cahirwpz/mimiker/blob/master/include/sleepq.h) as a starting point, please.

Darge:
How will we keep the pointer to the current thread structure?
Linux calculates its address using the stack pointer: [link](http://stackoverflow.com/questions/12193281/how-does-the-kernel-know-what-is-the-current-thread).

cahirwpz:
Your changes were merged after minor fixes with master. Nonetheless development of sleep queues must continue as the module is incomplete. Thanks for your work!

PR #56, opened 2016-05-26
implement sleepq

Darge:
Implements #40


Darge:
Merged with master.

I left TODO in places where threads are supposed to be put to sleep and woken up. We don't have a scheduler yet.

I also don't know if I should panic if `sleepq_signal` or `sleepq_broadcast` have no threads to wake up.

And what about the new field, `void *td_wchan` in thread.h?

PR #65, opened 2016-08-13
synchronize the scheduler

Darge:


cahirwpz:
This PR was an attempt to synchronize the scheduler. However there were no right tools to do so in place. Hence a second attempt must be devised.

I'm trying to put all prerequisites in place. Please have a look at 2daefcc7ac09ea98ec646135dee27693686b13a1. I'm now going to change context switching routines as discussed offline. Stay tuned.

PR #68, opened 2016-08-15
Mutexes

Darge:


critical_section.c
```diff
@@ -0,0 +1,24 @@
+#include <critical_section.h>
+#include <interrupts.h>
+#include <common.h>
+#include <libkern.h>
+
+volatile int cs_level = 0;
```
pdziepak: This `volatile` is redundant (and, actually, harmful). Disabling interrupts should already sufficiently protect the accesses to the variable on single-core platforms and on multi-core systems it is not enough anyway. So the only thing that `volatile` here does is that it, needlessly, prevents compiler optimisations.

critical_section.c
```diff
@@ -0,0 +1,24 @@
...
+
+void cs_leave() {
+  if (during_intr_handler())
+    return;
+
+  assert(cs_level > 0);
+
+  cs_level--;
+  if (cs_level == 0)
+    intr_enable();
```
pdziepak: This will break if someone enters critical section with interrupts disabled by direct call to `intr_disable`. General practice is to have interrupt disabling function to return the previous state so that it can be restored when the section requiring to be not interrupted is left.

include/interrupts.h
```diff
@@ -39,6 +39,8 @@ void intr_event_remove_handler(intr_handler_t *ih);
 void intr_event_execute_handlers(intr_event_t *ie);
 void run_event_handlers(unsigned irq);
 
+bool during_intr_handler();
+
 #define intr_disable() __extension__ ({ asm("di"); })
```
pdziepak: I'm aware that it's not part of this patch, but since it is in the context I'm going to comment anyway.
Interrupt disabling and enabling functions should also clobber memory so that the compiler is not tempted to reorder memory accesses.

include/mutex_yield.h
```diff
@@ -0,0 +1,13 @@
+#ifndef __MUTEX_YIELD_H__
+#define __MUTEX_YIELD_H__
+
+#include <common.h>
+#include <mips.h>
+
+typedef volatile uintptr_t mtx_yield_t;
```
pdziepak: As I mentioned in the comments to the other commit this `volatile` shouldn't be needed.

mutex_yield.c
```diff
@@ -0,0 +1,73 @@
...
+      return;
+    }
+    cs_leave();
+  }
+}
+void mtx_yield_unlock(mtx_yield_t *mtx) {
+  assert(mtx);
+
+  cs_enter();
+  *mtx = 0;
```
pdziepak: It doesn't look like it is necessary to enter critical section here. Such simple, aligned, writes are usually guaranteed to be atomic. All that is needed is a compiler barrier so that memory accesses are not reordered after releasing the mutex.

Also, it is quite sad that such small functions cannot be inlined.

critical_section.c
Darge: Take a look at the following situation:
`cs_enter(); cs_leave(); cs_enter(); cs_leave();`.
What if the compiler decides to inline all of these calls and read cs_level only once? Between the two middle function calls, an interrupt might happen that changes the context. That new context might modify the cs_level value. I suppose that the compiler might do such an optimization and I didn't want to take chances.

BTW my initial idea was to use that cs_level variable every time interrupts are enabled/disabled, including the interrupt-handling routines.  Then it would be a debugging tool too.

sleepq.c
```diff
@@ -0,0 +1,161 @@
...
+  for (int i = 0; i < SC_TABLESIZE; i++)
+    LIST_INIT(&sleepq_chains[i].sc_queues);
+}
+
+sleepq_t *sleepq_lookup(void *wchan) {
+  sleepq_chain_t *sc = SC_LOOKUP(wchan);
+
+  sleepq_t *sq;
+
+  LIST_FOREACH(sq, &sc->sc_queues, sq_entry) {
```
pdziepak: This doesn't really scale, does it? True, there is a hash table involved, but since it's size is fixed the complexity is still linear. Obviously, such implementation is nice and general, but rather sub-optimal for mutexes.
Perhaps mutexes could keep a pointer to their sleep queue so that lookups won't be necessary.

critical_section.c
Darge: As I have said above, my initial idea was to use these two functions every time interrupts are enabled/disabled (with the first two lines deleted of course). Then this issue wouldn't come up.

Yes, that's true that it would break in that situation, that's a nice catch, thank you!

critical_section.c
pdziepak: That's why interrupt disable and enable should also be a compiler barrier (as I mentioned in another comment) so that compiler won't misoptimise things. Right, now `cs_level` is indeed protected but other memory accesses in the critical section can still be reordered.

Counter for interrupt disable/enable seems like a good idea, it's similar to that saving previous state on interrupt disable I mentioned the other time.

include/interrupts.h
Darge: Oh, so this is how you wanted to deal with the issue that I hopefully solved with a "volatile" :)

pdziepak: The problem with `volatile` is that it is insufficient and too much at the same time.
Firstly, it protects only `cs_level` – accesses to all other variables may still be reordered out of the critical section. Secondly, it needlessly prevents reordering and merging of accesses to `cs_level` *inside* the critical section.

mutex_yield.c
Darge: I have another version of that mutex with atomic operations, but I delivered this one, because I wasn't completely sure what memory order I should use, what fences etc. One battle at a time :)

include/interrupts.h
Darge: You are right, forbidding the compiler from reordering `intr_disable()` and `intr_enable()` is indeed superior.
This project was meant to teach us something and you certainly helped in that.

mutex_yield.c
pdziepak: Well, this code already relies quite heavily on the fact that there is only a single core so most of the interesting stuff doesn't apply here.

The important things are that storing 0 in `mtx` is not interrupted and the memory accesses are not reordered by compiler (they can be reordered by the CPU since there is no other processor to see that). I am not very familiar with MIPS so this is something that should be checked, but I am pretty sure that `*mtx = 0` is already atomic (assuming mtx points to an aligned address, but it should, anyway). Obviously, using C11 atomics would be more explicit and safer. In that case `memory_order_relaxed` (i.e. the weakest one) is enough, since, as I mentioned earlier, there is no other CPU to see any reorderings.

As for compiler barriers, it is either `asm volatile("" : : : "memory")` – an empty inline assembly that tells GCC that memory was modified or `atomic_signal_fence` (again, the weakest one that only "synchronises the compiler") with `memory_order_seq_cst` to tell the compiler that nothing shall pass.

PR #71, opened 2016-08-21
Change a global array size from 0 to 1

Darge:


Darge:
I decided to simply push the change, sorry for trouble.

PR #72, opened 2016-09-30
Pmap remove pde

coodie:
I had a problem with debugging which turned out not to be a bug in my code but... in Makefile, after running `make clean` and `make` code started to work, someone (possibly having more experience in writing Makefiles) needs to look into Makefile because this consumed few hours of my time.

I'm still not 100% sure that code works, this needs some kind of testing, and I'll work on it. There are few points to make:

- Number of mapped entries is number of valid entries. When there are no valid PTE entries on PDE page I assume that PDE page can be deleted.
- Since I decided to count valid entries I needed space to do so. I had couple of ideas like
 - Add array with PDE_SIZE entries which would keep count for every PDE entry. This is obviously wrong because this would eat A LOT of space and increase initialization time.
 - Add extra structure which would keep reference to allocated page along with the count. I didn't like this idea, because this required me to manage memory of extra structure which i don't like doing (and pmap doesn't have any memory pool).
- The best Idea I could come up with was add extra fields in `vm_page`, note that because it is inside union, there is basically no extra space overhead.

### DIFF
--- a/include/vm.h
+++ b/include/vm.h
@@ -36,7 +36,7 @@ typedef struct vm_page {
     } obj;
     struct {
       TAILQ_ENTRY(vm_page) list;
-      uint16_t entries_cnt; /* number of active entries in pmap */
+      uint16_t valid_cnt; /* number of valid entries in pmap */
       uint16_t pde_index; /* pde index this page is associated with */
     } pt;
   };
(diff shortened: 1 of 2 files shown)


cahirwpz:
I've run into problem with Makefiles as well. I fully agree we have to change build system to something easier to manage. My latest attempts to befriend SCons were quite successful and I'm thinking of migration.

PR #75, opened 2016-10-07
Selectable debugger in launch script

rafalcieslak:
- Optional flag -D selects the gdb frontend to run in debug mode.
- Currently recognized values are gdb, cgdb, gdbtui, emacs and ddd.
- The default behavior of the launch script is unmodified.

### DIFF
--- a/launch
+++ b/launch
@@ -33,6 +33,8 @@ DEBUGGERS = {
     'gdbtui': '%(triplet)s-gdb -tui',
     'emacs' : 'emacsclient -c -e \'(gdb "%(triplet)s-gdb -i=mi %(kernel)s")\''
 }
+DEFAULT_DEBUGGER = 'cgdb'
+assert(DEFAULT_DEBUGGER in DEBUGGERS)
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


launch
```diff
@@ -60,10 +70,21 @@ if __name__ == "__main__":
...
 
     if args.debug:
+
+        if not args.debugger:
+            args.debugger = 'cgdb'
+
+        if not args.debugger in DEBUGGERS:
+            raise SystemExit('Unrecognized debugger %s selected.' % args.debugger)
+
+        dbgcmd = DEBUGGERS[args.debugger]
```
cahirwpz: Please use https://docs.python.org/2/library/stdtypes.html#string-formatting-operations instead of `replace`.

launch
```diff
@@ -60,10 +70,21 @@ if __name__ == "__main__":
         pass
 
     if args.debug:
+
+        if not args.debugger:
```
cahirwpz: I guess you can specify it in corresponding call to `add_argument`.

cahirwpz: ... or even use `choice` parameter!

launch
```diff
@@ -25,15 +26,25 @@ OVPSIM_OVERRIDES = {
...
         description='Launch OVPsim with Malta board.')
     parser.add_argument('kernel', metavar='KERNEL', type=str,
                         help='Kernel file in ELF format.')
     parser.add_argument('-d', '--debug', action='store_true',
-                        help='Start simulation under gdb using cgdb.')
+                        help='Start simulation under gdb.')
+    parser.add_argument('-D', metavar='DEBUGGER', type=str, dest='debugger',
+                        choices=DEBUGGERS.keys(), default='cgdb',
+                        help='Debugger to use in -d mode. Available options: '
+                        '"cgdb" (default), "gdb", "gdbtui", "ddd", "emacs".')
```
cahirwpz: If only that line could be generated from `DEBUGGER.keys()` then the change would be perfect :)

rafalcieslak: I have considered that, but it is tricky to mark which option is the default one. The best I can come up with looks slightly confusing:


    help='Debugger to use in -d mode. Available options: ' +
    ', '.join([x + " (default)" if x == DEFAULT_DEBUGGER else x for x in DEBUGGERS.keys()]) + "."

rafalcieslak: Oh wait, I can totally solve this by slightly changing the message format.

cahirwpz:
Ok. Please squash commits into single one and I'll merge the PR.

cahirwpz:
Ah... I have a possibility to do so by myself. Don't bother.

PR #76, opened 2016-10-11
Fixed TLB entry not invalidated due to bad asid mask

goniz:
Hello

I'm using your platform as a reference in my own hobby educational OS and I've found this typo while debugging..
It seems to cause pmap_map() to fail invalidating TLB entries and this affects mapping a PFN1 page while PFN0 is already mapped or the other way around.


coodie:
I might not be head of the project, but it's obvious bug and should be fixed asap. Thanks a lot for your contribution, I hope the debugging session didn't take too long :).

PR #77, opened 2016-10-11
Fixed C0_PAGEMASK garbage values in tlb_overwrite_random()

goniz:
tlb_overwrite_random() assumed to be given a forth argument to write to C0_PAGEMASK
but C function declaration did not reflect this argument and was called without it.
the PageMask register might be overwritten with garbage data


coodie:
Again - obvious bug, and again - thanks for fixing!

PR #78, opened 2016-10-14
Fixed wrong bitmask in status register in user_exc_enter()

goniz:
The status register SR_KSU bits are: 0 for kernel, 1 for supervisor, 2 for user
Before this commit, user_exc_enter() 'turned off' the kernel-bits instead of the user-bits

### DIFF
--- a/mips/exc.S
+++ b/mips/exc.S
@@ -73,7 +73,7 @@ user_exc_enter:
 
         # Turn off FPU, enter kernel mode,
         # drop exception level and disable interrupts.
-        li      $t1, ~(SR_CU1|SR_KSU_KERN|SR_EXL|SR_IE)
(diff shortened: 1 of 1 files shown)


cahirwpz:
I don't have access to header files from the toolchain right now. Could you tell me what are the values of `SR_KSU_*` definitions?

goniz:
`#define SR_KSU_MASK	0x00000018	/* ksu mode mask */
#define SR_KSU_SHIFT		 3
#define SR_KSU_BITS		 2
#define SR_KSU_USER	0x00000010	/* user mode */
#define SR_KSU_SPVS	0x00000008	/* supervisor mode */
#define SR_KSU_KERN	0x00000000	/* kernel mode */`

cahirwpz:
Please use `SR_KSU_MASK` instead of `SR_KSU_USER` then. Thanks for pointing out the bug!

goniz:
Good point!
Done.

PR #79, opened 2016-10-19
Imported used smallclib sources

rafalcieslak:
The most recent release of Imgtech's MIPS SDK does not include the MIPS Bare Metal SmallCLib, which is now provided as a separate download.

It will be convenient to import the parts of smallclib that we use into our source, instead of pulling these files from the toolchain during compilation. I did my best for this change to have minimal impact on our existing code.


cahirwpz:
Just a quick question – would it be possible to (quickly) replace those files with corresponding implementations from BSD libc?

PR #80, opened 2016-10-21
[WIP] Embedding user ELFs

rafalcieslak:
This branch is not to be merged. It is a temporary sketch/playground that I made for demonstrating how to correctly embed an ELF file as binary data within the kernel image. The exec() function demonstrates it can access the data.

By the way, I checked in 'Allow edits from maintainers' when creating this branch, which should let you push commits onto this branch, even though it originates in my fork.


rafalcieslak:
I also made a simple proof-of-concept ELF loading. At this point `exec()` is able to successfully parse the ELF, prepare a virtual memory map, load contents, enable memory protection* and update the context so that the loaded code gets executed.

The example program (`elf/prog.c`) uses a text string from `.data` to animate a marquee in an array in `.bss`. An easy way to see that it works is to run the program (test in `exec.c/elf`), interrupt the debugger, and see `print (char*)0x4000`, which is where the scrolling string is written.

DISCLAIMER: This is a prototype implementation, I have NOT yet thoroughly examined `exec` implementation in FreeBSD. I did this to investigate what challenges I would face, to gain a better understanding of the problem.

In particular, the loaded code is executed in kernel mode, on kernel stack. The created memory map is not associated with the thread (for it is a kernel thread, the memory map should be associated with a user-space process). So there is a lot more to be done here.

But anyway, I welcome feedback.

sys/exec.c
```diff
@@ -1,21 +1,162 @@
...
+        kprintf("[exec] Exec failed: ELF target architecture is not MIPS\n");
+        return -1;
+    }
+
+    // Take note of the entry point
+    kprintf("[exec] Entry point will be at 0x%08x.\n",
+            (unsigned int)eh->e_entry);
+
+    // Ensure prog header size compliance
+    if(eh->e_phentsize != sizeof(Elf32_Phdr)/sizeof(uint8_t)){
```
pdziepak: This is stricter than necessary. ELF format is designed so that it can be extended while maintaining backward compatibility. For example, a new field to segment entries can be easily added and old implementations will still be able to load the file, just ignoring the new fields.

`sizeof(uint8_t)` seems redundant, it is going to be 1.

sys/exec.c
```diff
@@ -1,21 +1,162 @@
...
+        case PT_SHLIB:
+            kprintf("[exec] Exec failed: ELF file contains a PT_SHLIB segment\n");
+            return -1;
+        case PT_LOAD:
+            kprintf("[exec] Processing a PT_LOAD segment: VirtAddr = %p, "
+                    "Offset = 0x%08x, FileSiz = 0x%08x, MemSiz = 0x%08x, "
+                    "Flags = %d\n", (void*)ph->p_vaddr, (unsigned int)ph->p_offset,
+                    (unsigned int)ph->p_filesz, (unsigned int)ph->p_memsz,
+                    (unsigned int)ph->p_flags);
+            if(ph->p_vaddr % PAGESIZE){
```
pdziepak: This is not an error. There is nothing that requires `PT_LOAD` segments to start at page boundary and if you look at the output of `readelf -l your_faviourite_elf_file` often they do not.

sys/exec.c
```diff
@@ -1,21 +1,162 @@
...
+    }
+
+    vm_map_dump(vmap);
+
+    // We need to correct the thread structure so that it can hold the
+    // correct $gp value for this execution context.
+    thread_t* th = thread_self();
+    th->td_kctx.gp = 0;
+    th->td_kctx.ra = 0xdeadbeef;
+    th->td_kctx.pc = eh->e_entry;
```
pdziepak: Note, that this is not enough. ELF files have specific requirements regarding their entry environment. For example, they need to be passed arguments list, environment variables and auxiliary vector. This should be sufficient for a very simple example, though.

elf/Makefile
```diff
@@ -2,7 +2,7 @@ all: prog.uelf.o
 
 include ../Makefile.common
 
-CFLAGS   = -std=gnu11 -O0 -Wall -Werror -fno-builtin -ffreestanding
+CFLAGS   = -std=gnu11 -Og -Wall -Werror -fno-builtin -ffreestanding
```
pdziepak: In my experience `-Og` does interfere with debugging, despite claiming not doing so. It is not as bad as `-O3`, obivously, but can be annoying. This was few GCC versions ago, though, maybe things have improved since then.

sys/exec.c
```diff
@@ -1,21 +1,162 @@
...
+            // TODO: What if segments overlap?
+            // Temporarily permissive protection.
+            vm_map_entry_t* segment = vm_map_add_entry(vmap, start, end,
+                                    VM_PROT_READ|VM_PROT_WRITE|VM_PROT_EXEC);
+            // Allocate pages backing this segment.
+            segment->object = default_pager->pgr_alloc();
+            // Copy data into the segment
+            memcpy((uint8_t*)start, elf_image + ph->p_offset, ph->p_filesz);
+            // Zero the rest
+            if(ph->p_filesz > ph->p_memsz){
```
pdziepak: I believe this is backwards. Memory at the end needs to be zeroed if `memsz` is larger than `filesz`.

elf/Makefile
rafalcieslak: It has not improved much, apparently, which is why I keep flipping between `-O0` and `-Og`. Debugging the loaded code is tricky, as the gdb can't access debugging information there - so I'm mostly stuck with `objdump -d -S prog.o`. `-Og` tends to do pretty confusing tricks at asm level, while `-O0` usually generates fat too much instructions to step through them conveniently.

sys/exec.c
rafalcieslak: You are correct. The fact that I did not spot this while testing probably means that new pages are already zeroed.

sys/exec.c
rafalcieslak: I wasn't sure why is the program header size specified in the ELF header, but what you say makes perfect sense.

elf/Makefile
pdziepak: You can use command `add-symbol-file` [[1]](https://sourceware.org/gdb/current/onlinedocs/gdb/Files.html#Files) to tell gdb that that there is a file is loaded at the specified address. With some scripting it can even be made to work for dynamically loaded binaries (including DSOs in user space).

rafalcieslak: Wow, this is absolutely brilliant and will save us a lot of trouble. Thanks!

sys/exec.c
cahirwpz: This will do for a prototype implementation. However it'd be nice to document the environment in TODO comment.

include/exec.h
```diff
@@ -0,0 +1,6 @@
+#ifndef __RTC_H__
```
cahirwpz: EXEC_H :)

sys/exec.c
```diff
@@ -0,0 +1,185 @@
...
+       eh->e_ident[EI_MAG1] != ELFMAG1 ||
+       eh->e_ident[EI_MAG2] != ELFMAG2 ||
+       eh->e_ident[EI_MAG3] != ELFMAG3){
+        kprintf("[exec] Exec failed: Incorrect ELF magic number\n");
+        return -1;
+    }
+    // Check ELF class
+    if(eh->e_ident[EI_CLASS] != ELFCLASS32){
+        kprintf("[exec] Exec failed: Unsupported ELF class (!= ELF32)\n");
+        return -1;
```
cahirwpz: Perhaps it's time for stealing `errno` definition?

cahirwpz:
Just a quick note on the PR – it's too big :)

If you're quite confident you'll need header files stolen from *BSD, please put them into another PR and state clearly the origin. I'll merge them immediately.

Please use types from `stdint.h` wherever makes sense to do so. Use `clang-format` to keep formatting consistent.

user/mimiker.ld
```diff
@@ -0,0 +1,55 @@
...
+PHDRS
+{
+    text   PT_LOAD FLAGS(5); /* read-only, executable */
+    data   PT_LOAD FLAGS(6); /* read-write */
+    rodata PT_LOAD FLAGS(4); /* read-only */
+    bss    PT_LOAD FLAGS(6); /* read-write */
+}
+SECTIONS
+{
+    .text 0x00001000:
```
cahirwpz: `mips-linux-gnu` toolchain places `.text` section at `0x0400000`. I guess that's a right choice.

```
PROVIDE (__executable_start = SEGMENT_START("text-segment", 0x0400000)); . = SEGMENT_START("text-segment", 0x0400000) + SIZEOF_HEADERS;
```

sys/exec.c
rafalcieslak: So that exec() might report an error value to the caller user process? As far as I know, `errno` is implemented using thread local storage, which wouldn't be straightforward to steal.

sys/exec.c
rafalcieslak: Interestingly, FreeBSD makes [the same assumption](http://bxr.su/FreeBSD/libexec/rtld-elf/map_object.c#121).

sys/exec.c
cahirwpz: FreeBSD kernel uses various `errno` values internally. You can easily find it out by looking for references to [errno.h](http://bxr.su/FreeBSD/sys/sys/errno.h)

sys/exec.c
cahirwpz: If `PT_LOAD` segments were not page aligned and not relocatable (which is the common case for ELF binaries) then our life would be miserable. No segment sharing or protection would work 100%. Am I wrong?

pdziepak: FreeBSD checks `p_align`, not `p_vaddr`. `p_align` is more or less page size assumed by the ELF file. Specifically, it is required that `p_vaddr ≡ p_offset (mod p_align)`. This, obviously, doesn't mean that `p_offset` or `p_vaddr` need to be aligned in any way.

I don't see how unaligned `p_vaddr` would be a significant problem. Since the offset inside the page is the same for `p_vaddr` and `p_offset` the segment loading code would just have to round the base address down to the page boundary. True, more data will be loaded from the disk than absolutely necessary, but it is not really a problem.

BTW With PIE being more and more common I wouldn't be so sure that fixed-position ELF binaries are the common case.

sys/exec.c
pdziepak: Another option is for the function to return -ERROR_CODE, no need for `errno` in the kernel space.

cahirwpz: Right! I didn't mean to use `errno` global variable, but just error codes.

sys/exec.c
cahirwpz: Let's just keep things simple. We all know PIEs benefits, but we don't need them for the time being.

pdziepak: Oh, I didn't mean to suggest supporting PIEs, at least at this point. But segments with unaligned `p_vaddr` are perfectly normal and common. Adding support for them should be quite trivial.

sys/exec.c
rafalcieslak: That seems fine. However, I suggest we implement `errno` in a separate branch. It will probably be useful in many functions throughout the entire source.

rafalcieslak:
I had to rebase this branch onto `master` to incorporate #83 and #84.

At this point `exec` is prepares a new stack for the user image, copies program argument vector onto that stack, and these values are correctly passed to user `main()`'s `argc` and `argv`. Some hints on how to test this are written down in `./exec.c` test case.

I have consulted FreeBSD implementation of `exec`, and while it does some interesting tricks that I will be happy to present at the seminar, not all of them are relevant for our case. I have corrected various details in my implementation according to theirs, as it helped me to clear up some misconceptions I had.

I have also improved the verbosity of various comments, wrote down the description of the environment in which user programs start, and (finally) formatted the source with using `.clang-format` style (can we maybe hook it to run automatically before commit?).

What I would like to do, but is currently impossible, is to start multiple (different) user programs, and see the scheduler switch between them. What is missing is a process description thread - in particular, the vm map that is associated with a process. When switching between user threads, the current user vm map should also be switched, which is not yet implemented. I am unsure whether this (large...) branch is a good place to work on implementing process structure. However, I have marked places in `do_exec` where interaction with process structure should happen.

rafalcieslak:
Rebased onto master to solve conflicts, updated return values to use errno.

sys/exec.c
```diff
@@ -0,0 +1,292 @@
...
+  vm_addr_t p = *stack_bottom_p - total_arg_size;
+  for (int i = 0; i < args->argc; i++) {
+    size_t n = strlen(args->argv[i]) + 1;
+    arg_vector[i] = p;
+    memcpy((uint8_t *)p, args->argv[i], n);
+    p += n;
+  }
+  assert(p == stack_end);
+  // Move the stack down and word-align it downwards
+  *stack_bottom_p = (*stack_bottom_p - total_arg_size) & 0xfffffffc;
```
cahirwpz: Please read [MIPS ABIs Described – Calling Conventions](https://dmz-portal.mips.com/mw/images/f/fe/MD00305-2B-ABIDESC-SPC-01.03.pdf).

I think that strings should be 4 bytes aligned, and top of the stack - 8 bytes aligned.

sys/exec.c
```diff
@@ -0,0 +1,292 @@
...
+extern uint8_t _binary_prog_uelf_end[];
+
+int get_elf_image(const exec_args_t *args, uint8_t **out_image,
+                  size_t *out_size) {
+  if (strcmp(args->prog_name, "prog") == 0) {
+    *out_image = _binary_prog_uelf_start;
+    *out_size = (size_t)_binary_prog_uelf_size;
+    return 0;
+  } else {
+    return -1;
```
cahirwpz: `errno` value here?

user/start.S
```diff
@@ -0,0 +1,33 @@
...
+
+    .set noreorder
+    .global _start
+
+# This is the starting procedude for user ELF images, used as their
+# entry point. It takes care of preparing argc and argc for main().
+# For details about environment in which this procedure starts, refer
+# to prepare_program_stack in exec.c.
+
+LEAF(_start)
```
cahirwpz: Consider using [NESTED](https://github.com/cahirwpz/mimiker/blob/master/include/mips/asm.h#L261) and `CALLFRAME_*` values here.

sys/exec.c
```diff
@@ -0,0 +1,292 @@
...
+  // This condition will be unnecessary when we take the vmap info
+  // from thread struct. The user thread calling exec() WILL
+  // certainly have an existing vmap before exec()ing.
+  if (old_vmap)
+    vm_map_delete(old_vmap);
+
+  // TODO: Assign the new vm map to the process structure
+
+  // We need to correct the thread structure so that it can hold the
+  // correct $gp value for this execution context.
```
cahirwpz: I don't think we have to care about it... please read about *gp in non-PIC code* from *MIPS ABIs Described*, page 7.

cahirwpz:
Looks mostly ok. Please have a look at my questions and apply fixes if needed.

exec.c
```diff
@@ -0,0 +1,30 @@
+#include <common.h>
+#include <exec.h>
+
+int main() {
+  // This is a simple demonstration of the exec functionality. It
```
rafalcieslak: I will. My comments are pretty verbose, so it's better to write them using a consistent style.

sys/exec.c
rafalcieslak: The linker gets confused when I use any non-zero value for `-G`, so I think we can afford not caring about small data.

user/start.S
rafalcieslak: To be honest, I don't really understand these, and I don't know how to choose them correctly.

rafalcieslak: I'm not sure I see any difference in how gdb behaves when I use NESTED.

rafalcieslak:
I pushed all fixes. I'm also wondering if it wouldn't make sense to move `prepare_program_stack` from `exec.c` to `mips/exec.c`, as it uses strictly machine-specific layout.

cahirwpz:
There're still `//` comments out there ;-) There should be no platform dependent code inside `sys` directory, so moving the code around makes sense - good point!

rafalcieslak:
Fixed and synced with `master`.

cahirwpz:
Good job!

PR #81, opened 2016-10-22
Use global var instead of C0_USERLOCAL for improved portability.

goniz:
A regular global variable is less cool, but much more portable then a c0 USERLOCAL register..

USERLOCAL behaves weird on QEMU and this is one of the changes needed for running mimiker on QEMU.

### DIFF
--- a/mips/exc.S
+++ b/mips/exc.S
@@ -11,6 +11,8 @@
         .globl kern_exc_leave
         .local user_exc_enter
         .local kern_exc_enter
+        .extern _current_thread
+
 
 # [$k0] must be set to value of C0_STATUS
 exc_enter:
@@ -20,7 +22,8 @@ exc_enter:
 
 user_exc_enter:
         # Fetch the context from thread control block.
-        mfc0    $k0, C0_USERLOCAL
+        la      $k0, _current_thread
+        lw      $k0, 0($k0)
         addu    $k0, TD_UCTX
 
         # Save all user registers...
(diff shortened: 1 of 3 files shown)


cahirwpz:
I understand your concern, but in fact we need a solution that fixes more general problem – i.e. how to keep global state of the operating system that needs to be shared between different subsystems.

We should definitely follow [SPOT rule](http://www.catb.org/esr/writings/taoup/html/ch04s02.html#spot_rule). In fact FreeBSD does that by using [pcpu](http://bxr.su/FreeBSD/sys/sys/pcpu.h#150) structure! Could you read the sources and try to figure out how *pcpu_space* is used throught different parts of the kernel? When done I guess you'll have much better idea how to approach removal of `C0_USERLOCAL`.

goniz:
Do we intend to implement SMP support?
as far as I can tell, the pcpu structure is used primary to allow the same code to run on multiple cores

cahirwpz:
In long-term SMP support is highly desired. For the time being we ignore it. However I see `pcpu` structure as a good place to keep, in one place, global variables related to CPU state. 

Please, have a look at [mips/pcpu.h](http://bxr.su/FreeBSD/sys/mips/include/pcpu.h) as well.

goniz:
I did look there already. 
For non smp systems it acts as a simple global structure with "api" macros for access. 
If that's what you have in mind then it's fairly simple to implement.

cahirwpz:
Well... yes. Just define a global structure and API to access it. That'll solve the problem with portability (and QEMU missing `C0_USERLOCAL` implementation) and help adding SMP support later.

goniz:
See pull request #85

PR #82, opened 2016-10-22
boot: use regular jr to transfer control to kernel_boot instead of eret

goniz:
QEMU does not like c0_errpc and freaks out

Please report if this breaks your non-qemu emulator.
And if its does, we'll need to think if we want to solve this in different compilation targets or in startup runtime.

### DIFF
--- a/mips/boot.S
+++ b/mips/boot.S
@@ -7,6 +7,7 @@
         .local kernel_go
 
 # Boot environment is described in MIPS® YAMON™ User's Manual, Chapter 5
+# ... and MIPS® YAMON™ Reference Manual, Chapter 8.3
 #
 # $a0 = argc
 # $a1 = argv
@@ -15,7 +16,8 @@
 #
 # $sp is initialized
 # $gp is not initialized
-# C0_STATUS(SR_IE) is cleared
+#
+# C0_STATUS is same as YAMON™ context, but interrupts are disabled.
 
 LEAF(_start)
         la      $t0, _memsize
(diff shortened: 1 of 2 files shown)


cahirwpz:
`QEMU` is buggy to the point, we decided not to use it and replace it with `OVPsim`.

Could spend some time reading YAMON documentation and add a comment which describes initial boot environment for kernel? I guess this code may be rewritten without using `eret` at all.

cahirwpz:
Please do following tasks:

 * identify corresponding code in [FreeBSD](http://bxr.su/FreeBSD/sys/) source tree and put links here,
 * under **gdb** dump `C0_STATUS` just after `_start`.

goniz:
http://bxr.su/FreeBSD/sys/mips/mips/locore.S

It seems to me, that they turn off most of the bits in SR, then enable fpu and go right ahead to calling platform init and such. 
After that platform_init + mi_startup sets up the rest of the subsystems required for threading and fixes the current execution environment to be the first thread 
When it's done,  it passes itself to the "swapper"  function to be scheduled in/out. 

I hope I got it right.. 😉

goniz:
c0_status value is 0x00400000 at _start

cahirwpz:
The bit is SR_BEV (boot exception vectors). Apparently YAMON starts the program in kernel mode with exceptions turned off – `eret` does not make sense at all here. Seems like `boot.S` needs to be rewritten.

Could you verify YAMON behaviour by reading the documentation (link in main README.md file)?

goniz:
Already working on it.
Will backup my implementation with the YAMON doc.

goniz:
YAMON docs were somewhat vague.. and it seems it did not require much changes to make it work..
Please report if its working as expected in OVPsim.

mips/boot.S
```diff
@@ -29,17 +31,21 @@ clear_bss:
...
 
 kernel_go:
-        # Status.ERL is set by the processor when a Reset, Soft Reset, NMI or
-        # Cache Error exception are taken. When ERL is set:
-        # - The processor is running in kernel mode
-        # - Hardware and software interrupts are disabled
-        # - The ERET instruction will use the return address held in ErrorEPC
-        #   instead of EPC
-
+	/*
```
cahirwpz: Please keep your coding style as similar to the surrounding code as possible.

cahirwpz:
AFAIK there's an option to allow owner of forked project to make changes in your branch? Could you make sure it's enabled - I'd like to put several comments into the file.

goniz:
Are you talking about "allow edits from maintainers"? It's enabled

goniz:
Already working on it.
Will backup my implementation with the YAMON doc.

PR #83, opened 2016-10-23
Imported headers with ELF type definitions from FreeBSD svn

rafalcieslak:
These headers define structures, symbols and constants that follow the ELF file format specification.

They were copied from FreeBSD SVN (as of 22 X 2016). Since they implement an old, solid standard, they were almost unchanged for 18 years, and we will not have to update them in the future.

### DIFF
--- a/include/elf/elf_common.h
+++ b/include/elf/elf_common.h
@@ -1,0 +1,1337 @@
+/*-
+ * Copyright (c) 2000, 2001, 2008, 2011, David E. O'Brien
+ * Copyright (c) 1998 John D. Polstra.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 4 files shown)


include/elf/elf_common.h
```diff
@@ -0,0 +1,1337 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ *
+ * $FreeBSD$
+ */
+
+#ifndef _SYS_ELF_COMMON_H_
+#define	_SYS_ELF_COMMON_H_ 1
+
+#define u_int32_t uint32_t
```
cahirwpz: Please remove it.

cahirwpz:
Would it make sense to run `clang-format` over those header files?

include/elf/elf_common.h
rafalcieslak: Should I also replace all occurrences of `u_int32_t` with `uint32_t`?

cahirwpz: Yes. Obviously :)

pdziepak: I am not so sure that (needlessly) modifying code imported from another project is a good idea. Undoubtedly, in the future there will be a need to sync it with the upstream. That may require more effort than it should if there are conflicts between versions.

rafalcieslak:
Running `clang-format` on these headers using our `.clang-format` style description seriously misaligns values `#define` blocks, which are numerous in these files. For example, it turns:

```
#define	STB_LOCAL	0	/* Local symbol */
#define	STB_GLOBAL	1	/* Global symbol */
#define	STB_WEAK	2	/* like global - lower precedence */
#define	STB_LOOS	10	/* Start of operating system reserved range. */
#define	STB_GNU_UNIQUE	10	/* Unique symbol (GNU) */
#define	STB_HIOS	12	/* End of operating system reserved range. */
#define	STB_LOPROC	13	/* reserved range for processor */
#define	STB_HIPROC	15	/*   specific semantics. */
```

into

```
#define STB_LOCAL 0       /* Local symbol */
#define STB_GLOBAL 1      /* Global symbol */
#define STB_WEAK 2        /* like global - lower precedence */
#define STB_LOOS 10       /* Start of operating system reserved range. */
#define STB_GNU_UNIQUE 10 /* Unique symbol (GNU) */
#define STB_HIOS 12       /* End of operating system reserved range. */
#define STB_LOPROC 13     /* reserved range for processor */
#define STB_HIPROC 15     /*   specific semantics. */
```

I do not know which version we prefer.

cahirwpz:
The latter, please.

rafalcieslak:
Done. Almost all changes done by `clang-format` are limited to trimming multiple spaces.

PR #84, opened 2016-10-23
Imported strcmp from smallclib

rafalcieslak:
This is fairly straightforward. This branch expands the set of basic functions we import from SmallCLib to include `strcmp`.

I came to need this function at 9861491 in `elf-playground`. Basically, if we intend to include multiple ELFs embedded in the kernel image, it will be handy to be able to refer to them by their name, so that `exec` interface can take a string ("path to file") as its argument.

This implementation of `strcmp` is very simple, has no dependencies. Including it has no impact on our existing sources.

### DIFF
--- a/stdc/Makefile
+++ b/stdc/Makefile
@@ -26,6 +26,7 @@ LIB_SOURCES += string/memset.c
 LIB_SOURCES += string/bzero.c
 LIB_SOURCES += string/memcpy.c
 LIB_SOURCES += string/strlen.c
+LIB_SOURCES += string/strcmp.c
 LIB_SOURCES += string/strspn.c
 LIB_SOURCES += stdlib/qsort.c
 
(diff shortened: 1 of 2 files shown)


cahirwpz:
This [one](http://bxr.su/FreeBSD/sys/libkern/strcmp.c) is simplier.

rafalcieslak:
It is indeed, although it is similar to the one from smallclib with `__PREFER_SIZE_OVER_SPEED__` defined. And I suppose I could quickly write an even simpler implementation. But is picking different (nicest?) implementations from various sources per each per stdlib function what we actually want? I supposed that the purpose of agreeing on smallclib was mostly so that we could stick with it and import whatever we need from there. Mixing various sources may become tedious in the future.

cahirwpz:
To be clear - I'll always prefer implementation that is simpler. AFAIR we initially chose `smallclib` over everything else, only because of `printf` simplicity. The choice was optimal at the time being... but I'm not convinced it's valid anymore.

Let's discuss it during seminar this week. I'll accept `strcmp` from `smallclib` for now.

PR #85, opened 2016-10-23
Introduced pcpu_t structure to hold global cpu state

goniz:
Initial proposal for the implementation of the pcpu_t structure used to hold global cpu state

* Used sys/mips/include/pcpu.h and sys/sys/pcpu.h as reference.
* Implemented both C and ASM api macros.
* In the future this will hold per-cpu cpu state, for now its a simple global cpu state structure.
* As of now, only `current_thread` is held in pcpu_t.


cahirwpz:
Good work! Thanks!

PR #87, opened 2016-10-28
Remove asid from vm_map and pmap interface

coodie:
While working on changing vm_map for user threads i realised it is better to keep asids away from interface. In future, when we have processes we might want to create asid or pid allocator, but for now simple incrementation until we run out of asids should be enough.

### DIFF
--- a/mips/pmap.c
+++ b/mips/pmap.c
@@ -23,7 +23,6 @@
 #define PTF_SIZE (PTF_ENTRIES * sizeof(pte_t))
 #define PT_ENTRIES (PD_ENTRIES * PTF_ENTRIES)
 #define PT_SIZE (PT_ENTRIES * sizeof(pte_t))
-#define MAX_ASID 255
 
 static const vm_addr_t PT_HOLE_START =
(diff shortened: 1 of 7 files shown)


cahirwpz:
Please remember to run `clang-format` on your changes :)

mips/pmap.c
```diff
@@ -23,7 +23,6 @@
 #define PTF_SIZE (PTF_ENTRIES * sizeof(pte_t))
 #define PT_ENTRIES (PD_ENTRIES * PTF_ENTRIES)
 #define PT_SIZE (PT_ENTRIES * sizeof(pte_t))
-#define MAX_ASID 255
```
coodie: No it's not. Asids are platform dependent, as far as I know some MIPS architectures have different max asids. I couldn't spot max asid macro in toolchain so i decided it's best to use ENTRY_HI mask.

rafalcieslak:
This is great! I will need this change for experimenting with loading multiple user programs.

coodie:
Running clang-format on mips/pmap.c pollutes this file with other formatting changes not associated with my change, because clang-format wasn't used on this file everytime it was changed.

PR #89, opened 2016-10-28
Ran clang-format over all .c and .h files

rafalcieslak:
Somebody has to do that eventually.

If you don't like this PR being so large and unverifiable by hand, please go ahead and run `clang-format` directly in master. If we demand all new changes to be formatted correctly, it will be much more convenient if existing code also is. Especially as I've configured my editor to run `clang-format` automatically each time I save a file - currently this introduces a lot of unrelated changes to my commits.

By the way, as we are sticking to `.clang-format` style definition, can we agree to remove the redundant file `astyle.options`?


cahirwpz:
Please remove `astyle.options` in another commit.

PR #90, opened 2016-10-28
User threads switch vm_map spaces

coodie:
I decided to introduce user space threads, which are in fact one thread processes. This is needed to test execution of recently introduced elf files and good introduction to processes. This will obviously need some changes, but it is a good start.

### DIFF
--- a/include/mips/ctx.h
+++ b/include/mips/ctx.h
@@ -15,7 +15,7 @@
 typedef struct ctx {
   reg_t s0, s1, s2, s3, s4, s5, s6, s7;
   reg_t gp, sp, fp, ra;
-  reg_t pc, sr, tcb;
+  reg_t pc, sr;
 } ctx_t;
 
 /* Following registers are saved only for user threads. */
--- a/mips/context.c
+++ b/mips/context.c
@@ -22,7 +22,6 @@ void ctx_init(thread_t *td, void (*target)()) {
   /* Initialize registers just for ctx_boot to work correctly. */
   kctx->pc = (reg_t)kern_exc_leave;
   kctx->sp = (reg_t)kframe;
-  kctx->tcb = (reg_t)td;
 
   /* This is the context that ctx_boot will restore. */
   kframe->pc = (reg_t)target;
--- a/mips/genassym.c
+++ b/mips/genassym.c
@@ -13,6 +13,7 @@ ASSYM(TD_KFRAME, offsetof(thread_t, td_kframe));
 ASSYM(TD_KCTX, offsetof(thread_t, td_kctx));
 ASSYM(TD_KSTACK, offsetof(thread_t, td_kstack));
 ASSYM(TD_FLAGS, offsetof(thread_t, td_flags));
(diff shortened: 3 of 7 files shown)


rafalcieslak:
I have been using this branch for a while now, testing it with my experimental parallel user program support, and it appears to be working well.

coodie:
I suspect this will be merged, therefore I'll work on turnstiles from this branchmerge. @cahirwpz  Please merge this ASAP.

coodie:
@cahirwpz Can you please review this? I don't feel like rebasing branches after every commit on master which creates merge conflicts.

sys/sched.c
```diff
@@ -59,8 +59,11 @@ void sched_switch(thread_t *newtd) {
 
   if (newtd == NULL) {
     newtd = runq_choose(&runq);
-    if (newtd)
+    if (newtd) {
+      if(newtd->td_type == TD_USER && newtd->user_map)
+          set_active_vm_map(newtd->user_map);
```
cahirwpz: Switching address space should be done in [ctx_switch] (https://github.com/cahirwpz/mimiker/blob/master/mips/switch.S#L37). For reference look into FreeBSD version of [ctx_switch](http://bxr.su/FreeBSD/sys/mips/mips/swtch.S#354).

coodie: Why should it be done in ctx_switch?

cahirwpz: When we enter new context we want to see memory that is tied to this context.

coodie: Shouldn't we introduce process structure? This shouldn't be that hard. Besides I don't see how to change vm_map without calling vm_map_change in ctx_switch, which would require change from LEAF to NESTED, is this a problem, or you have better idea?

cahirwpz: > Shouldn't we introduce process structure? This shouldn't be that hard. 

I don't think it's necessary to introduce process structure. How our kernel would benefit from it right now? After all, each thread may own its private user address space. Note that L4 microkernels don't have a concept of process at all – only threads and address spaces.

> Besides I don't see how to change vm_map without calling set_active_vm_map in ctx_switch, which would require change from LEAF to NESTED, is this a problem, or you have better idea?

Calling `set_active_vm_map` (which is our version of `pmap_activate`) seems to be fine. Can't tell if change to `NESTED` is a problem in our case – that's something to be verified.

Let's return to my previous post for a while. For sure, after doing a context switch between threads working in kernel-space, the chosen thread has to be able to modify user-space memory (e.g. copy data to user space in order to return from `read` syscall).

cahirwpz:
Is there any progress here? If not, I'll fix the PR taking into account my comments.

rafalcieslak:
Please go ahead. @coodie was recently busy and asked me to finish this
branch, so you may do this just as well.

3 lis 2016 10:28 "Krystian Bacławski" <notifications@github.com> napisał(a):

Is there any progress here? If not, I'll fix the PR taking into account my
comments.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/cahirwpz/mimiker/pull/90#issuecomment-258095791>, or mute
the thread

PR #92, opened 2016-10-29
Fix for user threads never getting preempted

rafalcieslak:
The finalization for for `kernel_exc_enter` already checks whether the current threads `TD_NEEDSSWITCH`, and switches context if it does. This patch adds the same logic to `user_exc_enter`. Without this fix any thread running in user mode may be interrupted, but the scheduler never gets a chance to preempt it.

This is part of preparation for experimental usermode support I'm working on.

### DIFF
--- a/mips/exc.S
+++ b/mips/exc.S
@@ -86,8 +86,16 @@ user_exc_enter:
 
         # Interrupts may be enabled here and that's ok.
 
+        LOAD_PCPU($t0)
+        lw      $t0, PCPU_CURTHREAD($t0)
+        lw      $t1, TD_FLAGS($t0)
+        andi    $t1, TDF_NEEDSWITCH
+        beqz    $t1, 1f
+        move    $a0, $sp
+        jal     exc_before_leave
+
         # Disable interrupts till ERET.
-        di
+1:      di
 
 user_exc_leave:
         # Fetch the context from thread control block.


cahirwpz:
Please don't work on it. I've already started fixing many issues around exception processing. Unfortunately I lost access to zeus server and I cannot continue.

rafalcieslak:
That's a shame, because I've been also working on this since early morning. In this case I'll wait for your patches and compare them with mine. Would you mind creating a PR with them instead of committing directly to `master` so that we can discuss your changes?

Zeus is generally very unreliable. More often than not, I find myself being forced to use a local copy of OVPsim with my personal license, and I am glad I have it configured for local use.

cahirwpz:
Roughly speaking, I've managed to run syscalls from user-space. I only needed to fix a problem with kernel-space stack in `user_exc_enter`. I'm afraid I won't have the access to patches till Monday :(

rafalcieslak:
My use case (still experimenting with running multiple user threads in parallel) required more fixes. I'll list them briefly as these bugs may affect syscalls as well! So besides the patch in this PR:

1) Kernel `$sp` needs setting up when entering exception handler from user mode. There's a `TODO` [comment](https://github.com/cahirwpz/mimiker/blob/master/mips/exc.S#L73) on that.
2) [`ctx_switch` loads status register](https://github.com/cahirwpz/mimiker/blob/master/mips/switch.S#L55) using incorrect base address (is: `a0`, should be: `a1`).
3) Entire `sched_switch` procedure needs some sync, otherwise `runq` may become corrupted if an interrupt happens while choosing a thread to run. Simple `cs_enter/leave` seems to solve this problem.

I have a hunch that my branch will be complimentary to your work on syscalls, and it will be interesting to cross-test them against each other.

cahirwpz:
Not an issue anymore. Fixed in 4d79fce33d0e7780ad726a0490a9fd7262d90899.

PR #94, opened 2016-10-29
Fixed pmap_type/vm_map_type confusion

rafalcieslak:
As of now, `pmap_type_t` and `vm_map_type_t` are used interchangeably. It works, because these two enums are identical, but it is a confusing example of implicit type conversions - and `clang` is very annoyed by how we confuse these types.

This branch fixes type sanity, by refactoring `vm_map` interface so that it only uses `vm_map_type`, and fixing enum values used on calls to `vm_map` functions.

### DIFF
--- a/include/vm_map.h
+++ b/include/vm_map.h
@@ -9,7 +9,7 @@
  * That's because in current implementation page table is always located in
  * KSEG2, while user vm_map address range contains no KSEG2 */
 
-typedef enum { KERNEL_VM_MAP, USER_VM_MAP } vm_map_type_t;
+typedef enum { KERNEL_VM_MAP, USER_VM_MAP, VM_MAP_TYPE_LAST } vm_map_type_t;
 
 typedef struct vm_map_entry vm_map_entry_t;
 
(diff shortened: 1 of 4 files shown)


coodie:
I really like this change, because I'm often confused whether I should PMAP_* or VM_MAP_* macros, and yes - we should also use some extra tool to discover more bugs.

cahirwpz:
I wonder if we're going to need more than two active `vm_map`'s. I tried to come up with a scenario where three or more need to be active at once... and nothing comes to my mind.

Perhaps we should just use:
```c
vm_map_t *get_user_vm_map();
vm_map_t *get_kernel_vm_map();
```

rafalcieslak:
By following this logic, we should also split:
```c
static vm_map_t *active_vm_map[VM_MAP_LAST];
```
into:
```c
static vm_map_t *active_user_vm_map;
static vm_map_t *active_kernel_vm_map;
```

cahirwpz:
I don't think `active_user_vm_map` has to be global. If we introduced a pointer to `vm_map` in `thread_t` structure, then `active_user_vm_map` could be looked up by accessing `currthread` from `pcpu` structure.

On the other hand kernel `vm_map` could be allocated statically in `.data` section.

rafalcieslak:
For the record, getting current user `vm_map` from thread struct can't be reliably implemented before #90 is done.

cahirwpz:
Resolved with 756b358a58270714e8564f49c593e2408ee8e1a7.

PR #95, opened 2016-10-29
Serious typeo in sched_switch code (& and ==)

coodie:
Shouldn't this be "==" instead of "&" ?

By the way, since me and @rafalcieslak managed to find some bugs in @cahirwpz implementation, wouldn't it be good idea to make pull requests so everyone can do code review?

### DIFF
--- a/sys/sched.c
+++ b/sys/sched.c
@@ -54,7 +54,7 @@ void sched_switch(thread_t *newtd) {
 
   td->td_flags &= ~(TDF_SLICEEND | TDF_NEEDSWITCH);
 
-  if (td->td_state && TDS_RUNNING)
+  if (td->td_state == TDS_RUNNING)
     sched_add(td);
 
   if (newtd == NULL) {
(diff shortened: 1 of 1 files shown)


sys/sched.c
```diff
@@ -54,7 +54,7 @@ void sched_switch(thread_t *newtd) {
 
   td->td_flags &= ~(TDF_SLICEEND | TDF_NEEDSWITCH);
 
-  if (td->td_state & TDS_RUNNING)
+  if (td->td_state && TDS_RUNNING)
```
cahirwpz: equivalent to `(td->td_state != 0) && (TDS_RUNNING != 0)`

sys/sched.c
```diff
@@ -54,7 +54,7 @@ void sched_switch(thread_t *newtd) {
 
   td->td_flags &= ~(TDF_SLICEEND | TDF_NEEDSWITCH);
 
-  if (td->td_state & TDS_RUNNING)
+  if (td->td_state == TDS_RUNNING)
```
cahirwpz: Right! Nice find!

cahirwpz:
I'm a human and I make mistakes – I'll ask both of you for review when my changes are non-trivial.

rafalcieslak:
You just saved me a lot of debugging!

But I doubt I could spot this if I were to review this code.

cahirwpz:
*"Given enough eyeballs, all bugs are shallow."* – Linus Torvalds ;-)

PR #96, opened 2016-10-30
Unsigned variable is always bigger than zero!

coodie:
No need to explain what's going on here. This causes bus error if runq is empty (which happened in my case).

### DIFF
--- a/sys/runq.c
+++ b/sys/runq.c
@@ -16,7 +16,7 @@ void runq_add(runq_t *rq, thread_t *td) {
 }
 
 thread_t *runq_choose(runq_t *rq) {
-  for (unsigned i = RQ_NQS - 1; i >= 0; i--) {
+  for (int i = RQ_NQS - 1; i >= 0; i--) {
     struct rq_head *head = &rq->rq_queues[i];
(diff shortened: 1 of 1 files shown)


cahirwpz:
Another nasty bug identified! Thanks!

cahirwpz:
This made me think... shouldn't compiler warn about it?

rafalcieslak:
GCC would warn us if we enabled `-Wextra`, or at least `-Wsign-compare`.

cahirwpz:
Then I guess we should start using `-Wextra` flag immediately.

How difficult would it be to run clang frontend on kernel sources to catch extra warnings?

rafalcieslak:
Fairly straightforward, I believe.

```
$ clang -fsyntax-only -Wall -Wextra -Iinclude --std=gnu11 sys/vm_map.c
```

In my case this outputs some 8 warnings for this file. The only tricky parts are that it (unlike `mips-mti-elf-gcc`) is not aware of header files we pull from the toolchain (`mips/cpu.h`, in particular), and that it has some trouble expanding preprocessor macros from `include/elf/*.h`.

PR #98, opened 2016-10-31
Implement syscalls from user space. Fix switching routines.

cahirwpz:
@rafalcieslak @coodie Please review hard ;-)

BTW. `vm_map.elf` and `pmap.elf` tests in master branch are broken.

### DIFF
--- a/mips/exc.S
+++ b/mips/exc.S
@@ -10,6 +10,7 @@
 
         .globl exc_enter
         .globl kern_exc_leave
+        .globl user_exc_leave
         .local user_exc_enter
         .local kern_exc_enter
 
@@ -70,8 +71,18 @@ user_exc_enter:
         SAVE_REG($t2, BADVADDR, $k0)
         SAVE_REG($t3, PC, $k0)
 
-        # TODO: switch to kernel SP!
+        # Load kernel global pointer.
         la      $gp, _gp
+
+        # Load kernel stack pointer (always stack bottom ?)
+        LOAD_PCPU($t0)
+        lw      $t0, PCPU_CURTHREAD($t0)
+        lw      $sp, TD_KSTACK+STK_BASE($t0)
+        lw      $t1, TD_KSTACK+STK_SIZE($t0)
+        add     $sp, $t1
+
+        # No exeception frame so set td_frame to NULL.
+        sw      $0, TD_KFRAME($t0)
 
         # Turn off FPU, enter kernel mode,
         # drop exception level and disable interrupts.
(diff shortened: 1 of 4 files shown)


mips/exc.S
```diff
@@ -70,9 +71,19 @@ user_exc_enter:
...
 
+        # Load kernel stack pointer (always stack bottom ?)
+        LOAD_PCPU($t0)
+        lw      $t0, PCPU_CURTHREAD($t0)
+        lw      $sp, TD_KSTACK+STK_BASE($t0)
+        lw      $t1, TD_KSTACK+STK_SIZE($t0)
+        add     $sp, $t1
+
+        # No exeception frame so set td_frame to NULL.
+        sw      $0, TD_KFRAME($t0)
```
cahirwpz: Is `td_kframe` useful at all if `exc_frame_t` is passed to exception / interrupt handler directly ?

mips/exc.S
```diff
@@ -70,9 +71,19 @@ user_exc_enter:
         SAVE_REG($t2, BADVADDR, $k0)
         SAVE_REG($t3, PC, $k0)
 
-        # TODO: switch to kernel SP!
+        # Load kernel global pointer.
         la      $gp, _gp
 
+        # Load kernel stack pointer (always stack bottom ?)
```
cahirwpz: Is there any reason for it not to be stack bottom? Could not find one so far...

mips/exc.S
```diff
@@ -85,30 +96,39 @@ user_exc_enter:
...
 
-        # Update status register held in user context (interrupt mask).
-        LOAD_REG($t0, SR, $k0)
-        li      $t1, ~SR_IMASK
-        and     $t0, $t1
-        mfc0    $t1, C0_STATUS
-        andi    $t1, SR_IMASK
+        # Turn on FPU, enter exception level, enter user mode
+        # and enable interrupts (last two applies only after ERET).
+        # TODO: What about interrupt mask?
```
cahirwpz: Is there any reason for entering user-space without unmasking all interrupts?

syscall.c
```diff
@@ -1,8 +1,44 @@
...
+
+  vm_map_dump(umap);
+
+  uint32_t *code = (uint32_t *)stext;
+
+  *code++ = 0x24020002; /* li	$v0, 2 */
+  *code++ = 0x0000000c; /* syscall   */
+  *code++ = 0x1000ffff; /* b .       */
+
+  /* TODO: clear caches ? */
```
cahirwpz: Most likely enabling full cache simulation will break this test. Must verify that.

mips/exc.S
rafalcieslak: I've been thinking about this, and I believe that it is a safe assumption. Here's how it works:

We restore the kernel `$sp` only when entering interrupt handler while a user-mode code executes, which is only the case for user threads that are executing user code. Kernel threads don't need this, because their correct stack pointer is already stored in their saved context. User threads that are in progress of handling another interrupt also don't need this, for the same reason. But user threads normally work on their dedicated stack in user memory, so we need to temporarily move them to a safe stack somewhere in kernel memory. There are exactly two scenarios when this happens:

- Either this interrupt we are handling is **not** the first interrupt this user thread experienced. In that case, what could be left on the kernel stack? The last thing it was used for was processing that previous interrupt, but, apparently, that last time we have returned to user mode (via `user_exc_leave`), so the entire stack that was used for handling that interrupt has rolled back, and is now empty. Since no data is left on the stack, we can safely restore `$sp` to kernel stack bottom.

- The other case is when this is the very first interrupt this thread receives after switching to user-mode. In that case kernel-stack was most recently used to perform the initial switch to user mode, with `fork_trampoline` or similar procedure. However, `fork_trampoline` is not expected to return, so it's caller (and the caller's caller etc.) do not expect to ever continue execution. Therefore, in this case, even though the stack is non-empty and does contain stored variables, they are no longer relevant and can never be accessed anymore, so it is safe to throw them away and treat the kernel stack as empty. So again, we can start at stack bottom.

By induction, it is always safe to restore `$sp` to kernel stack bottom - when the interrupt is received in user mode.

Note that these observations are unaffected by diverging contexts: if a new kernel thread is created, it starts with an empty kernel stack; if a user thread `fork()s`, it begins user-mode execution with kernel stack being used for `fork_trampoline`.

Also, the proposed behavior is compatible with recursive interrupts. The first interrupt received may require switching from user stack to kernel stack, but if a subsequent interrupt is received before we are done handling this one, we are already in kernel-mode, so the `$sp` won't be moved to kernel stack (as it already points somewhere inside it), and will be correctly used to recursively handle the new interrupt.

Maybe it would be a good idea to explain that briefly in a comment in sources.

mips/exc.S
rafalcieslak: I can't find any, at least for now. If one day we'll decide to implement ultra-priority threads which should not be preempted by some insignificant interrupts in order to improve their performance, then maybe it will be worth to experiment with interrupt masks. But for now we'll be fine handling everything immediately.

rafalcieslak:
I have been testing this branch + `vm_space_switch` for a while, and I have been able to correctly switch between user programs now(!) - so this branch does indeed correctly fix the issues I experienced. Very nice! Green light from me.

PR #99, opened 2016-11-02
Travis

dyniec:


cahirwpz:
Looks ok!

Have you tried to run the kernel with `qemu` provided by Imagination Technologies ?

dyniec:
Yes. And it works! I changed launch script. UART over TCP works. Tests give similar (if not the same) results. I'm not sure about gdb. 
Only difference I found(and TLB size):
`Vectored interrupts implemented : no\n
UserLocal register implemented  : no`
I hope i can change this in qemu settings.

Idea for automated tests(for Travis) :
`nm.expect('Test passed.')` in lauch_qemu, but not every `.elf` finishes with this status.

cahirwpz:
Could you start working on `launch` script modifications to enable `qemu` as an alternative simulator? Extra `-S` / `--simulator` option would be ok.

cahirwpz:
I'll merge the branch as is. Please start working on `launch` script modifications in another PR.

PR #101, opened 2016-11-05
User mode programs

rafalcieslak:
This branch concludes changes to `do_exec()` that are required to correctly start multiple user-mode threads. It implements `ctx_init_usermode` which is semantically similar to `fork_trampoline`, gets `exec()` to store `vmap` to thread structure and switch to usermode when switching to loaded program, cleans up `exec()` from machine-specific code and provides a simple demonstration.

My recommended quick-test procedure (until we have a "print" syscall):
1) Load `exec.elf` in debugger
2) `add-symbol-file user/prog.uelf 0x00400000`
3) `continue` multiple times to step through all `main()`s
4) Interrupt the execution `^C`, print `textarea`, continue further.
5) Interrupt again. Unless you are unlucky, a different user thread is now active, and it wrote different values to `textarea`.
6) Repeat to observe threads getting switched.

### DIFF
--- a/mips/context.c
+++ b/mips/context.c
@@ -32,16 +32,10 @@ void ctx_init(thread_t *td, void (*target)()) {
 }
 
 void ctx_init_usermode(vm_addr_t entry_point, vm_addr_t stack_pointer) {
-
   thread_t *td = thread_self();
-  td->td_uctx.gp = 0;
+  bzero(&td->td_uctx, sizeof(exc_frame_t));
+  td->td_uctx.gp = 0; /* Explicit. */
   td->td_uctx.pc = entry_point;
   td->td_uctx.sp = stack_pointer;
-  td->td_uctx.ra = 0;
-  /* TODO: Is there any reason to clear other registers? */
-
-  /* This will apply context, enter user mode and re-enable interrupts. */
-  user_exc_leave();
-
-  __builtin_unreachable();
+  td->td_uctx.ra = 0; /* Explicit. */
 }
--- a/mips/context.c
+++ b/mips/context.c
@@ -31,11 +31,11 @@ void ctx_init(thread_t *td, void (*target)()) {
   kframe->sr = (reg_t)sr;
 }
 
-void ctx_init_usermode(vm_addr_t entry_point, vm_addr_t stack_pointer) {
-  thread_t *td = thread_self();
+void uctx_init(thread_t *td, vm_addr_t pc, vm_addr_t sp) {
   bzero(&td->td_uctx, sizeof(exc_frame_t));
+
   td->td_uctx.gp = 0; /* Explicit. */
-  td->td_uctx.pc = entry_point;
-  td->td_uctx.sp = stack_pointer;
+  td->td_uctx.pc = pc;
+  td->td_uctx.sp = sp;
   td->td_uctx.ra = 0; /* Explicit. */
 }
--- a/mips/context.c
+++ b/mips/context.c
@@ -30,3 +30,22 @@ void ctx_init(thread_t *td, void (*target)()) {
   kframe->sp = (reg_t)sp;
   kframe->sr = (reg_t)sr;
 }
+
+void ctx_init_usermode(vm_addr_t entry_point, vm_addr_t stack_pointer) {
+  /* We are starting machine-specific thread context setup and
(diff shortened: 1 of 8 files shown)


mips/context.c
```diff
@@ -30,3 +30,22 @@ void ctx_init(thread_t *td, void (*target)()) {
   kframe->sp = (reg_t)sp;
   kframe->sr = (reg_t)sr;
 }
+
+void ctx_init_usermode(vm_addr_t entry_point, vm_addr_t stack_pointer) {
+  /* We are starting machine-specific thread context setup and
+     switch. It would be very unfortunate if threads switched before
+     we are done. */
+  intr_disable();
```
cahirwpz: I don't think `intr_disable()` is necessary here. Would lack of it render code below incorrect ?

mips/context.c
```diff
@@ -30,3 +30,22 @@ void ctx_init(thread_t *td, void (*target)()) {
...
+
+  thread_t *td = thread_self();
+  td->td_uctx.gp = 0;
+  td->td_uctx.pc = entry_point;
+  td->td_uctx.sp = stack_pointer;
+  td->td_uctx.ra = 0;
+  /* TODO: Is there any reason to clear other registers? */
+
+  /* This will apply context, enter user mode and re-enable interrupts. */
+  user_exc_leave();
```
cahirwpz: Look at `ctx_init`. It's a simple function that sole responsibility is to set up a context. `ctx_init_usermode` should behave similarly.

sys/exec.c
```diff
@@ -86,13 +87,16 @@ int do_exec(const exec_args_t *args) {
...
 
-  /* TODO: Get current process description structure */
+  thread_t *td = thread_self();
 
-  /* The current vmap should be taken from the process description! */
-  vm_map_t *old_vmap = get_user_vm_map();
-  /* We may not destroy the current vm map, because exec can still
-   * fail, and in that case we must be able to return to the
-   * original address space
+  /* Please don't switch to other threads while we prepare the address
```
cahirwpz: Is that really a problem? After all, even if a thread has got interrupted, the control will return with correct user vm space.

rafalcieslak: It is. `ctx_switch` will restore the vm space that is stored in `td->td_uspace`, which is the old one, not the new one we are preparing.

mips/context.c
rafalcieslak: You are correct. I forgot that `user_exc_leave` disables interrupts at it's very beginning anyway.

mips/context.c
rafalcieslak: Sorry, but I don't see how is this relevant. `ctx_init` sets up a different context, which has different requirements. What exactly should I do in a fashion similar to `ctx_init`?

cahirwpz: Do not use`user_exc_leave` here :)

sys/exec.c
cahirwpz: Please have a look at #103, would that be a solution?

rafalcieslak: I think it does fix the problem, but since #103 entirely messes up my way of imagining the distinction between user and kernel threads, I can't clearly see if it does not have any fancy side-effects.

mips/context.c
rafalcieslak: And how do I do that? The purpose of `ctx_init_usermode` is to actually enter user context, without returning. To do that, we need to exit interrupt handling in a (possibly) different way than we entered, so merely changing stack frames won't do. I could manually `eret` with correct registers, but that's exactly what `user_exc_leave` does, so I reused it.

sys/exec.c
cahirwpz: In fact, there's no distinction between kernel and user threads. Both are the same – at any given moment a thread can execute either kernel-space or user-space code (provided that a thread can safely enter user-mode). That's all.

According to classic OS textbooks our threads are Kernel Level Threads in 1-to-1 model.

mips/context.c
cahirwpz: I just ask you to separate context preparation step from context switch step. There's no need to merge both steps, that's all.

cahirwpz:
If you finished working on that branch, I'd like to possibly make some changes (in your repository) before merging with `master`. Is that ok with you?

rafalcieslak:
Go ahead.

PR #104, opened 2016-11-05
Templates for get_elf_image and generic ./user Makefile

rafalcieslak:
These proposed changes simplify adding multiple user programs to `./user`, by implementing a generic `Makefile` and templates for selecting embedded elf image in `get_elf_image`.


cahirwpz:
LGTM!

PR #106, opened 2016-11-05
Added qemu support to lauch script

dyniec:
QEMU works most of the time. In `rtc.elf`even first printed time is correct(OVPsim).
But sometimes (30% of runs) mimiker stops/crashes before main. Even errors are non-deterministic.
Last printfed message is `[startup] subsystems initialized` or `[thread] Activating 'main' {0x802be030} thread!`. When run with debugger attached it stops just after `kmalloc_init(td_pool)` in `thread_init`.
QEMU likes to exit with different messages: one of them is
 `dma: command fc not supported`
`dma: command ff not supported`
`qemu: unsupported keyboard cmd=0x78`
`qemu: unsupported keyboard cmd=0x14`
`qemu: unsupported keyboard cmd=0x10`
`qemu: unsupported keyboard cmd=0x80`
and another one
`PFLASH: Possible BUG - Write block confirm`.
When started with debugger only second one appears.
I'm not sure if we should merge this, but I would be happy if somebody could check if this is either mimiker's or qemu's fault.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,13 @@ language: c
 compiler: gcc
 before_install:
  - sudo apt-get -qq update
- - sudo apt-get install -y qemu-system-mips ctags cscope wget
+ - sudo apt-get install -y qemu-system-mips ctags cscope wget python3-pip
+ - sudo pip3 install -I pexpect
  - TOOLCHAIN=`mktemp -d toolchain.XXXXXX`
  - cd ${TOOLCHAIN}
  - wget -c http://codescape-mips-sdk.imgtec.com/components/toolchain/2016.05-03/Codescape.GNU.Tools.Package.2016.05-03.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz
  - tar -xf Codescape.GNU.Tools.Package.2016.05-03.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz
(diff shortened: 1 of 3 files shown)


goniz:
I have a lot of experience running on qemu, and 100% of the times I've encountered these errors, my bugs were to blame.. 

Take those crashes as warnings for the debugging sessions a ahead of you.  :)

goniz:
You might also want to specify the cpu family in the qemu flags using -cpu

cahirwpz:
FYI `OVPsim` simulate 24Ke processor by default.

dyniec:
QEMU reports support for these families:
 `MIPS '4Kc'`
`MIPS '4Km'`
`MIPS '4KEcR1'`
`MIPS '4KEmR1'`
`MIPS '4KEc'`
`MIPS '4KEm'`
`MIPS '24Kc'`
`MIPS '24Kf'`
`MIPS '34Kf'`
`MIPS '74Kf'`
`MIPS 'M14K'`
`MIPS 'M14Kc'`
`MIPS 'P5600'`
`MIPS 'mips32r6-generic'`
Those don't even boot properly:
 `MIPS '4Kc'`
`MIPS '4Km'`
`MIPS '4KEcR1'`
`MIPS '4KEmR1'`
`MIPS '4KEm'`
`MIPS 'M14K'`
`MIPS 'mips32r6-generic'`
I hope any of them(maybe 24Kf) is similar to 24Ke.

cahirwpz:
Right! I meant 24Kf, which is 24Kc + FPU.

PR #107, opened 2016-11-05
Hello world

rafalcieslak:
This proposal implements the bare minimum required for user-space programs to output some text via UART, for debugging purposes. It does not incorporate an elegant system call handler table, nor universal procedures for copying system call arguments - which I believe should be done as a separate change.

A lot of changes I did here may be done in various ways (I tried to make this change minimal), so I'm looking for feedback.

I chose arbitrary high syscall numbers 1001 and 1002, as to not interfere with any reserved value.

A new demonstration program is included, which prints out a "Hello world" when started.

### DIFF
--- a/sys/sys_uartprint.c
+++ b/sys/sys_uartprint.c
@@ -5,3 +5,10 @@ void sys_uart_print_char(syscall_args_t *sa) {
   char c = sa->args[0];
   kprintf("%c", c);
 }
+
+void sys_uart_print_str(syscall_args_t *sa) {
+  /* TODO: Copy string from user space? Is that even necessary for
+     this case? */
+  char *c = (char *)sa->args[0];
+  kprintf("%s", c);
+}
--- a/sys/sys_uartprint.c
+++ b/sys/sys_uartprint.c
@@ -1,0 +1,7 @@
+#include <syscall.h>
+#include <stdc.h>
+
+void sys_uart_print_char(syscall_args_t *sa) {
+  char c = sa->args[0];
+  kprintf("%c", c);
+}
--- a/user/syscall_test.c
+++ b/user/syscall_test.c
@@ -1,0 +1,17 @@
+
+void syscall_uart_print_char(char c) {
+  asm volatile("move $a0, %0\n"
+               "move $v0, 1001\n"
+               "syscall"
+               :
+               : "r"(c));
+}
+
+int main(int argc, char **argv) {
+  const char *str = "Hello world from a user program!\n";
+  const char *c = str;
+  while (*c)
+    syscall_uart_print_char(*c);
+  while (1)
+    ;
+}
--- a/user/syscall_test.c
+++ b/user/syscall_test.c
@@ -1,17 +1,28 @@
 
 void syscall_uart_print_char(char c) {
   asm volatile("move $a0, %0\n"
-               "lui $v0, 1001\n"
+               "li $v0, 1001\n"
+               "syscall"
+               :
+               : "r"(c));
+}
+
+void syscall_uart_print_str(const char *c) {
+  asm volatile("move $a0, %0\n"
+               "li $v0, 1002\n"
                "syscall"
                :
                : "r"(c));
 }
 
 int main(int argc, char **argv) {
+  /* Char-by-char output */
   const char *str = "Hello world from a user program!\n";
   const char *c = str;
   while (*c)
-    syscall_uart_print_char(*c);
+    syscall_uart_print_char(*c++);
+  /* String output */
+  syscall_uart_print_str(str);
   while (1)
     ;
 }
--- a/user/syscall_test.c
+++ b/user/syscall_test.c
@@ -1,7 +1,7 @@
 
 void syscall_uart_print_char(char c) {
   asm volatile("move $a0, %0\n"
-               "move $v0, 1001\n"
+               "lui $v0, 1001\n"
                "syscall"
                :
                : "r"(c));
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 # vim: tabstop=8 shiftwidth=8 noexpandtab:
 
-TESTS = callout.elf malloc.elf physmem.elf pmap.elf rtc.elf sched.elf \
-	sleepq.elf syscall.elf thread.elf vm_map.elf exec.elf
+TESTS = callout.elf malloc.elf physmem.elf pmap.elf rtc.elf sched.elf	\
+	sleepq.elf syscall.elf thread.elf vm_map.elf exec.elf				\
+	exec_syscall.elf
 SOURCES_C = 
 SOURCES_ASM = 
 
@@ -12,7 +13,7 @@ include Makefile.common
 LDLIBS += -Lsys -Lmips -Lstdc \
 	  -Wl,--start-group -lsys -lmips -lstdc -lgcc -Wl,--end-group
 
-LD_EMBED = user/prog.uelf.o
+LD_EMBED = user/prog.uelf.o user/syscall_test.uelf.o
 
 # Kernel runtime files
 KRT = stdc mips sys user
@@ -28,6 +29,7 @@ physmem.elf: physmem.ko $(KRT)
 sched.elf: sched.ko $(KRT)
 sleepq.elf: sleepq.ko $(KRT)
 exec.elf: exec.ko $(KRT)
+exec_syscall.elf: exec_syscall.ko $(KRT)
 
 libkernel.a: $(DEPFILES) $(OBJECTS)
 
--- a/exec_syscall.c
+++ b/exec_syscall.c
@@ -4,8 +4,8 @@
 int main() {
   exec_args_t exec_args;
   exec_args.prog_name = "syscall_test";
-  exec_args.argv = (char *[]){};
-  exec_args.argc = 0;
+  exec_args.argv = (char *[]){"syscall_test"};
+  exec_args.argc = 1;
 
   do_exec(&exec_args);
 
--- a/exec_syscall.c
+++ b/exec_syscall.c
@@ -1,0 +1,13 @@
+#include <common.h>
+#include <exec.h>
+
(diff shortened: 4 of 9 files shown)


rafalcieslak:
@cahirwpz: I'm considering writing a simple syscall vector. Should I create a separate PR for it, or do you wish me to add it to this branch?

cahirwpz:
Please implement syscall vector in separate PR. Use `include/sysent.h` and `sys/sysent.c` files for that. As the toolchain from `toolchain/mips` contains `newlib` without syscall interface, I'd like you to investigate what is missing there and what needs to be implemented.

[syscall.h](https://sourceware.org/git/gitweb.cgi?p=newlib-cygwin.git;f=libgloss/syscall.h;hb=HEAD) from `newlib` lists small number of syscalls that are needed to use C standard library functionality. Please add dummy implementation for:
```c
#define SYS_exit        1
#define SYS_open        2
#define SYS_close       3
#define SYS_read        4
#define SYS_write       5
#define SYS_lseek       6
#define SYS_unlink      7
#define SYS_getpid      8
#define SYS_kill        9
#define SYS_fstat       10
```

sys/sys_uartprint.c
```diff
@@ -0,0 +1,14 @@
...
+void sys_uart_print_char(syscall_args_t *sa) {
+  char c = sa->args[0];
+  kprintf("%c", c);
+}
+
+void sys_uart_print_str(syscall_args_t *sa) {
+  /* TODO: Copy string from user space? Is that even necessary for
+     this case? */
+  char *c = (char *)sa->args[0];
+  kprintf("%s", c);
```
cahirwpz: Passing unbounded buffer to kernel is evil!

Please reimplement both syscalls in terms of SYS_write with file descriptor no. 1.

user/syscall_test.c
```diff
@@ -0,0 +1,28 @@
...
+}
+
+int main(int argc, char **argv) {
+  /* Char-by-char output */
+  const char *str = "Hello world from a user program!\n";
+  const char *c = str;
+  while (*c)
+    syscall_uart_print_char(*c++);
+  /* String output */
+  syscall_uart_print_str(str);
```
cahirwpz: If you pass here invalid argument (pointer to unmapped region) the thread will crash inside kernel, which makes me quite unhappy.

Perhaps it's time to focus on making the separation between kernel-mode and user-mode more effective? Or shall we ignore the issue for now?

cahirwpz:
Dummy syscalls implementation are given in `libnosys.a`:

```
$ cd ~/local/mips-unknown-elf/mips-unknown-elf/sysroot/lib
$ mips-unknown-elf-nm libnosys.a | grep T | cut -d ' ' -f 3 | sort
_chown
close
execve
_exit
fork
fstat
getpid
gettimeofday
isatty
kill
link
lseek
open
read
_readlink
sbrk
stat
_symlink
times
unlink
wait
write
```

rafalcieslak:
What's interesting about those stubs is that they store error code at `0(gp)`:

```
write.o:     file format elf32-littlemips


Disassembly of section .text:

00000000 <write>:
_DEFUN (_write, (file, ptr, len),
        int   file  _AND
        char *ptr   _AND
        int   len)
{
  errno = ENOSYS;
   0:	24020058 	li	v0,88
   4:	af820000 	sw	v0,0(gp)
  return -1;
}
   8:	03e00008 	jr	ra
   c:	2402ffff 	li	v0,-1
```

Is that a convention?

cahirwpz:
I guess the code above assumes `errno` to be placed in global data area. Excerpt from [MIPS Assembly Language Programmer’s Guide](http://www.cs.unibo.it/~solmi/teaching/arch_2002-2003/AssemblyLanguageProgDoc.pdf),  page 11-23:

> The global data area is part of the data segment of an executable program. It
contains short data items which can be addressed by the gp register relative
addressing mode. The global data area comprises all the sections with the
SHF_MIPS_GPREL attribute.

> The compilers generate short-form (one machine instruction) gp relative
addressing for all data items in any of these sections with the
SHF_MIPS_GPREL attribute. The compilers must generate two machine
instructions to load or store data items outside the global data area. A program
executes faster if more data items are placed in the global data area.

>The size of the global data area is limited by the addressing constraints on gp
relative addressing, namely plus or minus 32 kilobytes relative to gp. This
limits the size of the global data area to 64 kilobytes.

rafalcieslak:
Yes, naturally. But there is an issue I have been facing with the small-data section for a while: when anything (literally anything, can be as small as 4 bytes) is placed into that section, the linker gets annoyed, and outputs a message similar to this:

```
ld: small-data section exceeds 64KB; lower small-data size limit (see option -G)
test.o: In function `test_func':
test-file.c:28:(.text+0xc): relocation truncated to fit: R_MIPS_GPREL16 against `some_variable'
```

This message surprises me, because there is no way I exceed 64KB with a single int variable. But `ld` insists. I tried lowering the limit of what data is considered "small" by using `-G8` or `-G1`, but even then, `ld` still complains. The only way I managed to get this working is to use `-G0`, which effectively disables small data entirely, then it finally gets "under 64 KB".

This is the reason why makefiles in `./user` compile and link sources with `-G0`. So far it looked like a fair solution, but I've encountered another.

I was trying to somehow experiment with `newlib`. I successfully compiled it for `mipsel` with our custom toolchain, and I can link user programs against newlib, so some functions (say, `strlen`) work fine! However, there are various functions which use global variables - for example, `errno`. Naturally, `errno` lands in small-data, which triggers the same linker problem:

```
/mipsel-unknown-elf/lib/gcc/mipsel-unknown-elf/5.2.0/../../../../mipsel-unknown-elf/bin/ld: small-data section exceeds 64KB; lower small-data size limit (see option -G)
libnosys.a(execve.o): In function `execve':
/newlib-cygwin/build-mips/mipsel-unknown-elf/libgloss/libnosys/../../../../libgloss/libnosys/execve.c:21:(.text+0xc): relocation truncated to fit: R_MIPS_GPREL16 against `errno'
```

This completely stops me from using newlib. I see these options we have for solving this problem:

1) Compile newlib with `-G0`. This is not entirely straight-forward, because newlib's `configure` script does not expose an option to append arbitrary flags for the target compiler, but it's probably doable by editing newlib's makefiles. However, I am suspicious about this option. In the previous comment I've noticed that - at least - nosyslib assumes that `errno` is in small-data, as it is accessed by `0(gp)`. So if we force `errno` to `.data`, nosyslib will break. What about the rest of newlib? Maybe the assumption of `errno` being in `.sdata` is also used elsewhere? I'm not sure if that is the case, but if it is, then we can't just `-G0`.

2) Fix ld, convince it that `.sdata` used in input `.o` files is not that large (it isn't! simple `readelf` confirms that). Maybe the problem is with our custom linker scripts? There isn't much information about this MIPS-specific problem on the internet, but all mentions that I was able to find generally boil down to using `-G0`, so I wasn't able to find anything explaining why does this happen.

This is the n-th time I spent a lot of time investigating this issue, and I don't seem to be getting anywhere. I would really appreciate some assistance with this problem.

rafalcieslak:
Okay, this was a fun night. I think I have figured out what is going on.

**TL;DR: Minor, rare bug in GNU ld.bfd. Easy to avoid.**

So, being stubborn, I ended up investigating how does to linker process `.sdata` and related, and how exactly is `gp` set. Now this is generally confusing, because the source files of `ld.bfd` are difficult to read. So instead I looked into `ld.gold`, and learned a lot about `.sdata` by reading what gold does.

So here are some facts about `$gp` that I deduced from `binutils-2.25.1/gold/mips.cc` (sorry for no links, but I can't find online-browsable binutils sources):

- It is recommended to place `.sdata` immediately after `.got`, if the latter is present.
- `gp` should be 16-byte alligned.
- `gp` should point at a 15-bit offset from `.got` (or close to, to fulfill 16-byte alignment requirement).

Thus, by default, `gold` will place `gp` at `.got + ALIGN(16) + 0x7ff0`. The offset is chosen so that the the furthest negative relative reference from `gp` will be nearby `.got`. If `.got` is missing (as it will be, in our case), `gold` uses `.sdata` instead of `.got`.

Interestingly enough, a comment in `gold` sources states that **no** modern MIPS elf architectures use `gp` for addressing `.sdata`, only `.got`. Luckily they still support such architectures.

Now: if the linker script provides a `_gp` symbol, `gold` will use it's address for `gp`, regardless of where `.got` is placed.

So what is wrong with `ld.bdf`? It expects the `_gp` symbol to be present. And it is in the default builtin linker script. When that symbol is not present, `ld.bdf` does not emit any error message, instead it arbitrarily sets `gp` to `0x12345678` (`binutils-2.25.1/ld/elfxx-mips.c:12820`), and continues. Naturally, later on this causes some `.sdata` references that use `gp` fail, as they can't reach data using such pointless `gp`. Thus `ld` assumes that since data is too far from `gp`, then it must have exceeded 64KB limit - which explains the confusing message.

Apparently `ld` did not expect we would write a custom linker script. Also continuing with a random value in `gp` is clearly a bug - this is an uninitialized value, hard-coded only to suppress compiler warning.

So to solve this problem we should ~~switch to gold~~ define a hidden `_gp` symbol in the userspace linker script. I've tested that it does indeed solve the issue, but I still have to verify if the elf build this way is correct - also, now `exec` would have to set `gp` correctly, to match the `.sdata` section. I'll look into it and prepare a patch soon.

cahirwpz:
Nice findings! I really appreciate your perseverance in pinning down the culprit. Now it seems to be pretty obvious that `_gp` symbol was missing, but I guess at first glance it was deeply confusing. Unfortunately `ld.bfd` doesn't follow good engineering practice to fail quickly and not propagate error.

If the problem is really solved then I guess userspace C standard library is mostly a problem of the past. This brings us closer to a kernel we can show off, which is HUGE!

cahirwpz:
@rafalcieslak I guess we can close this PR. IIUC you'll work on the problem in #125 and consecutive PRs ?

PR #108, opened 2016-11-05
Initial gdb commands for thread debugging

coodie:
Since I'm stuck debugging turnstiles I decided to implement some debugging framework to help me with it. Toolchain gdb doesn't support python by default so I had to compile my own with python support. This is example on how to use python API to get some commands into gdb. I added just one which is 'kernel-threads' which prints runnable threads and currently running one. I'll find commands which track context_switches and new threads appearing in kernel useful and I'll definitely add them. They are mostly for my use, but someone might find them helpful.

### DIFF
--- a/threads-gdb.py
+++ b/threads-gdb.py
@@ -26,6 +26,7 @@ def current_thread():
 class KernelThreads (gdb.Command):
     def __init__(self):
         super(KernelThreads, self).__init__("kernel-threads", gdb.COMMAND_USER)
+
     def invoke(self, args, from_tty):
         argv = gdb.string_to_argv(args)
         if len(argv) != 0:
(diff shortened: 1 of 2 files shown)


rafalcieslak:
This looks very handy. Could you guide me how should I compile my gdb to test it? I assume the toolchain's version is not just a plain standard gdb? Do I somehow enable mips support?

coodie:
It actually is gdb compiled from sources, if you run it you will encounter following message:

```
This GDB was configured as "--host=x86_64-pc-linux-gnu --target=mips-mti-elf"
```

In general if you want to have gdb with python support you want to download the gdb sources and run command:

./configure --with-python --target=mips-mti-elf

If you want to have gdb store it's data (libraries, compiled binary) after make install in other place, you want to run it as

./configure --with-python --target=mips-mti-elf --prefix=/path/to/directory

cahirwpz:
This looks just awesome! I hope we'll create a nice set of GDB extensions that will make our debugging sessions much more productive. Please prepare a short speech about python gdb scripts for next seminar meeting.

W.r.t. python formatting... please use three tools to check your code `pep8`, `pylint` and `pyflakes`. All of them are available as a part of [python-mode](https://github.com/klen/python-mode) for Vim.

It'd be even better if you could configure git pre-commit hook to run those tools automatically.

threads-gdb.py
```diff
@@ -0,0 +1,110 @@
...
+            self.newthreadbp.delete()
+            self.newthreadbp = None
+        else:
+            print('new thread trace on')
+            self.newthreadbp = NewThreadTracerBP()
+
+
+KernelThreads()
+CtxSwitchTracer()
+NewThreadTracer()
```
coodie: Python interface supports command autocompletion. 

Please keep in mind that making really good interface might take some time, at this moment I'm focused on my needs (debugging the scheduler and turnstiles).

threads-gdb.py
```diff
@@ -0,0 +1,110 @@
...
+        to = gdb.parse_and_eval('(struct thread*)$a1').dereference()
+        print 'context switch from ', pretty_thread(frm), ' to ', pretty_thread(to)
+        return self.stop_on
+
+    def set_stop_on(self,arg):
+        self.stop_on = arg
+
+class CtxSwitchTracer (gdb.Command):
+    def __init__(self):
+        super(CtxSwitchTracer, self).__init__('cs-trace', gdb.COMMAND_USER)
```
coodie: Yes, this name might be confusing :).

Tracing threads (having function similar to info threads) is quite difficult, since we don't have 'global threads list' in system, This can be solved by adding breakpoint on every `thread_create` and `thread_delete` and keeping list of all threads in memory. I don't like this because this: makes code run under debugger slower, makes scripts harder to write because system has two states: gdb state and running state and we have to keep them somehow consistent. I'd rather if we had global queue for threads and every time we create or delete thread we add it to the queue.

cahirwpz:
More recent version of this PR is #124

PR #109, opened 2016-11-06
Implemented vm_map_findspace

rafalcieslak:
This branch introduces `vm_map_findspace`, which finds a free space in virtual memory large enough to fit requested length. It is roughly based on FreeBSD's `vm_map_findspace`. This branch also includes some formatting fixes unrelated to `findspace`, because *one of us* is not running `make format` before each commit. I've also included a pretty compelling test case which should be enough for now.

I need this function to correctly implement anonymous mappings for `mmap` syscall. And I need that syscall in order to implement `fork`, because our current "no-processes" thread model requires me to actually implement `clone`, with adjustable level of memory sharing - and traditionally it is the caller to clone who has to provide a new stack for child thread, so it will need `mmap` to create space for such stack. And since `mmap` is pretty easy to implement, I thought it would be a nice place to start.

### DIFF
--- a/sys/vm_map.c
+++ b/sys/vm_map.c
@@ -7,6 +7,7 @@
 #include <vm_pager.h>
 #include <vm_object.h>
 #include <vm_map.h>
+#include <errno.h>
 
 static vm_map_t kspace;
 
@@ -23,8 +24,12 @@ vm_map_t *vm_map_activate(vm_map_t *map) {
   return old;
 }
 
-vm_map_t *get_user_vm_map() { return thread_self()->td_uspace; }
-vm_map_t *get_kernel_vm_map() { return &kspace; }
+vm_map_t *get_user_vm_map() {
+  return thread_self()->td_uspace;
+}
+vm_map_t *get_kernel_vm_map() {
+  return &kspace;
+}
 
 static bool in_range(vm_map_t *map, vm_addr_t addr) {
   return map && (map->pmap->start <= addr && addr < map->pmap->end);
@@ -32,9 +37,9 @@ static bool in_range(vm_map_t *map, vm_addr_t addr) {
 
 vm_map_t *get_active_vm_map_by_addr(vm_addr_t addr) {
   if (in_range(get_user_vm_map(), addr))
-      return get_user_vm_map();
+    return get_user_vm_map();
   if (in_range(get_kernel_vm_map(), addr))
-      return get_kernel_vm_map();
+    return get_kernel_vm_map();
   return NULL;
 }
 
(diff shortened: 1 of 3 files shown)


sys/vm_map.c
```diff
@@ -132,10 +137,57 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
...
 }
 
+int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
+                     vm_addr_t /*out*/ *addr) {
+
+  assert(is_aligned(length, PAGESIZE));
+  /* Bounds check */
+  if (start < map->pmap->start)
+    start = map->pmap->start;
+  if (start + length >= map->pmap->end)
```
cahirwpz: Looks like this code will never allocate the last page.
Shouldn't you check for `(start + length > map->pmap->end)` here?

sys/vm_map.c
```diff
@@ -132,10 +137,57 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
...
+    if (next->start - it->end >= length) {
+      /* We will fit into this gap. */
+      *addr = it->end;
+      return 0;
+    }
+  }
+
+  /* Finally, check for free space after end. */
+  vm_map_entry_t *last = TAILQ_LAST(&map->list, vm_map_list);
+  if (map->pmap->end - last->end > length) {
```
cahirwpz: Again I can see a problem with the last page... `(map->pmap->end - last->end >= length)` ?

cahirwpz:
LGTM!

sys/vm_map.c
```diff
@@ -23,18 +24,22 @@ vm_map_t *vm_map_activate(vm_map_t *map) {
   return old;
 }
 
-vm_map_t *get_user_vm_map() { return thread_self()->td_uspace; }
-vm_map_t *get_kernel_vm_map() { return &kspace; }
+vm_map_t *get_user_vm_map() {
```
rafalcieslak: Just add a `.git/hooks/pre-commit` bash script. It will be run in repo root directory, so you can just call `make format`. Note however that hooks are not versioned, so they are **not** shared between repository clones.

I use another approach, as I like to manually review changes done by automatic formatting. I configured my editor to run the formatter each time before I save a file. I don't know how to do that with vi, but - for the record - this is how I did it with emacs:

```elisp
;; Call clang-format before saving C files
(add-hook 'c-mode-hook 
   (lambda () 
      (add-hook 'before-save-hook 'clang-format-buffer nil 'make-it-local)))
```

sys/vm_map.c
rafalcieslak: Correct. Fixed both with a new commit.

PR #110, opened 2016-11-06
Turnstile

coodie:
I finally got something that looks like it is working. I'm still unsure about some places but I need to make better tests and think of more places where context_switches might occur. This could possibly be merged, but the code isn't clean yet.

### DIFF
--- a/include/thread.h
+++ b/include/thread.h
@@ -24,7 +24,6 @@ typedef struct thread {
   enum { TDS_INACTIVE = 0x0, TDS_WAITING, TDS_READY, TDS_RUNNING } td_state;
   uint32_t td_flags;           /* TDF_* flags */
   volatile uint32_t td_csnest; /* critical section nest level */
-  vm_map_t *user_map;           /* User space memory map */
   /* thread context */
   exc_frame_t td_uctx;    /* user context (always exception) */
   fpu_ctx_t td_uctx_fpu;  /* user FPU context (always exception) */
--- a/include/thread.h
+++ b/include/thread.h
@@ -22,9 +22,9 @@ typedef enum
 } td_type_t;
 
 typedef struct thread {
-  TAILQ_ENTRY(thread) td_runq;   /* a link on run queue */
-  TAILQ_ENTRY(thread) td_sleepq; /* a link on sleep queue */
-  TAILQ_ENTRY(thread) td_lock;
+  TAILQ_ENTRY(thread) td_runq;    /* a link on run queue */
+  TAILQ_ENTRY(thread) td_sleepq;  /* a link on sleep queue */
+  TAILQ_ENTRY(thread) td_lock;    /* a link on turnstile */
   const char *td_name;
   /* thread state */
   enum { TDS_INACTIVE = 0x0, TDS_WAITING, TDS_READY, TDS_RUNNING } td_state;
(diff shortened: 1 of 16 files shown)


include/thread.h
```diff
@@ -15,13 +16,15 @@ typedef struct vm_map vm_map_t;
...
-  TAILQ_ENTRY(thread) td_sleepq; /* a link on sleep queue */
+  TAILQ_ENTRY(thread) td_runq;    /* a link on run queue */
+  TAILQ_ENTRY(thread) td_sleepq;  /* a link on sleep queue */
+  TAILQ_ENTRY(thread) td_lock;    /* a link on turnstile */
   const char *td_name;
   /* thread state */
   enum { TDS_INACTIVE = 0x0, TDS_WAITING, TDS_READY, TDS_RUNNING } td_state;
   uint32_t td_flags;           /* TDF_* flags */
   volatile uint32_t td_csnest; /* critical section nest level */
+  vm_map_t *user_map;           /* User space memory map */
```
rafalcieslak: There is already `td_uspace` for that.

cahirwpz:
This PR is huge! We'll have to split it into smaller ones.

Please create another PR with your attempt to fix synchronization issues around the scheduler.

mips/atomic.c
```diff
@@ -0,0 +1,56 @@
...
+#define __MIPS_PLATFORM_SYNC_NOPS ""
+static __inline void mips_sync(void) {
+  __asm __volatile(".set noreorder\n"
+                   "\tsync\n" __MIPS_PLATFORM_SYNC_NOPS ".set reorder\n"
+                   :
+                   :
+                   : "memory");
+}
+
+static __inline uint32_t atomic_cmpset_32(__volatile uint32_t *p,
```
cahirwpz: GCC gives a number of different synchronization primitives described [here](https://gcc.gnu.org/onlinedocs/gcc-5.4.0/gcc/_005f_005fsync-Builtins.html#g_t_005f_005fsync-Builtins). I've checked our new toolchain and indeed `__sync_bool_compare_and_swap` is defined in `libgcc.a`. The disassembly gives following code:

```
00000000 <__sync_bool_compare_and_swap_4>:
   0:   0000000f        sync
   4:   c0830000        ll      v1,0(a0)
   8:   14650006        bne     v1,a1,24 <__sync_bool_compare_and_swap_4+0x24>
   c:   24020000        li      v0,0
  10:   00c00821        move    at,a2
  14:   e0810000        sc      at,0(a0)
  18:   1020fffa        beqz    at,4 <__sync_bool_compare_and_swap_4+0x4>
  1c:   24020001        li      v0,1
  20:   0000000f        sync
  24:   03e00008        jr      ra
  28:   00000000        nop
```

coodie: I copied these from FreeBSD, thus I'm pretty sure they work. I think this is also quite educational to see how to implement these functions properly, so it's not bad idea to leave them, but this isn't a good argument to keep this ;).

Anyway I don't like using internal gcc functions too much because we can never know if they are compiled or not. What if we change toolchain again and in new implementation of toolchain there won't be these operations? That's another few hours (potentially) of wasted time trying to compile/link etc. get everything to work. Also I like to have my functions documented inside code and have possibility to see implementation, if I was new to the project and saw something like `__sync_bool_compare_and_swap` I'd probably waste few minutes trying to find something on stackoverflow, while having documentation in *.h or *.c file makes everything obvious instantly, and if someone is bad at writing documentation I have implementation inside and I can check what it does. 

Also, here is a reason why FreeBSD uses their own compare and swap, right?

PR #111, opened 2016-11-06
Remove mtx_t

coodie:
With upcoming turnstiles I find old mtx_t unnecessary. We should follow FreeBSD convention and make mtx_t good, old blocking mutex.

### DIFF
--- a/sys/malloc.c
+++ b/sys/malloc.c
@@ -9,20 +9,17 @@ extern uint8_t __ebss[];
 static struct {
   void *ptr; /* Pointer to the end of kernel's bss. */
   void *end; /* Limit for the end of kernel's bss. */
-  mtx_t lock;
   bool shutdown;
-} sbrk = {__ebss, __ebss + 512 * PAGESIZE, MTX_INITIALIZER, false};
+} sbrk = {__ebss, __ebss + 512 * PAGESIZE, false};
 
 void kernel_brk(void *addr) {
   if (sbrk.shutdown)
     panic("Trying to use kernel_brk after it's been shutdown!");
-  mtx_lock(sbrk.lock);
   void *ptr = sbrk.ptr;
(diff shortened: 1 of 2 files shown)


sys/malloc.c
```diff
@@ -9,43 +9,36 @@ extern uint8_t __ebss[];
...
   void *end; /* Limit for the end of kernel's bss. */
-  mtx_t lock;
   bool shutdown;
-} sbrk = {__ebss, __ebss + 512 * PAGESIZE, MTX_INITIALIZER, false};
+} sbrk = {__ebss, __ebss + 512 * PAGESIZE, false};
 
 void kernel_brk(void *addr) {
   if (sbrk.shutdown)
     panic("Trying to use kernel_brk after it's been shutdown!");
-  mtx_lock(sbrk.lock);
```
cahirwpz: I agree that current `mtx_lock` implementation should be removed. However leaving `sbrk` calls without synchronization seems to be light-minded. Is it possible to replace `mtx_*` with critical section primitive in this case?

coodie: Funny thing is we have way much bigger synchronisation problems like... scheduler runs without critical section which sometimes can mess up run queue (turnstile PR fixes that).

Look at this piece of code (startup.c):
```
pm_init();
intr_init();
```

All uses of kernel_sbrk and kernel_brk are done before intr_init with kernel_sbrk_shutdown being called in pm_init, so all of these functions are run in disable interrupts mode, so we can safely remove mtx_t in this case. Actually it's even funnier because this code was actually ENABLING interrupts (after unlocking) before we even initialized them.

cahirwpz: Right! The code implicitely assumes interrupts are not running. I'm not sure what to do with it. I'll just leave synchronization markers for the time being.

PR #112, opened 2016-11-07
Implemented mmap

rafalcieslak:
This is a very basic `mmap` implementation that only supports anonymous mappings. Generally, it is merely a fancy wrapping for `vm_map_findspace` and `vm_map_add_entry`. It is not yet hooked to a syscall handler, but eventually it will be useful for userspace programs.

Incidentally I discovered some minor bugs in `vm_map_findspace`, so this branch fixes them as well.


cahirwpz:
As far as I can tell `malloc` from `newlib` can be configured to use `mmap`. But since there's not a single `mmap` user for the time being let's postpone this commit. However fixes to `vm_map_findspace` are of high value right now, so move them to another PR, which will get merged very soon.

rafalcieslak:
I've cleaned up this branch and updated it to use new syscall interface.

cahirwpz:
I'll not accept `mmap` and `munmap` syscall interface until kernel-side interface is well tested.

Having kernel threads gracefully crash on bad address access and `copyin` and `copyout` routines in place, it's time to provide thorough testing for `vm_map` interface. With that being done and added implementation of signals we can return to `mmap` and family implementation.

rafalcieslak:
Do we really need signals before a simple `mmap`? I still believe `mmap` will be useful for `fork` - user programs typically use `mmap` to allocate the stack of a new thread when asking `fork` to share memory map between parent/child.

sys/sysent.c
```diff
@@ -42,6 +43,7 @@ int sys_exit(thread_t *td, syscall_args_t *args) {
 
 /* clang-format hates long arrays. */
 sysent_t sysent[] = {
-  {sys_nosys}, {sys_exit},  {sys_open},  {sys_close}, {sys_read},  {sys_write},
-  {sys_lseek}, {sys_nosys}, {sys_nosys}, {sys_nosys}, {sys_fstat}, {sys_sbrk},
+  {sys_nosys}, {sys_exit},  {sys_open},  {sys_close}, {sys_read},
```
cahirwpz: Please use `[SYS_MMAP] = {sys_mmap}` extended C syntax to avoid problems with entry renumbering.

sys/sys_mmap.c
```diff
@@ -0,0 +1,101 @@
...
+      vm_map_add_entry(vmap, addr, addr + length, VM_PROT_READ | VM_PROT_WRITE);
+    /* Assign a pager. */
+    entry->object = default_pager->pgr_alloc();
+
+  } /* Release vm_map rwlock. */
+
+  log("Created entry at %p, length: %zu", (void *)addr, length);
+
+  if (flags & MMAP_FLAG_ANONYMOUS) {
+    /* Calling bzero here defeats the purpose of paging on demand, but I do not
```
cahirwpz: Well... we should assume it zeros the pages. However that seems to be broken, please fix [default_pager_fault](http://mimiker.ii.uni.wroc.pl/source/xref/mimiker/sys/vm_pager.c#16). Call to  `vm_map_protect` won't be necessary then.

sys/sys_mmap.c
```diff
@@ -0,0 +1,101 @@
...
+    log("mmap called by a thread with no user space.");
+    /* I have no idea what would be a good error code for this case. */
+    *error = ENOENT;
+    return MMAP_FAILED;
+  }
+
+  {
+    rw_scoped_enter(&vmap->rwlock, RW_WRITER);
+
+    if (addr >= vmap->pmap->end) {
```
cahirwpz: I guess this and next check could be moved outside the scope.

sys/sys_mmap.c
```diff
@@ -0,0 +1,101 @@
...
+      log("Non-anonymous memory mappings are not yet implemented.");
+      *error = EINVAL;
+      return MMAP_FAILED;
+    }
+
+    length = roundup(length, PAGESIZE);
+
+    if (addr == 0x0) {
+      /* We will choose an appropriate place in vm map. */
+      if (vm_map_findspace_nolock(vmap, MMAP_LOW_ADDR, length, &addr) != 0) {
```
cahirwpz: Lines 56-59 are same as 69-73. Shouldn't you consider `addr != 0` case first?

sys/sys_mmap.c
```diff
@@ -0,0 +1,101 @@
...
+           space. */
+        if (vm_map_findspace_nolock(vmap, MMAP_LOW_ADDR, length, &addr) != 0) {
+          /* Still no memory found. */
+          *error = ENOMEM;
+          return MMAP_FAILED;
+        }
+      }
+    }
+
+    /* TODO: Check if the range of addresses we are about to map is not already
```
cahirwpz: Umm... what... why?

user/include/sys/mman.h
```diff
@@ -0,0 +1,12 @@
+#include <stdint.h>
+
+/* Newlib does not provide mmap prototype, so we need to use our own. */
+
+#define MMAP_FLAG_ANONYMOUS 1
```
cahirwpz: Please use same values as mentioned in [mmap(2)](http://netbsd.gw.com/cgi-bin/man-cgi?mmap), ie. `MAP_ANONYMOUS`, `PROT_WRITE` and so on.

sys/sys_mmap.c
```diff
@@ -0,0 +1,101 @@
...
+    return -error;
+  return result;
+}
+
+vm_addr_t do_mmap(vm_addr_t addr, size_t length, vm_prot_t prot, int flags,
+                  int *error) {
+  thread_t *td = thread_self();
+  vm_map_t *vmap = td->td_uspace;
+
+  if (!vmap) {
```
cahirwpz: This should be a fatal error. Translate to `assert` please.

rafalcieslak:
All done.

I've also realized that with VFS interface in place, it will be pretty simple to implement support for non-anonymous (file-backed) mappings. All it would take is adding a new pager which keeps a reference to an open `file_t`.

sys/sys_mmap.c
```diff
@@ -0,0 +1,79 @@
...
+      /* No memory was found following the hint. Search again entire address
+         space. */
+      if (vm_map_findspace_nolock(vmap, MMAP_LOW_ADDR, length, &addr) != 0) {
+        /* Still no memory found. */
+        *error = ENOMEM;
+        return MMAP_FAILED;
+      }
+    }
+
+    /* Create new vm map entry for this allocation. Temporarily use permissive
```
cahirwpz: I think this comment is outdated.

sys/sys_mmap.c
```diff
@@ -0,0 +1,79 @@
...
+  if (!flags & MMAP_FLAG_ANONYMOUS) {
+    log("Non-anonymous memory mappings are not yet implemented.");
+    *error = EINVAL;
+    return MMAP_FAILED;
+  }
+
+  {
+    rw_scoped_enter(&vmap->rwlock, RW_WRITER);
+
+    length = roundup(length, PAGESIZE);
```
cahirwpz: Only acquire lock for actions that require it. Please move lines 47-54 outside the scope.

sys/Makefile
```diff
@@ -28,6 +28,7 @@ SOURCES_C  = \
 	startup.c \
 	sync.c \
 	sysent.c \
+	sys_mmap.c \
```
cahirwpz: If `sys_*.c` files are meant to contain syscall implementations I'd like this idea to be applied consistently throught code base.

rafalcieslak:
The Travis test issue is a false positive, and I propose a fix for it in #250.

PR #113, opened 2016-11-07
Fix for vm_map_findspace start strictness

rafalcieslak:
Extracted from #112 per [request](https://github.com/cahirwpz/mimiker/pull/112#issuecomment-258784549). `vm_map_findspace` would sometimes return an address lower than `start`, this patch fixes it to never consider spaces that would begin before `start`.

### DIFF
--- a/include/vm_map.h
+++ b/include/vm_map.h
@@ -47,10 +47,6 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
 vm_map_entry_t *vm_map_add_entry(vm_map_t *map, vm_addr_t start, vm_addr_t end,
                                  vm_prot_t prot);
 
-/* Searches for a space in memory map where length would fit, starting search
-   from start address. On success, returns 0 and sets *addr. If set, *addr is
-   guaranteed to be not smaller than start. Start and length arguments must be
-   page-aligned. */
 int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
                      vm_addr_t /*out*/ *addr);
 
--- a/include/vm_map.h
+++ b/include/vm_map.h
@@ -47,6 +47,10 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
 vm_map_entry_t *vm_map_add_entry(vm_map_t *map, vm_addr_t start, vm_addr_t end,
                                  vm_prot_t prot);
 
+/* Searches the provided memory map for an empty space of at least the given
+   length. It only considers areas that begin at start address, thus addr is
+   guaranteed to be not smaller than start. On success, returns 0 and sets
+   *addr. Requires space and length to be page-aligned. */
 int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
                      vm_addr_t /*out*/ *addr);
 
--- a/include/vm_map.h
+++ b/include/vm_map.h
@@ -47,6 +47,9 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
 vm_map_entry_t *vm_map_add_entry(vm_map_t *map, vm_addr_t start, vm_addr_t end,
                                  vm_prot_t prot);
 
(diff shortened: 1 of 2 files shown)


sys/vm_map.c
```diff
@@ -166,16 +166,23 @@ int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
     vm_map_entry_t *next = TAILQ_NEXT(it, map_list);
     if (!next)
       continue;
-    if (next->start - it->end >= length) {
-      /* We will fit into this gap. */
+    if (start < it->end && next->start - it->end >= length) {
```
cahirwpz: Can't you simplify both cases to a single expression: `(start >= it->end) && (start + length <= next->start)`?

sys/vm_map.c
```diff
@@ -139,7 +139,7 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
 
 int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
```
cahirwpz: Could you add a comment that would describe in a few words what that function is supposed to do? Is `start` argument only a hint?

sys/vm_map.c
```diff
@@ -166,16 +166,23 @@ int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
...
+      /* Start points to inside this gap, and we'll fit there. */
+      *addr = start;
+      return 0;
     }
   }
 
   /* Finally, check for free space after end. */
   vm_map_entry_t *last = TAILQ_LAST(&map->list, vm_map_list);
-  if (map->pmap->end - last->end >= length) {
+  if (start > last->end && map->pmap->end - start >= length) {
```
rafalcieslak: I'm not sure I follow. The condition you propose does not consider the situation where `start` points inside the last mapped region and there is enough space after it to fit `length`.

cahirwpz: Ok. That changes a lot :)

include/vm_map.h
```diff
@@ -47,6 +47,10 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
 vm_map_entry_t *vm_map_add_entry(vm_map_t *map, vm_addr_t start, vm_addr_t end,
                                  vm_prot_t prot);
 
+/* Searches the provided memory map for an empty space of at least the given
+   length. It only considers areas that begin at start address, thus addr is
```
cahirwpz: Statement "at start address" is actually misleading. I think an explanation you give in another comment is much clearer.

cahirwpz:
Please review my changes. I think it's much more streamlined than your implementation and easier to track. When you write complex logic please always attempt to reduce number of special cases under consideration.

I moved the comment close to the implementation. The idea is that you keep all information required to understand the logic of a procedure in one place.

rafalcieslak:
Originally I have considered using something similar to `next ? next->start : map->pmap->end;` to unify these cases, but I was worried you would consider it a *trick*, so I implemented another variant which was slightly more complex, but also more explicit. I always find it difficult to find the right spot between *hacks* and simplifying code, that will be to your liking.

I really dislike not having interface documented in header files. Genrerally I agree that the information required to understand the implementation should be nearby that implementation, but for functions that are meant to be used externally (say, in other TUs), the function specification must be present in the header file. This way 
- a developer who wishes to use the function only has to read the header
- in order to learn about an interface one only has to browse through functions available in the header file
- various tools can automatically pull the docs from included header files

The way I see it, details about how implementation works should be placed in/near the implementation, but function specification, including argument constraints and semantics, must be available in the header.

cahirwpz:
I always prefer shorter and simpler implementations. I believe you have enough experience and coding culture not to worry about my bias.

As you mention comments are kinda anti-pattern and I agree – it's much better to spend time making the procedure interface easy to use rather than document all quirks. Comments explaining implementation details should be scarce and close to the implementation.

If you see no bugs in the code, I'm ready to merge it.

rafalcieslak:
I'm positive this branch is ready to hit `master`.

PR #117, opened 2016-11-07
New toolchain support in make rules

rafalcieslak:
This tiny change to `Makefile.common` enables the custom toolchain - **if available** at host. Since this would be a major breaking change for other contributors (and for travis), for some transitional period both toolchains are supported. Make simply checks whether `mipsel-unknown-elf-gcc` is available, and if not, it falls back to `mips-mti-elf-*`. In both cases compilation succeeds, and the produced kernel image operates normally.

Also I had to disable `-Wformat`. gcc 5.2 is very confused about one particular detail that appears in numerous places in our sources, namely printing a `vm_addr_t` (or similar address types) with a `%lx`. `vm_addr_t` is AKA `unsigned int`, so gcc complains that we should not use the `l` prefix for printing. I might go ahead and correct all `%lx` to `%x`, but since they are there *for some reason*, I assume we want such address types to be realized as `long`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ SOURCES_ASM =
 all: tags cscope mips stdc sys $(TESTS)
 
 include Makefile.common
+$(info Using CC: $(CC))
 
 LDLIBS += -Lsys -Lmips -Lstdc \
 	  -Wl,--start-group -lsys -lmips -lstdc -lgcc -Wl,--end-group
--- a/Makefile.common
+++ b/Makefile.common
@@ -29,7 +29,7 @@ SYSROOT  = $(realpath $(dir $(shell which $(P)gcc))/..)
 GENASSYM = $(TOPDIR)/script/genassym.py $(NM)
 
 ASFLAGS  =
-CFLAGS   = -std=gnu11 -Og -Wall -Werror -Wno-format -fno-builtin -ffreestanding
+CFLAGS   = -std=gnu11 -Og -Wall -Werror -fno-builtin -ffreestanding
 CPPFLAGS = -Wall -Werror -DDEBUG -I$(TOPDIR)/include
 LDLIBS   =
 LDFLAGS  = -nostdlib -T malta.ld
--- a/Makefile.common
+++ b/Makefile.common
@@ -6,7 +6,16 @@ SOURCES = $(SOURCES_C) $(SOURCES_ASM)
 OBJECTS = $(SOURCES_C:.c=.o) $(SOURCES_ASM:.S=.o)
 DEPFILES = $(SOURCES_C:%.c=.%.D) $(SOURCES_ASM:%.S=.%.D)
 
-P        = mips-mti-elf-
+TRIPLET_CUSTOM  = mipsel-unknown-elf-
+TRIPLET_IMGTECH = mips-mti-elf-
+
+CUSTOM_TOOLCHAIN_FOUND = $(shell which $(TRIPLET_CUSTOM)gcc > /dev/null; echo $$?)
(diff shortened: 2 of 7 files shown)


cahirwpz:
I guess `vm_addr_t` should be defined as `intptr_t`.

rafalcieslak:
We define `vm_addr_t` as `uintptr_t`, in `common.h`. That, in turn, becomes `unsigned int`. What I end up with is a ton of:
```
/include/common.h:81:13: error: format '%lx' expects argument of type 'long unsigned int', but argument 4 has type 'vm_addr_t {aka unsigned int}' [-Werror=format=]
     kprintf("[panic] %s:%d " FMT "\n", __FILE__, __LINE__, ##__VA_ARGS__);     \
             ^
pmap.c:359:5: note: in expansion of macro 'panic'
     panic("No virtual address space defined for %08lx!", vaddr);
     ^
```

cahirwpz:
If you defined `vm_addr_t` as `unsigned long` would that solve the problem?

rafalcieslak:
A quick test confirms that `sizeof(unsigned long) == sizeof(void*)`. This makes sense, `unsigned long` is guaranteed to be at least 32 bits in size, while `unsigned int` may use just 16. Indeed defining `vm_addr_t` as `unsigned long` fixes the issue, do you want me to add this to this branch?

cahirwpz:
Yes, please do so... and reenable -Wformat.

rafalcieslak:
Done. By the way, is it not a problem that now `size_t` is `unsigned int` and not `unsigned long`? This is starting to confuse me.

cahirwpz:
Grepping through toolchain files I could only find `size_t` defined as `unsigned long` and that's exactly what I'd expect on 32-bit architecture.

pdziepak:
The standard way to print types like `uintptr_t` and `[u]int{8, 16, 32, 64}_t` family is to use macros from `inttypes.h`. However, `size_t`, `intmax_t` and `ptrdiff_t` have special `z`, `j` and `t` length modifiers. For example:
```c
uintptr_t pointer;
size_t size;
printf("This is my uintptr_t = %" PRIuPTR, value);
printf("...and this is my size_t = %zu, value);
```
This is the most future-proof and machine independent way to deal with format strings. For custom types like `vm_addr_t` it is probably a good idea to define their own macros.

BTW Nice thing about `unsigned long` that it can be `size_t` and `uintptr_t` for both all 32-bit architectures and LP64 ones (which is almost everything except Windows).

rafalcieslak:
But when I do:
```
$ mipsel-unknown-elf-gcc -mips32r2 -EL -g -std=gnu11 -Og -Wall -Werror -fno-builtin -ffreestanding -Wall -Werror -DDEBUG -I/mimiker/include  exec.c -E | grep size_t -C1
# 216 "~/local/mipsel-unknown-elf/lib/gcc/mipsel-unknown-elf/5.2.0/include/stddef.h" 3 4
typedef unsigned int size_t;
[...]
```
So I look around that `stddef.h` file:
```
#ifndef __SIZE_TYPE__
#define __SIZE_TYPE__ long unsigned int
#endif
#if !(defined (__GNUG__) && defined (size_t))
typedef __SIZE_TYPE__ size_t; // <- this is line 216
```
and `__SIZE_TYPE__` does not appear anywhere else in the file (which does not `#include` any other files). So if it is not using `long unsigned int`, then `__SIZE_TYPE__` must be a preprocessor built-in.

[And apparently it is](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html). 

How do we influence that?

cahirwpz:
Ignore my previous comment. Just use `%zu` as @pdziepak suggested.

PR #119, opened 2016-11-08
Basic atomic operations (compare and swap, atomic store)

coodie:
This PR adds atomic operations stolen from FreeBSD. They are used to implement mutexes on other branch, thus tested and proved to be working.

### DIFF
--- a/include/atomic.h
+++ b/include/atomic.h
@@ -1,0 +1,25 @@
+#ifndef __ATOMIC_H__
+#define __ATOMIC_H__
+
+#include <stdc.h>
+
+/*
+ * Atomically compare the value stored at *p with cmpval and if the
+ * two values are equal, update the value of *p with newval. Returns
+ * zero if the compare failed, nonzero otherwise.
+ * Pseudocode:
+ * if *p == cmpval:
+ *  *p = newval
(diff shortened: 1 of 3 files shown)


pdziepak:
Generally, it is better to prefer standard solutions to anything else and if they are not available then compiler builtins are always better than custom implementation. The reason for that is that if the compiler exactly knows what the function is doing it can better optimise it.

In this case there are obviously C11 atomics which can be implemented using GCC builtins [1]. C11 atomics have the additonal advantage of being more flexible in terms of memory ordering specification that most custom implementations (including this one).

Another thing to consider is that kernels without multicore support should very rarely need atomic operations (actually, I am very tempted to say "should never need them", but surely someone will come up with some obscure counterexample ;). Atomics are usually expensive even if there is no contention (this obviously can vary depending on the architecture/cpu model – worth checking the actual values for MIPS) and single core kernel can do all synchronisation by disabling interrupts or just disabling preemption (often cheaper than disabling interrupts).

[1] https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#g_t_005f_005fatomic-Builtins

PR #122, opened 2016-11-11
Add blocking mutexes

coodie:
This PR adds mutexes which make use of turnstile data structure.

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,3 +1,6 @@
 target remote localhost:1234
 break main
 continue
+
+source debug/ktrace.py
+source debug/kdump.py
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,6 +1,8 @@
+set tcp connect-timeout 30
 target remote localhost:1234
 break main
 continue
 
 source debug/ktrace.py
 source debug/kdump.py
+source debug/mutex.py
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ SOURCES_ASM =
 all: tags cscope mips stdc sys $(TESTS)
 
 include Makefile.common
+$(info Using CC: $(CC))
 
 LDLIBS += -Lsys -Lmips -Lstdc \
 	  -Wl,--start-group -lsys -lmips -lstdc -lgcc -Wl,--end-group
--- a/Makefile
+++ b/Makefile
@@ -39,17 +39,9 @@ tags:
 	find -iname '*.[ch]' -not -path "*/toolchain/*" | ctags --language-force=c -L- -e -f etags
 	find -iname '*.S' -not -path "*/toolchain/*" | ctags -a --language-force=asm -L-
 	find -iname '*.S' -not -path "*/toolchain/*" | ctags -a --language-force=asm -L- -e -f etags
-	find $(SYSROOT)/mips-mti-elf/include -type f -iname 'mips*' \
-		| ctags -a --language-force=c -L-
-	find $(SYSROOT)/mips-mti-elf/include -type f -iname 'mips*' \
-		| ctags -a --language-force=c -L- -e -f etags
-	find $(SYSROOT)/lib/gcc/mips-mti-elf/*/include -type f -iname '*.h' \
-		| ctags -a --language-force=c -L-
-	find $(SYSROOT)/lib/gcc/mips-mti-elf/*/include -type f -iname '*.h' \
-		| ctags -a --language-force=c -L- -e -f etags
 
-FORMATTABLE_EXCLUDE = include/elf stdc/smallclib include/mips/asm.h
-FORMATTABLE = $(shell find -type f -name '*.c' -or -name '*.h' | grep -v $(FORMATTABLE_EXCLUDE:%=-e %))
+FORMATTABLE_EXCLUDE = include/elf stdc/smallclib include/mips/asm.h include/mips/m32c0.h
+FORMATTABLE = $(shell find -type f -not -path "*/toolchain/*" -and \( -name '*.c' -or -name '*.h' \) | grep -v $(FORMATTABLE_EXCLUDE:%=-e %))
 format:
 	@echo "Formatting files: $(FORMATTABLE:./%=%)"
 	clang-format -style=file -i $(FORMATTABLE)
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # vim: tabstop=8 shiftwidth=8 noexpandtab:
 
-TESTS = callout.elf malloc.elf physmem.elf pmap.elf rtc.elf sched.elf \
-	sleepq.elf syscall.elf thread.elf vm_map.elf exec.elf mutex.elf
+TESTS = callout.elf malloc.elf physmem.elf pmap.elf rtc.elf sched.elf		\
+	sleepq.elf syscall.elf thread.elf vm_map.elf exec.elf exec_syscall.elf mutex.elf
 SOURCES_C = 
 SOURCES_ASM = 
 
@@ -13,7 +13,7 @@ $(info Using CC: $(CC))
 LDLIBS += -Lsys -Lmips -Lstdc \
 	  -Wl,--start-group -lsys -lmips -lstdc -lgcc -Wl,--end-group
 
-LD_EMBED = user/prog.uelf.o
+LD_EMBED = user/prog.uelf.o user/syscall_test.uelf.o
 
 # Kernel runtime files
 KRT = stdc mips sys user
@@ -29,6 +29,7 @@ physmem.elf: physmem.ko $(KRT)
 sched.elf: sched.ko $(KRT)
 sleepq.elf: sleepq.ko $(KRT)
 exec.elf: exec.ko $(KRT)
+exec_syscall.elf: exec_syscall.ko $(KRT)
 mutex.elf: mutex.ko $(KRT)
 
 libkernel.a: $(DEPFILES) $(OBJECTS)
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # vim: tabstop=8 shiftwidth=8 noexpandtab:
 
 TESTS = callout.elf malloc.elf physmem.elf pmap.elf rtc.elf sched.elf \
-	sleepq.elf syscall.elf thread.elf vm_map.elf exec.elf mutex.elf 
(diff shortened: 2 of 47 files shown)


sys/sched.c
```diff
@@ -13,7 +13,7 @@ static thread_t *idle_thread;
 static runq_t runq;
 static bool sched_active = false;
 
-#define SLICE 10
+#define SLICE 1
```
coodie: I decided to change slice to minimum value because I want the kernel to make as much context switches as possible at this moment, so there is bigger chance for race condition, while testing I couldn't get the race condition using slice of size 10, but it was very easy to get the race condition using slice of size 1. In future when this implementation looks OK, it's good idea to just change it to longer slice,

Makefile
```diff
@@ -29,6 +29,8 @@ physmem.elf: physmem.ko $(KRT)
 sched.elf: sched.ko $(KRT)
 sleepq.elf: sleepq.ko $(KRT)
 exec.elf: exec.ko $(KRT)
+turnstile.elf: turnstile.ko $(KRT)
+mutex.elf: mutex.ko $(KRT)
```
rafalcieslak: It looks like you're missing `mutex.c` file. Did you forget to `git add` it?

sys/mutex.c
```diff
@@ -0,0 +1,30 @@
...
+}
+
+void mtx_init(mtx_t *mtx) {
+  mtx->mtx_state = MTX_UNOWNED;
+  turnstile_init(&mtx->turnstile);
+}
+
+void mtx_lock(mtx_t *mtx) {
+  assert(!mtx_owned(mtx)); // No recursive mutexes for now
+  while (!mtx_try_to_lock(mtx)) {
```
cahirwpz: Imagine the code gets preempted between line 22 and 23. What will happen? Can line 28 be executed by other thread in the meantime ?

coodie: I think you are right. There is possibility that one thread will wait forever.

sys/mutex.c
```diff
@@ -0,0 +1,30 @@
...
+    turnstile_wait(&mtx->turnstile);
+  }
+}
+
+void mtx_unlock(mtx_t *mtx) {
+  atomic_store(&mtx->mtx_state, MTX_UNOWNED);
```
cahirwpz: Can you elaborate on that ?

sys/mutex.c
```diff
@@ -0,0 +1,44 @@
...
+
+  while (!mtx_try_lock(mtx)) {
+    cs_enter();
+    /* Check if the mutex got unlocked since a call to mtx_try_lock */
+    if (mtx->mtx_state == MTX_UNOWNED) {
+      cs_leave();
+      continue;
+    }
+    assert(mtx_owned(mtx));
+    turnstile_wait(&mtx->turnstile);
```
cahirwpz: In the implementation of [turnstile_wait]( https://github.com/cahirwpz/mimiker/blob/master/sys/turnstile.c#L10), `cs_enter` happens before `sched_yield`. Isn't that creating a window of opportunity for race condition?

sys/mutex.c
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+void mtx_unlock(mtx_t *mtx) {
```
cahirwpz: I think we need `assert(mtx_owner(mtx) == thread_self())` here.

sys/mutex.c
coodie: I have carefully examined that, and I think it does not. After `cs_leave` and before `sched_yield` if there is unvoluntary sched_yield() the thread won't get added to run queue, but will wait to get signaled. The funny part is that it will yield again after that and that is actually pretty pointless yield, I think we need scheduler function which yields AND leaves critical section.

sys/mutex.c
coodie: That's actually a good point. FreeBSD actually uses atomic compare_and swap where compared value is currently running thread, thus only owner of mutex may unlock it.

sys/mutex.c
```diff
@@ -0,0 +1,44 @@
...
+  atomic_store(&mtx->mtx_state, MTX_UNOWNED);
```
coodie: I have also thought about this, and you are right about violating some mutex queue ordering, however I was focused on getting rid of potential race conditions. Apparently I won't be able to push 'just quite correct mutexes' to the master repo :).

cahirwpz: A pair of `cs_enter` and `cs_leave` should solve the problem in this case, I suppose.

sys/mutex.c
cahirwpz: Well... it's a very good question if we should allow to call `sched_switch` while `thread_self()->td_csnest > 0`. Time to take the dog on a evening walk - I'll think of it.

cahirwpz: It's ok to call `sched_switch` within critical section. `ctx_switch` will properly save and restore `SR_IE` bit.

include/turnstile.h
```diff
@@ -10,8 +10,7 @@ void turnstile_init(turnstile_t *);
 
 /* This puts currently running thread on the turnstile,
  * puts it in waiting state and yields. While waiting on turnstile
- * thread cannot be run. This function has to be called with critical section,
- * and exits critical section upon exection. */
+ * thread cannot be run. This function is done under critical section. */
```
coodie: @cahirwpz 
Why did you change the function invariant? You broke the code!

include/turnstile.h
```diff
@@ -10,7 +10,8 @@ void turnstile_init(turnstile_t *);
 
 /* This puts currently running thread on the turnstile,
  * puts it in waiting state and yields. While waiting on turnstile
- * thread cannot be run. This function is done under critical section. */
+ * thread cannot be run. This function has to be called with critical section,
```
cahirwpz: Please don't change the semantics of this function. This is a bad practice that violates [Principle of Least Astonishment](http://www.catb.org/esr/writings/taoup/html/ch01s06.html#id2878339):

> The flip side of the Rule of Least Surprise is to avoid making things superficially similar but really a little bit different. This is extremely treacherous because the seeming familiarity raises false expectations. It's often better to make things distinctly different than to make them almost the same.

A user expects `turnstile_*` functions to be synchronized internally. But then... you make an exception for `turnstile_wait`, instead of just fixing the caller. If the user forgets to call `cs_enter` before, all hell will break loose without an apparent reason.

BTW: I hoped you'll ask about it instead of just reverting back to the original proposal.

sys/turnstile.c
```diff
@@ -8,12 +8,11 @@ void turnstile_init(turnstile_t *ts) {
 }
 
 void turnstile_wait(turnstile_t *ts) {
-  cs_enter();
```
cahirwpz: Please restore this line and call `cs_leave` after call to `turnstile_wait`.

debug/mutex.py
```diff
@@ -0,0 +1,57 @@
...
+def get_threads_blocked_on_mutex(mtx):
+    ts = mtx['turnstile']
+    return map(lambda t: t['td_name'].string(), tailq.collect_values(ts['td_queue'], 'td_lock'))
+
+
+def pretty_list_of_string(lst):
+    res = "["
+    for i in range(0, len(lst)):
+        res = res + str(lst[i])
+        if(i != len(lst) - 1):
```
cahirwpz: Parens after `if` or `while` in Python are mostly regarded as C-like habits.

.gdbinit
```diff
@@ -1,6 +1,8 @@
+set tcp connect-timeout 30
 target remote localhost:1234
 break main
 continue
 
 source debug/ktrace.py
 source debug/kdump.py
+source debug/mutex.py
```
cahirwpz: GDB extensions look ok. Please move to separate PR.

debug/mutex.py
```diff
@@ -0,0 +1,57 @@
...
+            res = res + ", "
+    res = res + "]"
+    return res
```
dyniec: Did you mean:
```python
return "[" + ", ".join(lst) + "]"
```

cahirwpz: EDIT: In fact just `return repr(lst)` works as, I suppose, intended.

coodie: I had to change string representation of list slightly because doing `print lst` printed the list like this:

`[u't1', u't2', u't3']`

And I wanted the list to be printed like this:

`[t1, t2, t3]`

Same goes for `repr(lst)`. I totally didn't think of functional way to create this function. I like @dyniec's solution the most since it's the most readable to me.

include/turnstile.h
coodie: Sorry to disappoint you about not asking ;).

I like this solution more. I just want to mention that FreeBSD counterpart requires we enter the `turnstile wait` with spinlock on turnstile, which in our case is equivalent to entering this function in with critical section on.

cahirwpz: > Sorry to disappoint you about not asking ;)

No problem ;-)

Yeah. You're right – it's even explicitely mentioned in [the comment](http://bxr.su/FreeBSD/sys/kern/subr_turnstile.c#674). Frankly, I cannot justify the reason behind doing so. IMO a reader will get confused when they don't see `cs_enter` / `cs_leave` symmetry. Let's hold on consistency as much as we can :)

cahirwpz:
Let's finish it up today! We're almost there ;-)

include/turnstile.h
coodie: I think I can justify why they do it like this. When you have SMP architecture disabling interrupts isn't enough (other processor might still change the value), but blocking spin lock is enough. So the question is where to put the spin lock. Obviously we should lock the spin lock when trying to modify the turnstile, but the spinlock is inside the turnstile itself so should we exctract it and lock it outside, or just embed it in function and change invariants. Note that since FreeBSD uses hashing on turnstile they have more locks to lock than just one. Also similar situations with race condition and missed wake-ups might occur with implementation of other types of mutexes, these weird invariants might be a hint that one should remember about possible race conditions.

.gdbinit
coodie: I realised it is better not to do force pushing because this messes up github history. Since these changes work, are tested by me, reviewed, and obviously associated with mutexes it is OK I guess to put them in one pull request.

I think it is good idea to add GDB scripts to every new subsystem associated PR. So this is something like 'system + scripts' or nothing, because in case something happens we might have easier time debugging it, because we instantly get the scripts.

cahirwpz:
Good job! Finally, we have a basic implementation of mutexes!

PR #125, opened 2016-11-11
Enabled libc for user programs

rafalcieslak:
Thanks to crosstool-ng already building newlib for us, it turned out that enabling libc is as simple as adding `-lc` to linker flags - so this branch has a working libc demo.

Note that disabling `-nostdlib` is not the same as adding `-ld`. In particular, `-nostdlib` pulls `crti.o` and `crtbegin.o` from compiler's resources, and we are definitely not at the point where we might use them instead of our own `start.S`.

I'll continue working on this branch to add system call wrappers stubs. Kernel-side system call support will be done in a separate branch.

### DIFF
--- a/user/Makefile
+++ b/user/Makefile
@@ -6,6 +6,7 @@ PROGRAMS_C = prog.c
 
 CFLAGS   = -std=gnu11 -O0 -Wall -Werror -fno-builtin -ffreestanding
 LDFLAGS  = -nostdlib -T mimiker.ld
+LDLIBS   = -lc
 
 PROGRAMS_OBJS = $(PROGRAMS_C:%.c=%.o)
 PROGRAMS_UELFS = $(PROGRAMS_C:%.c=%.uelf)
@@ -21,7 +22,7 @@ all: $(PROGRAMS_UELF_OS) $(PROGRAMS_UELFS)
 # Linking the program according to the provided script
 %.uelf: %.o start.o mimiker.ld
 	@echo "[LD] $(DIR)$< -> $(DIR)$@"
-	$(CC) $(LDFLAGS) start.o -o $@ $<
+	$(CC) $(LDFLAGS) start.o -o $@ $< $(LDLIBS)
 
 # Treating the liked program as binary data, placing it in .rodata
 # section, and embedding into an object file
--- a/user/Makefile
+++ b/user/Makefile
@@ -3,6 +3,7 @@ all: prog.uelf.o
 include ../Makefile.common
 
 PROGRAMS_C = prog.c
+SYSCALLS_C = syscalls.c
 
 CFLAGS   = -std=gnu11 -O0 -Wall -Werror -fno-builtin -ffreestanding
 LDFLAGS  = -nostdlib -T mimiker.ld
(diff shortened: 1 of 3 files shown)


cahirwpz:
I'm ready to merge this branch as is.

rafalcieslak:
Travis uses the old toolchain, which, apparently, comes with a different `unistd.h`, so implementations for system call wrappers do not match their prototypes. I may remove `#include <unistd.h>` to fix this mismatch, but it would need restoring once we move to new toolchain entirely.

cahirwpz:
Would you be able to `ifdef` things out based on preprocessor builin defines?

rafalcieslak:
This is ridiculous. `libc` provided by `mips-mti-elf` uses `-mhard-float`, while `libc` provided by `mipsel-unknown-elf` uses `-msoft-float`, so I can't produce universal `prog.o` that would link correctly with `libc`s from both toolchains. 

To solve this, I would need to introduce extra makefile variables that control compilation flags depending on the used toolchain. Should I do that? We'll be dropping support for "old toolchain" eventually (soon?).

goniz:
yes, the mti-elf toolchain provides super weird newlib libc versions only with hard float while the mti-linux has soft float..
by my taste, the crosstool-ng generated toolchain is preferable as you can always update the toolchain components when the need arises and the toolchain defaults can be adjusted to fit your products needs..

cahirwpz:
Enabled hardware FPU in the toolchain with b36621f11b891164995f10339d22deccb6a13095.

rafalcieslak:
The toolchain accepts new configuration, and indeed passes `-mhard-float` when building *some* components, but not all. This results in linker printing numerous warnings when linking `prog.uelf`, that are similar to this:
```
~/local/mipsel-unknown-elf/lib/gcc/mipsel-unknown-elf/5.2.0/../../../../mipsel-unknown-elf/bin/ld:
Warning: prog.uelf uses -mhard-float (set by start.o),
~/local/mipsel-unknown-elf/lib/gcc/mipsel-unknown-elf/5.2.0/../../../../mipsel-unknown-elf/lib/libc.a(lib_a-abort.o)
uses -msoft-float
```
For the record, crostool compiled `abort.c` from newlib with this command:
```
mipsel-unknown-elf-cc -B/mimiker/toolchain/mips/.build/mipsel-unknown-elf/build/build-libc/mipsel-unknown-elf/newlib/ -isystem /mimiker/toolchain/mips/.build/mipsel-unknown-elf/build/build-libc/mipsel-unknown-elf/newlib/targ-include -isystem /mimiker/toolchain/mips/.build/src/newlib-2.2.0/newlib/libc/include -B/mimiker/toolchain/mips/.build/mipsel-unknown-elf/build/build-libc/mipsel-unknown-elf/libgloss/mips -L/mimiker/toolchain/mips/.build/mipsel-unknown-elf/build/build-libc/mipsel-unknown-elf/libgloss/libnosys -L/mimiker/toolchain/mips/.build/src/newlib-2.2.0/libgloss/mips    -DPACKAGE_NAME="newlib" -DPACKAGE_TARNAME="newlib" -DPACKAGE_VERSION="2.2.0" -DPACKAGE_STRING="newlib 2.2.0" -DPACKAGE_BUGREPORT="" -DPACKAGE_URL="" -I. -I/mimiker/toolchain/mips/.build/src/newlib-2.2.0/newlib/libc/stdlib -Os -D__NO_SYSCALLS__ -DNO_FLOATING_POINT -DMISSING_SYSCALL_NAMES -fno-builtin      -g -Os -c -o lib_a-abort.o `test -f 'abort.c' || echo '/mimiker/toolchain/mips/.build/src/newlib-2.2.0/newlib/libc/stdlib/'`abort.c
```

So user programs do successfully compile, but the linker reminds us to be careful about mixing soft and hard floats. Even though newlib objects like `abort.o` don't actually use floating point calculations.

PR #126, opened 2016-11-11
Syscalls vector and sys_write stub

rafalcieslak:
This is an improved variant of #107. Introduces a minimal syscall vector, and implements `sys_write` stub, which may be used by user programs for writing to temporary stdout/stderr dummies.

### DIFF
--- a/exec_syscall.c
+++ b/exec_syscall.c
@@ -1,0 +1,13 @@
+#include <common.h>
+#include <exec.h>
+
+int main() {
+  exec_args_t exec_args;
+  exec_args.prog_name = "syscall_test";
+  exec_args.argv = (char *[]){"syscall_test"};
+  exec_args.argc = 1;
+
+  do_exec(&exec_args);
+
+  return 0;
+}
--- a/include/sysent.h
+++ b/include/sysent.h
@@ -33,7 +33,7 @@ extern sysent_t sysent[];
 
 int syscall(int n);
 
-/* Empty syscall handler, for uninmplemented and depracated syscall numbers. */
+/* Empty syscall handler, for unimplemented and deprecated syscall numbers. */
 int sys_nosys(thread_t *, syscall_args_t *, int *error);
 
 #endif /* !_SYS_SYSENT_H_ */
--- a/include/sysent.h
+++ b/include/sysent.h
@@ -1,16 +1,39 @@
 #ifndef _SYS_SYSENT_H_
 #define _SYS_SYSENT_H_
+#include <stdint.h>
+#include <mips/mips.h>
 
 typedef struct thread thread_t;
 
-typedef int syscall_t(thread_t *);
+#define SYSCALL_ARGS_MAX 4
+
+typedef struct syscall_args {
+  uint32_t code;
+  reg_t args[SYSCALL_ARGS_MAX];
+} syscall_args_t;
+
+typedef int syscall_t(thread_t *, syscall_args_t *, int *error);
 
 typedef struct { syscall_t *call; } sysent_t;
 
 extern sysent_t sysent[];
 
-#define SYS_LAST 0
+#define SYS_EXIT 1
+#define SYS_OPEN 2
+#define SYS_CLOSE 3
+#define SYS_READ 4
+#define SYS_WRITE 5
+#define SYS_LSEEK 6
+#define SYS_UNLINK 7
+#define SYS_GETPID 8
+#define SYS_KILL 9
+#define SYS_FSTAT 10
 
-int syscall(unsigned num);
+#define SYS_LAST 10
+
+int syscall(int n);
+
+/* Empty syscall handler, for uninmplemented and depracated syscall numbers. */
+int sys_nosys(thread_t *, syscall_args_t *, int *error);
 
 #endif /* !_SYS_SYSENT_H_ */
--- a/include/sysent.h
+++ b/include/sysent.h
@@ -12,7 +12,7 @@ typedef struct syscall_args {
   reg_t args[SYSCALL_ARGS_MAX];
 } syscall_args_t;
 
-typedef int syscall_t(thread_t *, syscall_args_t *, int *error);
+typedef int syscall_t(thread_t *, syscall_args_t *);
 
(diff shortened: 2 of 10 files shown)


exec_syscall.c
```diff
@@ -0,0 +1,13 @@
+#include <common.h>
+#include <exec.h>
+
+int main() {
```
cahirwpz: Split it out to another PR, where you add `exec_launch` that can start arbitrary embedded process image based upon kernel parameter.

include/sysent.h
```diff
@@ -1,16 +1,39 @@
...
+#define SYS_GETPID 8
+#define SYS_KILL 9
+#define SYS_FSTAT 10
+
+#define SYS_LAST 10
+
+int syscall(int n);
 
-int syscall(unsigned num);
+/* Empty syscall handler, for uninmplemented and depracated syscall numbers. */
```
cahirwpz: Can you configure emacs to spell-check comments?

exec_syscall.c
rafalcieslak: I didn't do it just yet, because there indeed seems to be some problem with passing arguments to kernel. It appears to me as if OVPsim ignored `--argv` flag. I know @mkaim intended to investigate this today, so maybe he knows more defails.

sys/sysent.c
```diff
@@ -0,0 +1,37 @@
...
+    /* We need a local copy of the string to null-terminate it after @count
+       bytes, because the only printing facility we have is kprintf. */
+    char buf2[count + 1];
+    memcpy(buf2, buf, count);
+    kprintf("%.*s", (int)count, buf);
+
+    return count;
+  }
+
+  *error = EBADF;
```
cahirwpz: Is there any reason why you've chosen to return errno value and -1 separately?

sys/sysent.c
```diff
@@ -0,0 +1,37 @@
...
+int sys_write(thread_t *td, syscall_args_t *args, int *error) {
+  int fd = args->args[0];
+  const char *buf = (const char *)(uintptr_t)args->args[1];
+  size_t count = args->args[2];
+
+  /* TODO: copyout string from userspace */
+  if (fd == 1 || fd == 2) {
+    /* We need a local copy of the string to null-terminate it after @count
+       bytes, because the only printing facility we have is kprintf. */
+    char buf2[count + 1];
```
rafalcieslak: Oh yes, I forgot to remove this! See below that `buf2` is not even used.

exec_syscall.c
mkaim: Well, I've tried to figure it out but with no luck - I found out that "The semihost libraries for some processors do not support this feature" in [OVPsim user guide](http://www.ovpworld.org/documents/Simulation_Control_of_Platforms_and_Modules_User_Guide.pdf) (page 30) and nothing really helpful on the internet.

include/sysent.h
rafalcieslak: Certainly!

sys/sysent.c
rafalcieslak: Mostly to conform to a generic template, where system calls may return any value and set errno. Do I understand you correctly that you would lik to merge these into a single return value, with negative results standing for error codes, and positive ones for successful return values? I am not sure whether all possible syscalls always use -1 to signalize error.

user/syscall_test.c
```diff
@@ -0,0 +1,25 @@
...
+  asm volatile("move $a0, %2\n"
+               "move $a1, %3\n"
+               "move $a2, %4\n"
+               "li $v0, 5\n" /* SYS_WRITE */
+               "syscall\n"
+               "move %0, $v0\n"
+               "move %1, $v1\n"
+               : "=r"(retval), "=r"(errval)
+               : "r"(fd), "r"(buf), "r"(count)
+               : "4", "5", "6", "2", "3");
```
rafalcieslak: I admit, I am not exactly confident how exactly these asm inlines work. What I intended to express is that the compiler shouldn't use registers `a0`, `a1`, `a2`, `v0`, and `v1` for input/output operands, because I use these registers internally (otherwise the compiler placed `count` in `$a0`, so the first assembly instruction would overwrite it...). It didn't accept `"$a0"` nor `"a0"` as input, so I used register numbers, and then it immediately started working as I intended - `fd` got placed in `$a3`, `buf` in `$t0` and `count` in `$t1`.

I suppose there may be a more elegant way to do this, but the GCC inline asm documentation confuses me a lot. I would even prefer to write this in pure asm.

exec_syscall.c
cahirwpz: I think `command` parameter carries kernel arguments. Please read following document: `ImperasLib/source/mips.ovpworld.org/peripheral/SmartLoaderLinux/1.0/doc/OVP_Peripheral_Specific_Information_SmartLoaderLinux.pdf`.

sys/sysent.c
rafalcieslak: Fragment from `man 2 getpriority`:
```
RETURN VALUE
       Since getpriority() can legitimately return the value -1, it is necessary
to clear the external variable errno prior to the call, then check it afterward
to determine if -1 is an error or a legitimate value.
```

rafalcieslak: Oh, though in case of Linux, this behavior is only used by library function `getpriority()`, not the system call itself - which returns `20 - priority`, so that negative values are still used for errors. As we have control over both system call handlers AND library wrappers, we can always find a similar trick. No function should need entire 32-bit range for return values, even the address space is not that large.

Thus I think I'll go ahead and use negative return values for error codes.

user/syscall_test.c
cahirwpz: Ok. My bad. I got confused by the fact you use register numbers and not symbolic names.

To avoid readability issues I suggest to rewrite the routine in pure assembly.

rafalcieslak:
Merged with `master`, so that `write` implementation now lands in `user/syscalls.c`, like other wrapper stubs. I modified error value processing so that negative return values represent errors. Also, it turns out I can use `"%a0"` for register clobbers (silly intel-oriented gcc), so that asm inline is much more readable now.

PR #127, opened 2016-11-12
Expanding user syscall stubs

rafalcieslak:
This branch effectively glues #125 and #126, so that the standard library can actually use system calls.

Good news is that `printf` in userspace actually works now, even though we don't have `sbrk`. I don't know how does it do string formatting (all libc's internal memory allocation fails) but somehow it does.


user/syscalls.c
```diff
@@ -56,9 +100,22 @@ ssize_t write(int fd, const void *buf, size_t count) {
   return retval;
 }
 
-_off_t lseek(int __fildes, _off_t __offset, int __whence) {
-  errno = ENOSYS;
-  return (off_t)-1;
+_off_t lseek(int fd, _off_t offset, int whence) {
```
cahirwpz: Please rewrite this file to assembly. In most cases the code will be shorter and clearer:

```
#define DO_SYSCALL(num)         \
        li      $v0, num        \
        syscall                 \
        bgez    $v0, 1f         \
        move    $v1, $0         \
        sub     $v1, $0, $v0    \
        addi    $v0, $0, -1     \
1:      jr      $ra             \
        sw      $v1,_errno($gp) \

LEAF(_lseek)
      DO_SYSCALL(SYS_LSEEK)
END(_lseek)
```

rafalcieslak: I've considered using assembly, but I'm not sure if it's the right choice. The problem is with `errno`. Newlib implements it as a preprocessor macro, which expands to `(*__errno())`. The function `__errno` ensures reentrancy and returns an indirect pointer to where error value should be written. So that easily works in C, but not in assembly. I might reproduce this behavior in assembly, but it will make us incompatible with other standard libraries, and I am not sure whether we with to commit to newlib entirely. I am also not sure if a differently configured newlib won't change how `errno` is implemented.

rafalcieslak: So I've tried writing an asm macro that calls `__errno`. It needs stack to save `$ra` and the value returned from syscall. I did **not** test it, but it should be something along the lines of:
```
/* This implementation uses newlib-specific errno processing. */
#define DO_SYSCALL(num) \
    addiu $sp, $sp, -16 \
    sw $ra, 12($sp)     \
    li $v0, num         \
    syscall             \
    bgez $v0, 1f        \
    nop                 \
    sw $v0, 8($sp)      \
    jal __errno         \
    nop                 \
    lw $v1, 8($sp)      \
    negu $v1, $v1       \
    sw $v1, 0($v0)      \
    li $v0, -1          \
1:  lw $ra, 12($sp)     \
    addiu $sp, $sp, 16  \
    jr $ra              \
    nop
```
If you still consider it shorter and cleaner, then I'm fine with rewriting `syscals.c` to assembly.

rafalcieslak: I made a commit in an experimental branch which has a working implementation: 4b210edc44e8bcd96f6c9973ea11cd608aef0480. It's not as bad as I expected it to be. I'll merge it into this branch if you want.

cahirwpz: It looks much better and I think we can make it even shorter.

user/syscalls.S
```diff
@@ -0,0 +1,67 @@
...
+    lw $ra, 4($sp);     \
+    addiu $sp, $sp, 8;  \
+    jr $ra;             \
+    nop;
+
+#define SYSCALL(name, num) \
+    LEAF(name)      \
+    DO_SYSCALL(num) \
+    END(name)
+#define NO_SYSCALL(name) \
```
cahirwpz: I think you can define it only once (and locally) and then use `ALEAF` macro from `include/mips/asm.h`.

user/syscalls.S
```diff
@@ -0,0 +1,67 @@
+#include <mips/asm.h>
+#include <errno.h>
+
+    .set noreorder
+    .set noat
+
+/* This implementation uses newlib-specific errno processing. */
+#define DO_SYSCALL(num)  \
```
cahirwpz: What about that:

```asm
#define DO_SYSCALL(num)       \
    li $v0, num;              \
    syscall;                  \
    bltz $v0, syscall_failed; \
    nop;                      \
    jr $ra;                   \
    nop

SNESTED(syscall_failed, 24, $ra)
    addiu $sp, $sp, -24
    sw $ra, 20($sp)
    sw $v0, 16($sp)
    jal __errno
    nop
    lw $v1, 16($sp)
    negu $v1, $v1
    sw $v1, 0($v0)
    li $v0, -1
    lw $ra, 20($sp)
    jr $ra
    addiu $sp, $sp, 24
SEND(syscall_failed)
```

Please note that minimum stack frame size for non-leaf function is 24 according to [MIPS o32 Calling Convention Examples](http://cs.unm.edu/~jeffk/cs341f09/_media/o32callingconvention.pdf)

user/syscalls.S
rafalcieslak: I really can't figure out how to use `ALEAF`. It looks to me that it may only be used to declare a procedure and it's *single* alias, right?

user/syscalls.S
rafalcieslak: I believe this works well. I'm not sure if it's more clear, especially with how `bltz $v0, syscall_failed;` does not set `$ra` so that `syscall_failed` returns from 2 nested calls at once... but that's just my opinion.

I'm pushing a commit which applies this change, but can you please finish the work on this branch? It doesn't seem I'm doing anything useful here at this point.

PR #128, opened 2016-11-12
Add kthread command

coodie:
This PR adds kthread command to gdb. Everything should be documented and easy to use.

### DIFF
--- a/debug/threads.py
+++ b/debug/threads.py
@@ -25,6 +25,7 @@ def current_thread():
 
 
 class CtxSwitchTracerBP(gdb.Breakpoint):
+
     def __init__(self):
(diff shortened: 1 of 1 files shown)


debug/threads.py
```diff
@@ -167,25 +163,22 @@ def invoke(self, args, from_tty):
...
+                print(tds[0])
                 return
 
         for t in get_all_threads():
             if thread_id(t) == args:
                 print(t)
                 return
 
-        print "Can't find thread with '", args, "' id or name"
+        print("Can't find thread with '", args, "' id or name")
```
coodie: Note that this is Python2 and this will be printed as tuple.

cahirwpz: Right! We need extra line `from __future__ import print_function` for Python 2.x!

PR #129, opened 2016-11-12
Implemented sbrk system call

rafalcieslak:
Here's a simple yet complete `sbrk` implementation. It doesn't actually use `.data` (because we actually don't know where it is, and there may be no free space next to it), but that shouldn't matter. `libc` appears to be using it fine, even `malloc` in user program works now! There is a basic test provided in `user/prog.c`. I needed to add a `vm_map_expand`  function, which is used for increasing the length of a `vm_map_entry`.


user/prog.c
```diff
@@ -19,15 +20,44 @@ void marquee(const char *string, int offset) {
...
   if (argc < 1)
     abort();
 
+  sbrk_test();
+
   /* Test some libstd functions. They will all fail, because system calls are
      not hooked up yet, but this should at least compile and link
      successfully. */
-  printf("Hello libc!");
+  printf("Hello libc!\n");
```
rafalcieslak: As soon as `newlib` noticed that `sbrk` is working, it started buffering output... so I had to add a `\n` here to flush the buffer.

user/prog.c
```diff
@@ -19,15 +20,44 @@ void marquee(const char *string, int offset) {
...
+  /* And again. */
+  char *a3 = sbrk(50);
+  assert(a3 == a2 + 40);
+  /* Now expand it a lot, much more than page size. */
+  char *a4 = sbrk(0x5000);
+  assert(a4 == a3 + 50);
+  /* Test write access. */
+  memset(a4, -1, 0x5000);
+  /* See that previous data is unmodified. */
+  assert(*(a1 + 5) = 1);
```
goniz: Whoops.. did you mean  '==',  my friend?

rafalcieslak: Yes indeed! Nice catch, thanks.

include/thread.h
```diff
@@ -41,6 +42,10 @@ typedef struct thread {
   /* scheduler part */
   td_prio_t td_prio;
   int td_slice;
+  /* sbrk support */
```
cahirwpz: `vm_map` is a place to keep this information. Note, that several threads can share one `vm_map` instance.

sys/sysent.c
```diff
@@ -24,8 +27,58 @@ int sys_write(thread_t *td, syscall_args_t *args) {
...
+   memory map. But it does not matter much, because no application would assume
+   that we are actually expanding .data, it will use the pointer returned by
+   sbrk. */
+int sys_sbrk(thread_t *td, syscall_args_t *args) {
+  size_t increment = (size_t)args->args[0];
+
+  kprintf("[syscall] sbrk(%zu)\n", increment);
+
+  if (!td->td_uspace) {
+    log("sbrk called by a thread with no userspace map");
```
cahirwpz: This should never happen. In such cases `assert` is preferred method to indicate error.

user/prog.c
```diff
@@ -19,15 +20,44 @@ void marquee(const char *string, int offset) {
   textarea[TEXTAREA_SIZE - 1] = 0;
 }
 
+/* Temporarily required, because I don't want to mix printf with sbrk_test. */
+void assert(int cond) {
+  if (!cond)
```
cahirwpz: Please add dummy `sys_exit` implementation that will just set `td_state` to `TDS_INACTIVE` and switch to another thread. @psie will provide full implementation later.

sys/sysent.c
```diff
@@ -24,8 +27,58 @@ int sys_write(thread_t *td, syscall_args_t *args) {
...
+    return -ENOMEM;
+  }
+
+  if (!td->td_brk_entry) {
+    /* sbrk was not used before by this thread. */
+    size_t size = roundup(increment, PAGESIZE);
+    vm_addr_t addr;
+    if (vm_map_findspace(td->td_uspace, 0x08000000, size, &addr) != 0)
```
cahirwpz: Magic constant alert! Please name it and provide a comment why you chose this specific value.

sys/vm_map.c
```diff
@@ -181,6 +181,18 @@ int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
   return 0;
 }
 
+int vm_map_expand(vm_map_t *map, vm_map_entry_t *entry, vm_addr_t new_end) {
```
cahirwpz: I think `sbrk` can force a `vm_map_entry` to shrink if passed negative argument. Thus I suggest to rename this routine to `vm_map_entry_resize` and handle `new_end < entry->end` case.

rafalcieslak:
All requests done. I'm wondering whether we should invalidate some TLB entries when a `vm_map_entry` gets shrinked so that the old data wouldn't be accessible anymore. Any program that accesses the freed memory area triggers UB, so we are free to do whatever we want. In case of `sbrk`, it is not even a security issue, because the data that's left there belonged to the same program anyway... so I suggest we leave it for now, as correctly invalidating TLB may be tricky to implement.

rafalcieslak:
Now that I think about it: maybe it's not a security issue, but we need to clean TLB anyway - because if the same address space is used soon after for another mapping (say, `mmap`, or next `sbrk`), then the TLB entries may be still pointing to the old location.

Same problem happens when a `vm_map_entry` is removed from the map (`unmmap`) - the TLB must have corresponding entries invalidated.

include/vm_map.h
```diff
@@ -22,6 +27,9 @@ typedef struct vm_map {
   SPLAY_HEAD(vm_map_tree, vm_map_entry) tree;
   size_t nentries;
   pmap_t *pmap;
+  /* sbrk support */
+  vm_map_entry_t *td_brk_entry; /* The entry where brk segment resides in. */
```
cahirwpz: `brk` and `sbrk` change data segment size, so I suppose those two fields should be named `data_entry`, `data_size`.

sys/sysent.c
```diff
@@ -24,8 +28,69 @@ int sys_write(thread_t *td, syscall_args_t *args) {
...
+   that we are actually expanding .data, it will use the pointer returned by
+   sbrk. */
+int sys_sbrk(thread_t *td, syscall_args_t *args) {
+  size_t increment = (size_t)args->args[0];
+
+  kprintf("[syscall] sbrk(%zu)\n", increment);
+
+  assert(td->td_uspace);
+
+  if (!td->td_uspace->td_brk_entry) {
```
cahirwpz: In normal circumstances `td_uspace->data_entry` should be set up by call to `execve`.

cahirwpz: I don't think there's such thing as *break segment*. Here's an excerpt from *brk(2)* manual:

> brk() and sbrk() change the location of the program break, which defines the end of the process's data segment (i.e., the program break is the first location after the end of the uninitialized data segment).

I'm not sure if there's any software that relies on the fact that *program break* should start immediately after `_end` linker symbol. There's no trace of `sbrk` in FreeBSD's [vmspace](http://bxr.su/FreeBSD/sys/vm/vm_map.h#234). OTOH in Linux [mm_struct](http://lxr.free-electrons.com/source/include/linux/mm_types.h#L396) holds `start_brk` and `brk` fields.

Short program running on Linux suggests *sbrk* doesn't have to be mapped after `.bss`:
```c
#include <unistd.h>
#include <stdio.h>

extern char _edata[0];
extern char _end[0];

int main() {
  void *_ebrk = sbrk(0);
  printf("edata = %p\n", _edata);
  printf("end = %p\n", _end);
  printf("ebrk = %p\n", _ebrk);
  pause();
  return 0;
}
```

```
# ./test
edata = 0x5595e6b77040
end = 0x5595e6b77048
ebrk = 0x5595e856f000
```

```
# pmap $(pgrep test)
18096:   ./test
00005595e6976000      4K r-x-- test
00005595e6b76000      4K r---- test
00005595e6b77000      4K rw--- test
00005595e856f000    132K rw---   [ anon ]
```

sys/vm_map.c
```diff
@@ -181,6 +181,25 @@ int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
...
+    /* Expanding entry */
+    vm_map_entry_t *next = TAILQ_NEXT(entry, map_list);
+    vm_addr_t gap_end = next ? next->start : map->pmap->end;
+    if (new_end > gap_end)
+      return -ENOMEM;
+  } else {
+    /* Shrinking entry */
+    if (new_end < entry->start)
+      return -ENOMEM;
+    /* TODO: Invalidate tlb? */
```
cahirwpz: Certainly we have to unmap and free pages that are in released region. `pmap` interface should deliver abstractions that do not require explicit operations on TLB. However `pmap` is  incomplete in many aspects.

user/syscalls.c
```diff
@@ -17,8 +17,16 @@ struct timeval;
 #include <sys/types.h>
 #endif
 
-void _exit(int __status) {
-  /* Exit may not return! */
+void _exit(int status) {
```
cahirwpz: Please move the implementation to `syscalls.S`.

sys/sysent.c
rafalcieslak: For the record, FreeBSD [does not implement `sbrk` at all](http://bxr.su/FreeBSD/sys/vm/vm_mmap.c#121).

IIRC historically, back when expanding the data segment was the only way of allocating memory, data segment and process stack were placed in the same segment, so a movable pointer *break* was used to choose the point that divided these two. So I suppose what *brk(2)* says is mostly a historical convention. The test you did is probably a valid proof that applications don't generally require `sbrk` to expand `.data`, because if this was a case, such programs would not function correctly on Linux.

rafalcieslak:
I've updated this branch so that it works with recent additions to the syscall vector and glue. I've also disabled negative arguments to `sbrk`, marking them with comments in source and warnings in runtime.

PR #134, opened 2016-11-14
Major makefile cleanup

rafalcieslak:
I finally got down to cleaning up the main Makefile. My changes include:
- Correcting dependencies so that they point to files and not their parent directories
- Simplifying frequently edited variables
- Automatically generating as much as possible to reduce human-error risk (for example, creating a new test `*.c` required adding the file in two places in `Makefile`, and as some tests were present in only one of these places, the success of build varied with the order of dependency evaluation...)
- Providing slightly more documentation, as makefiles tend to get difficult to read. Minor formatting fixes.

As a pleasant effects of these changes:
- `make` no longer pointlessly rebuilds kernel `.elf`s and some other files when nothing was changed
- Fixed dependencies finally allow for parallelized building.

I've tested this new `Makefile` by manually touching each `*.{c,h,S}` file and verifying whether all targets rebuild correctly and in order, both with `make` and `make -j32`, so I am fairly confident everything works well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,40 +14,27 @@ TESTS = \
 	syscall.elf \
 	thread.elf \
 	vm_map.elf
-SOURCES_C = 
-SOURCES_ASM = 
+SOURCES_C =
+SOURCES_ASM =
 
-all: tags cscope mips stdc sys $(TESTS)
+all: tags cscope $(TESTS)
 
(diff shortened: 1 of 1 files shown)


cahirwpz:
Make metaprogramming FTW! It looks way better, thanks!

PR #140, opened 2016-11-18
Imported some clib string functions

rafalcieslak:
I need these to create a simple kernel args parser.


rafalcieslak:
It turns out that smallclib's implementation of `strsep` is incorrect, because and it behaves just like `strtok`. Therefore I imported `strsep` from FreeBSD's *libkern*, and it seems to work correctly.

cahirwpz:
This is first time we've run into problems with smallclib. My hunch was that it would happen and will happen again. That's why I proposed to take **all** routines (or as much as possible) from *BSD systems. We know they're widely used, throughly tested and behave well on variety of systems.

Moreover, not all C standard library functions are considered to be well designed. `strtok` is such an example – manual pages:
* state that the function is obsoleted by `strsep(3)` in [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=strtok&sektion=3&apropos=0&manpath=freebsd),
* or warn a user and give extensive list of bugs in [Linux](http://man7.org/linux/man-pages/man3/strtok.3.html#BUGS).

Thus, it's time to make a policy on importing new functions into `stdc`:

* If a function to be imported is not a part of *BSD libkern, one must provide good justification for its use.
* Always prefer simpler implementation over optimized one.
* Format new files and remove everything that the reader should not be bothered about.

I'd keep current way the sources are organized, i.e. directories: `stdlib`, `string` and so on.

@rafalcieslak please give me your opinion.

rafalcieslak:
To be fair, I wasn't accurate when I said that smallclib's `strsep` is incorrect. `strsep` is non-standard function, and actually smallclib does not provide it - but it has an incomplete stub, which - when completed, makes `strsep` an alias for `strtok`. So, strictly speaking, smallclib doesn't do anything wrong here. The problem we encountered this time is not with smallclib at all, but with **my whim to use a non-standard function**. IIRC, previous problems we had with smallclib should also be attributed to our inability to use it correctly.

But I don't mind switching to e.g. libkern. It provides a reasonable subset of standard functions with their "safer" replacements, and various additional utilities we may find useful at some point. Hopefully it would save us from revisiting this topic every month or so.

If we import sources from elsewhere, I suggest to keep our changes to minimum (including cleanup!). We don't want to fork and maintain our own version of the standard library, which might happen if we start modifying and simplifying copied files.

PR #141, opened 2016-11-18
Basic kernel arguments support

rafalcieslak:
**Note:** This branch incorporates changes from #140, so please process it before this one.

Here's a simple interface for accessing kernel args. These proposed changes include:
 - parsing kernel arguments (splitting them by spaces)
 - providing utilities for convenient access to arguments, namely:
```
const char *kernel_args_get(const char *key);
int kernel_args_get_flag(const char *flag);
```
The new test `args.c` is pretty self-explanatory. The `launch` script was updated, so that it may be run like this:
```
./launch -d args.elf arg1=value1 flag1 arg2=value2
```
Works with both OVPsim and QEMU.

@mkaim: I suppose you will find this useful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 # vim: tabstop=8 shiftwidth=8 noexpandtab:
 
 TESTS = \
+	args.elf \
 	callout.elf \
 	exec.elf \
 	exec_syscall.elf \
--- a/args.c
+++ b/args.c
@@ -1,0 +1,14 @@
+#include <args.h>
+#include <stdc.h>
+
+int main() {
+  const char *arg1 = kernel_args_get("arg1");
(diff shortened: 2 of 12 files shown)


cahirwpz:
I acknowledge the fact that we need global kernel runtime configuration. FreeBSD solves that using `kenv` inteface, i.e. `getenv` and `setenv` functions. Please read [kern_environment.c](http://bxr.su/FreeBSD/sys/kern/kern_environment.c), find similar solution in other kernels and compare them, try to capture all requirements and propose an interface (header file).

mkaim:
Well, it was my job to handle argv and now there is my [pull request](https://github.com/cahirwpz/mimiker/pull/142) doing almost same work.

PR #142, opened 2016-11-19
Kernel args

mkaim:
This pull request includes:

- platform_init function in mips/malta.c that clears bss section and fixes argv
- very simple argv parser - almost copy-paste from @rafalcieslak [pull request](https://github.com/cahirwpz/mimiker/pull/141)

~~argv fix function properly handle quoted strings.~~ - it's no longer true after code simplification.

Hope that `args.c` file is self-explanatory.


cahirwpz:
Let's move `launch` script to separate PR and discuss the changes there.

mips/malta.c
```diff
@@ -0,0 +1,64 @@
+#include <mips/malta.h>
+#include <malloc.h>
+
+void clear_bss() {
+  extern unsigned int __bss[];
+  extern unsigned int __ebss[];
+  for (unsigned int *p = __bss, *end = __ebss; p < end; ++p) *p = 0;
```
cahirwpz: Please **do use** functions from [stdc.h](https://github.com/cahirwpz/mimiker/blob/master/include/stdc.h) (`bzero` in this case). If there're functions from C standard library you'd like to use inside the kernel, please ask about them.

mips/malta.c
```diff
@@ -0,0 +1,64 @@
...
+}
+
+#define ISSPACE(a) ((a) == ' ' || ((a) >= '\t' && (a) <= '\r'))
+
+void fix_argv(int *p_argc, char ***p_argv) {
```
cahirwpz: Can you add a comment that explains what are you trying to fix? Show example input data.

Makefile
```diff
@@ -13,7 +13,8 @@ TESTS = \
 	sleepq.elf \
 	syscall.elf \
 	thread.elf \
-	vm_map.elf
+	vm_map.elf \
+	args.elf
```
rafalcieslak: Can we keep such lists alphabetically ordered? (see also `sys/Makefile`)

mips/malta.c
```diff
@@ -0,0 +1,70 @@
...
+  if (*p_argc < 2)
+    return;
+
+  char *cmd = (*p_argv)[1];
+  char *s = cmd, *p = cmd;
+  int argc;
+  char **argv;
+
+  // argv[1] string is modified - all whitespaces are overwritten
```
rafalcieslak: I'm not sure, but wouldn't it be a good idea to try and fix all arguments, not just argv[1]? If we possess the capability of splitting "incorrectly merged" arguments, then why not apply it everywhere, just in case? It probably (?) wouldn't hurt us, may hypothetically solve issues on other platforms (or in case we change simulator configuration and for some reason it changes the order/numer of arguments), and would be easier to document (*fixing all args just in case* vs. *fixing a particular arg for a particular platform*).

Though maybe it's just that I have a personal preference for generic code.

mkaim: Don't focus too much to that argv fix - we have much more problems to solve in this project. I know that my solution is not great but for now it should be enough. Just do not waste time trying to fix problems that has not even appeared yet and are unlikely to happen in near future. If they appear I will improve this function, but for now just keep it as it is.

mips/malta.c
```diff
@@ -0,0 +1,81 @@
...
+}
+#endif
+
+void platform_init(int argc, char **argv, char **envp, unsigned memsize) {
+  /* clear BSS section */
+  bzero(__bss, __ebss - __bss);
+  
+  _memsize = memsize;
+
+  _kenv.argc = argc;
```
cahirwpz: Please reimplement `fix_argv` function so that it goes over all `argv` and `envp` entries, tokenizes them and puts them into new `argv` array.

```
argc=3;
argv={"test.elf", "arg1 arg2=val   arg3=foobar  ", "  init=/bin/sh "};
envp={"memsize=128MiB", "uart.speed=115200"}
```

... should be transformed into:
```
_kenv.argc=7;
_kenv.argv={"test.elf", "arg1", "arg2=val", "arg3=foobar", "init=/bin/sh", "memsize=128MiB", "uart.speed=115200"}
```

mkaim: Your example of `envp` looks incorrect: it should be:
```
envp={"memsize", "128MiB", "uart.speed", "115200"}
```
You want me to join every key-value pair of `envp` into one word and put it into _kenv.argv?

mips/malta.c
```diff
@@ -0,0 +1,85 @@
...
+
+void platform_init(int argc, char **argv, char **envp, unsigned memsize) {
+  /* clear BSS section */
+  bzero(__bss, __ebss - __bss);
+  
+  _memsize = memsize;
+
+  fix_argv(&argc, &argv);
+  _kenv.argc = argc;
+  _kenv.argv = argv;
```
mkaim: `envp` contains key-value pairs - do you want me to translate these pairs into arguments (`"<key>=<value>"` strings) and put them into `_kenv.argv`? Or maybe we just keep `envp` as it is?

cahirwpz: Please convert `envp` to `argv` format. That'll do for now.

PR #152, opened 2016-11-27
Basic file descriptor support

rafalcieslak:
Here is my proposed interface to files and descriptor tables.

I realize this is a not as small PR. I did my best to implement as little as possible and keep changes to minimum, but it's still not a trivial change. I believe it will be very difficult to split it into smaller parts, because pretty much everything here is needed even to get the simplest example working. So if anyone has some ideas on how to trim this PR, I'm all ears - just take note of how these new functions depend on each other. I suppose the primary reason of why this change is big is because all logic is split into a large number of tiny functions, and all of them are very useful on their own. Anyway, I suppose it won't be hard to review, since everything here is mostly straight-forward.

I've based my implementation on FreeBSD, but it is *severly simplified*, and supports only the very basic operations. What I find confusing about FreeBSD's implementation is that they name various concepts using the same term `fd`. So sometimes "closing fd" refers to removing the file descriptor table, sometimes to invalidating a particular descriptor, and sometimes to closing the underlying file. To clean up the terms, I'm using:

- `file_desc_table` referring to the per-thread mapping from integers to files
- `file_desc` then I mean a particular descriptor (number) in the table
- `file` when I refer to the underlying file.

As we agreed, currently there is a hard limit of 20 files open per-process.

As for now, it is difficult to test this implementation thoroughly, and it will stay this way up until we implement `fork/clone` and/or `dup`. I suggest we test this as much as we can with currently available features, and re-investigate this when new features make use of this interface.

Because I removed the previous way we processed `write()` to files 1 and 2, usermode programs no longer have a way of printing output. My suggestion for fixing this is to implement a stub `/dev/uart` pseudo-file, in a way almost identical to how I implemented `/dev/null`. It's going to be very simple, and will also help testing descriptors a lot!

I am not yet finished with this branch, I will be cleaning up the implementation a bit, and will prepare some more exhaustive tests. Comments on the interface and initial implementation are very welcome!


sys/mutex.c
```diff
@@ -1,6 +1,7 @@
 #include <sync.h>
 #include <atomic.h>
 #include <mutex.h>
+#include <thread.h>
```
rafalcieslak: I've run into some trouble with how `mutex.c` and `turnstile.h` were using features included by a long chain of indirect `#includes`. Maybe we should use IWYU or a similar tool to analyze our sources.

sys/mutex.c
```diff
@@ -45,3 +46,7 @@ void mtx_unlock(mtx_t *mtx) {
   turnstile_signal(&mtx->turnstile);
   cs_leave();
 }
+
+bool mtx_is_locked(mtx_t *mtx) {
```
rafalcieslak: This function isn't useful by itself, but I found it convenient to use it for some assertions. After cleaning the code, it turns out I only use it once, so I can remove it, if you want - but we may find it useful in other places as well.

rafalcieslak:
At this point `fd_test.c` is fairly complex and uses all basic descriptor operations.

user/fd_test.c
```diff
@@ -0,0 +1,91 @@
...
+  int error = 0;
+  int fd0, fd1, fd2;
+  char buf[100];
+
+  /* Open a /dev/null */
+  fd0 = open("/dev/null", 0, O_WRONLY);
+  assert(fd0 == 0);
+  /* Open an invalid file */
+  fd1 = open("/proc/cpuinfo", 0, NULL);
+  assert(fd1 < 0);
```
cahirwpz: Please use proper `errno` values.

include/filedesc.h
```diff
@@ -0,0 +1,50 @@
...
+#define __FILEDESC_H__
+#include <file.h>
+#include <bitstring.h>
+#include <mutex.h>
+
+typedef struct thread thread_t;
+
+typedef struct {
+  file_t **fdt_files; /* Open files array */
+  bitstr_t *fdt_map;  /* Bitmap of used fds */
```
cahirwpz: Please allocate statically, for simplicity, if you know its size is bounded and small.

rafalcieslak:
I've updated this branch so that `sys_{read,write,open}` now correctly copyin/copyout data from/to user space. I've also added extra tests to `fd_test.c` which make sure that kernel correctly handles invalid pointers in syscall arguments.

cahirwpz:
Let's work on trimming down this PR.

* Firstly... move the test and simple fixes to their correspoding PRs. This will get merged promptly.
* Secondly... I'll ask you to read https://github.com/cahirwpz/mimiker/commit/8d875302e72dd129200dbde878f168600c1a0e3a and share your opinion on it. I'd like you to implement `devnull` as the first device, after you move it to `dev/null.c` location. Later a wrapper for `rtc.c` could be provided as well.

As a preparation for the implementation of first device you should write [uiomove](http://netbsd.gw.com/cgi-bin/man-cgi?uiomove+9+NetBSD-current), which should be relatively straightforward. Please have a look at [subr_copy.c](http://bxr.su/NetBSD/sys/kern/subr_copy.c#99).

rafalcieslak:
I'm not sure if moving `fd_test.c` away from this branch is a good idea. Whenever I update anything here or add new features and/or possible fail scenarios, I will update the tests, so keeping them in a separate branch will be tedious.

I've investigated how `uiomove` works, and clearly NetBSD's version is much simpler than the one used in FreeBSD (which I failed to comprehend). Also please note that NetBSD does not use `uiomove` for `/dev/null` at all, and actually, in order to use it I would need to implement e.g. a `/dev/zero`. And, naturally, implementing 8d87530 and related utilities will grow this branch even more, instead of trimming it down... Should I work on basic devices and `uiomove` in a separate branch?

My notes about 8d87530:
- I believe the `uio_vmspace` pointer in `uio_t` is unnecessary, as we currently assume that the user vm map corresponding to currently running thread is always active.
- I don't think we will want to provide device `mmap` support at this point, so maybe we should postpone adding `d_mmap`.

exec_fd_test.c
```diff
@@ -0,0 +1,13 @@
+#include <common.h>
+#include <exec.h>
+
+int main() {
```
cahirwpz: Since @mkaim implemented proper kernel parameter passing, could we have a single launcher that passes `run=elfname;arg1;arg2;...` to the kernel?

rafalcieslak: I would love to, but this is still not enough! In order to do this elegantly, a getter function that looks up kernel flags ([such as this one](http://bxr.su/FreeBSD/sys/kern/kern_environment.c#451)) would be very helpful, and I am under impression that we are considering such interface and maybe even @mkaim is working on it (I honestly don't know). I would have implemented such launcher ages ago, but I abstain from it until the interface to kernel args ("environment") is ready.

cahirwpz:
> Should I work on basic devices and uiomove in a separate branch?

Yes, please. We probably don't need `/dev/zero`, however we could add `dev/malta/uart_cbus.c` (i.e. interrupt driven **tty2** driver), since `sleepqueue` implementation is just behind the corner (`sched_yield` will do for now).

> the `uio_vmspace` pointer in `uio_t` is unnecessary

Other choice was `thread_t`. This unambiguously identifies address space which we want to copy to / from. I reckon the field is not particularly useful but I'd keep it there for the record – just to make it easier to debug tens or hundreds `uio`'s in flight.

> we should postpone adding `d_mmap`

I agree.

user/fd_test.c
```diff
@@ -0,0 +1,137 @@
+#include <unistd.h>
+#include <fcntl.h>
+#include <assert.h>
+#include <string.h>
+#include <errno.h>
+
+int main(int argc, char **argv) {
```
cahirwpz: Could you split it into a set of functions each testing specific feature?

rafalcieslak:
> We probably don't need `/dev/zero`

I completely agree that we don't *need* it for anything, but I think its implementation would be good for documentation purposes, as it would be a non-quite-trivial example of using the device interface. A `/dev/zero` would also help in preparing meaningful tests, since `/dev/null` doesn't even use `uiomove`. For this reason, if I were to implement writing `/dev/zero`, I would actually copy the data into some junkyard buffer - merely to demonstrate/test the interface.

What is handy about `uio_vmspace` is that it may be used to recognize whether the move is performed from/to user space, or kernel space. I imagine some drivers may want to use `uiomove` for their own purposes, in such case copying data should not fail on writing to kernel space addresses. See [`copyout_vmspace`](http://bxr.su/NetBSD/sys/kern/subr_copy.c#235).

cahirwpz:
> I'm not sure if moving `fd_test.c` away from this branch is a good idea.

I reckon it is worth it, as it will get merged promptly. Even if you modify the file later, the PR will still be smaller.

rafalcieslak:
I'm glad I took my time to prepare a complex `fd_test.c` - this way porting this branch to use VFS totally felt like test-driven development!

rafalcieslak:
I'm done with the final refactoring I wanted to do - this branch is again ready to be reviewed.

cahirwpz:
Please move changes that are not specific to file / file descriptor management to a separate PR.

cahirwpz:
LGTM

I remember you complaining on current method of conducting code merge. It's obvious I won't pull ~50 commits into the master. How should we proceed to save you unnecessary work?

rafalcieslak:
I believe this would require a detailed discussion, so for now let's merge this as you normally would - I don't have many branches that incorporate these changes anyway, so I should be fine.

PR #153, opened 2016-11-28
Safe data copy between kernel and user space

cahirwpz:
This PR adds `copyin`, `copyout`, `copyinstr`, `fuword32`, `suword32` that deal with safe copying of data between user and kernel space.

To show that it actually works `user/syscall_test.c` was recycled into `user/misbehave.c` and I fixed a kernel crash on following user code:
```c
assert(write(STDOUT_FILENO, (char *)0xDEADC0DE, 100) == -1);
assert(errno == EFAULT);
```

The issue is not fully fixed, but all tools are right in place. To detect if an exception happened one must check `td_onfault` field which stores program counter of failing routine continuation.

@rafalcieslak Please review the code.


mips/copy.S
```diff
@@ -0,0 +1,155 @@
...
+#include "assym.h"
+
+	.set	noreorder		# Noreorder is default style!
+
+/*
+ * int copystr(const void *kfaddr, void *kdaddr, size_t len, size_t *lencopied)
+ *
+ * Copy a NIL-terminated string, at most maxlen characters long.  Return the
+ * number of characters copied (including the NIL) in *lencopied.  If the
+ * string is too long, return -ENAMETOOLONG; else return 0.
```
rafalcieslak: I'm not sure I like this error code, `ENAMETOOLONG` refers to filenames. Maybe `ENOMEM` ("Not enough space") would be better?

sys/sysent.c
```diff
@@ -14,20 +15,29 @@ int sys_nosys(thread_t *td, syscall_args_t *args) {
...
+  log("sys_write(%d, %p, %zu)", fd, buf, count);
 
   /* TODO: copyout string from userspace */
   if (fd == 1 || fd == 2) {
-    kprintf("%.*s", (int)count, buf);
+    char kbuf[80];
+    size_t done = min(count, sizeof(kbuf));
 
-    return count;
+    retval = copyin(buf, kbuf, done);
```
rafalcieslak: I suppose we should also add a test case (to `misbehave.c`?) where `buf` does not fit in `kbuf` - even though I realize this *should* work.

sys/sysent.c
```diff
@@ -14,20 +15,29 @@ int sys_nosys(thread_t *td, syscall_args_t *args) {
...
 int sys_write(thread_t *td, syscall_args_t *args) {
+  int retval;
   int fd = args->args[0];
   const char *buf = (const char *)(uintptr_t)args->args[1];
   size_t count = args->args[2];
 
-  kprintf("[syscall] write(%d, %p, %zu)\n", fd, buf, count);
+  log("sys_write(%d, %p, %zu)", fd, buf, count);
 
   /* TODO: copyout string from userspace */
```
rafalcieslak: This comment should be removed now.

include/common.h
```diff
@@ -10,6 +10,10 @@
 typedef unsigned long vm_addr_t;
 typedef unsigned long pm_addr_t;
 
+/* Wrapper for various GCC attributes */
+#define __nonnull(x)    __attribute__((__nonnull__(x)))
+#define __returns_twice __attribute__((__returns_twice__))
```
rafalcieslak: You introduce this macro, but never use it - seems like you forgot to apply it to some function.

sys/sysent.c
cahirwpz: We cannot do much until buffering for I/O operations in kernel is added. The behavior is correct according to `write(2)` manual:

> On success, the number of bytes written is returned (zero indicates nothing was written).  It is not an error if this number is  smaller  than  the number of bytes requested; this may happen for example because the disk device was filled.

cahirwpz:
@rafalcieslak Is the PR ready to merge in your opinion?

sys/sysent.c
rafalcieslak: Yes, I understand, this **is** correct, but an extra test case confirming that `sys_write` does not fail in such scenario might be useful. Though I might just as well add such test case to #152, so do go ahead and merge this branch.

mips/copy.S
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * Copy specified amount of data from user space into the kernel.
+ */
+NESTED(copyin, CALLFRAME_SIZ, ra)
+	PTR_SUBU sp, sp, CALLFRAME_SIZ
+	.mask	0x80000000, (CALLFRAME_RA - CALLFRAME_SIZ)
+	LA	v0, copyerr
+	blt	a0, zero, copyerr       # make sure address is in user space
+	REG_S	ra, CALLFRAME_RA(sp)
+	LOAD_PCPU(v1)
```
rafalcieslak: I see what's wrong! You forgot `lw $k1, PCPU_CURTHREAD($k1)` here, and in other places where you reference `LOAD_PCPU`. Otherwise instead of setting current thread's `td_onfault`, you write to some random address.

cahirwpz: Just found it out (independently) and fixed that! Nonetheless, good find! :)

rafalcieslak:
I tested `copyin`, `copyout` and `copyinstr` with #152 and an even larger testcase set, and everything appears to be working well now.

PR #154, opened 2016-11-28
Implemented uiomove (and imported device interface)

rafalcieslak:
This was indeed straight-forward. I hope this is what you meant!

I imported 8d87530 with very minor changes.

The provided test case verifies that both move directions work, multiple io vectors are supported, as well as offsets.

### DIFF
--- a/sys/subr_copy.c
+++ b/sys/subr_copy.c
@@ -1,0 +1,71 @@
+#include <uio.h>
+#include <systm.h>
+#include <stdc.h>
+#include <vm_map.h>
+
+static int copyin_vmspace(vm_map_t *vm, const void *restrict udaddr,
+                          void *restrict kaddr, size_t len) {
+  if (vm == get_kernel_vm_map()) {
+    memcpy(kaddr, udaddr, len);
+    return 0;
+  } else if (vm == get_user_vm_map()) {
+    return copyin(udaddr, kaddr, len);
+  } else {
+    /* copyin on non-active vm maps is not supported */
+    assert(false);
+    __builtin_unreachable();
+  }
+}
+static int copyout_vmspace(vm_map_t *vm, const void *restrict kaddr,
+                           void *restrict udaddr, size_t len) {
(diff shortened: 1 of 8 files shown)


sys/subr_copy.c
```diff
@@ -0,0 +1,71 @@
...
+int uiomove(void *buf, size_t n, uio_t *uio) {
+  char *cbuf = buf + uio->uio_offset;
+  assert(uio->uio_op == UIO_READ || uio->uio_op == UIO_WRITE);
+  int error = 0;
+  while (n > 0 && uio->uio_resid > 0) {
+    /* Take the first io vector */
+    iovec_t *iov = uio->uio_iov;
+    size_t cnt = iov->iov_len;
+    if (cnt == 0) {
+      /* If no data left to move in this vector, proceed to the next io vector,
```
cahirwpz: `s/vector/block/` ?

sys/subr_copy.c
```diff
@@ -0,0 +1,71 @@
...
+         or finish moving data if this was the last vector. */
+      if (uio->uio_iovcnt == 0)
+        break;
+      uio->uio_iov++;
```
rafalcieslak: In such case, the caller must have another pointer to the allocation. In case `uiomove` fails, the passed `uio` shall represent I/O operations that were not performed.

sys/subr_copy.c
rafalcieslak: I'm pretty sure `iovec` stands for "I/O vector", so `uio_iov` is "an array of I/O vectors", and a single element of this array is a "vector". But this is just a comment, so feel free to edit is as you wish.

cahirwpz: Right. It's even called this way in [uiomove(9)](http://netbsd.gw.com/cgi-bin/man-cgi?uiomove+9+NetBSD-current).

sys/subr_copy.c
cahirwpz: Ok, I get it.

Issue #156, opened 2016-11-29
Interrupts are enabled before threads are initialized


rafalcieslak:
Having heard that @dyniec has problems with QEMU, I spent this day investigating what happens when running mimiker on QEMU, and I've found some pretty interesting stuff. Obviously, as it always turns out, it is by mere luck that the system runs okay-ish on OVPsim.

So the first problem is something we were already aware of: QEMU always jumps to `ebase + 0x200` whenever any interrupt arrives. Vectored interrupts are incorrectly supported in QEMU (see: [helper.c:674](https://github.com/qemu/qemu/blob/master/target-mips/helper.c#L674)), and we can either fix that ourself (shouldn't be difficult, but I doubt we want to maintain our qemu fork), or prepare two variants of our exception handlers - one for QEMU, using a single interrupt address `ebase + 0x200`.

As for now, this is how we process such interrupt:
```
irq0:
1:      j       b1
        nop
```
which causes an endless loop whenever QEMU emits an interrupt. So in order to *temporarily* fix this, I changed it into:
```
irq0:
1:      j       irq7
        nop
```
since the clock interrupt is the only interrupt we currently use anyway.

But that led to even more confusing problems. To explain their cause, please have a look at how we currently do `kernel_boot`, taking note on when are interrupts enabled:

```c
  cpu_init();
  pcpu_init();
  pci_init();
  pm_init();
  intr_init();      // <------- This calls enable_interrupts
  callout_init();
  tlb_init();
  pmap_init();
  vm_object_init();
  vm_map_init();
  sched_init();
  sleepq_init();
  mips_clock_init();  // <------ This enables clock interrupt, also disables and re-enables interrupts
  kprintf("[startup] subsystems initialized\n"); // <----- This takes, relatively, a long time
  thread_init((void (*)())main, 2, argc, argv);
```

Thus it is totally possible for a clock interrupt to arrive while we are printing "subsystems initialized", and it happens quite often when running mimiker under QEMU. Suppose that happens. To process an interrupt, we start by calling `kern_exc_enter`, which stores some data onto the stack, but also accesses current thread data (`LOAD_PCPU` and `PCPU_CURTHREAD`). However, `_pcpu_data.currthread` was not set yet, because `thread_init` wasn't called yet! This, naturally, lead to immediate crash by dereferencing a null pointer.

Thus we need to either:
[...]

PR #157, opened 2016-12-01
cdev, devfs stubs, /dev/null and /dev/zero

rafalcieslak:
I decided that there is no point (yet) in having `devfs` use its own list of registered devices, for now it looks for cdevs in the list of all allocated character devices. This made `devfs` stub so simple it's silly.

The next steps will be to add some more devices (`uart`, `kmem`), and link devfs with #152.

### DIFF
--- a/sys/cdev.c
+++ b/sys/cdev.c
@@ -1,0 +1,37 @@
+#include <cdev.h>
+#include <malloc.h>
+#include <string.h>
+#include <mutex.h>
+
+cdev_list_t all_cdevs;
+static mtx_t all_cdev_mtx;
+
+static MALLOC_DEFINE(cdev_pool, "cdevs pool");
+
+void cdev_init() {
+  TAILQ_INIT(&all_cdevs);
+
+  kmalloc_init(cdev_pool);
+  kmalloc_add_arena(cdev_pool, pm_alloc(1)->vaddr, PAGESIZE);
+
+  mtx_init(&all_cdev_mtx);
+
+  /* Initialize some core devices */
+  dev_null_init();
+}
+
+cdev_t *make_dev(cdevsw_t cdevsw, const char *name) {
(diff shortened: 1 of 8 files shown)


sys/cdev.c
```diff
@@ -0,0 +1,37 @@
...
+
+cdev_t *make_dev(cdevsw_t cdevsw, const char *name) {
+  cdev_t *cdev = kmalloc(cdev_pool, sizeof(cdev_t), M_ZERO);
+  strlcpy(cdev->cdev_name, name, CDEV_NAME_MAX);
+  cdev->cdev_sw = cdevsw;
+
+  /* TODO: This function may get called before the threads system is
+     initialized. Calling mtx_lock before threads started crashes, because
+     thread_self() is not ready yet! See issue #156. Once we fix this, enable
+     mtx_lock and mtx_unlock below. */
```
rafalcieslak: Marking a reference to #156 with this comment.

sys/dev_null.c
```diff
@@ -0,0 +1,61 @@
...
+      len = PAGESIZE;
+    error = uiomove((void *)junk_page->vaddr, len, uio);
+  }
+  return error;
+}
+
+static cdevsw_t dev_null_sw = {.d_open = dev_null_open,
+                               .d_close = dev_null_close,
+                               .d_read = dev_null_read,
+                               .d_write = dev_null_write};
```
rafalcieslak: `clang-format` applies fancy styles to struct initializers. Can we do something about it?

rafalcieslak:
In FreeBSD, the functions that implement a layer translating from vnode operations to cdev operations are listed [here and below](http://bxr.su/FreeBSD/sys/fs/devfs/devfs_vnops.c#1692). These functions don't do much, but they are not trivial either.

rafalcieslak:
This branch is superseded by #167.

PR #158, opened 2016-12-03
Split kernel startup into two phases

### LINKED ISSUE #156: Interrupts are enabled before threads are initialized

cahirwpz:
This PR fixes various issues with kernel startup phase. Firstly it splits previous `kernel_init` into `platform_init` and `kernel_init`. The former does various hardware dependant initialization steps and creates a thread that will execute `kernel_init` procedure. The latter does machine independent stuff, runs scheduler and never returns. Secondly makes PCI initialization use regular memory allocator. Thirdly it fixes some mistakes in sw & hw interrupts handling routines. 

The outcome of this effort is that tests run with QEMU, which might be of interest to @goniz

EDIT: The intent of this PR is to fix #156


cahirwpz:
@rafalcieslak Please review :)

rafalcieslak:
Sorry for the delay. These changes look good, I've run all tests under QEMU and they seem to work fine on my end. Since we've already discussed this PR, I don't have more questions or comments.

PR #163, opened 2016-12-06
Minor fix to how C0_STATUS is handled when transitioning from/to user mode

rafalcieslak:
Some recent changes made us never actually enter user mode, and completely invalidate `C0_STATUS` when we leave it.

### DIFF
--- a/mips/exc.S
+++ b/mips/exc.S
@@ -90,6 +90,7 @@ user_exc_enter:
 
         # Turn off FPU, enter kernel mode,
         # drop exception level and disable interrupts.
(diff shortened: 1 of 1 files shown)


mips/exc.S
```diff
@@ -90,6 +90,7 @@ user_exc_enter:
 
         # Turn off FPU, enter kernel mode,
         # drop exception level and disable interrupts.
+        mfc0    $t0, C0_STATUS
```
cahirwpz: Oh... wow... how did I miss that?!

Issue #161, opened 2016-12-06
Is it okay to tlb_print in tlb_exception_handler?


rafalcieslak:
Currently, in `void tlb_exception_handler(exc_frame_t *frame)` we do:

```c
log("%s at $%08x, caused by reference to $%08lx!", exceptions[code],
    frame->pc, vaddr);
tlb_print();
```

On numerous occasions already this has caused me problems with `tlb_print` looping infinitely (recursively?). It seems as if a tlb exception was triggered while printing. Could this be the case? From my experience, the address referenced (`vaddr`) in such troublesome case is usually in range `0xc*******`. Isn't this where we keep our page table?

I've ignored this problem until today, when I found out that removing `tlb_print` from `tlb_exception_handler` apparently fixes the issues I've experienced - the tlb gets refilled, and any further data access appears correct.

So should we strictly abstain from printing tlb status in the exception handler, or is there something I'm missing?

cahirwpz:
Can you explain how to reproduce this bug ?

rafalcieslak:
I can't exactly pinpoint the problem, so the test case I can provide is not simple. You need to `malloc` 6MiB (as a userspace program). For some reason it has to be exactly 6MiB, many other sizes I've tested do not trigger as many bugs as 6MiB - in fact, I've been using this test for weeks, and it already helped me discover numerous bugs.

So to reproduce this issue, use this userspace program (or add this to the beginning of an existing one):

```c
int main(){
  int *ptr = malloc(6 * 1024 * 1024);
  free(ptr);
}
```
`malloc` uses `sbrk` and then writes some data into the allocated memory.

rafalcieslak:
With #166 merged, I hope this is fixed. If I find another way to trigger problems with TLB, I'll reopen or create a new issue.

PR #166, opened 2016-12-07
Fixed indices in tlb_exception_handler

rafalcieslak:
The issue I found in `tlb_exception_handler` is that the PTE validity check:

```c
if (is_valid(pmap->pde[index]) || is_valid(pmap->pde[index + 1])) {
```
is not accurate. It may happen (and that's probably what I observed in #161) that a mapping created ends at an even page, and some other mapping, created after the first one is accessed, starts at a corresponding odd page.

In that case the PDE will have a valid entry for `index`, but not for `index+1`. If we happen to access data that is backed by `index+1`, then the test quoted above will succeed, because `pde[index]` is valid. So we end up updating TLB with an invalid entry for `index+1`, while we should actually call `vm_page_fault`, to have `pde` updated.

My proposed solution is to test not whether **any** of pages in pair has valid entry in `pde`, but to test whether **the one accessed** has. In order to do that, I use variable `index` to store the index of the page accessed, while `base_index` stores the index rounded down to an even number.

*Also, by the way: is tlb_exception_handler the right place to check `td->td_onfault`? What if `copyin/copyout` tries to access a valid address that is accessible by the user program, but just happens to be not present in TLB? Shouldn't we check for `td->td_onfault` iff `vm_page_fault` fails?*


cahirwpz:
There's something wrong with `exec` and I cannot successfully run `exec_misbehave.elf` test:

```
[pmap.c:186] Add mapping for page 00408000 (PTE at c0001020)
[exec.c:132] Processing a PT_LOAD segment: VirtAddr = 0x407020, Offset = 0x00008020, FileSiz = 0x00000000, MemSiz = 0x0000001c, Flags = 6
[exec.c:134] Exec failed: Segment p_vaddr is not page alligned
```

Please note that proper spelling is `aligned` not `alligned`!

rafalcieslak:
I think that was fixed by 7e9b8bf535b1a8d160ffe3e865eb07f6cac5a07b in `master`.

cahirwpz:
Apparently it did not solve the problem, the binary is still failing, though git history says:

```
2016-12-10 15:43 Krystian Bacławski     o [fix_tlb_pmap_indices2] {origin/fix_tlb_pmap_indices2} Handle copyin/copyout fault if everything else failed.
2016-12-07 15:01 Rafał Cieślak          o Renamed indices in tlb_exception_handler
2016-12-07 14:41 Rafał Cieślak          o Fixed indices in tlb_exception_handler
2016-12-07 08:55 Rafał Cieślak          o Fix C0_STATUS handling on transitioning from/to user mode (#163)
2016-12-06 15:19 Rafał Cieślak          o Fixed .sbss section for user programs (#160)
```

rafalcieslak:
I cannot reproduce this bug. I tested this branch in a new, clean working directory, and `misbehave` runs correctly.

rafalcieslak:
I've merged from master, to pull in the changes to the `./user` build rules - maybe you have some intermediate build files left from another branch?

cahirwpz:
Your last commit (merge) did the trick, thanks!

PR #167, opened 2016-12-09
Initial VFS interface

rafalcieslak:
I wrote down the interface we discussed today. I've also gave a closer look at how these mechanisms cooperate in FreeBSD, and I am now pretty confident to how it works overall.

This branch is not complete, I created this PR just to share the interface ASAP. There will be no way to test this branch until I re-implement `devfs` using this new interface, which I hope to do soon - so this branch will progressively grow bigger with new features - but you can ask me to stop at any point. In particular, I will probably need to reimplement `file_t` in a fashion pretty much identical to #152 (or maybe I can avoid this, temporarily, by not returning a user-accessible `file` from `VOP_OPEN`).

I also hope that by providing this very basic implementation I may help others understand the core mechanics of filesystems and mounting.


rafalcieslak:
This is taking a nice shape! And it took much less effort to get this working than I anticipated, apparently the research we did yesterday was on point.

At this point there is a filesystem-type interface present, a single filesystem type `devfs` is implemented and it gets registered during startup. Later, an instance of `devfs` gets mounted on `/dev`. Currently `/dev` provides two device implementations, namely `zero` and `null`. These are accessible by their path, as `vfs_lookup` translates a pathname into the underlying vnode, taking care of filesystem boundaries (see `./vfs.c` for example access).

`zero` and `null` may be written to and read from as vnode ops, but they do not support `open` operation - it will only be useful for #152, so I suggest we implement it later. Also none of vnodes I provided implement `readdir`, it's also unnecessary at this moment.

I'm seriously considering implementing a `embedfs` file system type - for the files we embed in kernel image, just to demonstrate filesystem type diversity. However, I don't want to inflate this branch.

rafalcieslak:
For the record, I also have a working `/dev/uart`, but I am not adding it to this branch (unless you ask me to) to keep this PR small.

Makefile.common
```diff
@@ -34,7 +34,14 @@ CPPFLAGS = -Wall -Werror -DDEBUG -I$(TOPDIR)/include
 LDFLAGS  = -nostdlib -T $(TOPDIR)/mips/malta.ld
 
 LDLIBS += -L$(TOPDIR)/sys -L$(TOPDIR)/mips -L$(TOPDIR)/stdc \
-	  -Wl,--start-group -lsys -lmips -lstdc -lgcc -Wl,--end-group
+	  -Wl,--start-group \
+        -Wl,--whole-archive \
+          -lsys \
```
rafalcieslak: `-lsys` must be linked as `--whole-archive`. Otherwise linker set entries declared in translation units that do not use the linker set are stripped from the final kernel image, because linker notices that these symbols are never referenced and therefore do not need to be provided in the final binary. Therefore set entries declared in the same translation unit as the one that is iterating over the set would be the only ones present. In order to prevent the linker from emptying `linker_set_*` sections, I ask it to provide the entire `-lsys` in the final image. There is no linker flag to fine-tune this behavior (e.g. ask to provide all symbols from a specific section).

sys/vfs.c
```diff
@@ -0,0 +1,226 @@
...
+    if (v->v_mountedhere != NULL) {
+      mount_t *m = v->v_mountedhere;
+      vnode_t *v_mntpt;
+      int error = VFS_ROOT(m, &v_mntpt);
+      vnode_release(v);
+      vnode_unlock(v);
+      if (error != 0)
+        return error;
+      vnode_lock(v_mntpt);
+      v = v_mntpt;
```
cahirwpz: Isn't `vnode_hold` missing here?

sys/vfs.c
```diff
@@ -0,0 +1,226 @@
...
+    }
+    /* Look up the child vnode */
+    vnode_t *v_child;
+    int error = VOP_LOOKUP(v, component, &v_child);
+    vnode_release(v);
+    vnode_unlock(v);
+    if (error)
+      return error;
+    vnode_lock(v_child);
+    v = v_child;
```
cahirwpz: Ditto. Possibly missing `vnode_hold`.

sys/vnode.c
```diff
@@ -0,0 +1,54 @@
...
+void vnode_hold(vnode_t *v) {
+  assert(mtx_owned(&v->v_mtx));
+  ++v->v_refcnt;
+}
+
+void vnode_release(vnode_t *v) {
+  assert(mtx_owned(&v->v_mtx));
+  if (--v->v_refcnt == 0) {
+    /* Ignore the mutex here, we are the only reference to this vnode left */
+    kfree(vnode_pool, v);
```
cahirwpz: What if there's a thread waiting (`vnode_lock`) on the mutex?

include/vnode.h
```diff
@@ -0,0 +1,103 @@
...
+  /* Type-specific fields */
+  union {
+    mount_t *v_mountedhere; /* The mount covering this vnode */
+  };
+
+  int v_refcnt; /* Reference count */
+  mtx_t v_mtx;
+} vnode_t;
+
+typedef struct vattr {
```
cahirwpz: This maps well onto attributes that can be extracted from cpio archives.

include/mount.h
```diff
@@ -0,0 +1,100 @@
...
+  vfs_vget_t *vfs_vget;
+  vfs_init_t *vfs_init;
+} vfsops_t;
+
+/* Description of a filesystem type. There is one instance of this struct per
+ * each filesystem type in the kernel (e.g. tmpfs, devfs). */
+typedef struct vfsconf {
+  char vfc_name[VFSCONF_NAME_MAX]; /* Filesystem type name */
+  vfsops_t *vfc_vfsops;            /* Filesystem operations */
+  int vfc_refcnt; /* Number of mounted filesystems of this type */
```
rafalcieslak: Note that this is *not* a reference count, and naming it `refcnt` may be confusing. This field is used to count the number of filesystem instances of this type - the number of actual references may be higher. In particular, when a file system type is registered but not mounted anywhere, this counter stays at 0 - even though VFS does have a reference to this `vfsconf`. Also since there is no need to ever free `vfsconfs`, a reference counter wouldn't be useful.

sys/vfs.c
rafalcieslak: My original intention was that `VFS_ROOT` and `VOP_LOOKUP` should `hold` the vnode before returing it to the caller - since that's where the reference is copied (depending on implementation, a `VOP_LOOKUP` may return a reference without copying it, in that case it would not need to `hold` it again). For some reason, I forgot to implement that! Do you like this approach, or would you rather manually `vnode_hold(v_mntpt)` here?

sys/vnode.c
rafalcieslak: Can it happen? Such other thread would need to have another reference to the same vnode, before it could lock it (note that the ref counter should be increased immediately when a pointer to vnode is copied). Thus in such scenario the ref counter won't reach zero, and we will not free the vnode.

sys/vfs.c
cahirwpz: My expectations would be that vnodes returned by `VFS_ROOT` and `VOP_LOOKUP` won't get recycled till we call `vnode_release`, i.e. first option is ok.

include/mount.h
cahirwpz: Ok then, please`s/vfc_refcnt/vfc_mountcnt/`.

sys/vnode.c
cahirwpz: Perhaps locking strategy you're trying to implement is unclear to me and requires some comment. Is following scenario possible:

```c
T1 {
  /* T1 intention is to release v */
  vnode_lock(v);
  /* assume v->refcnt == 1 */
}
T2 {
  /* T2 intention is to obtain v and perform some operations on it */
  vnode_lock(v); /* goes to sleep */
}
T1 {
  vnode_release(v);
}
T2 {
  /* crash ? */
}
```

rafalcieslak: The way I want this to work is to ensure that each of these threads has their own reference to `v`, thus `v->refcnt` would be 2 in this scenario. Threads can't just take a copy of the pointer used by another thread. Any transfer of pointers to vnodes **must** involve `vnode_hold`. [The only exception are global vnodes, such as `vfs_root_vnode`, but these are never freed as their reference counter will never reach zero.]

cahirwpz: FYI NetBSD's [vnode](http://bxr.su/NetBSD/sys/sys/vnode.h#141) structure contains [mutex(9)](http://netbsd.gw.com/cgi-bin/man-cgi?mutex+9+NetBSD-current), [condvar(9)](http://netbsd.gw.com/cgi-bin/man-cgi?condvar+9+NetBSD-current) and [rwlock(9)](http://netbsd.gw.com/cgi-bin/man-cgi?rwlock+9+NetBSD-current). While FreeBSD's one uses only [mutex(9)](https://www.freebsd.org/cgi/man.cgi?query=mutex&sektion=9) and [lock(9)](https://www.freebsd.org/cgi/man.cgi?query=lock&sektion=9).

cahirwpz:
If you don't find my last commit disturbing, I'm ready to accept the PR.

rafalcieslak:
Looks good! One detail though: can we have a comment in `vnode_unref` stating that (or why) we're not `free`ing vnodes yet - just so that it wouldn't confuse other readers (or us in the future)?

Issue #168, opened 2016-12-09
Import minimal implementation of linker sets


cahirwpz:
Please import linker sets either from [FreeBSD](http://bxr.su/FreeBSD/sys/sys/linker_set.h) or [DragonFly](http://bxr.su/DragonFly/sys/sys/linker_set.h) – whatever you'll find simpler and cleaner. If you need to introduce gcc specific variable attributes please wrap them up and put at the beginning of [common.h](https://github.com/cahirwpz/mimiker/blob/master/include/common.h#L14). Perhaps you'll have to modify linker script as well. Please provide a test that will present how to use linker sets.

PR #170, opened 2016-12-09
Linker sets

rafalcieslak:
Very simple and straightforward implementation of linker sets, based on DragonFly BSD.

ref: #168

### DIFF
--- a/malta.ld
+++ b/malta.ld
@@ -26,7 +26,7 @@ SECTIONS
     . = ALIGN(4096);
   } : text
 
-  .data ADDR(.text) + SIZEOF (.text):
+  .data :
   {
     __data = ABSOLUTE(.);
     _gp = .;
@@ -37,7 +37,7 @@ SECTIONS
     __edata = ABSOLUTE(.);
   } : data
 
-  .bss ADDR (.data) + SIZEOF (.data) (NOLOAD) :
+  .bss :
   {
     __bss = ABSOLUTE(.);
     *(.sbss .scommon)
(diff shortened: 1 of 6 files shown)


malta.ld
```diff
@@ -26,7 +26,7 @@ SECTIONS
     . = ALIGN(4096);
   } : text
 
-  .data ADDR(.text) + SIZEOF (.text):
+  .data :
```
rafalcieslak: The reason for this change is that if we "manually" compute section locations, the linker heuristic responsible for placing orphaned sections decides to place them after `.text`, so such section and our `.data` overlap. The link sets need to be an orphaned section, because only then will the linker provide `__start_*` and `__end_*` symbols.

For consistence, I also removed this address computation from `.bss`, I don't think we need to explicitly specify the address - the linker will use the correct one anyway.

cahirwpz: Looks good, I did not know about linker's behaviour w.r.t. orphaned sections. Please note that at some point we will have to investigate why FreeBSD's [ldscript.mips](http://bxr.su/FreeBSD/sys/conf/ldscript.mips) is so complex, i.e. if we're missing sth important.

cahirwpz:
Please investigate header dependency issue. Otherwise the PR is ready for merge.

malta.ld
rafalcieslak: For the same reason empty linker sets will never work correctly - but I don't think we should worry about that.

rafalcieslak:
I think the reason is pretty obvious: we do not generate dependency data for test sources. The tests only depend on `lib{mips,sys,stdc}.a`, and are rebuild when any of these subcomponents changes - so as long as `./linkerset.c` is the only user of `./include/linker_set.h`, touching the header will not rebuild the test, but once we start using that header anywhere in `sys`, the test will start rebuilding correctly. We might start emitting dep rules for test sources, but I don't think this is a big issue.

PR #172, opened 2016-12-10
Moved test sources to ./tests subdirectory

rafalcieslak:
We've been trying to do that for some time already - I decided to investigate, and here's my solution. 

What's cool about this change is that it happens to significantly simplify our makefiles! The root `./Makefile` no longer deals with any sources, and does not require meta-programming anymore. Test building in `./tests/` follows `./Makefile.common`. The root directory is not longer cluttered.

I've also fixed various bugs in Makefiles - in particular, I took the chance to (again) get parallelized builds to work - we're having more and more source files so running `make` with `-j` starts bringing benefits. Supporting `-j` builds requires carefully specified dependencies and build order, so I fixed the parts where we were wrong.

@coodie, I know you've given up working on this a long while ago, so you may be interested in how I did this.


rtc.test
```diff
@@ -1,24 +0,0 @@
-#!/usr/bin/python3
```
cahirwpz: Is there any particular reason for removal of this file ?

cahirwpz:
Please move `$(TOPDIR)/malta.ld` to `$(TOPDIR)/mips/malta.ld` as well.

rtc.test
cahirwpz: It definitely should work!

Issue #102, opened 2016-11-05
Machine independent function call construction


cahirwpz:
I believe [prepare_program_stack](https://github.com/cahirwpz/mimiker/blob/master/mips/stack.c#L35) procedure can be made platform independant. Given `exc_frame_t` and `call_ctor_t` structure (latter doesn't exist yet) one should be able to fix [thread.c:32](https://github.com/cahirwpz/mimiker/blob/master/sys/thread.c#L23). Proposal follows:

```c
  call_ctor_t call;
  call_init(&call, td->td_kframe);
  for (int i = 0; i < n; i++)
    call_add_argument(&call, va_arg(ap, reg_t));
  call_finalize(&call);
```

cahirwpz:
This could be an extension of proposed machine dependent stack handling API:
```c
typedef struct user_stack {
  vaddr_t us_user_bottom; /* UVA of stack bottom */
  void *us_limit; /* the highest KVA of stack bottom (constant) */
  void *us_bottom; /* KVA of stack bottom (grows down) */
  void *us_top; /* KVA of stack top (constant) */
  bool us_finalized; /* stack is properly aligned and cannot grow anymore */
} user_stack_t;

user_stack_t *usrstk_alloc(vaddr user_bottom, size_t capacity);
void usrstk_free(user_stack_t *us);

/* Stack is constructed from top to bottom manner
 * \return ENOMEM if stack is full
 * \return EBUSY if stack is finalized */
int usrstk_reserve_string(user_stack_t *us, char ** /*out*/ str_p, size_t len);
int usrstk_reserve_ptr(user_stack_t *us, vaddr_t * /*out*/ kva_p);
int usrstk_push_int(user_stack_t *us, int value);

/*! \brief Calculate UVA from KVA between us_bottom and us_top.
 * \note Assumes that stack will not grow anymore. */
vaddr_t usrstk_user_addr_of(user_stack_t *us, void *ptr);

int usrstk_copy(user_stack_t *us);
void usrstk_finalize(user_stack_t *us);
```

cahirwpz:
I don't think we have a use scenario for this interface anymore.

PR #173, opened 2016-12-10
Fixed exec arg alignment and argv[0]

rafalcieslak:
- Fixed arguments in `./exec.c` so that we follow the convention of passing the program name as the first argument.
- Fixed alignment of arguments on the program stack when the number of arguments is even.


cahirwpz:
Does it make sense to leave only machine dependent bits in `mips/stack.c` and preprare initial program stack in machine independent manner? Please have a look at #102

rafalcieslak:
I'm worried this may bring little benefit. The problem I see is that we need to know all arguments in advance, to correctly figure out where to place the first argument. Therefore such `call_add_argument` (at least the MIPS variant) wouldn't do anything besides storing the argument, and `call_finalize` would do exactly everything `prepare_program_stack` does now.

cahirwpz:
What about a machine dependent routine that knows how to put a (1) string, (2) pointer, (3) integer onto the stack ?

PR #176, opened 2016-12-10
Simplify mutex implementation, add recursive capability.

cahirwpz:
@rafalcieslak, @coodie Please review.

I've noticed some locking patterns while reviewing #167 that made me wondering if it's right time to add recursive mutexes. This is my first attempt. I've completely got rid of `atomic_cmp_exchange ` as we run whole OS on single core - critical section will do for now.

### DIFF
--- a/include/mount.h
+++ b/include/mount.h
@@ -1,6 +1,7 @@
 #ifndef _SYS_MOUNT_H_
 #define _SYS_MOUNT_H_
 
+#include <common.h>
 #include <queue.h>
 #include <mutex.h>
 
--- a/include/mutex.h
+++ b/include/mutex.h
@@ -6,8 +6,8 @@
 
 typedef struct thread thread_t;
 
-#define MT_DEF 0
-#define MT_RECURSE 1
+#define MTX_DEF 0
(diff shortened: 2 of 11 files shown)


rafalcieslak:
This looks good to me, but I'm not the right person to review this. I know you've spend a very long time with @coodie polishing mutex implementation, so replacing it with a major rewrite without reviewing it just as hard as the previous version would invalidate your efforts. But let's wait for @coodie's input on this.

As a side-note, can we keep the `mtx_islocked`? In some operations I would prefer to `assert(mtx_islocked)` instead of locking recursively. For example, anyone who calls `vnode_hold` (from #167) will want to modify the vnode, so by using such assertion in `vnode_hold` we make it much less likely to accidentally forget locking the mutex. This assertion is a common pattern in FreeBSD, and is used to ensure that the caller remembered to lock the appropriate mutex.

Also my suggestion is that in case this PR takes a longer while to review properly, we shouldn't wait for it to get merged in order to work on #167. The impact of recursive mutices on #167 is quite low, can be easily added at any time, and we'll have to correct the way we use locks in the entire source anyway.

PR #179, opened 2016-12-12
embedfs

rafalcieslak:
~**WARNING**: This branch uses #178 as a prerequisite, and therefore it pulls all changes introduced by #178. It will wait until #178 is merged.~

To go in pair with #177, this branch demonstrates what does it take to implement a new filesystem type! This one is not particularly sophisticated, it only implements `read` and comes with no directories - but the implementation is very easy to follow.

This file system provides convenient access to files embedded in the kernel image. We currently use this mechanism for exposing user programs to `exec`. This branch makes such data accessible as read-only files residing in `/embed`. Adding a new embedded file requires adding it's corresponding `*.eo` to `LD_EMBED` in `Makefile.common`, and declaring anywhere in a source file:
```c
EMBED_FILE("filename.xyz", filename_binary)
```
where `filename_binary` is the symbol linker used for marking the file. Such file can be then accessed as a vnode with a `vfs_lookup("/embed/filename.xyz", &v)`.

We haven't discussed this proposal much, so I don't know if we even want to merge such filesystem. I reckon it will be quite useful to *have some data in the filesystem* before we have a working ramdisk. We may want to link `exec` with `embedfs` later on, so that it uses VFS to access program .elf images!

Regardless of whether we're interested in applying this PR or not, I will be using this filesystem for my own experiments, and it serves as a nice demo of what we've been working on during the weekend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 *.ko
 *.a
 
+# Embeddables as object files
+*.eo
+
 # Executables
 *.elf
 *.uelf
--- a/Makefile.common
+++ b/Makefile.common
@@ -43,11 +43,17 @@ LDLIBS += -L$(TOPDIR)/sys -L$(TOPDIR)/mips -L$(TOPDIR)/stdc \
         -lgcc \
       -Wl,--end-group
 
-# Files that need to be embedded alongside kernel image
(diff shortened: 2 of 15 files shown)


rafalcieslak:
With `cpio` support in place, this branch is no longer relevant or useful.

PR #181, opened 2016-12-15
Changes to: cond_var, sleepq, tests

czapiga:
sleepq interface and implementation changes
added condition variables interface and implementation
added producer_consumer test
removed old sleepq test


czapiga:
Seems like our hiding of condvar struct implementation didn't go so well

PR #183, opened 2016-12-19
Add initrd section to linker script

coodie:
There are few problems with the way the OVPSim maps initial ramdisk to kernel space. Looking at the code of SmartLoader, initial ramdisk is always placed at the end of .data - or in general at the last section with PT_LOAD attribute. We could probably hack around and try to place ramdisk behind .bss section, and this could be easily done with setting .bss to PT_LOAD, but we don't want to do that.

There are few solutions I could come up with to this problem:
- Copy ramdisk to other place before .bss is zeroed. This is a little bit more difficult.
- Maybe PT_LOAD .bss isn't that bad at all.
- Add non-PT_LOAD section containing fixed amount of space for ramdisk. This isn't that bad for now, because we won't need big ramdisks for some time probably, it is extremly easy to add and requires no changes to existing code.

### DIFF
--- a/include/malloc.h
+++ b/include/malloc.h
@@ -21,7 +21,6 @@
  */
 void kernel_brk(void *addr);
 void *kernel_sbrk(size_t size) __attribute__((warn_unused_result));
-void *kernel_sbrk_shutdown();
 
 /*
  * General purpose kernel memory allocator.
--- a/include/physmem.h
+++ b/include/physmem.h
@@ -3,19 +3,30 @@
 
 #include <vm.h>
 
+typedef struct pm_seg pm_seg_t;
+
 /* Platform independant initialization of physical memory manager. */
 void pm_init();
 
+/* Calculates the size of data structure that should be allocated to describe
+ * segment of @size bytes. */
+size_t pm_seg_space_needed(size_t size);
 /*
- * This adds segment to be managed by the vm_phys subsystem.
- * In this function system uses kernel_sbrk function to allocate some data to
- * manage pages. After kernel_sbrk_shutdown this function shouldn't be used.
+ * Initialize data structure describing physical memory segment.
+ *
  * Note that this system manages PHYSICAL PAGES. Therefore start and end,
- * should be physical addresses. vm_offset determines initial virt_addr
- * for every page allocated from system. All pages in this segment will have
- * their default vm_addresses in range (start + vm_offset, end + vm_offset).
+ * should be physical addresses. @offset determines initial virtual address
+ * offset for pages, i.e. all pages in this segment will be visible under
+ * addresses (start + offset, end + offset) by default.
  */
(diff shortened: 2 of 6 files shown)


cahirwpz:
Please do not modify the linker script. Use the same [technique](https://github.com/cahirwpz/mimiker/blob/master/sys/exec.c#L13) we're using to embed ELF files into kernel image. Please ask @rafalcieslak how to use it.

coodie:
I could try embed ramdisk inside kernel, but do we really want to recompile the kernel everytime we want to use other ramdisk?

cahirwpz:
You don't need to compile whole kernel. Just perform linking once again. But that has to be done in both cases – i.e. your original solution and ramdisk embedding, AFAIUC.

coodie:
We don't have to recompile or relink the kernel in case of this solution. We just have 20kb linker space for ramdisk. 

However there is better solution to that: move bss section before data. Then ramdisk would be placed after data (as it always is in the case of ovpsim --ramdisk command). There is one problem however, we need some early allocation mechanism and currently we do this by extending .bss section. But what is the problem of allocating by extending the pointer to the end of ramdisk?

Note that all of these problems are caused by impossibility of moving ramdisk around in the kernel from ovpsim command line. I think in general we want to be able to run any program in user space without having to compile it into kernel (as it is currently), but rather put it into ramdisk and then run kernel with ramdisk given.

rafalcieslak:
I completely agree that the ramdisk should be provided separately, not embedded into the kernel image. We will want to configure the ramdisk independently from the kernel. That is one of the most useful features of a ramdisk: one can modify it without touching kernel sources. I imagine eventually we'll be using a single kernel image, and a ramdisk with `init` for PID 1.

Just in case you missed the point of this PR: it's **not** trying to place the ramdisk data in the linker script. What it is supposed to do is to reserve some free space between `.data` and `.bss` where `SmartLoaderLinux` will place the ramdisk provided with `initrd` argument.

cahirwpz:
What about moving ramdisk to the end of memory before we clear `.bss`? Then at `pmap` initalization phase we can reserve those pages.

cahirwpz:
Another idea: could we add zero sized PT_LOAD segment a few megabytes after end of `.bss` section?

coodie:
`Another idea: could we add zero sized PT_LOAD segment a few megabytes after end of .bss section?` That's also great idea, but we need to allocate memory somehow :).

cahirwpz:
Could you check final size of `.bss` section by printing `sbrk.end` address in `kernel_sbrk_shutdown`?

rafalcieslak:
Here you go. At start:
```
sbrk.ptr = 0x801aea14
sbrk.end = 0x803aea14
```
At  shutdown: 
```
sbrk.ptr: 0x8032eb14
sbrk.end: 0x8032f000
```
These results are for current `master`.

PR #184, opened 2016-12-22
Add cpio based ramdisk

coodie:
**THIS PR IS NOT DONE YET!**

This PR is created to show problems associated with ramdisk implementation and current state of work. Comment secition under PR also creates some space for discussion.
- I found it is way much easier to decode crc formatted cpio archive (because in classic cpio bytes are octal), however there are few issues  with paddings so i had to look at cpio source code and play a bit to be able to decode cpio correctly (still not 100% sure it works, but I tested it few times and couldn't find non-working version). Just pass --format=crc when running cpio command.
- I did some kernel args parsing, to be thrown away layer after @mkaim implements this.
- There is better way to load ramdisk into the kernel, rather than allocate fixed size space between .data and .bss.

Almost everything is done. Missing stuff:
- Slight refactor (at least running clang format).
- vnode_readdir (to be done in another PR)
- vnode_open, this one however depends on file descriptors (to be done in another PR)


rafalcieslak:
> I did some kernel args parsing, to be thrown away layer after @mkaim implements this.

Congratulations, you are the third person to implement this independently.

@cahirwpz Could you please write down (as a new issue) some minimum requirements you want our `sysctl` implementation to fulfill? I suppose making a working stub should be of highest priority now.

sys/initrd.c
```diff
@@ -0,0 +1,193 @@
...
+    kprintf("c_uid: %d\n", stat->c_uid);
+    kprintf("c_gid: %d\n", stat->c_gid);
+    kprintf("c_nlink: %zu\n", stat->c_nlink);
+    kprintf("c_mtime: %ld\n", stat->c_mtime);
+    kprintf("c_filesize: %ld\n", stat->c_filesize);
+    kprintf("c_dev_maj: %ld\n", stat->c_dev_maj);
+    kprintf("c_dev_min: %ld\n", stat->c_dev_min);
+    kprintf("c_rdev_maj: %ld\n", stat->c_rdev_maj);
+    kprintf("c_rdev_min: %ld\n", stat->c_rdev_min);
+    kprintf("c_namesize: %u\n", stat->c_namesize);
```
rafalcieslak: Travis still uses the `mips-mti-elf` toolchain, which as a different definition of `size_t`. To avoid issues with printing `size_t`s, use the `%zu` format.

include/cpio.h
```diff
@@ -1,58 +1,73 @@
...
-#define C_ISVTX		0001000
-#define C_ISDIR		0040000
-#define C_ISFIFO	0010000
-#define C_ISREG		0100000
-#define C_ISBLK		0060000
-#define C_ISCHR		0020000
-#define C_ISCTG		0110000
-#define C_ISLNK		0120000
+/* Extended cpio format from POSIX.1.
+   Copyright (C) 1992, 2005, 2007, 2010, 2014-2015 Free Software
```
cahirwpz: Do not import GPL licensed source files into our project! MIT & BSD licenses are ok, though

cahirwpz:
I created archive with following command:

```
find . -depth -print | cpio -H crc -ov > ../initrd
```

... where directory structure is listed below:

```
empty_file
file1
file2
file3
file4
just_another_dir/file6
some_dir/nested_dir
some_dir/nested_dir/file5
some_dir/nested_dir/message
```

... and I was not able to make the test pass by running `launch --ramdisk initrd test=ramdisk`. Could you provide a script that creates a ramdisk for testing purposes?

coodie:
I commited everything and I think this is ready to be merged with master.

There is also initial ramdisk for testing purposes in tests/initrd.cpio.

Here is command which works (yours probably works too):

find -depth -print | cpio --format=crc -o > ../initrd.cpio

Here is how you launch this:
./launch --ramdisk tests/initrd.cpio test=ramdisk

Note that for now initrd is mounted in /initrd/ which means that every path of the directory you've put into ramdisk is going to start with /initrd/ inside kernel.

sys/initrd.c
```diff
@@ -0,0 +1,246 @@
...
+  uid_t c_uid;
+  gid_t c_gid;
+  dev_t c_rdev;
+  off_t c_size;
+  time_t c_mtime;
+
+  const char *c_path;
+  void *c_data;
+  
+  vnode_t *c_vnode;
```
cahirwpz: An object pointed by `c_vnode` may be freed without notice. Because of that, inside `cpio_node` there must be no references to objects other than another `cpio_node`.

cahirwpz:
Please build tree of `cpio_node`'s as we discussed during our last meeting. Note that each archived  directory has corresponding `cpio_node` structure. You'll have to extend `cpio_node` structure with two extra fields:

```c
cpio_list_t c_children; /* head of list that stores direct descendants */
TAILQ_ENTRY(cpio_node) c_siblings; /* nodes that have the same parent */
```

coodie:
@cahirwpz 

I've done tree building, what is left to be done? Would you like me to implement remaining operations? (vnode_open, vnode_readdir)

cahirwpz:
Please write a kernel-space test that will use VOP interface to perform operation similar to `find / -ls` command.

sys/initrd.c
```diff
@@ -0,0 +1,279 @@
...
+    ai++;
+    while (*ai && *ai != '/')
+      ai++;
+    if (*ai == '\0')
+      return true;
+    return false;
+  }
+}
+
+void initrd_construct_c_name(cpio_node_t *cn) {
```
cahirwpz: Remember to declare function as `static` if you don't want to expose them to other modules in the kernel.

sys/initrd.c
```diff
@@ -0,0 +1,279 @@
...
+  int c_path_len = strlen(cn->c_path);
+  cn->c_name = cn->c_path + c_path_len;
+  while (cn->c_name != cn->c_path) {
+    if (*cn->c_name == '.' || *cn->c_name == '/') {
```
cahirwpz: Could you add a comment that explains what do you do here?

sys/initrd.c
```diff
@@ -0,0 +1,279 @@
...
+    if (!read_cpio_header(&tape, node) ||
+        strcmp(node->c_path, CPIO_TRAILER) == 0) {
+      kfree(mp, node);
+      break;
+    }
+    TAILQ_INSERT_TAIL(&initrd_head, node, c_list);
+  }
+}
+
+static bool is_direct_descendant(const char *A, const char *B) {
```
cahirwpz: Hmm... could you use [string](http://netbsd.gw.com/cgi-bin/man-cgi?string) functions from standard library to perform this task - there's a lot of string processing magic below that is difficult to read.

If one you'd like to use is not in the kernel please feel free to take one from NetBSD / FreeBSD - I can help you with that.

sys/initrd.c
```diff
@@ -0,0 +1,279 @@
...
+  const char *ai = A, *bi = B;
+  /* if B is empty or dot, A is of form entry_name where
+   * entry_name has no "/" character */
+  if (*bi == '\0' || *bi == '.') {
+    if (*bi == '.')
+      bi++;
+    while (*ai && *ai != '/')
```
cahirwpz: [strchr](http://netbsd.gw.com/cgi-bin/man-cgi?strchr+3) ?

sys/initrd.c
```diff
@@ -0,0 +1,279 @@
...
+      ai++;
+    if (*ai == '\0')
+      return true;
+    return false;
+  }
+  /* order to check if A is direct descendant of B
+   * it is enough to check if A is of the form "B/entry_name"
+   * where entry_name does not contain "/" character */
+  else {
+    while (*ai && *ai == *bi) {
```
cahirwpz: [strncmp](http://netbsd.gw.com/cgi-bin/man-cgi?strcmp+3) ?

coodie:
Won't bash be better, easier and more readable for generating directories? It's basically a bunch of `mkdir`s and `echo`s... in python it's actually more difficult. And I doubt we are going to run this on windows.

cahirwpz:
Bash script is fine. Indeed, Python may be an overkill for the task.

run_tests.py
```diff
@@ -111,6 +112,8 @@ def test_seed(seed, sim='qemu', repeat=1, retry=0):
     if args.ovpsim:
         sim = 'ovpsim'
 
+    # Build initial ramdisk before running tests
+    subprocess.call('./build_cpio.sh')
```
rafalcieslak: Are you certain this is what you wanted to do? This way the preparation of the ramdisk isn't a part of the build process, but becomes a testing feature instead (...?). It will regenerate the ramdisk every time the test runner script is used, even if no changes were done. It will **not** regenerate the ramdisk when launching the kernel normally (`./launch ...`). I suppose the `./build_cpio.sh` script should be referenced in the main `Makefile`, so that it becomes a part of the build process, ramdisk image gets regenerated on demand etc.

launch
```diff
@@ -141,6 +145,8 @@ if __name__ == '__main__':
                               (', '.join(sim_choices.keys()), sim_default)))
     parser.add_argument('-t', '--test', action='store_true',
                         help='Use current stdin & stdout for simulated UART.')
+    parser.add_argument('--ramdisk',
+                        help='Give initial ramdisk to OVPSim. '),
```
rafalcieslak: Would it be difficult to add support for passing ramdisk to QEMU as well? QEMU seems to support `-initrd file` flag, and [this line of malta-specific QEMU code](https://github.com/qemu/qemu/blob/a1cf5fac2b929ffa2abd1285401f2535ff8c6fea/hw/mips/mips_malta.c#L869) suggests that it also uses `rd_start` and `rd_size`.

coodie:
For some non-obvious reason ramdisk doesn't work with Qemu (there is wrong magic number). Also, qemu gives rd_start as 64 bit string, meaning that I have change parsing algorithm (not a big problem, but pollutes code anyway). This is probably going to take me while to figure this out. I'd merge this PR at this moment, and move adding qemu support to other branch.

launch
cahirwpz: Right! Otherwise we cannot run tests on Travis, that require ramdisk.

@coodie Please make sure ramdisk works with QEMU as well!

rafalcieslak:
There are some minor problems with the build process, but please don't focus on them. I've came up with a very elegant and extensive solution which will let us nicely integrate ramdisk content definition and gathering with the build process, but it will require a lot of makefile meta-programming. So let's just make sure the ramdisk is build anyhow (and it is already), and we'll improve the build process soon afterwards - this branch is waaaaay overdue anyway. I'll either propose a solution in the next two days, or will discuss it during the next meeting.

One significant problem I've noticed concerns launching the kernel without any ramdisk - `parse_rd_start` crashes as it is given a null pointer, and it tries to `strlen` it. I think it should at least crash gracefully with a message informing that the ramdisk is missing. However, I don't expect we'll be starting the kernel without a ramdisk at all once this is merged, so an `assert` ensuring that there is a ramdisk provided should be fine too.

Also I don't expect we'll be using multiple ramdisks for different purposes, so hardcoding the one we use into the `launch` script so that the kernel **always** has a ramdisk seems viable - and it would also solve the above problem.

sys/initrd.c
```diff
@@ -0,0 +1,266 @@
...
+void ramdisk_init() {
+  unsigned rd_size = ramdisk_get_size();
+
+  TAILQ_INIT(&initrd_head);
+  kmalloc_init(mp);
+  kmalloc_add_pages(mp, 2);
+
+  if (rd_size) {
+    initrd_ops.v_lookup = initrd_vnode_lookup;
+    initrd_ops.v_read = initrd_vnode_read;
```
cahirwpz: `v_readdir` and `v_getattr` are missing, but this PR is large enough to ignore this issue for now.

cahirwpz:
I accept this PR on a premise that you'll implement `v_open`, `v_readdir` and `v_getattr` operations for ramdisk vnodes and test the access to ramdisk files from userspace in consecutive PR(s).

PR #191, opened 2017-01-19
Initial rwlock implementation

czapiga:
This implementation is based on condvar and mutex. In **near** future I will replace condvar and mutex by sleepqueue and critical_*() functions.


cahirwpz:
Please dig through some books, *BSD tests and internet resources to find nice use case for our `rwlock`. I want the implementation to be tested before it gets merged with the master.

PR #192, opened 2017-01-20
Changes in mutex implementation

czapiga:
I think this is important to have critical_*() around `m->m_count++` and `m->m_count--`

### DIFF
--- a/sys/mutex.c
+++ b/sys/mutex.c
@@ -17,13 +17,14 @@ void mtx_init(mtx_t *m, unsigned type) {
 }
 
 void mtx_lock(mtx_t *m) {
+  critical_enter();
   if (mtx_owned(m)) {
     assert(m->m_type & MTX_RECURSE);
     m->m_count++;
+    critical_leave();
     return;
   }
 
-  critical_enter();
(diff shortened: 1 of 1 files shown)


cahirwpz:
I don't think so. Can you prove it? Note that only the owner can set `m_owner` field to `thread_self()`.

In `mtx_lock` the program will enter first conditional scope only if the thread locked the mutex previously. Otherwise `m_owner` field will be set to something else than `thread_self()` and nobody will change it to `thread_self()` in the meantime.

Similarly `mtx_unlock` we assume explicitly (by `assert` statement) that the only thread that calls `mtx_unlock` is the one that locked it previously. 

Please note that recursive mutex state transitions look as follows:
```
{unowned} mtx_lock {owned, count=0} mtx_lock {owned, count=1} ...
... {owned, count=1} mtx_unlock {owned, count=0} mtx_unlock {unowned}
```

If you find this explanation valid, then please add it in the comment. They will be valuable for the reader if they get confused.

However... because you sowed the seeds of doubt, perhaps the implementation should be analyzed for interaction with interrupt context of owner's thread.

PR #194, opened 2017-01-20
Basic lseek and fstat interface

rafalcieslak:
Cherry-picked from the [`doom_test3`](https://github.com/rafalcieslak/mimiker/tree/doom_test3) branch, slightly updated to reflect recent changes (error return value sign, some renamed functions).

As usual, some files I edited weren't formatted correctly.

Note that as for now no filesystems support `VOP_GETATTR`, but `cpio` eventually will.

### DIFF
--- a/sys/file.c
+++ b/sys/file.c
@@ -4,6 +4,7 @@
 #include <errno.h>
 #include <mutex.h>
 #include <thread.h>
+#include <vnode.h>
 
 static MALLOC_DEFINE(file_pool, "file pool");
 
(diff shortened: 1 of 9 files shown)


sys/file.c
```diff
@@ -68,8 +69,11 @@ static int badfo_close(file_t *f, struct thread *td) {
...
-fileops_t badfileops = {
-  .fo_read = badfo_read,
-  .fo_write = badfo_write,
-  .fo_close = badfo_close,
-};
+static int badfo_getattr(file_t *f, struct thread *td, vattr_t *buf) {
+  return -EBADF;
+}
+
+fileops_t badfileops = {.fo_read = badfo_read,
```
cahirwpz: I could not find a way to teach `clang-format` to break line after brace if structure definition does not fit single line. Perhaps you know how to fix it.

PR #195, opened 2017-01-26
Richer tests and CI support

rafalcieslak:
This branch introduces better kernel test support:

- Canonical failure/success messages for easy grepping over kernel output
- Per-test flags: `KTEST_FLAG_NORETURN`, `KTEST_FLAG_DIRTY`, `KTEST_FLAG_USERMODE`.
- Automatically running all tests (those which can be tested one after another) when command-line argument `test=all` is present
- `ktest_assert` macro and `KTEST_FAILURE` return code
- Displaying test name if it has failed
- When running in all tests mode and a test fails, the full list of tests that were ran previously is displayed, with the failing test marked

This branch is not finished yet, I'm working on shuffling the test list so that the tests are ran in an arbitrary order. The shuffle RNG will be initialized with seed passed as command-line argument, and if a test fails, the seed will be displayed alongside the list of tests, so that this particular test order may be easily reproduced - by reusing the seed number.

Also, it turned out that a large number of tests we use fails if some other test was previously performed - for some tests this is natural, but there are many tests that should not break kernel state, yet they do. I've temporarily marked these tests with the `KTEST_FLAG_BROKEN` flag, and I'm planing to fix them.


rafalcieslak:
Here's an example of use. I start the kernel with `./launch -t test=all seed=2`. The tail of the output is:

```
  [...]

Running test uiomove.
Assertion 'res == 1' at uiomove.c:58 failed!
[TEST FAILED]
Failure while running multiple tests.
  strtol
  linker_set
  malloc
  findspace
  uiomove  <---- FAILED
The seed used for this test order was: 2. Start kernel with `test=all seed=2` to reproduce this test case.
[ktest.c:49] Halting kernel on failed test.
```

I find this quite convenient. And this mechanism has already helped me find some bugs that were not discovered by running each test separately!

mips/pmap.c
```diff
@@ -347,7 +348,7 @@ void tlb_exception_handler(exc_frame_t *frame) {
       log("Address %08lx not mapped by any active pmap!", orig_vaddr);
       goto fault;
     }
-    if (is_valid(pmap->pde[index])) {
+    if (pmap->pde && is_valid(pmap->pde[index])) {
```
rafalcieslak: I believe **this** is the actual fix for that TLB problem I've been trying to reproduce for some longer time (see #161)!

tests/callout.c
```diff
@@ -17,7 +18,7 @@ int test_callout() {
     callout_process(i);
   }
 
-  return 0;
+  return KTEST_SUCCESS;
 }
 
-KTEST_ADD(callout, test_callout);
+KTEST_ADD_FLAGS(callout, test_callout, KTEST_FLAG_BROKEN);
```
rafalcieslak: Running this test causes a crash the next time a clock interrupt arrives. I'd rather not fix this on my own as I'm not familiar with callouts, so I've excluded this test from running in `test=all` mode by marking it as `BROKEN`.

rafalcieslak:
I've also added a `run_tests.py` script which automatically runs kernel tests using some randomly selected seeds. It reports the nature of failure as well as the failing test order (for easy reproduction), returns 0/1 exit code on success/failure. I've hooked it to `make test` target.

Here's how it works in practice:
```
$ make test
Using CC: mipsel-unknown-elf-gcc -mips32r2 -EL -g
[DEP] tests/.broken.D
[CC] tests/broken.c -> tests/broken.o
[AR] tests/broken.o tests/callout.o tests/crash.o tests/exec.o tests/exec_misbehave.o tests/exec_fd_test.o tests/linker_set.o tests/malloc.o tests/mutex.o tests/physmem.o tests/pmap.o tests/producer_consumer.o tests/rtc.o tests/sched.o tests/strtol.o tests/syscall.o tests/thread.o tests/uiomove.o tests/vm_map.o tests/vfs.o -> tests/libtests.a
[LD] Linking kernel image: mimiker.elf
Testing seed 1484965626...
Test failure reported!

Failure while running multiple tests.
  linker_set
  pmap_user
  vm
  vfs
  uiomove
  findspace
  strtol
  malloc
  thread
  pmap_kernel
  broken_test  <---- FAILED
The seed used for this test order was: 1484965626. Start kernel with `test=all seed=1484965626` to reproduce this test case.
[ktest.c:51] Halting kernel on failed test.

[thread.c:91] Thread 'main' {0x8033e208} has finished.

Makefile:38: recipe for target 'test' failed
make: *** [test] Error 1
```

And here's a successful run:
```
$ make test
Using CC: mipsel-unknown-elf-gcc -mips32r2 -EL -g
[DEP] tests/.broken.D
[CC] tests/broken.c -> tests/broken.o
[AR] tests/broken.o tests/callout.o tests/crash.o tests/exec.o tests/exec_misbehave.o tests/exec_fd_test.o tests/linker_set.o tests/malloc.o tests/mutex.o tests/physmem.o tests/pmap.o tests/producer_consumer.o tests/rtc.o tests/sched.o tests/strtol.o tests/syscall.o tests/thread.o tests/uiomove.o tests/vm_map.o tests/vfs.o -> tests/libtests.a
[LD] Linking kernel image: mimiker.elf
Testing seed 69759498...
Testing seed 2792287138...
Testing seed 257390304...
Testing seed 4157792272...
Testing seed 3604868030...
Tests successful!
```

Another nice feature is that the `ktest` framework is now capable of detecting invalid memory address correctly, so that such faults within kernel tests will be considered a proper failure instead of looping endlessly.

rafalcieslak:
As you can see, the tests are now running on Travis, and it correctly reports test failures (for example, see [this build](https://travis-ci.org/cahirwpz/mimiker/builds/195843777)). I consider this branch ready for review.

mips/pmap.c
cahirwpz: How come? We always create `pmap` with `pde` assigned to it!

cahirwpz: BTW. Yesterday, during a meeting, we identified a bug for which a quick workaround would be to flush TLB during `pmap_activate`. We've worked out a better solution, but @strzkrzysiek changes are not ready to integrate with master.

rafalcieslak: I've met with at `pmap` that had it's fields set to 0 at the point of this `if`. This happened on an (intentionally) invalid memory access into user memory, which triggered a recursive TLB exception. I'll see if I can trigger this again to investigate why was the pmap empty.

rafalcieslak: I expect this issue will never appear unless kernel misbehaves, so it does not affect this branch much. And since there are major changes to `pmap` coming soon anyway, I believe this PR is not a right place for considering this hypothesized fix - so I'll revert it for now, and possibly discuss it as a separate feature.

include/ktest.h
```diff
@@ -2,16 +2,59 @@
...
+  const char test_name[KTEST_NAME_MAX];
   int (*test_func)();
+  uint32_t flags;
 } test_entry_t;
 
-#define KTEST_ADD(name, func)                                                   \
-  test_entry_t name##_test = {#name, func};                                    \
+void ktest_main(const char *test);
+
+#define KTEST_ADD_FLAGS(name, func, flags)                                     \
```
cahirwpz: I've got confused while reading tests – I though that you have another linker set for test flags.

Couldn't we just use `KTEST_ADD` with extra argument for flags – it's just extra 3 characters?

rafalcieslak: I only kept both of these to stay compatible - I agree this is confusing.

sys/vfs.c
```diff
@@ -215,7 +215,8 @@ int vfs_lookup(const char *path, vnode_t **vp) {
       if (error)
         return error;
       v = v_mntpt;
-      vnode_ref(v);
+      /* vnode_ref(v); No need to ref this vnode, VFS_ROOT already did it for
```
cahirwpz: I'd like to see this fixed in another PR, that I'll merge immediately.

run_tests.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/python3
+import argparse
+import pexpect
+import sys
+import random
+
+N_SIMPLE = 5
+N_THOROUGH = 200
+
+parser = argparse.ArgumentParser(description='Automatically performs kernel tests.')
```
cahirwpz: Please use:

```python
if __name__ == '__main__':
  ...
```

... to capture top-level program logic.

sys/ktest.c
```diff
@@ -0,0 +1,175 @@
...
+/* Stores currently running test data. */
+static test_entry_t *current_test = NULL;
+/* A null-terminated array of pointers to the tested test list. */
+static test_entry_t **autorun_tests;
+
+int ktest_test_running_flag = 0;
+
+/* We only need this pool to allocate some memory for performing operations on
+ * the list of all tests. */
+static MALLOC_DEFINE(test_pool, "test data pool");
```
cahirwpz: I'd prefer the test framework not to rely on dynamic memory allocation, which is going to be test subject on its own. Could you introduce `test_entry_swap(a, b)` routine and perform `qsort` on the linker set with not-runnable tests move to the bottom of `tests` array?

cahirwpz:
This change is highly desired QA feature, I hope it'll be ready for merge with master ASAP.

sys/ktest.c
rafalcieslak: I agree the framework should not use malloc if possible - I originally suspected I'll need more memory space than just that list of pointers. However, I would suggest a different solution - a statically allocated space for the array. It doesn't have to be very big, it's just ~1 ptr per each test. The kernel will check if the current size is large enough to store all pointers (we would manually increase it if required). This way the linker set will stay `const`, array management will be much simpler, and we'll be able to e.g. run each test multiple times in a shuffled order (a feature I'll implement in a subsequent PR).

cahirwpz:
LGTM Should I merge it ?

rafalcieslak:
I believe it's ready.

PR #199, opened 2017-01-28
Various ktest fixes

rafalcieslak:
I'll be using this PR to investigate spuriously failing tests on Travis.


rafalcieslak:
The result of [this build](https://travis-ci.org/cahirwpz/mimiker/builds/196139783) suggests that QEMU did not even start within the timeout! `pexpect` received no output from the simulation at all. I've seen this happen on my local machine a few times, but it was quite rare. This sort of problem could happen when there is a daemon listening for TCP at port 8000, but I don't think this is the case in our Travis sandbox.

I'll have `run_tests.py` restart the simulation several times in case no output was received at all.

rafalcieslak:
I don't know how to interpret [this result](https://travis-ci.org/cahirwpz/mimiker/builds/196142922). It looks like the kernel hangs on `thread_switch_to`. Unsurprisingly, I am not able to reproduce this scenario on my local machine. Note that these threads are not registered in the scheduler, so this is not an unlucky preemption. I'll try to come up with a way of debugging this non-interactively on Travis - assuming I will be able to trigger this problem again.

rafalcieslak:
One problem I've identified was that most memory tests were dirty, i.e. they did not restore original state when completed. Also sometimes it just happens that QEMU takes a very long time to start, or outputs an EOF immediately after starting - I've modified `run_tests.py` so that it will retry the test (up to 5 times) when such condition is detected, and it seems to have helped!

I'll keep restarting a recent build to (hopefully) identify more problems.

rafalcieslak:
Okay, so here's a summary of changes in this branch:

 - Memory tests now restore original state.
 - If QEMU for some reason fails to start at all, `run_tests.py` will retry the test several times.
 - Simulator process is now forcefully terminated when a test run completes of times out.
 - Printing `[TEST_PASSED]` now happens in a critical section, to avoid begin preempted whilst printing.

I am not entirely certain this fixes the problems we've observed, but it certainly fixes **some** problems. I've restart the recent build on Travis 10 times already, and all test runs were successful, so maybe this fix is on point.

I consider this PR ready for review.

rafalcieslak:
Wrapping `kprintf` in a critical section is controversial, so I've disabled it for now. After all, it is very unlikely this problem would cause a false-negative.

PR #200, opened 2017-01-30
Usermode kernel tests

rafalcieslak:
~**Note: This branch is built on top of #198 and #199, so please review these two first!**~

This branch extends the `ktest` framework so that usermode programs may be treated as tests. In particular, we would like to run the `fd_test` program as proper test, I expect that we'll have many more similar tests in the future. The best way to explain the interface is to have a peek at the new version of `exec_fd_test.c`:

```c
static void test_fd_user_thread(void *arg) {
  exec_args_t exec_args;
  exec_args.prog_name = "fd_test";
  exec_args.argv = (const char *[]){"fd_test"};
  exec_args.argc = 1;

  do_exec(&exec_args);
}

static int test_exec_fd_test() {
  thread_t *user_thread = thread_create("fd_test", test_fd_user_thread, NULL);
  sched_add(user_thread);
  ktest_wait_for_user_thread(user_thread, 1000);
  return KTEST_SUCCESS;
}
```

The `ktest_wait_for_user_thread` function is a blocking call which waits until `user_thread` exits, and tests its exit code - if it is non-zero, it will be reported as test failure. If 1000 ms have elapsed and `user_thread` did not finish, or if it performs invalid memory operation, it will be also treated as test failure.

The implementation is very simple and it uses a combination of a conditional variable in `thread_t` and a removable `callout`.

~~Thanks to this feature, I've been able to find a specific bug in `vnode` reference counting - I've fixed it in 62baf712d9719. I'll cherry-pick this commit in a separate PR so that it can be merged quickly.~~ Already merged.


sys/ktest.c
```diff
@@ -187,3 +190,33 @@ void ktest_main(const char *test) {
...
+}
+
+void ktest_wait_for_user_thread(thread_t *td, int timeout_ms) {
+  callout_t timeout_callout;
+  if (timeout_ms > 0)
+    callout_setup_relative(&timeout_callout, timeout_ms, ktest_usermode_timeout,
+                           NULL);
+
+  sleepq_wait(td, "usermode ktest");
+  int status = td->td_ktest_status;
```
cahirwpz: If `td_ktest_status` plays a role of exit code then just name it this way (i.e. `td_exitcode`) and make it a regular member of thread structure (and not `ktest` specific).

cahirwpz:
Just a random question... couldn't we rename `ktest_usermode_exit` to `thread_cancel` or `thread_exit`; `ktest_wait_for_user_thread` to `thread_join_timeout`; and make those belong to thread interface?

rafalcieslak:
Funny thing, this branch got trimmed down from 250 diff lines to barely 10.

It's should be simple to review it now :-)

PR #201, opened 2017-01-30
Fixed vnode ref counting on vfs_open and vnode_generic_close

rafalcieslak:
This is a fairly simple fix for vnode ref counts that I've discovered while working on #200.


rafalcieslak:
Note: This travis check failure is most likely fixed by #199.

PR #207, opened 2017-03-17
Implemented ktests with a random integer

rafalcieslak:
Yet another ktest feature: Tests added with `KTEST_ADD_RANDINT` will be passed a random (deterministic) integer as an argument. The range [0..max] may be specified as an extra argument when registering the test.


cahirwpz:
LGTM. Please run `clang-format` on the change.

PR #208, opened 2017-03-17
Convenient ktest_assert alias

rafalcieslak:
I've recently noticed @czapiga using `assert` when writing new tests, while he should choose `ktest_assert` instead, as this variant correctly notifies the testing framework about a failure.

We can avoid this inconvenience by defining `assert` to be an alias of `ktest_assert` when compiling sources in `./tests` directory.

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,6 +1,7 @@
 set tcp connect-timeout 30
 target remote localhost:1234
 break main
+break assert_fail
 continue
 
 source debug/ktrace.py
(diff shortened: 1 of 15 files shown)


cahirwpz:
How about redirecting regular calls to `assert` (e.g. in `pmap` module to `ktest_assert`)? I'd like the test framework to capture failed assertions in all code called by a test.

rafalcieslak:
This seems like a good idea.

```c
kprintf("Assertion .... failed");
if(ktest_test_running_flag)
  ktest_failure();
else
  panic(...);
```
As this logic requires peeking into `ktest_running_test_flag`, I'll have to implement it as a proper function and not within the `assert` macro (which should still provide condition check so that we pay the cost of a function call only if the assertion actually fails). I'm not sure which `.c` file should this implementation belong to, is it okay if I create `assert.c`?

cahirwpz:
Yes, it is ok to create `assert.c` file.

PR #211, opened 2017-03-17
Various callout fixes

rafalcieslak:
~I created this PR only so that I can debug travis test results, please ignore it for now.~

### DIFF
--- a/sys/callout.c
+++ b/sys/callout.c
@@ -62,6 +62,7 @@ void callout_stop(callout_t *handle) {
  * position.
 */
 void callout_process(realtime_t time) {
+  kprintf("Processing callout time %ld\n", (long int)time);
   int now = time % CALLOUT_BUCKETS;
   bool done = false;
 
--- a/sys/callout.c
+++ b/sys/callout.c
@@ -2,8 +2,8 @@
 #include <callout.h>
 
 /* Note: If the difference in time between ticks is greater than the number of
-   buckets, some callouts may be skipped and/or called out-of-order! */
-#define CALLOUT_BUCKETS 10
+   buckets, some callouts may be called out-of-order! */
+#define CALLOUT_BUCKETS 64
 
 #define callout_set_active(c) ((c)->c_flags |= CALLOUT_ACTIVE)
 #define callout_clear_active(c) ((c)->c_flags &= ~CALLOUT_ACTIVE)
--- a/sys/callout.c
+++ b/sys/callout.c
@@ -64,16 +64,17 @@ void callout_stop(callout_t *handle) {
 */
 void callout_process(realtime_t time) {
   kprintf("Processing callout time %ld\n", (long int)time);
-  int now = time % CALLOUT_BUCKETS;
-  bool done = false;
+  unsigned int current_bucket = ci.last % CALLOUT_BUCKETS, last_bucket;
+  if (time - ci.last > CALLOUT_BUCKETS) {
+    /* Process all buckets */
+    last_bucket = (ci.last - 1) % CALLOUT_BUCKETS;
+  } else {
+    /* Process only buckets in time range ci.last to time */
+    last_bucket = time % CALLOUT_BUCKETS; 
+  }
 
-  while (!done) {
-    int last = ci.last++ % CALLOUT_BUCKETS;
-
-    if (last == now)
-      done = true;
-
-    callout_head_t *head = &ci.heads[last];
+  while (1) {
+    callout_head_t *head = &ci.heads[current_bucket];
     callout_t *elem = TAILQ_FIRST(head);
     callout_t *next;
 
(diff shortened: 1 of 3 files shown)


rafalcieslak:
It appears that callout's `ci.last` may lag behind actual time quite a lot! This is because `ci.last` is incremented not until it equals `time`... but until it equals `time` **modulo** `CALLOUT_BUCKETS` (10). Therefore, it a clock interrupt arrives late (e.g. we spent a long time in a critical section), `ci.last` will be permanently lower than actual current time - and in result registering a callout with `callout_setup_relative` computes a too small `c_time`, so the callout gets triggered far too early (or even immediatelly).

And that causes a second problem. When multiple callouts from the same bucket are triggered at once, their order of execution is the same as the order of registration - and thus it is not guaranteed that action A, scheduled for a time earlier than B, will actually execute before B. So the `callout_order` test fails. (We would have to sort scheduled events before executing them, but sorting is `O(nlogn)`, and we don't want to spend much time under an interrupt!)

The `callout_stop` test fails when a clock interrupt arrives after a "bad" callout was added, but before it was removed. This should not happen, because the "bad" callout is scheduled at +10ms, which should be much more than enough. However, if `ci.last` is more than 10ms behind actual time, the "bad" callout will fire as soon as any clock interrupt arrives, which may happen way before the callout is unregistered.

The way I see it, these observations prove that our current callout implementation is based on a _completely flawed design_, and will require a major rewrite.

rafalcieslak:
Now, I can fix the `callout_process` routine so that `ci.last` is correctly updated, and thus fix the problems that appear on Travis. However, I am not sure if it is worth it: the callout implementation will be still broken due to how it calls multiple triggers in the same bucked according to registration order. And the modulo comparison between actual time and `ci.last` may cause some callouts to be ignored when processing a late (delayed) clock interrupt.

So I can provide a valid workaround for the issue we experience on Travis, but we would still have to consider callouts as *utterly broken*.

rafalcieslak:
One more note: The reason why this problem does not appear on ovpsim is because we receive clock interrupts there once every 2 milliseconds (which is lower than `CALLOUT_BUCKETS`, so the modulo comparison never lets `ci.last` drift too far away), while on qemu we receive them once per 2-20 milliseconds.

rafalcieslak:
I've cleaned up this branch so that the changes it introduces are actually useful.

Summary:
 - Rewritten the main loop of `callout_process` so that it never skips any triggers even if time delta is large, and always correctly updates `ci.last`.
 - Increased the number of callout buckets. The purpose of these buckets is such that we don't want to iterate over many registered callouts each time `callout_process` is called, and therefore we limit our search to valid buckets. Thus, the more buckets we use, the less work each `callout_process` has to do (ignoring cache behavior), and we totally can afford 64 buckets. What is important is that callouts may be triggered out-of-order only when the time delta between subsequent `callout_process` calls is greater than the number of buckets. 10 buckets was definitely too little (as I've observed occasional deltas >25ms on Travis), yet it would be pretty weird if we were to not receive any clock interrupt for more than 64ms.
- Wrapped callout tests preparation in a critical section, so that interrupts won't interfere with relative timing calculations.

Still, in case of a >64ms lag, some callouts may be triggered out-of-order. Note that this can't be solved by using an RB-tree in each bucket, since in the case of such a large time difference all buckets are browsed.

sys/callout.c
```diff
@@ -62,16 +62,17 @@ void callout_stop(callout_t *handle) {
...
-  bool done = false;
-
-  while (!done) {
-    int last = ci.last++ % CALLOUT_BUCKETS;
-
-    if (last == now)
-      done = true;
-
-    callout_head_t *head = &ci.heads[last];
+  unsigned int current_bucket = ci.last % CALLOUT_BUCKETS, last_bucket;
```
cahirwpz: `, last_bucket` confused me a bit. Please try not to not mix declaration and definition on a single line.

sys/callout.c
```diff
@@ -90,5 +91,12 @@ void callout_process(realtime_t time) {
 
       elem = next;
     }
+    if (current_bucket == last_bucket)
+      break;
+
+    current_bucket = (current_bucket + 1) % CALLOUT_BUCKETS;
   }
+
+  /* Update ci.last. */
```
cahirwpz: Well... that's obvious... could you add a comment to `last` field explaining its purpose?

rafalcieslak: Yeah, what I actually wanted to express was something along the lines of "`ci.last` was not updated above".

sys/callout.c
rafalcieslak: Would
```
unsigned int last_bucket, current_bucket = ci.last % CALLOUT_BUCKETS;
```
be better?

cahirwpz: Let's set up a rule then... multiple declarations on single line is ok, but only one definition, e.g.:

```C
unsigned int last_bucket, first_bucket;
unsigned int current_bucket = ci.last % CALLOUT_BUCKETS;
```

sys/callout.c
cahirwpz: Nonetheless when I had a look at this code I was missing the meaning of `ci.last` – hence I believe adding a comment to `ci` structure would have a positive impact on readability.

PR #185, opened 2017-01-12
[WIP] Worker thread

psie:
This is a draft of worker thread concept. Includes a use case of cleaning up after a thread that exited. Awaiting comments.

Issue #213, opened 2017-03-24
callout improvements


cahirwpz:
Callouts must be made thread-safe! We have to consider where to execute timeout callbacks. Probably high-priority worker thread is the best place.

Additionally we need following changes to the interface:

```C
void callout_schedule(callout_t *c, unsigned time);
void callout_set_func(callout_t *c, timeout_t fn, void *arg);
void callout_reset(callout_t *c, unsigned time, timeout_t fn, void *arg);
```

Please note that `callout_setup*` functions are replaced by `callout_schedule` and `callout_set_func`. Callout functions use `time` measured in milliseconds. Functions accept time relative to the time kept by `hardclock`.

For exact meaning of each function please look at [callout(9)](http://netbsd.gw.com/cgi-bin/man-cgi?callout+9+NetBSD-current).

**IMPORTANT** I do not expect timeout callbacks to be called in order!

rafalcieslak:
Callout implementation improvements are currently blocked by #216 (Worker thread).

cahirwpz:
Seems like callouts were greatly improved in c64a9ac72726999483b499b27e3f38d8ec14af10, 5ce803bc4dd0ae064b936dc8b449eec7f076d7c9. IMO at the moment callout API is in good condition.

PR #216, opened 2017-03-24
Worker thread

rafalcieslak:
This is a continuation of #185 moved to a new branch and rebased on top of current master.

New features:
 - `thread_join`
 - Taskqueues/workqueue
 - Cleaning up after a thread has exitted (`thread_delete` was unused until now)
 - Thread exit code

Again, this PR incorporates some code formatting fixes in unrelated files.

I should probably merge `workqueue.*` files into `taskqueue.*`, would you like me to do that?


rafalcieslak:
This current Travis test failure is addressed by #217.

include/thread.h
```diff
@@ -28,6 +28,7 @@ typedef struct thread {
   enum { TDS_INACTIVE = 0x0, TDS_WAITING, TDS_READY, TDS_RUNNING } td_state;
   uint32_t td_flags;           /* TDF_* flags */
   volatile uint32_t td_csnest; /* critical section nest level */
+  int td_exitcode;
```
cahirwpz: We should discuss the meaning of this field. Initial idea is to map kernel exit codes (i.e. `errno` and received signal) to negative numbers, whereas positive numbers are free to use by userspace. What do you think?

sys/taskqueue.c
```diff
@@ -0,0 +1,62 @@
...
+  mtx_lock(&tq->tq_mutex);
+  while (!STAILQ_EMPTY(&tq->tq_list)) {
+    task = STAILQ_FIRST(&tq->tq_list);
+    task->func(task->arg);
+    STAILQ_REMOVE_HEAD(&tq->tq_list, tq_link);
+  }
+  mtx_unlock(&tq->tq_mutex);
+}
+
+task_t *task_create(void (*func)(void *), void *arg) {
```
cahirwpz: We can safely remove this function. I presume that the caller will be responsible to provide memory for `task_t` structure.

sys/taskqueue.c
```diff
@@ -0,0 +1,62 @@
...
+}
+
+void taskqueue_add(taskqueue_t *tq, task_t *task) {
+  mtx_lock(&tq->tq_mutex);
+  STAILQ_INSERT_TAIL(&tq->tq_list, task, tq_link);
+  cv_signal(&tq->tq_nonempty);
+  mtx_unlock(&tq->tq_mutex);
+}
+
+void taskqueue_run(taskqueue_t *tq) {
```
cahirwpz: I think you can safely merge `taskqueue_worker_thread` into `taskqueue_run` and make it `static` function. We don't need to expose it.

include/thread.h
rafalcieslak: I suppose that would be okay, although the POSIX standard requires `waitid()` to recover the full number passed to `void exit(int status)`, which would mean we need all 32 bits for userspace exit codes.

sys/taskqueue.c
rafalcieslak: Allocating a `task_t` is one thing, but it also needs to be freed after it is called. This could be easily done from `taskqueue_run` if the task was allocated in `tq_pool` (yeah, I did not remember to do that), but if we allow other components to allocate tasks on their own, they will be also responsible for cleaning up. Thus the function called by a task should also get a pointer to itself as an argument, and the task handler would need to free the task first. That seems like a needlessly complicated interface. Having all tasks allocated in a pool managed by `taskqueue.c` would make thing simpler.

sys/taskqueue.c
rafalcieslak: If we don't want to expose `taskqueue_run`, then I will need to implement a `taskqueue_wait_until_empty` instead. This is needed for testing the taskqueue, we have to wait until all scheduled tasks were completed before we can verify test success.

sys/taskqueue.c
```diff
@@ -0,0 +1,62 @@
...
+  task_t *task;
+  mtx_lock(&tq->tq_mutex);
+  while (!STAILQ_EMPTY(&tq->tq_list)) {
+    task = STAILQ_FIRST(&tq->tq_list);
+    task->func(task->arg);
```
cahirwpz: The whole idea of worker queues is that I can quickly postpone the work to other thread. These function calls may take some time while the mutex is being hold, effectively blocking thread (or interrupt) that calls `taskqueue_add`.

sys/taskqueue.c
cahirwpz: Ok. Then leave an annotation in the header file, that one should not use the function.

sys/taskqueue.c
cahirwpz: By looking at FreeBSD's [taskqueue](https://www.freebsd.org/cgi/man.cgi?query=taskqueue) and NetBSD's [workqueue](http://netbsd.gw.com/cgi-bin/man-cgi?workqueue+9+NetBSD-current) I can see they both assume task's memory will be managed outside. While I value separation of concern I can also see valid point in your comment.

The question is if `workqueue` is right tool to cleanup after a thread is done. I need to think it through and read some sources.

sys/taskqueue.c
```diff
@@ -48,8 +48,16 @@ void taskqueue_run(taskqueue_t *tq) {
   mtx_lock(&tq->tq_mutex);
   while (!STAILQ_EMPTY(&tq->tq_list)) {
     task = STAILQ_FIRST(&tq->tq_list);
+    mtx_unlock(&tq->tq_mutex);
+
     task->func(task->arg);
```
cahirwpz: Instead of repeatedly locking and unlocking the mutex it's better to copy head structure aside and reinitialize the original list head.

cahirwpz:
Please merge `workqueue` and `taskqueue` as you mentioned in the initial comment.

cahirwpz:
Apparently we don't need to implement taskqueue to clean up after a thread exits. Please have a look at FreeBSD's [thread_reap](http://bxr.su/search?q=&defs=&refs=thread_reap&path=kern_thread.c&project=FreeBSD) – it's called from `thread_alloc` and `thread_wait`, which make sense! NetBSD provides similar procedure called [lwp_free](http://bxr.su/search?q=lwp_free&defs=&refs=&path=kern_lwp.c&project=NetBSD) which again is called in `lwp_create` and `lwp_wait`, though FreeBSD's implementation seems to be cleaner.

They both use a queue of threads to be destroyed (zombies), instead of taskqueue.

My proposal is to split this PR into one that solely focuses on `thread_join` and `thread_exit`, and other that implements taskqueue's. I'll review thread related PR first. I'm waiting for your feedback.

cahirwpz:
I have to evaluate how actually we want `taskqueue`'s to be used in the kernel. By looking at FreeBSD kernel source it's not yet apparent to me.

rafalcieslak:
We've decided to reimplement it so that all memory is allocated outside work queue interface.

rafalcieslak:
@cahirwpz Can I please get a status update on this one? Workqueues are required to implement callout improvements (#213), which are necessary for better timeout support, and `nanosleep`-like systemcalls.

cahirwpz:
@rafalcieslak This task has relatively low priority. ATM it's important for me to finish review of #271 and #276, and continue my work towards more complete driver infrastructure.

cahirwpz:
I've decoupled thread management from taskqueue. Though there's no immediate need to use this interface, I'll integrate it with the kernel. I hope it'll be a good starting point for message ports (as known from microkernels) implementation.

PR #218, opened 2017-03-25
Basic fork

rafalcieslak:
This is my most recent attempt at implementing a basic `fork()`. It almost seems too simple to work! Did I miss anything important?

This implementation shares the userspace memory between parent and child (!!!) which is, naturally, a major issue, as both threads will use the same user stack. We're not ready to perform COW yet, so I suppose I'll just have to clone the entire `td_uspace` vm_map (which is probably okay for now, but please confirm before I put effort into writing such mechanism).

I am super hyped by this branch, as this `fork` uncovered a major bugs in descriptor table management! I'll extract the fix as a separate PR soon.

Some parts of `do_fork` are very platform-dependent, but I can't come up with some elegant ways of abstracting them, so I could use some hints.

I believe this is the high time we've decided on our fork interface. Do we want a richer (`clone`-like maybe) interface? If yes, what options shall we support? Shared/cloned vm, shared/cloned descriptor table, optional vfork come to mind. Do we wish to implement such richer interface as another syscall, separate from plain simple `fork` (although they both could share implementation, just with a different set of parameters)?


cahirwpz:
I'm in the process of reading `fork()` implementation in both FreeBSD and NetBSD. In principle is rather trivial. I guess we can start with simple interface and then gradually modify it having more sophisticated requirements.

I'll have a look at platform dependant parts of fork implementation. I remember seeing [fork_trampoline](http://bxr.su/FreeBSD/sys/mips/mips/swtch.S#74) to be written in assembly.

I guess just copying `td_uspace` will do for now, although I have to look at this mechanism in *BSD kernels to make sure we don't make a mistake.

rafalcieslak:
What happens during `fork_trampoline` is pretty much the same was what we do with `user_exc_leave`.

sys/fork.c
```diff
@@ -0,0 +1,64 @@
...
+  thread_t *td = thread_self();
+
+  /* Cannot fork non-user threads. */
+  assert(td->td_uspace);
+
+  thread_t *newtd = thread_create(td->td_name, NULL, NULL);
+
+  /* Prevent preemption, so that the copied thread won't run while we're copying
+     it. */
+  critical_enter();
```
cahirwpz: You perform several steps with interrupts turned off. For now they seem to be rather cheap, however I cannot conceive of copying address space under critical section. This has to be addressed.

rafalcieslak: Actually, I think we might drop the critical section here altogether. The nature of `fork` is such that copying happens within the context of the source thread, so its state will not change while `fork` is performed!

Copying address space is different, because it may be already shared between multiple threads, so some other thread could modify address space layout while it is being copied. Using a mutex to lock access to the vm map structure will not suffice - it would protect structure integrity, but the actual contents, the data in address space, could still be modified.

We would need to pick all threads that share the copied address space, and ask the scheduler not to run them until we are done. This would have to be done under a critical section, but would take little time. (Same goes for unblocking these threads when we are done copying). Would that be feasible?

sys/fork.c
```diff
@@ -0,0 +1,57 @@
...
+     thread's kernel context to a fresh one so that it will continue execution
+     starting from user_exc_leave (which serves as fork_trampoline). */
+  ctx_init(newtd, user_exc_leave, NULL);
+
+  /* Share userspace memory with parent. TODO: (optionally?) clone user
+     memory. */
+  newtd->td_uspace = td->td_uspace;
+
+  /* Copy the parent descriptor table. */
+  newtd->td_fdtable = fdtab_copy(td->td_fdtable);
```
cahirwpz: [fdtab_copy](http://mimiker.ii.uni.wroc.pl/source/xref/mimiker/sys/filedesc.c#97) should perform a shallow copy, ie. corresponding entries from original and copy descriptor table should point to the same file. `fdtab_copy` may automatically set reference counter to 1. Hence file descriptor tables and files are now shared, we need to add some synchronization to those structures.

sys/fork.c
```diff
@@ -0,0 +1,57 @@
...
+  fdtab_ref(newtd->td_fdtable);
+
+  newtd->td_sleepqueue = td->td_sleepqueue;
+  newtd->td_wchan = td->td_wchan;
+  newtd->td_wmesg = td->td_wmesg;
+
+  newtd->td_prio = td->td_prio;
+  newtd->td_slice = td->td_slice;
```
cahirwpz: I reckon fresh process should get a full time slice.

sys/fork.c
```diff
@@ -0,0 +1,57 @@
...
+  /* Share userspace memory with parent. TODO: (optionally?) clone user
+     memory. */
+  newtd->td_uspace = td->td_uspace;
+
+  /* Copy the parent descriptor table. */
+  newtd->td_fdtable = fdtab_copy(td->td_fdtable);
+  fdtab_ref(newtd->td_fdtable);
+
+  newtd->td_sleepqueue = td->td_sleepqueue;
+  newtd->td_wchan = td->td_wchan;
```
cahirwpz: Child process starts immediately and parent is running, so those `td_wchan` and `td_wmesg` are `NULL`.

sys/fork.c
```diff
@@ -0,0 +1,57 @@
...
+
+  /* Share userspace memory with parent. TODO: (optionally?) clone user
+     memory. */
+  newtd->td_uspace = td->td_uspace;
+
+  /* Copy the parent descriptor table. */
+  newtd->td_fdtable = fdtab_copy(td->td_fdtable);
+  fdtab_ref(newtd->td_fdtable);
+
+  newtd->td_sleepqueue = td->td_sleepqueue;
```
cahirwpz: There should be exactly as many sleepqueue structures as there're threads in the system.

sys/fork.c
```diff
@@ -0,0 +1,57 @@
...
+
+  /* The new thread already has a new kernel stack allocated. There is no need
+     to copy its contents, it will be discarded anyway. We just prepare the
+     thread's kernel context to a fresh one so that it will continue execution
+     starting from user_exc_leave (which serves as fork_trampoline). */
+  ctx_init(newtd, user_exc_leave, NULL);
+
+  /* Share userspace memory with parent. TODO: (optionally?) clone user
+     memory. */
+  newtd->td_uspace = td->td_uspace;
```
cahirwpz: Please do copy whole user address-space. It's memory ineffective, but at least we get a `fork()` that has anticipated semantics.

sys/fork.c
```diff
@@ -0,0 +1,57 @@
...
+
+int do_fork() {
+  thread_t *td = thread_self();
+
+  /* Cannot fork non-user threads. */
+  assert(td->td_uspace);
+
+  thread_t *newtd = thread_create(td->td_name, NULL, NULL);
+
+  /* Clone the thread. Since we don't use fork-oriented thread_t layout, we copy
```
cahirwpz: According to POSIX.1 standard, the only thread that passes to child process is the one that called `fork()`.

sys/fork.c
rafalcieslak: I wanted to implement a `vm_map_clone` function, but (until we have more sophisticated scheduler synchronization mechanisms) the entire copying process will need to happen in a critical section. Would that be okay for now?

sys/fork.c
rafalcieslak: Setting the slice is actually irrelevant, it will be immediately overwritten by `sched_add`. And it shouldn't matter anyway, this field is not meaningful for threads that aren't running yet.

sys/fork.c
cahirwpz: Just assume we're implementing a system where there's only one thread in a process. This way you don't need to worry about synchronization.

sys/fork.c
cahirwpz: But then set it to zero, so a reader is not confused.

sys/fork.c
rafalcieslak: So that means that no threads ever share their memory space, right?

sys/fork.c
rafalcieslak: I removed this line entirely.

sys/fork.c
cahirwpz: FTTB, yes!

rafalcieslak: Actually... I don't know how to copy memory contents. Only one `vm_map` is active at a time, so I would need to copy out contents somewhere (where? kernel memory? how much space?), and switch between `vm_map`s back and forth... which seems like a very primitive solution.

I suppose no reference OS implements such mechanism (they all prefer COW + a hierarchy of shadow vm objects etc.), so I don't think I can find help there.

The best idea I have so far is to create a new pager type, which clones pages on demand. It would have to serve both the parent and child process, to avoid inconsistencies.

Or maybe this is the time we've started working on a simple layered hierarchy of vm_objects?

cahirwpz: Please note that all pages of physical memory are mapped in `KSEG0` and each `vm_page` stores `paddr` physical address. You can simply allocate pages in virtual address space, but copy content using physical addresses. This is a hack, but acceptable.

sys/fork.c
```diff
@@ -0,0 +1,55 @@
...
+     all fields one-by one for clarity. The new thread is already on the
+     all_thread list, has name and tid set. */
+
+  newtd->td_state = td->td_state;
```
cahirwpz: Two threads in running state? Leave it to `sched_add`.

sys/fork.c
```diff
@@ -0,0 +1,55 @@
...
+  newtd->td_flags = td->td_flags;
+  newtd->td_csnest = td->td_csnest;
```
cahirwpz: You should assume calling thread is not in critical section.

sys/fork.c
```diff
@@ -0,0 +1,55 @@
...
+  assert(td->td_uspace);
+
+  thread_t *newtd = thread_create(td->td_name, NULL, NULL);
+
+  /* Clone the thread. Since we don't use fork-oriented thread_t layout, we copy
+     all fields one-by one for clarity. The new thread is already on the
+     all_thread list, has name and tid set. */
+
+  newtd->td_state = td->td_state;
+  newtd->td_flags = td->td_flags;
```
cahirwpz: I see no reason to copy it from the parent. It will be probably zeroed out by `sched_add`.

sys/vm_map.c
```diff
@@ -239,6 +240,37 @@ void vm_map_dump(vm_map_t *map) {
   }
 }
 
+/* This entire function is a nasty hack, but we'll live with it until proper COW
+   is implemented. Since*/
```
cahirwpz: Have you forgotten to add / remove something here?

PR #221, opened 2017-03-25
Fixed test/server inconsiscency in launch script

rafalcieslak:
This bug was originally discovered by @strzkrzysiek.

When launching the kernel using `./launch` script with `-t` option (which is useful for redirecting kernel output to another terminal or file) and `-d` option simultaneously, both qemu monitor and gdb interface will hook to the same terminal, and will share input/output, making it impossible to enter gdb commands.

This patch disables the qemu monitor in test mode - we do not use the monitor interface anyway.

### DIFF
--- a/launch
+++ b/launch
@@ -52,7 +52,7 @@ def configure_ovpsim(args):
                '--wallclock',
                '--kernel', args.kernel]
 
-    if args.test:
+    if args.test or args.server:
         OVPSIM_OVERRIDES['uartCBUS/console'] = 0
 
     for item in OVPSIM_OVERRIDES.items():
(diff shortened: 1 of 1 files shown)


rafalcieslak:
A closer investigation let me to an observation that we are starting `nc` incorrectly, so that in case qemu is not ready when it opens a connection, it permanently becomes a zombie process. I'll try to rewrite this part to correct this issue.

rafalcieslak:
I went a step further and fixed the inconsistency of the `--test` (`-t`) option of the launch script. I've split it into `--test` and `--server` (`-s`). The former will redirect simulator output to stdout (no-op for qemu). The other will expose the output via TCP server at port 8000. This way it becomes clear where the output actually goes. Option `-s` is particularly useful when running a debugger, this way simulator output does not interleave with debugger output.

Incidentally, this feature actually simplifies `./launch` script implementation a bit, because we no longer need to start `nc` when using qemu.

PR #222, opened 2017-03-25
Automatically inspect kernel state on test deadlock

rafalcieslak:
This is a particularly nifty feature: When test results are inconclusive because neither success nor failure were reported within timeout, the `run_tests.py` script will launch `gdb` to get some information (`info registers`, `backtrace`, `kdump threads`, etc.) about kernel state, which may give some hints on why did it hang.

### DIFF
--- a/launch
+++ b/launch
@@ -70,7 +70,8 @@ def configure_qemu(args):
                '-machine', 'malta',
                '-cpu', '24Kf',
                '-kernel', args.kernel,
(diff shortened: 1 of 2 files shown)


rafalcieslak:
I've restarted the test build until it [failed](https://travis-ci.org/cahirwpz/mimiker/builds/215033868) so that its output demonstrates this feature in action.

PR #223, opened 2017-03-26
Implemented thread_join and thread_reap

rafalcieslak:
See #216.


rafalcieslak:
The test results indicate that the idle thread is incorrectly marked as `RUNNING`. I've fixed this issue in #224.

sys/thread.c
```diff
@@ -14,11 +14,17 @@ typedef TAILQ_HEAD(, thread) thread_list_t;
 /* TODO: Synchronize access to the list */
 static thread_list_t all_threads;
 
+mtx_t zombie_threads_mtx;
```
cahirwpz: Shouldn't it be defined static?

sys/thread.c
```diff
@@ -14,11 +14,17 @@ typedef TAILQ_HEAD(, thread) thread_list_t;
 /* TODO: Synchronize access to the list */
```
cahirwpz: Could you fix that in this PR as well?

tests/thread_exit.c
```diff
@@ -0,0 +1,52 @@
+#include <stdc.h>
+#include <clock.h>
+#include <thread.h>
+#include <sched.h>
+#include <ktest.h>
+
+int exit_time[] = {100, 200, 150};
```
cahirwpz: All (?) global variables in test modules should be defined static to avoid symbol clashes.

PR #224, opened 2017-03-26
Fixed idle thread always marked as running

rafalcieslak:
The idle thread would always stay marked as `TDS_RUNNING`, this patch fixes this problem.

### DIFF
--- a/sys/sched.c
+++ b/sys/sched.c
@@ -22,10 +22,13 @@ void sched_init() {
 void sched_add(thread_t *td) {
   // log("Add '%s' {%p} thread to scheduler", td->td_name, td);
 
+  td->td_state = TDS_READY;
+
+  /* Idle thread does not get inserted to the runqueue, and it does not require
+     increasing its time slice. */
   if (td == PCPU_GET(idle_thread))
     return;
 
-  td->td_state = TDS_READY;
   td->td_slice = SLICE;
   critical_enter();
 


rafalcieslak:
[The test results for this commit have enlightened me!](https://travis-ci.org/cahirwpz/mimiker/builds/215206603)

This branch fixes the problem where the idle thread (`kernel-main`) would stay marked as running even though it was not. But the test results are that kernel got stuck in the middle of `thread` test (typical). The threads state is:

```
 kdump threads
----------------------------------
| Id | Name        | State       |
----------------------------------
| 0  | kernel-main | TDS_RUNNING |
| 1  | main        | TDS_READY   |
| 2  | first       | TDS_RUNNING |
| 3  | second      | TDS_READY   |
----------------------------------
```
so there are still more than 1 threads marked as `RUNNING` at the same time (wasn't this branch supposed to fix that?).

The problem is that this time it is actually the idle thread that is in fact executing, so the `first` thread (one of threads created within the `thread` test) is the one incorrectly marked as running. Why is that? The thing is, the `first` and `second` thread are not managed by the scheduler at all, they are not inserted into the runqueue etc, because this is a legacy test back from when we did not have a working scheduler.

The test uses an old function `thread_switch_to` (which forces a context switch independent from the scheduler), to switch back and forth between `first` and `second` threads. This function marks one thread as `RUNNING`, and the other as `READY`. `thread_switch_to` is not uses anywhere else, it exists only to serve this test.

This test used to work well back when we had no scheduler, but we can now run into problems if this test gets preempted, which naturally will cause interference. The `first` and `second` threads are not on the runqueue, but `main` is!

In this particular case observable of the aforementioned failed test run the nature of the deadlock is such that the `first` thread was preempted, and somehow (some time later on) we see the idle thread is executing (`sched_run () at sched.c:99   td->td_flags |= TDF_NEEDSWITCH;`), and there is no way we'll return to the `first` thread as it was (correctly) not registered in the scheduler.

My conclusion is that the `thread` test is severely outdated and no longer makes any sense with the presence of a scheduler, even if its threads are not registered for scheduling. Same goes for `thread_switch_to`, as it itself interferes with scheduler logic. Thus I propose two possible solutions:

- Removing the `thread` test entirely (as well as `thread_switch_to` function). Soon we'll implement proper `thread_join`, which will allow us to prepare much better and more natural tests for threads behavior. 
- Wrapping the entire `thread` test in a critical section. I did not test if that works, but it should disable the scheduler for the duration of the test.

cahirwpz:
That is lenghty explanation! My answer will be much shorter... We can safely remove the piece of code which is a remnant of times where we did not have the scheduler. As you correctly noticed there's no much use of keeping `thread_switch_to` routine and the test for it.

rafalcieslak:
#225 solves this problem by removing outdated code.

PR #228, opened 2017-03-27
rwlock for vm_map

rafalcieslak:
We're starting to think about accessing a `vm_map` from multiple threads, so it's the high time we implemented synchronization for the `vm_map_t` structure. This branch uses `rwlock` for that purpose. This is convenient, because most access done to a `vm_map` is due to page faults, and as these do not modify the structure, we're happy to allow more than one thread to look up pages in the map simultaneously.

The `rwlock` locks the map structure and it's entries, but - naturally - not the data within address space represented by the map.

To implement this correctly, I had to do some minor refactoring:

- The `pmap_t *pmap` field in `vm_map_t` is now a constant pointer, as it is only set when the map is created, and it never gets changed. This way it is explicitly marked as immutable, and thus access to `pmap` does not require locking the `rwlock`.
- `sbrk` implementation details (which do a lot of operation on the underlying `vm_map`) are moved to `vm_map_sbrk` in `vm_map.c`, so that the internal structure of a `vm_map` (and synchronisation strategy) does not leak outside `vm_map.c`.
- I moved around some condition checks in `rw_enter`, so that a recursive lock would allow recursive READERS while it is WLOCKED. Previously, a writer-locked recursive rwlock would only allow recursive writers to enter, but I see no reason why should it not let extra recursive readers in. A good example of a use-case is `vm_map_sbrk`, which starts by locking the rwlock for WRITING, does some work, and then calls `vm_map_findspace`, which has to enter the lock for READING. Since the used lock is recursive, it seems to me that this is a valid scenario.

### DIFF
--- a/include/vm_map.h
+++ b/include/vm_map.h
@@ -58,17 +58,18 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
 vm_map_entry_t *vm_map_add_entry(vm_map_t *map, vm_addr_t start, vm_addr_t end,
                                  vm_prot_t prot);
 
+int vm_map_resize(vm_map_t *map, vm_map_entry_t *entry, vm_addr_t new_end);
+
 /* Looks up a gap of @length size in @map. The search starts at @start address.
  * On success, returns 0 and sets *addr. @start and @length arguments must be
  * page-aligned. */
 int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
                      vm_addr_t /*out*/ *addr);
+int vm_map_findspace_nolock(vm_map_t *map, vm_addr_t start, size_t length,
+                            vm_addr_t /*out*/ *addr);
 
 void vm_map_dump(vm_map_t *vm_map);
 
 int vm_page_fault(vm_map_t *map, vm_addr_t fault_addr, vm_prot_t fault_type);
-
-/* sbrk support in vm_map. */
-vm_addr_t vm_map_sbrk(vm_map_t *map, intptr_t increment);
 
 #endif /* _VM_MAP_H */
--- a/include/vm_map.h
+++ b/include/vm_map.h
@@ -58,8 +58,6 @@ void vm_map_protect(vm_map_t *map, vm_addr_t start, vm_addr_t end,
 vm_map_entry_t *vm_map_add_entry(vm_map_t *map, vm_addr_t start, vm_addr_t end,
                                  vm_prot_t prot);
 
-int vm_map_resize(vm_map_t *map, vm_map_entry_t *entry, vm_addr_t new_end);
-
 /* Looks up a gap of @length size in @map. The search starts at @start address.
  * On success, returns 0 and sets *addr. @start and @length arguments must be
  * page-aligned. */
@@ -67,6 +65,10 @@ int vm_map_findspace(vm_map_t *map, vm_addr_t start, size_t length,
                      vm_addr_t /*out*/ *addr);
 int vm_map_findspace_nolock(vm_map_t *map, vm_addr_t start, size_t length,
                             vm_addr_t /*out*/ *addr);
+
+/* Tries to resize an entry, by moving its end if there
+   are no other mappings in the way. On success, returns 0. */
+int vm_map_resize(vm_map_t *map, vm_map_entry_t *entry, vm_addr_t new_end);
 
 void vm_map_dump(vm_map_t *vm_map);
 
--- a/include/vm_map.h
+++ b/include/vm_map.h
@@ -4,6 +4,7 @@
 #include <queue.h>
 #include <pmap.h>
 #include <vm.h>
+#include <rwlock.h>
 
 typedef struct vm_map_entry vm_map_entry_t;
 
@@ -26,7 +27,9 @@ typedef struct vm_map {
   TAILQ_HEAD(vm_map_list, vm_map_entry) list;
   SPLAY_HEAD(vm_map_tree, vm_map_entry) tree;
   size_t nentries;
-  pmap_t *pmap;
+  pmap_t *const pmap;
+  rwlock_t rwlock; /* The recursive RW lock guarding vm_map structure and all
+                      its entries. */
   /* program segments */
   vm_map_entry_t *sbrk_entry; /* The entry where brk segment resides in. */
   vm_addr_t sbrk_end;         /* Current end of brk segment. */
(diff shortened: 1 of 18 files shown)


sys/vm_map.c
```diff
@@ -212,36 +239,90 @@ void vm_map_dump(vm_map_t *map) {
...
     kprintf("[vm_map] * %08lx - %08lx [%c%c%c]\n", it->start, it->end,
             (it->prot & VM_PROT_READ) ? 'r' : '-',
             (it->prot & VM_PROT_WRITE) ? 'w' : '-',
             (it->prot & VM_PROT_EXEC) ? 'x' : '-');
     vm_map_object_dump(it->object);
   }
+  rw_leave(&map->rwlock);
+}
+
+vm_addr_t vm_map_sbrk(vm_map_t *map, intptr_t increment) {
```
cahirwpz: This procedure does too much. Please move `sbrk` segment creation to `exec` syscall code. Then I'll revisit `sbrk` implementation.

include/vm_map.h
```diff
@@ -26,7 +27,9 @@ typedef struct vm_map {
   TAILQ_HEAD(vm_map_list, vm_map_entry) list;
   SPLAY_HEAD(vm_map_tree, vm_map_entry) tree;
   size_t nentries;
-  pmap_t *pmap;
+  pmap_t *const pmap;
```
cahirwpz: What do we gain by declaring this pointer as `const`? To circumvent constness you use `*((pmap_t **)(&x)) = p;` assignment, which looks pretty ugly.

sys/sysent.c
```diff
@@ -32,38 +32,7 @@ int sys_sbrk(thread_t *td, syscall_args_t *args) {
...
-    /* Shrink or expand the vm_map_entry */
-    vm_addr_t new_end = roundup(brk + increment, PAGESIZE);
-    if (vm_map_resize(td->td_uspace, brk_entry, new_end) != 0) {
-      /* Map entry expansion failed. */
-      return -ENOMEM;
-    }
-    td->td_uspace->sbrk_end += increment;
-    return brk;
-  }
+  return vm_map_sbrk(td->td_uspace, increment);
```
cahirwpz: I really dislike moving `sbrk` syscall handler code to `vm_map` module. This looks like a kludge. Can we do better?

sys/vm_map.c
rafalcieslak: How about I create a separate file which would implement the `sbrk` interface (that is: sbrk segment creation, called from `exec`, and sbrk fragment resizing)? The reason I moved it away from `sysent.c` is that `sbrk_entry` is a field of `vm_map_t`, and therefore I believe all operations on it should be wrapped in an interface implemented either by `vm_map` implementation, or maybe a separate `sbrk`-related file.

Another option might be to move `sbrk_entry` to the `thread_t` structure, although I think we've discussed this decision some time ago and dismissed it.

cahirwpz: Your proposal for `sys/sbrk.c` file seems to be fine.

I conform that I still believe that moving `sbrk_entry` to `thread_t` is a bad idea.

include/vm_map.h
rafalcieslak: By explicitly marking that the pointer is const we signify that it's value never changes during a `vm_map_t` lifetime, and thus it may be safely dereferenced with no synchronization mechanisms. As C does not have structure constructors, in order to set a `const`-qualified member field it must be accessed via a non-`const` pointer - which is okay during structure initialization, since it *does not* break the semantics of "never changes during lifetime". The one-time assignment would look less "ugly" if I did not choose to perform it in a single line (compare: `pmap_t** q = (pmap_t**)&x;    *q = p;`), but still it is a common C idiom. We could wrap it in a macro: `CONST_ASSIGN(pmap_t*, x, p);` if this looks better to you. By the way, such const-evading assignment is valid in C, but not in C++, where it triggers undefined behavior.

sys/vm_map.c
```diff
@@ -91,22 +94,31 @@ static bool vm_map_insert_entry(vm_map_t *vm_map, vm_map_entry_t *entry) {
 
 vm_map_entry_t *vm_map_find_entry(vm_map_t *vm_map, vm_addr_t vaddr) {
   vm_map_entry_t *etr_it;
-  TAILQ_FOREACH (etr_it, &vm_map->list, map_list)
-    if (etr_it->start <= vaddr && vaddr < etr_it->end)
+  rw_enter(&vm_map->rwlock, RW_READER);
+  TAILQ_FOREACH (etr_it, &vm_map->list, map_list) {
+    if (etr_it->start <= vaddr && vaddr < etr_it->end) {
+      rw_leave(&vm_map->rwlock);
```
cahirwpz: With many return points from function it's easier to forget to do unlock - it's probably better to call `rw_leave` in one place.

sys/vm_map.c
```diff
@@ -91,22 +94,31 @@ static bool vm_map_insert_entry(vm_map_t *vm_map, vm_map_entry_t *entry) {
...
 static void vm_map_remove_entry(vm_map_t *vm_map, vm_map_entry_t *entry) {
+  rw_assert(&vm_map->rwlock, RW_WLOCKED);
   vm_map->nentries--;
   vm_object_free(entry->object);
   TAILQ_REMOVE(&vm_map->list, entry, map_list);
   kfree(mpool, entry);
 }
 
 void vm_map_delete(vm_map_t *map) {
+  rw_enter(&map->rwlock, RW_WRITER);
```
cahirwpz: Aww... It'd be so nice to have magic macro to enable us to write:

```c
with_rw_lock(&map->rw_lock, RW_WRITER) {
 ...
}
```

... or alternatively use cleanup attribute (provided `rw_enter` returns pointer to locked object) like so:

```c
#define rw_auto_leave \
  rw_lock_t *__rw_ ## __LINE__ ## __attribute__((cleanup(rw_leave))) =
...
rw_auto_leave rw_enter(&map->rw_lock, RW_WRITER);
```

include/vm_map.h
cahirwpz: Ok. I'm going to leave with it :-)

sys/vm_map.c
rafalcieslak: The first variant would look cool, but I don't think there is any way to implement this, as the macro would not control neither the beginning nor the end of the scope.

The second style could be improved even a bit more:

```c
void rw_guard_leave(rwlock_t **rwlock) {
  /* Dereference required... */
  rw_leave(*rwlock);
}

#define rw_lock_guard(prwlock, mode)                                           \
  rwlock_t *rw_lock_guard_##__LINE__                                           \
    __attribute__((cleanup(rw_guard_leave))) = prwlock;                        \
  rw_enter(prwlock, mode);
```

and would be used with bare:
```c
rw_lock_guard(&map->rwlock, RW_WRITER);
```

We can also have a similar macro for `mtx_t`.

cahirwpz:
LGTM. Should I merge this branch before I start reviewing #238 ?

rafalcieslak:
Yes.

PR #234, opened 2017-03-27
Fix deadlock on zombie_threads_mtx

rafalcieslak:
During `thread_exit`, the mutex on `zombie_threads` must be unlocked before critical section is left. Otherwise we risk preemption before the mutex is unlocked, and as the thread is TDS_INACTIVE, it will never resume execution - permanently blocking any other threads from locking on the mutex.

I've observed this problem several times on my local machine, but I've only managed to trigger it with QEMU, and thus I cannot provide a seed for reproducing the issue.

### DIFF
--- a/include/thread.h
+++ b/include/thread.h
@@ -6,6 +6,8 @@
 #include <context.h>
 #include <exception.h>
 #include <sleepq.h>
+#include <mutex.h>
+#include <condvar.h>
 
 typedef uint8_t td_prio_t;
 typedef uint32_t tid_t;
@@ -25,6 +27,9 @@ typedef struct thread {
   TAILQ_ENTRY(thread) td_zombieq; /* a link on zombie queue */
   char *td_name;
   tid_t td_tid;
+  /* Locks*/
+  mtx_t td_lock;
+  condvar_t td_waitcv;
   /* thread state */
   enum { TDS_INACTIVE = 0x0, TDS_WAITING, TDS_READY, TDS_RUNNING } td_state;
   uint32_t td_flags;           /* TDF_* flags */
--- a/include/thread.h
+++ b/include/thread.h
@@ -21,15 +21,16 @@ typedef struct fdtab fdtab_t;
 #define TDF_NEEDSWITCH 0x00000002 /* must switch on next opportunity */
 
 typedef struct thread {
+  /* Locks*/
+  mtx_t td_lock;
+  condvar_t td_waitcv; /* CV for thread exit, used by join */
+  /* List links */
   TAILQ_ENTRY(thread) td_all;     /* a link on all threads list */
   TAILQ_ENTRY(thread) td_runq;    /* a link on run queue */
   TAILQ_ENTRY(thread) td_sleepq;  /* a link on sleep queue */
   TAILQ_ENTRY(thread) td_zombieq; /* a link on zombie queue */
   char *td_name;
   tid_t td_tid;
-  /* Locks*/
-  mtx_t td_lock;
-  condvar_t td_waitcv;
   /* thread state */
   enum { TDS_INACTIVE = 0x0, TDS_WAITING, TDS_READY, TDS_RUNNING } td_state;
   uint32_t td_flags;           /* TDF_* flags */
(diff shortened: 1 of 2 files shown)


cahirwpz:
Locking strategy of code that starts in [sys/thread.c:133](https://github.com/rafalcieslak/mimiker/blob/1e9b0989bc5634cddc59f40da53674030829858c/sys/thread.c#L133) looks like utter mess. A brief look at NetBSD's `proc` and `lwp` (FreeBSD's `proc` and `thread` reveal similar structure) shows that they use at least one mutex ([l_mutex](http://bxr.su/NetBSD/sys/sys/lwp.h#89), [p_lock](http://bxr.su/NetBSD/sys/sys/proc.h#234)) for whole task structure and additional conditional variable ([l_waitcv](http://bxr.su/NetBSD/sys/sys/lwp.h#134), [p_waitcv](http://bxr.su/NetBSD/sys/sys/proc.h#237)) to aid `exit` and `join` calls implementation.

Please add `td_lock` and `td_waitcv` to `thread` structure and provide cleaner solution to fix the bug.

sys/thread.c
```diff
@@ -70,6 +70,9 @@ thread_t *thread_create(const char *name, void (*fn)(void *), void *arg) {
   td->td_kstack.stk_base = (void *)PG_VADDR_START(td->td_kstack_obj);
   td->td_kstack.stk_size = PAGESIZE;
 
+  mtx_init(&td->td_lock, MTX_DEF);
+  cv_init(&td->td_waitcv, "wait");
```
cahirwpz: `"td_waitcv"` for better diagnostics

sys/thread.c
```diff
@@ -131,16 +136,18 @@ noreturn void thread_exit(int exitcode) {
...
 
   critical_enter();
-  TAILQ_INSERT_TAIL(&zombie_threads, td, td_zombieq);
-  td->td_exitcode = exitcode;
-  sleepq_broadcast(&td->td_exitcode);
-  td->td_state = TDS_INACTIVE;
+  {
+    td->td_exitcode = exitcode;
+    td->td_state = TDS_INACTIVE;
+    cv_signal(&td->td_waitcv);
```
cahirwpz: `cv_broadcast` to wake up all thread that awaits `td` exit?

sys/thread.c
```diff
@@ -131,16 +136,18 @@ noreturn void thread_exit(int exitcode) {
   fdtab_release(td->td_fdtable);
 
   mtx_lock(&zombie_threads_mtx);
+  TAILQ_INSERT_TAIL(&zombie_threads, td, td_zombieq);
+  mtx_unlock(&zombie_threads_mtx);
 
   critical_enter();
```
cahirwpz: Seems like we don't need `critical_enter` here as long as scheduler respects `td_lock` mutex. Please have a look at preferred solution [kern_thread.c:488](http://bxr.su/FreeBSD/sys/kern/kern_thread.c#488). You'll need to add `sched_abandon` function that removes a thread from scheduler and makes sure it will not ever be returned to.

include/thread.h
```diff
@@ -25,6 +27,9 @@ typedef struct thread {
   TAILQ_ENTRY(thread) td_zombieq; /* a link on zombie queue */
   char *td_name;
   tid_t td_tid;
+  /* Locks*/
+  mtx_t td_lock;
+  condvar_t td_waitcv;
```
cahirwpz: Please move these to the top of the structure and describe the purpose of `td_waitcv`.

sys/thread.c
rafalcieslak: I don't think I understand the solution you've linked to. The entire function is marked with a comment:
```
Always called with scheduler locked.
```
which suggests that there is an external mechanism which prevents the scheduler from running while a thread is exiting. The thread mutex gets locked only at the very last moment, which also hints at a different synchronization mechanism besides this mutex.

Also, I do not understand what do you mean by:

> as long as scheduler respects `td_lock` mutex.

Do we want to prevent the scheduler from switching context to threads whose mutex is locked? Should the scheduler lock (or, actually, try-lock) the thread mutex when picking next thread? For this to work well w/o obvious deadlock, we would also have to prevent preemption of threads who own their own thread mutex. I'm under the impression that this could lead to complex problems.

I'd appreciate some detailed description of the idea you want me to implement.

cahirwpz:
Ok. I believe there's no simple and clean way to fix the problem tackled by this PR. While I consider the change to be hack, we cannot wait any longer for it being integrated.

@rafalcieslak please look at issue #242, I'll put more information there soon.

Issue #214, opened 2017-03-24
Automatically verifying code formatting


rafalcieslak:
We frequently push code to `master` which is not formatted with `make format`. We've discussed using hooks or other mechanisms for automating this process, but it's tricky as each developer has a different work-flow, and we will easily run into inconsistencies.

My proposed solution is to start treating misformatted code as build failure. Travis will run `make format`, and if `git diff-index --quiet HEAD` returns non-zero exit code, it would report a failed build.

The reason why I want to be so strict about it is that I'm sick of being unable to format changes in a new branch because it would also correct unrelated unformatted files, introducing unwanted changes into my branch.

PR #235, opened 2017-03-29
Automatically verify source code formatting

rafalcieslak:
As we discussed. References #214.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ etags
 assym.h
 
 # toolchain build artifacts
-toolchain/
+toolchain
 
 # Editor swap files
 *.swp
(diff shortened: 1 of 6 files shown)


rafalcieslak:
This was tricky to set up, because Travis uses some weird configuration where it's difficult to use `llvm != 3.5`, and our `.clang-format` file only works correctly with `clang >= 3.8`.

[Here's an example build where formatting was incorrect, notice the diff & message at the end of the build log.](https://travis-ci.org/cahirwpz/mimiker/builds/216384805) This branch is finally ready for review.

cahirwpz:
LGTM. Implements #214

PR #237, opened 2017-04-02
Added synchronization to malloc module and created few basic tests

czapiga:
I'm going to create more complex (multithreaded) test in next two days.


include/malloc.h
```diff
@@ -36,6 +38,9 @@ typedef struct malloc_pool {
   uint32_t mp_magic;                /* Detect programmer error. */
   const char *mp_desc;              /* Printable type name. */
   struct ma_list mp_arena;          /* Queue of managed arenas. */
+  mtx_t mutex;                      /* Mutex protecting structure */
```
cahirwpz: Please use common prefix for structure fields, i.e. `mp_`.

Fields naming could also be made more reasonable:
* `mutex` -> `mp_lock`
* `pages_used` -> `mp_pages_min`
* `pages_limit` -> `mp_pages_max`

Actual number of used pages should somewhere between `min` and `max` values.

include/malloc.h
```diff
@@ -36,6 +38,9 @@ typedef struct malloc_pool {
...
+  uint32_t pages_used;              /* Current number of pages */
```
cahirwpz: I think declaring those as `unsigned` should be ok.

sys/malloc.c
```diff
@@ -17,6 +17,7 @@ static struct {
   void *end;
 } sbrk = {__ebss, __kernel_end};
 
+/*
```
cahirwpz: To comment out larger chunks of code please use following construct:
```c
#if 0
...
#endif
```

sys/malloc.c
```diff
@@ -171,11 +170,22 @@ static mem_block_t *try_allocating_in_area(mem_arena_t *ma,
...
+  TAILQ_INIT(&mp->mp_arena);
+  mtx_init(&mp->mutex, MTX_RECURSE);
+  kmalloc_add_pages(mp, pages);
+  mp->pages_used = pages;
+  mp->pages_limit = pages_limit;
+}
+
 void *kmalloc(malloc_pool_t *mp, size_t size, uint16_t flags) {
   size_t size_aligned = align(size, MB_ALIGNMENT);
+  if (size == 0 || size_aligned == 0)
```
cahirwpz: Following is true: `size_aligned == 0` <=> `size == 0`, so checking them both is superfluous.

sys/malloc.c
```diff
@@ -184,12 +194,20 @@ void *kmalloc(malloc_pool_t *mp, size_t size, uint16_t flags) {
...
       if (flags == M_ZERO)
         memset(mb->mb_data, 0, size);
-
+      mtx_unlock(&mp->mutex);
       return mb->mb_data;
     }
   }
-
   /* Couldn't find any continuous memory with the requested size. */
+  if (mp->pages_used < mp->pages_limit) {
```
cahirwpz: If flag `M_NOWAIT` is passed to `kmalloc` you should not attempt to allocate new pages. In future `kmalloc_add_pages` may trigger page replacement algorithm, that may force a thread go sleep!

czapiga: Fixed

sys/malloc.c
czapiga: Fixed

sys/malloc.c
czapiga: Fixed

include/malloc.h
czapiga: I don't need this `mp_pages_min` value for now since pages are only allocated when needed and never freed. I would rather keep just `mp_pages_used` and `mp_pages_max`.

include/malloc.h
czapiga: Changed

tests/malloc.c
```diff
@@ -1,65 +1,100 @@
...
-  void *ptr6 = kmalloc(mp, 2000, M_NOWAIT);
-  assert(ptr6 == NULL);
+#define ALLOCATIONS_PER_THREAD 1000
+#define THREADS_NUMBER 10
 
-  pm_free(page);
+static void threads_function(void *arg) {
+  for (int i = 0; i < ALLOCATIONS_PER_THREAD; i++) {
+    void *ptr = kmalloc(arg, 1 + (i * (int)arg) % 32, 0);
+    assert(ptr != NULL);
```
cahirwpz: Threads are crashing in this line?

czapiga: Yes, but size is != 0, there is no `M_NOWAIT` and I can't find panic message "memory exhausted in ".

czapiga: Fixed. Problem was with `(int)arg` - arg pointer casted to int resulted in negative value. I just changed it to `(unsigned)` and everything works.

tests/malloc.c
```diff
@@ -1,65 +1,100 @@
...
 
-  kmalloc_init(mp);
-  kmalloc_add_arena(mp, page->vaddr, PAGESIZE);
+static int malloc_multiple_allocations(void) {
+  MALLOC_DEFINE(mp, "testing memory pool");
+  kmalloc_init(mp, 1, 1);
+  const int n = 50;
+  void *ptrs[n];
+  for (int i = 0; i < n; i++)
+    assert(NULL != (ptrs[i] = kmalloc(mp, i + 1, M_NOWAIT)));
```
cahirwpz: Please do not use expression with side effects as arguments to macros.

tests/malloc.c
```diff
@@ -1,65 +1,100 @@
...
+}
 
-  void *ptr6 = kmalloc(mp, 2000, M_NOWAIT);
-  assert(ptr6 == NULL);
+#define ALLOCATIONS_PER_THREAD 1000
+#define THREADS_NUMBER 10
 
-  pm_free(page);
+static void threads_function(void *arg) {
+  for (int i = 0; i < ALLOCATIONS_PER_THREAD; i++) {
```
cahirwpz: Right now each thread has at most one block allocated at any given moment. This will not cause much lock contention. It'd be much more interesting to see each thread to have at least hundred of memory blocks active.

run_tests.py
```diff
@@ -7,7 +7,7 @@
 
 N_SIMPLE = 5
 N_THOROUGH = 100
-TIMEOUT = 5
+TIMEOUT = 25
```
rafalcieslak: If your test really needs 25 seconds to complete, then something seems off. Note that the [test results from before you increased timeout](https://travis-ci.org/cahirwpz/mimiker/builds/220722118) completed successfully for many seeds, and the time required to run all tests should't differ that much from a seed to another! Examining that output I see that the main thread is in `INACTIVE` state and only the idle thread is running - so that test run would never finish successfully, no mater how much time it was given.

cahirwpz:
Please move multithreaded tests to separate PR and I'll accept this one and merge with master.

rafalcieslak:
Now that the *long* test is gone, can we please revert the test timeout from 25 to 5 seconds?

PR #238, opened 2017-04-03
Lock guards

rafalcieslak:
**WARNING: This branch incorporates changes from #228 (which is close to merging).**

Inspired by [this suggestion](https://github.com/cahirwpz/mimiker/pull/228#discussion_r109361512) I've implemented `lock_guard`s for both `mtx_t` and `rwlock_t`. This branch also enables the use of these guards thorough the entire codebase. I've only switched to `lock_guards` in places where it actually simplifies source code, so the normal `mtx_lock` etc. are still widely used.

The diff will grow up a bit once #228 is merged, because this branch also slightly simplifies logic introduced by #228.

My implementation is mildly inspired by [systemd's implementation](https://github.com/systemd/systemd/blob/13b498f967c5117a88d72304bed1f8c0b9c1bb87/src/basic/macro.h) of a similar mechanism.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ etags
 assym.h
 
 # toolchain build artifacts
-toolchain/
+toolchain
 
 # Editor swap files
 *.swp
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,16 @@ language: c
 compiler: gcc
 before_install:
  - sudo apt-get -qq update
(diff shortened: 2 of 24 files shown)


rafalcieslak:
I don't like to invent names on my own, so I borrowed the [C++ name for the same RAII concept: `std::lock_guard<T>`](http://en.cppreference.com/w/cpp/thread/lock_guard). (Note that there also is a `scoped_lock` in the fresh C++17 standard, but it has a slightly different meaning!)

PR #239, opened 2017-04-05
Klog

laky55555:
Added logging mechanism.
Currently logging **is not** thread safe.
There is problem when logging message without any parameters.
Logging is done with **klog(format, params)** or **klog_mask(costum_mask, format, params)**.
Printing saved logs is done with **klog_dump_all();**
During starting user can add custom mask for logging messages with **klog-mask=hex_number**.


tests/klog.c
```diff
@@ -0,0 +1,168 @@
+#define KL_LOG KL_ALL
+
+#include <stdc.h>
+#include <ktest.h>
+#define _KLOG_PRIVATE
+#include <klog.h>
+
+static int test_klog_shorter() {
```
cahirwpz: By looking at test names it's difficult to say what were your intentions.

tests/klog.c
```diff
@@ -0,0 +1,168 @@
...
+
+  return 0;
+}
+
+static int test_klog_long() {
+  int number_of_logs = 55 * KL_SIZE + 1;
+  kprintf("Testing %d logs in array size %d\n", number_of_logs, KL_SIZE);
+  for (int i = 1; i <= number_of_logs; ++i)
+    klog("Testing longer; Message %d, of %d", i, number_of_logs);
+  assert(((klog.first + KL_SIZE - 1) & (KL_SIZE - 1)) ==
```
cahirwpz: There's a lot of code repetition in this file. This is a sign of bad coding practice. Please clean up the code.

tests/klog.c
```diff
@@ -0,0 +1,168 @@
...
+  return 0;
+}
+
+// static int test_seven_params() {
+//     klog("Message with %d, %d, %d, %d, %d, %d, %d parameters.", 7, 7, 7, 7,
+//     7, 7, 7, 7);
+//     return 0;
+// }
+
+static int test_klog() {
```
cahirwpz: Please write down a testing strategy (consider all important cases) and then implement tests. Testing procedure should not be slowed down by printing messages.

PR #240, opened 2017-04-05
Pooled memory allocator

MoloToha:
Uploaded basic prototype for pooled memory allocator along with a very basic test to just see if it works at all. I will try to write a more complex test and resolve TODO's in the nearest future.


prototypes/palloc/pool_alloc.c
```diff
@@ -0,0 +1,212 @@
...
+#define panic(text, args ...)                                                 \
+  __extension__({                                               \
+    fprintf(stderr, "FATAL ERROR: %s:%d:%s(): " text, __FILE__, __LINE__, __func__, ##args);     \
+    exit(EXIT_FAILURE);                                         \
+  })
+  
+int page_size;
+
+typedef struct pool_item {
+    uint64_t pi_guard_number; //0xcafebabecafebabe by default, normally isn't changed, so if it has another value, memory corruption has taken place
```
rafalcieslak: When picking magic guard values, it is a good practice to pick one that does not contain any (long) substring more than once. The one you picked is particularly unfortunate: `cafebabe` repeats twice at a word-boundary! So if one happens to write this value to address `X` and afterwards at `X+4`, they overlay perfectly, and it may seem that `X` still holds a valid guard number, even though it was overwritten.

prototypes/palloc/pool_alloc.c
```diff
@@ -0,0 +1,212 @@
...
+    DEBUG_PRINT("Entering create_slab\n");
+    pool_slab_t *slab = malloc(page_size); //TODO: Change this during implementation into kernel
+    slab -> ph_nused = 0;
+    slab -> ph_ntotal = (page_size-sizeof(pool_slab_t)-7)/(sizeof(pool_item_t)+size+1); // sizeof(pool_slab_t)+n*(sizeof(pool_item_t)+size+1)+7 <= page_size, 7 is maximum number of padding bytes for a bitmap 
+    
+    slab -> ph_nfree = slab -> ph_ntotal;
+    slab -> ph_start = sizeof(pool_slab_t)+align_to_binary_pow(slab -> ph_ntotal, 3);
+    memset(slab -> ph_bitmap, 0, slab -> ph_ntotal);
+    for (int i = 0; i < slab -> ph_ntotal; i++) {
+        pool_item_t *curr_pi = (pool_item_t*) ((uint64_t) slab+(slab -> ph_start)+i*(size+sizeof(pool_item_t)));
```
rafalcieslak: You perform this computation (slab address to pool_item address) many times - consider wrapping it in a convenience macro, it would increase readability and decrease the chance of mistakes.

prototypes/palloc/pool_alloc.c
```diff
@@ -0,0 +1,212 @@
...
+
+    pool_slab_t *new_slab;
+    if (pool -> pp_nitems) {
+        if (LIST_EMPTY(&(pool -> pp_part_slabs))) {
+            new_slab = LIST_FIRST(&(pool -> pp_empty_slabs));
+            LIST_REMOVE(new_slab, ph_slablist);
+        } else {
+            new_slab = LIST_FIRST(&(pool -> pp_part_slabs));
+            LIST_REMOVE(new_slab, ph_slablist);
+        }
```
rafalcieslak: These two `if` cases have almost identical content. Maybe it would be simpler to write:

```c
pool_slab_list_t* slab_list_to_use = pool->pp_nitems ? &pool->pp_empty_slabs : &pool->pp_part_slabs;
new_slab = LIST_FIRST(slab_list_to_use);
LIST_REMOVE(new_slab, ph_slablist);
```
or maybe even:
```c
new_slab = LIST_FIRST(pool->pp_nitems ? &pool->pp_empty_slabs : &pool->pp_part_slabs);
LIST_REMOVE(new_slab, ph_slablist);
```
Similar transformation can be used to simplify many of your `if` statements - but I am not sure if the above suggestion fits our coding style.

prototypes/palloc/pool_alloc.c
```diff
@@ -0,0 +1,212 @@
...
+    it = LIST_FIRST(&(pool -> pp_full_slabs));
+    while(it) {
+        pool_slab_t *next = LIST_NEXT(it, ph_slablist);
+        LIST_REMOVE(it, ph_slablist);
+        destroy_slab(it, pool);
+        it = next;
+    }
+
+    for (size_t i = 0; i < sizeof(pool_t)/8; i++) {
+        tmp[i] = 0xdeadbeefdeadbeef;
```
rafalcieslak: Am I correct that this causes writes to memory regions that were already freed? It might be better to first perform these writes (e.g. while checking the `tmp[i]` values at the beginning of `pool_destroy`), and destroy slabs afterwards.

rafalcieslak:
Also: remember to run `make format` periodically - it will help you clean up your code.

Additionally, you can skip some parenthesis:

`(x->y)++` is the same as `x->y++`,

`&(x->y)` is the same as `&x->y`.

MoloToha:
Thank you for suggestions! I will implement them ASAP.

cahirwpz:
@MoloToha Please address code changes outlined by @rafalcieslak. The suggestions can be viewed by clicking *show outdated* on this page. If I hadn't run `clang-format` they wouldn't be hidden – sorry for inconvenience.

prototypes/palloc/pool_alloc.c
MoloToha: This piece of code writes to a pool_t object which is located at client's side (so it can't be freed for the second time), slabs are destroyed separately before this operation.

prototypes/palloc/pool_alloc.c
MoloToha: I usually write this kind of code for readability purposes. We can discuss this at the meeting.

MoloToha:
The last commit is my attempt to integrate this component into kernel, everything works (at least on my computer), test runs correctly, currently this test is set up to cause memory corruption, so you should comment line containing memset (which causes this memory corruption) in order to pass the test.

rafalcieslak:
I see this is shaping up nicely. Note that throughout the codebase we prefer:
```c
/* C-style comments */
```
over:
```c++
// C++-style comments.
```

If you want to prepare a test which intentionally crashes (to demonstrate something), you can pass `KTEST_FLAG_BROKEN` to `KTEST_ADD` - this way the test will not be launched automatically during automatic testing, but you will still be able to run it by specifying `test=...`.

MoloToha:
This memset writes zeroes to space that is occupied by other items inside a slab, thus creating an artificial memory corruption

tests/pool_alloc_corruption.c
```diff
@@ -0,0 +1,49 @@
...
+#include <common.h>
+
+void another_int_constr(void *buf, __unused size_t size) {
+  int *num = buf;
+  *num = 0;
+}
+void another_int_destr(__unused void *buf, __unused size_t size) {
+}
+
+int test_pool_alloc_corruption() {
```
rafalcieslak: You can avoid code duplication by merging this function into one: `test_pool_alloc(int corrupt_memory)` and then using both variants from proxy functions:
```
int test_pool_alloc_corruption(){
  return test_pool_alloc(1);
} ... etc.
```

sys/pool_alloc.c
```diff
@@ -0,0 +1,235 @@
...
+#define GET_PI_AT_IDX(slab, i, size)                                           \
+  (pool_item_t *)((unsigned long)slab + (slab->ph_start) +                     \
+                  (i) * (size + sizeof(pool_item_t)))
+
+#define ALIGN_TO_BINARY_POW(x, a) ((((x) >> (a)) + 1) << (a))
+
+#define PALLOC_DEBUG 0
+
+#if defined(PALLOC_DEBUG) && PALLOC_DEBUG > 0
+#define PALLOC_DEBUG_PRINT(text, args...)                                      \
```
rafalcieslak: We don't need yet another macro for displaying messages - just use `log`. I see you wanted to quickly enable/disable pool allocator debug messages and that is indeed useful, but #239 (klog) should be ready very soon and it will enable kernel-wide message filtering. Also, generally, your debug messages are very verbose. Keeping track of the general state of the allocator is great, but displaying messages on every single allocation seems like a bit too much. Most of your messages seem to track control flow ("Entering..."), it it simpler to track it using breakpoints in a debugger. Now that you've tested the allocator and it seems to be working okay, you may want to trim down these messages a bit.

sys/pool_alloc.c
```diff
@@ -0,0 +1,235 @@
...
+
+#define PI_MAGIC_WORD 0xcafebabe
+
+#define PP_FREED_WORD 0xdeadbeef
+
+#define GET_PI_AT_IDX(slab, i, size)                                           \
+  (pool_item_t *)((unsigned long)slab + (slab->ph_start) +                     \
+                  (i) * (size + sizeof(pool_item_t)))
+
+#define ALIGN_TO_BINARY_POW(x, a) ((((x) >> (a)) + 1) << (a))
```
rafalcieslak: Correct me if I'm wrong, but I think that if `x` is already aligned, then this macro will return a needlessly larger value. Also you may want to use `align` from `common.h`.

sys/pool_alloc.c
MoloToha: Okay, I will change this as soon as klog is ready

tests/pool_alloc_corruption.c
MoloToha: Done

sys/pool_alloc.c
```diff
@@ -0,0 +1,235 @@
...
+typedef struct pool_item {
+  unsigned long pi_guard_number; /* PI_MAGIC_WORD by default, normally isn't */
+  /* changed, so if it has another value, memory */
+  /* corruption has taken place */
+  pool_slab_t *pi_slab; /* pointer to slab containing this item */
+  unsigned long pi_data[0];
+} pool_item_t;
+
+pool_slab_t *create_slab(size_t size, pool_t *pool,
+                         void (*constructor)(void *, size_t)) {
```
rafalcieslak: Is the `constructor` argument necessary? The constructor is stored within `pool`, so it seems redundant.

include/pool_alloc.h
```diff
@@ -0,0 +1,34 @@
...
+typedef struct pool_slab {
+  LIST_ENTRY(pool_slab) ph_slablist; /* pool slab list */
+  vm_page_t *ph_page;                /* page containing this slab*/
+  uint16_t ph_nused;                 /* # of items in use */
+  uint16_t ph_nfree;  /* # of free (and available) items, a bit redundant but
+                         there would be padding abyway*/
+  uint16_t ph_ntotal; /* total number of chunks*/
+  uint16_t ph_start;  /* start offset in page */
+  bitstr_t ph_bitmap[0];
+} pool_slab_t;
```
rafalcieslak: The definition of the contents of this structure is only needed inside `pool_alloc.c`. Moving the structure definition there will make it invisible from other files, which will clean up the interface a bit, and will prevent people from accessing slab metadata.

sys/pool_alloc.c
```diff
@@ -0,0 +1,235 @@
...
+    (sizeof(pool_item_t) + size +
+     1); /* sizeof(pool_slab_t)+n*(sizeof(pool_item_t)+size+1)+7+sizeof(unsigned
+          long)
+   <= PAGESIZE, 7 is maximum number of padding bytes
+   for a bitmap */
+
+  slab->ph_nfree = slab->ph_ntotal;
+  slab->ph_start =
+    sizeof(pool_slab_t) + ALIGN_TO_BINARY_POW(slab->ph_ntotal, 2);
+  memset(slab->ph_bitmap, 0, slab->ph_ntotal);
```
rafalcieslak: I think you are clearing too much with this memset, the size of a bitmap with n entries is `bitstr_size(n)`.

sys/pool_alloc.c
```diff
@@ -0,0 +1,235 @@
...
+
+void *
+pool_alloc(pool_t *pool,
+           __unused uint32_t flags) { /* TODO: implement different flags */
+  /*
+  if (there’s an object in the cache)
+      take it (no construction required);
+  else {
+      allocate memory;
+      construct the object;
```
rafalcieslak: Is this comment accurate? No object construction happens in this function. Maybe you meant slab creation? Anyway, *interleaving* the code with comments makes it simpler to read than providing the outline at the beginning of some functions. This way, when one reads thorough the code, the comments explain which part of the code is responsible for which part of your logic. A general outline before the actual code is only helpful if the code is very complex and the outline significantly simplifies the code.

tests/pool_alloc.c
```diff
@@ -0,0 +1,70 @@
+#include <pool_alloc.h>
+#include <stdc.h>
+#include <malloc.h>
+#include <ktest.h>
+
+void int_constr(void *buf, __attribute__((unused)) size_t size) {
```
rafalcieslak: `common.h` provides `__unused`.

tests/pool_alloc.c
```diff
@@ -0,0 +1,70 @@
...
+  int *num = buf;
+  *num = 0;
+}
+void int_destr(__attribute__((unused)) void *buf,
+               __attribute__((unused)) size_t size) {
```
rafalcieslak: I have a feeling it will be common for many simple objects allocated with the pool allocator to have no destructor. Well, maybe. In that case I would need to provide such pointless destructor function just to satisfy the interface. It might be convenient to have a no-op `pool_default_destructor` defined in `pool_alloc.c` so that one might:
```c
pool_init(&my_pool, size, my_constr, pool_default_destructor);
```
without creating additional empty destructor. We do a similar thing with `vnode` operations: a number of trivial `vnode_generic_*` is provided so that one does not have to write their own empty functions.

sys/pool_alloc.c
```diff
@@ -0,0 +1,235 @@
...
+
+typedef struct pool_item {
+  unsigned long pi_guard_number; /* PI_MAGIC_WORD by default, normally isn't */
+  /* changed, so if it has another value, memory */
+  /* corruption has taken place */
+  pool_slab_t *pi_slab; /* pointer to slab containing this item */
+  unsigned long pi_data[0];
+} pool_item_t;
+
+pool_slab_t *create_slab(size_t size, pool_t *pool,
```
rafalcieslak: `create_slab`, `destroy_slab` and `slab_alloc` are not a part of the interface, right? Make sure to mark these functions as `static`, so that they are inaccessible from other files.

sys/pool_alloc.c
```diff
@@ -0,0 +1,235 @@
...
+  void *p = slab_alloc(slab_to_use, pool->pp_itemsize);
+  pool->pp_nitems--;
+  pool_slab_list_t *slab_list_to_insert =
+    slab_to_use->ph_nfree ? &pool->pp_part_slabs : &pool->pp_full_slabs;
+  LIST_INSERT_HEAD(slab_list_to_insert, slab_to_use, ph_slablist);
+
+  return p;
+}
+void pool_free(pool_t *pool,
+               void *ptr) { /* TODO: implement empty slab management */
```
rafalcieslak: Remember that `TODO` comments must stay informative for a long time. Some (decent!) empty slab management is already implemented, so this comment does not tell me what exactly is missing. Imagine it's year 2020, you don't remember any of this code, and someone tasked you with resolving this `TODO` :) Could you make it a bit more verbose?

tests/pool_alloc.c
```diff
@@ -0,0 +1,70 @@
...
+}
+
+void test_pool_alloc(
+  int flag) { /* 0 - regular test, 1 - memory corrruption, 2 - double free */
```
rafalcieslak: I know, I know, this is just a simple test - but these flags would be much simpler to read and understand if you made them a `#define` or `enum`.

rafalcieslak:
So I gave this branch a close examination and (besides my previous comments) I believe this implementation is sound and solid. Good job :+1:! As you see, most of my remarks are about code style - I really don't have much to complain about how the pool allocator functions. I believe this branch will be ready for merging very soon, although certainly @cahirwpz will have much more stylistic remarks than I do.

MoloToha:
I addressed all issues, the last one remaining is proper calculation of ph_ntotal (total # of items in slab), because atm it is still done by the old formula which uses size of the old bitmap (8 times bigger than the current one) and uses the worst case value for bitmap padding (which is 3 bytes).

sys/pool.c
```diff
@@ -0,0 +1,203 @@
...
+  unsigned long pi_data[0];
+} pool_item_t;
+
+static pool_item_t *get_pi_at_idx(pool_slab_t *slab, unsigned i, size_t size) {
+  return (pool_item_t *)((intptr_t)slab + slab->ph_start +
+                         i * (size + sizeof(pool_item_t)));
+}
+
+static pool_slab_t *create_slab(pool_t *pool) {
+  /* log("create_slab: pool = %p, pp_itemsize = %d", pool, pool->pp_itemsize);
```
cahirwpz: Instead of commenting it out I suggest to use `debug` macro:
```c
#ifdef DEBUG
#define debug(...) log(__VA_ARGS__)
#else
#define debug(...)
#endif

File: sys/pool.c [hunk 18]
```diff
@@ -0,0 +1,203 @@
...
+  }
+  return slab;
+}
+
+static void destroy_slab(pool_t *pool, pool_slab_t *slab) {
+  log("destroy_slab: pool = %p, slab = %p", pool, slab);
+  for (int i = 0; i < slab->ph_ntotal; i++) {
+    pool_item_t *curr_pi = get_pi_at_idx(slab, i, pool->pp_itemsize);
+    if (pool->pp_dtor)
+      pool->pp_dtor(curr_pi->pi_data, pool->pp_itemsize);
```
cahirwpz: Is it necessary to pass `pool->pp_itemsize` here and in other places? What if it the value was present in `pool_slab` structure as well?

sys/pool.c
```diff
@@ -0,0 +1,203 @@
...
+static pool_slab_t *create_slab(pool_t *pool) {
+  /* log("create_slab: pool = %p, pp_itemsize = %d", pool, pool->pp_itemsize);
+   */
+  vm_page_t *page_for_slab = pm_alloc(1);
+  pool_slab_t *slab = (pool_slab_t *)page_for_slab->vaddr;
+  slab->ph_page = page_for_slab;
+  slab->ph_nused = 0;
+  /* sizeof(pool_slab_t) + n * (sizeof(pool_item_t) + size)
+   * + ((n + 31) / 32) * 4 <= PAGESIZE */
+  slab->ph_ntotal = (8 * (PAGESIZE - sizeof(pool_slab_t)) - 31) /
```
cahirwpz: Could you think of expressing it in a simpler way?

cahirwpz:
We're almost there... I'll have another detailed read this week and apply some cosmetic changes.

rafalcieslak:
I am not 100% sure, but it appears as though the last [failed test result](https://travis-ci.org/cahirwpz/mimiker/builds/225681324) was due to spurious qemu lag. Unless we keep experiencing this particular problem (deadlock after ` Sleep 'main' thread on 'callout_good'`), I suggest we ignore it - the callouts interface will be getting a major upgrade soon anyway.

cahirwpz:
@MoloToha Please do not modify this PR anymore if not requested. I'm taking it over in order to clean it up for merge with master.

PR #243, opened 2017-04-06
Formatting fix

czapiga:
Formatting fix after last commit. Actually fixing travis to see if it detects issues on other pull requests


rafalcieslak:
@cahirwpz, this is why we need to fix Travis tests ASAP: it [found the problem](https://travis-ci.org/cahirwpz/mimiker/builds/219243457) in your recent commit, but we're starting to ignore the test results, as they keep reporting the same problem (#234).

PR #252, opened 2017-04-07
Make debugging session with OVPsim and QEMU work without glitches

cahirwpz:
I had experienced many problems trying to create a proper debugging session under gdb / gdbtui or cgdb. This change fixes most of them, but leaves some space for improvements.

### DIFF
--- a/launch
+++ b/launch
@@ -7,15 +7,18 @@ import signal
 import subprocess
 import shlex
 import shutil
+import time
 from pexpect import popen_spawn
 
 PORT = 8000
+
+XTERM = ['x-terminal-emulator', '-geometry', '120x40']
 
 DEBUGGERS = {
     'gdb': '%(triplet)s-gdb %(kernel)s',
     'cgdb': 'cgdb -d %(triplet)s-gdb %(kernel)s',
     'ddd': 'ddd --debugger %(triplet)s-gdb %(kernel)s',
-    'gdbtui': '%(triplet)s-gdb -tui',
+    'gdbtui': '%(triplet)s-gdb -tui %(kernel)s',
     'emacs': 'emacsclient -c -e \'(gdb "%(triplet)s-gdb -i=mi %(kernel)s")\''
 }
 DEFAULT_DEBUGGER = 'cgdb'
@@ -37,7 +40,7 @@ def configure_ovpsim(args):
         'rtc/timefromhost': 1,
         'uartCBUS/console': 1,
         'uartCBUS/portnum': PORT,
-        'command' : '"' + ' '.join(args.args) + '"'
+        'command': '"' + ' '.join(args.args) + '"'
     }
 
     try:
@@ -52,7 +55,7 @@ def configure_ovpsim(args):
                '--wallclock',
                '--kernel', args.kernel]
 
-    if args.test or args.server:
+    if args.test or args.debug:
         OVPSIM_OVERRIDES['uartCBUS/console'] = 0
 
     for item in OVPSIM_OVERRIDES.items():
@@ -77,7 +80,7 @@ def configure_qemu(args):
                 '-serial', 'null',
                 '-serial', 'null']
 
-    if args.server:
+    if args.debug:
         options += ['-serial', 'tcp:127.0.0.1:%d,server,wait' % PORT]
     else:
         options += ['-serial', 'stdio']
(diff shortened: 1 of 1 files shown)


rafalcieslak:
**TL;DR:** You force developers to spawn xterm, ouch.

I'm sorry, but this change completely breaks my workflow, and decreases the flexibility of the launch script. Spawning an xterm is a nice trick, but it should be optional. With this change in place you make it impossible to start a debugging session without creating a new xterm window for kernel UART output, and I would find it very inconvenient.

Let me describe my current workflow so that the limitations imposed by your idea become apparent. I use two terminals (or, usually, two tabs in the same terminal window). One of them keeps running `while true; do nc localhost 8000; done`. The other one starts the `launch` script (with `-s` flag, which redirects UART output via TCP server at 8000). This way the second terminal controls the debugger, while kernel output is visible in the first terminal.

I'm fairly certain I'm not the only student who uses this approach, I believe it was originally popularized by @strzkrzysiek. So what's convenient about it?

 - I can quickly filter kernel output by piping `nc` to `grep` in the first terminal. This seems unnecessary for simple kernel tests, but when running a larger application it is tremendously helpful to track only a couple of messages that I'm currently interested in. Note that this won't be solved by `klog`, because I don't necessarily want to filter messages by their assigned kernel component - grep allows me for a more fine-grained choice.
To achieve the same with this fix in place, I would need to modify the launch script every time I want to change grep arguments. Ouch.

- Similarly, it is frequently useful to store the kernel output into a file (e.g. to compare it with a different run). Using `nc` I can simply redirect it's output `>` into a file, using xterm I would need to modify the command passed to xterm.

- Starting `xterm` opens a new window. For those who open up multiple terminal windows that doesn't seem like a big deal, they keep `alt+tab`-ing between all these windows anyway. But when you use a drop-down style terminal like [tilda](https://github.com/lanoxx/tilda) or [guake](http://guake-project.org/), you want to organize all your shells in tabs (or muxes), because this way switching between them is much faster and predictable, as it does not involve your window manager (and again, it's not just me who uses their computer this way, @coodie is also a fan of a drop-down terminal). More, sometimes it's convenient to run a shell within your text editor. Any extra windows that would appear then are an obstruction. (Also, when I work using my super-sluggish netbook, I frequently don't run an X server at all - but you could argue that this is an extreme case we won't support). Using a TCP connection with `nc` managing where the output will pop up is entirely up to me.

So I'm not really happy with this change, not because it reverts my recent changes, but because it forces the developers to work in one particular way.

However, I believe this can be fixed by slightly changing the direction of your proposed fix.

What I recently wanted to achieve (and the introduction of `-s` flag was as step in this direction) was to separate the debugging options from kernel output options.

The current situation is that we have three ways of dealing with UART messages:

1a. Using no flags, the simulator default behavior is used. In case of ovpsim, this means opening a new terminal (bah!). In case of qemu, the behavior is similar to `-t`. That's kind of inconsistent.
1b. Using `-s` flag kernel output is exposed via TCP server.
1c. Using `-t` flag kernel output is passed to current stdio. This is what Travis uses.

The exact problem you're trying to solve is that when running the debugger via `-d` flag with QEMU and no kernel output redirection, both qemu output and the debugger will share stdio, which leads to all kinds of glitches. I see it not as a problem in **how we execute the debug `-d` mode**, but in **how we interpret kernel output options**. As you see, the options we currently have (1a - 1c) are slightly inconsistent. I suppose we could fix them so that there is:

2a. An option to pass kernel output to an `xterm` window (this may be the default behavior).
2b. An option to expose kernel output via TCP server.
2c. An option to pass kernel output to current stdio.

If these options were **independent** from whether a debugger gets started or not, then this would solve the problems you've been experiencing (well, unless you start a debugger with 2c, but you would need to explicitly request such combination). It would also allow the developers to choose their preferred workflow. It would fix OOTB experience, by starting `xterm` by default (kind of like ovpsim does). And it would improve the consistency of the launch script, by simplifying 1a into 2a.

So what do you think? I understand this would require some extra refactoring of the launch script, but this would solve the targeted problems once and for all, while not limiting developers to one particular way of managing terminals and kernel output.

launch
```diff
@@ -162,15 +163,25 @@ if __name__ == '__main__':
...
                               {'kernel': args.kernel, 'triplet': triplet})
 
+        # start_new_session creates new process group in fresh session,
+        # so when the terminal emulator sends a signal on CTRL+C,
+        # SIGINT will not propagate it to *sim* and *uart* processes
         sim = subprocess.Popen(sim_cmd, start_new_session=True)
-        gdb = subprocess.Popen(dbg_cmd, start_new_session=True)
+        # FIXME Because popen_spawn.PopenSpawn cannot accept start_new_session,
+        # we need to resort to sleep synchronization :(
+        time.sleep(1)
```
rafalcieslak: I suppose this `sleep` might be moved to inside `xterm`, which would achieve the same result, while making the script seem more responsive, as the debugger could start earlier - it's only `nc` who needs to wait.

cahirwpz:
Up to this moment I was not aware of your and other people workflow. Having that in mind I see my approach to solve the problem is flawed.

Nonetheless, as you perhaps noticed, my change was motivated by inconsistent `launch` script behaviour and clash on `stdout` between a simulator and a debugger. The solution you propose seems to leave the most freedom, yet yields consistent user experience.

Just a minor remark - with `-t` option I'd disallow to launch a debugger.

Please fix the script to match proposed behaviour (you can ditch my changes). When it's done, I'll check if my workflow is not affected (and apply fixes if needed) and we'll ask other developers if they experience any problems.

Issue #246, opened 2017-04-07
Basic accounting for threads


cahirwpz:
Following statistics should be recorded in thread's structure:

```c
uint32_t td_runtime; /* ticks spent running */
uint32_t td_slptime; /* ticks spent sleeping */
uint64_t td_nctxsw;  /* total number of context switches */
```

Tick time unit is 1ms.

cahirwpz:
Implemented with #253.

PR #253, opened 2017-04-09
Basic accounting for threads

czapiga:
These changes implements #246


czapiga:
After NetBSD code analysis I think there is little change needed here. I will update this ASAP

cahirwpz:
@rafalcieslak Could you investigate what caused a Travis failure – it's somehow related to `run_tests.py` script...

rafalcieslak:
All it says is that kernel output had a non-ascii byte. That means either there are some non-ascii characters in our source code (some text editors automatically perform fancy substitutions for some characters), or kernel started dumping binary data (unterminated string, invalid ptr). I'm not sure if that's easily fixable by improving `run_tests.py`, I don't know any good way to dump raw binary data to std output in Python.

rafalcieslak:
I hope I'll be able to tell more about this problem once #304 is merged.

rafalcieslak:
I've found the reason for non-ascii characters in output log:

```
Running test ramdisk.
file /usr/include/sys/errno.h:
/* errno is not a global variable, because that would make using it
   non-reentrant.  Instead, its address is returned by the function
   __errno.  */

#ifndef _SYS_ERRNO_H_
#ifdef __cplusplus
extern "C" {
#endif
#define _SYS_ERRNO_H_

#include <sys/reent.h>

#ifndef _REENT_ONLY
#define errno (*__errno())
extern int *__errno _PARAMS ((void));
#endif

/* Please don't use these variables directly.
   Use strerror instead. */
extern __IMPORT _CONST char * _CONST _sys_errlist[];
extern __IMPORT intPÂÂ€
```

Apparently the ramdisk tests reads some invalid data from include headers used for demonstration. I'll open a separate issue addressing the problem.

I have no idea what causes the test to fail, though. `ktest_failure()` does get called, but it gets stuck in
```c
    for (test_entry_t **ptr = autorun_tests; *ptr != NULL; ptr++) {
```
??? `autorun_tests` should never be modified after tests have started, so it looks like some major memory corruption (or Travis output is bonkers).

cahirwpz:
Blocked by #331 as we need to measure time with higher resolution for the tests to pass.

czapiga:
I don't understand travis output.. I did run `make format` and I'm aligned with upstream/master..

czapiga:
@rafalcieslak please, can you look at this from travis POV

tests/thread_stats.c
```diff
@@ -0,0 +1,80 @@
...
+  }
+  for (int i = 0; i < THREADS_NUMBER; i++) {
+    thread_join(threads[i]);
+  }
+  for (int i = 0; i < THREADS_NUMBER; i++) {
+    klog("Thread:%d runtime:%u.%u sleeptime:%u.%u context switches:%llu", i,
+         threads[i]->td_rtime.tv_sec, threads[i]->td_rtime.tv_usec,
+         threads[i]->td_slptime.tv_sec, threads[i]->td_slptime.tv_usec,
+         threads[i]->td_nctxsw);
+    if ((threads[i]->td_rtime.tv_sec == 0 &&
```
cahirwpz: Could you rewrite it in terms of `timeval_isset`?

tests/thread_stats.c
```diff
@@ -0,0 +1,80 @@
+#include <thread.h>
+#include <ktest.h>
+#include <klog.h>
+#include <sched.h>
+#include <runq.h>
+
+#define THREADS_NUMBER 10
+#define TEST_TIME 500
```
cahirwpz: 500 µs? 500ms? 500s? Please use `timeval_t` values to explicitely express time unit:
```c
const timeval_t test_time = TIMEVAL(0.5);
```

tests/thread_stats.c
```diff
@@ -0,0 +1,80 @@
...
+
+static void thread_nop_function(void *arg) {
+  while (tv2st(get_uptime()) < TEST_TIME + tv2st(*(timeval_t *)arg))
```
cahirwpz: Why so complicated? Isn't it just better to calculate everything using `timeval_t` type?

cahirwpz:
@czapiga which version of `clang-format` do you use locally?

cahirwpz:
@czapiga If you can reformulate `timeval_t` API to make it more convenient to use and render code that uses it more concise, then go for it!

czapiga:
@cahirwpz I use clang-format 3.9.1

czapiga:
I think tests can sometimes fail due to bug found during  #342. I will try to finish  #342 ASAP.

czapiga:
@cahirwpz I think it's ready to merge

PR #255, opened 2017-04-09
Launcher script cleanup

rafalcieslak:
I decided to clean up the launch script so that it is easier to maintain and behaves more consistently.

There is still a lot of work ahead - there are some major problems wrt. how interrupts are managed, but I just wanted to demo the direction I'm taking.


rafalcieslak:
Some interesting challenges I've faced:

The `start_new_session` option for `subprocess.Popen` seemed cool, because it creates a new process group for the new process, and thus SIGINT sent to the parent (child) will not be delivered to the child (parent). We need this feature, because we want to sent `Ctrl+C` to the debugger without interrupting the simulator.

However, the problem with using process groups for all subprocesses is that when the parent is SIGTERMed, the children are not, so we need to take care of them in a signal handler. But a SIGKILL to the launch script is a worse problem: it does not kill children, and parent can't react on this! This leaves `qemu` running in background, which is unacceptable for many obvious reasons.

Solving this issue was quite a challenge, but eventually I came up with a simple solution: The debugger becomes the foreground process of the TTY. This way all events such as `Ctrl+C` are delivered directly to the debugger, which is convenient, and lets us manage processes in a sane manner.

Another problem, one which I am unable to fully solve, is that `cgdb` ignores SIGTERM. It registers a universal handler for some signals, including SIGINT, SIGHUP and SIGTERM, but the handler only processes some of these signals, not SIGTERM. That means the only simple way to stop the debugger (e.g. when the user closes `xterm` window) is to SIGKILL it, which gives it no chance to reset terminal configuration. I use `stty sane` to return to some reasonable config, but it's not perfect - and I shouldn't have to deal with such problem in the first place. Bad `cgdb`.

rafalcieslak:
Okay, should be cleaned up and ready for review. Note that this is pretty much rewrite from scratch, so the diff is not the recommended way of reading changes.

Simulator choice, debugger and output direction are now configurable separately. All options conform to a simple shared interface, which simplifies the main script a lot, and makes it much easier to maintain all  options.

Available options, in order of preference:
- `-S` - Simulators: `ovpsim, qemu`
- `-O` - Output: `xterm, server, stdio`
- `-D` - Debuggers: `cgdb, ddd, emacs, gdbtui, gdb`

I've tested all combinations except for `ddd`, `emacs` and `gdbtui`. That includes normal operation, as well as closing/killing processes.

Some convenience flags `-s`, `-t`, `-d` are supported and have kept unchanged meaning, for compatibility. In fact, this change should be entirely backwards-compatible - except for that by default qemu will also use xterm for output.

launch
```diff
@@ -4,196 +4,98 @@ import argparse
...
-    if not sim_default:
-        raise SystemExit('No simulator found!')
-
-    return sim_default, sim_choices
+if __name__ == '__main__':
+    # Check if a cross-compiler is present
+    common.find_toolchain()
 
+    # Probe which tools are available.
+    simulators, sim_default = common.find_available(SIMULATORS)
```
cahirwpz: There's probably no need to expose `SIMULATORS`, instead you could just call `simulators.find_available()` function.

rafalcieslak: That would require exposing the `simulators` module instead! I'll create a `launcher.find_simulators()` proxy in `__init__`.

launcher/common.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, name, needs_foreground=False):
+        self.name = name
+        self.process = None
+        self.cmd = None
+        self.needs_foreground = needs_foreground
+
+    def probe(self):
+        raise NotImplementedError('This launchable is not probable')
+
+    def configure(self, kernel, args="", debug=False, uart_port=8000):
```
rafalcieslak: TIL.

PR #256, opened 2017-04-10
Building initrd

rafalcieslak:
This is my proposed solution for building the initrd with the kernel. The idea is to place some `initrd.rules` files in any place in the source tree. These files define what should be placed in the ramdisk - their format is self-explanatory, see e.g. `user/initrd.rules`:

```
prog/prog.uelf /bin/prog
fd_test/fd_test.uelf /bin/fd_test
test_fork/test_fork.uelf /bin/test_fork
misbehave/misbehave.uelf /bin/misbehave
```

The build script gathers these files and prepares the ramdisk according to these rules. It also generates a set of dependencies for `make`, so that it correctly updates the ramdisk when required.

This method is extensible, universal and simple to maintain.

@coodie, how do you like this script?


coodie:
I don't really like using  Makefile (because It is old and hard to use), and I'd rather avoid using Makefile as much as possible. I'd rather have small bash script which builds initial ramdisk every time it should be built. In this case we have a problem where you want to load `.uelf` files which are inside our build system, therefore ramdisk should be rebuild every time they are changed. Although I hate Makefiles, I must admit that this solution is probably the cleanest, and your script is actually pretty readable.

cahirwpz:
@coodie Please finish the review and approve changes if you don't have any more remarks.

update_initrd.sh
```diff
@@ -0,0 +1,43 @@
...
+rm -rf $BUILDDIR
+mkdir -p $BUILDDIR
+
+SOURCEFILES=
+
+# Gather all files with rules, and copy files to the build dir
+for rulefile in $RULEFILES; do
+    echo "Processing file $rulefile"
+    sourcedir=$(dirname $rulefile)
+    rules=$(cat $rulefile)
```
cahirwpz: You don't use this variable?

update_initrd.sh
```diff
@@ -0,0 +1,43 @@
...
+    while read -r line; do
+        source=$sourcedir/$(echo "$line" | cut  -d' ' -f1)
```
cahirwpz: I prefer to use following syntax to access words within a line:

```
$ cat blah
foo bar boo
$ var=($(<blah)); echo "$var[1]:$var[2]:$var[3]"
foo:bar:boo
```

Splitting line into words is controlled by `IFS` shell variable (Internal Field Separator).

update_initrd.sh
```diff
@@ -0,0 +1,43 @@
...
+DEPFILE=".initrd.D"
+CPIOFILE="initrd.cpio"
+
+### TODO: It is possible to implement this script so that it would only copy the
+### modified files, and incrementally update the ramdisk. However, this is
+### difficult, and we don't keep thousands of files in the ramdisk, so
+### recreating entire cpio archive will be good for now.
+
+# Clean/create the build directory
+rm -rf $BUILDDIR
```
cahirwpz: Should not you clean up `BUILDDIR` after the job is done as well?

rafalcieslak:
I've updated this branch to address mentioned problems, and to reflect on changes introduced by merging cpio support.

update_initrd.sh
```diff
@@ -0,0 +1,42 @@
...
+### recreating entire cpio archive will be good for now.
+
+SOURCEFILES=
+
+# Gather all files with rules, and copy files to the build dir
+for rulefile in $RULEFILES; do
+    echo "Processing file $rulefile"
+    sourcedir=$(dirname $rulefile)
+    while read -r line; do
+        read source target <<< $line
```
cahirwpz: Nice! I did not know this syntax!

rafalcieslak:
So do we merge this?

PR #257, opened 2017-04-13
Implemented dev_uart_read

rafalcieslak:
This is just filling in the missing glue between `uart_cbus` and `dev_uart`, so that user programs can get some bytes from uart input. 

A proper solution will require device infrastructure and terminal support - but we're far away from that.

### DIFF
--- a/sys/dev_uart.c
+++ b/sys/dev_uart.c
@@ -22,12 +22,28 @@ static int dev_uart_write(vnode_t *t, uio_t *uio) {
 }
 
 static int dev_uart_read(vnode_t *t, uio_t *uio) {
+  char buffer[UART_BUF_MAX];
+  unsigned curr = 0;
+  while(curr < UART_BUF_MAX && curr < uio->uio_resid){
+    buffer[curr] = uart_getchar();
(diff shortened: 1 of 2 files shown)


sys/dev_uart.c
```diff
@@ -22,12 +22,28 @@ static int dev_uart_write(vnode_t *t, uio_t *uio) {
 }
 
 static int dev_uart_read(vnode_t *t, uio_t *uio) {
+  char buffer[UART_BUF_MAX];
+  unsigned curr = 0;
+  while (curr < UART_BUF_MAX && curr < uio->uio_resid) {
+    buffer[curr] = uart_getchar();
```
cahirwpz: This should not busy-loop but rather use `sleepq_wait`, while interrupt handler should call `sleepq_signal`.

rafalcieslak: I wanted this to be a temporary solution until we implement support for configuring interrupts.

cahirwpz: Ok. I'm currently working on intr-sketch branch to implement basic functionality for handling interrupts.

PR #258, opened 2017-04-13
Experiments with driver interface

rafalcieslak:
Here are my initial attempts at preparing a driver interface. I've decided to open this PR very early, so that you can comment on whether my direction is sound.

This interface is mostly inspired by LDD3 (https://lwn.net/Kernel/LDD3/) with some details based on FreeBSD implementation. My version is stripped down to absolute minimum, so that we have a solid hierarchy of data structures and their responsibilities, without cluttering the interface with features we won't be using right now.

At this point there is some partial support for PCI bus (for the most part I've extracted platform-independent parts from `mips/pci.c` to `sys/pci_bus.c`), PCI drivers (an example driver for `82371EB` is present, but it doesn't do anything at all) and device-driver matching and probing.

I will continue this work so that the `82371EB` actually does something useful - ideally I would use it to access the 16550 beneath it.

Note that I am **NOT** using any source code from Linux.

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -2,6 +2,7 @@ set tcp connect-timeout 30
 target remote localhost:1234
 break main
 break assert_fail
+break kernel_init
 continue
 
 source debug/ktrace.py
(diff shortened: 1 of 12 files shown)


cahirwpz:
The sooner we meet and discuss our ideas for driver infrastructure the better is for the project. I hope we'll smoothly converge at reasonable solution that could be quickly implemented.

rafalcieslak:
This branch is obsolete and no longer provides useful insight.

PR #259, opened 2017-04-14
Removed files embedded in kernel image

rafalcieslak:
We no longer need the silly trick of embedding files into kernel image - we can use ramdisk for providing files. This simplifies build and linking process.

- Removed make rules related to `.uelf.o`.
- Moved `/initrd` mount point to `/`.
- Renamed references to user programs to paths in `/bin`.
- Linked `do_exec` with `vfs` interface.
- Implemented absolutely minimal stub of `getattr` for `initrd` - @coodie, I suppose the rest will be up to you.
- Fixed a minor bug in intird where it would return invalid errno code.


sys/exec.c
```diff
@@ -107,12 +107,35 @@ int do_exec(const exec_args_t *args) {
...
+  assert(eh.e_phentsize < 128);
+  /* We allocate this buffer on stack. It's confirmed to be small. A statically
+     allocated buffer won't do, because it would need restrictive
+     synchronization. Using a malloc pool would be probably an overkill for this
+     case. */
+  size_t phs_size = eh.e_phoff * eh.e_phentsize;
+  char phs[phs_size];
+
+  /* Read program headers. */
+  uio.uio_op = UIO_READ;
```
cahirwpz: This piece of code looks familiar and apparently occurs in several places in the kernel. Could we wrap it with `uio_make(&uio, UIO_READ, ...)` function?

rafalcieslak: I think it's likely that @coodie will be doing some `uio` improvements due to his findings at #261, so let's introduce such convenience function separately.

PR #260, opened 2017-04-14
Add initrd_open but mostly tests

coodie:
I took a look at current implementation of initrd and realised that open operation was missing. However it turns out it was very easy to add, because of `vnode_open_generic`.


rafalcieslak:
Please merge from master - the ramdisk is now mounted in the root `/` directory.

coodie:
With #261 addition this PR is no longer necessary, because reading from initrd is already tested there, and vnode_generic_open is added to initrd file system operations.

PR #261, opened 2017-04-17
Uiomove offset

coodie:
While I've been working on getdirents syscall I had 'elegancy problem'. My implementation of getdirents system call was incredibly ugly and I couldn't really come up with elegant way of fixing it.

 I have looked into NetBSD code and noticed that it is indeed fault of our implementation of 'uiomove'. In our case `uio_offset` is IN argument of uio, while in fact this argument is 'OUT' argument supposed to show how much of the entire UIO operation was shifted. This argument is also used in setting proper file reading offsets. I can explain it in more details on seminary. Here is link to the NetBSD's implementaiton of uiomove:

http://bxr.su/NetBSD/sys/kern/subr_copy.c#99

Here is example usage in tmpfs in getdirents operation:

http://bxr.su/NetBSD/sys/fs/tmpfs/tmpfs_subr.c#807

Note how they check `uio->uio_offset` in their implementation and based on that they do something. I haven't looked into other operations, but I'm pretty confident `uio_offset` is also used there in similar way.

### DIFF
--- a/include/uio.h
+++ b/include/uio.h
@@ -22,5 +22,6 @@ typedef struct uio {
 } uio_t;
 
 int uiomove(void *buf, size_t n, uio_t *uio);
+int uiomove_frombuf(void *buf, size_t buflen, struct uio *uio);
 
 #endif /* !_SYS_UIO_H_ */
--- a/sys/initrd.c
+++ b/sys/initrd.c
@@ -195,7 +195,7 @@ static int initrd_vnode_lookup(vnode_t *vdir, const char *name, vnode_t **res) {
 static int initrd_vnode_read(vnode_t *v, uio_t *uio) {
   cpio_node_t *cn = (cpio_node_t *)v->v_data;
   int count = uio->uio_resid;
-  int error = uiomove(cn->c_data + uio->uio_offset, cn->c_size, uio);
+  int error = uiomove_frombuf(cn->c_data, cn->c_size, uio);
 
   if (error < 0)
     return -error;
--- a/sys/initrd.c
+++ b/sys/initrd.c
@@ -195,7 +195,7 @@ static int initrd_vnode_lookup(vnode_t *vdir, const char *name, vnode_t **res) {
 static int initrd_vnode_read(vnode_t *v, uio_t *uio) {
   cpio_node_t *cn = (cpio_node_t *)v->v_data;
   int count = uio->uio_resid;
-  int error = uiomove(cn->c_data+uio->uio_offset, cn->c_size, uio);
+  int error = uiomove(cn->c_data + uio->uio_offset, cn->c_size, uio);
 
   if (error < 0)
     return -error;
--- a/sys/initrd.c
+++ b/sys/initrd.c
@@ -245,6 +245,7 @@ void ramdisk_init() {
   if (rd_size) {
     initrd_ops.v_lookup = initrd_vnode_lookup;
     initrd_ops.v_read = initrd_vnode_read;
+    initrd_ops.v_open = vnode_open_generic;
     log("parsing cpio archive of %zu bytes", rd_size);
     read_cpio_archive();
     initrd_build_tree_and_names();
--- a/sys/initrd.c
+++ b/sys/initrd.c
@@ -195,7 +195,7 @@ static int initrd_vnode_lookup(vnode_t *vdir, const char *name, vnode_t **res) {
 static int initrd_vnode_read(vnode_t *v, uio_t *uio) {
   cpio_node_t *cn = (cpio_node_t *)v->v_data;
   int count = uio->uio_resid;
-  int error = uiomove(cn->c_data, cn->c_size, uio);
+  int error = uiomove(cn->c_data+uio->uio_offset, cn->c_size, uio);
 
   if (error < 0)
     return -error;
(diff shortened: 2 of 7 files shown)


rafalcieslak:
I appreciate your time investigating this, and I admit our current handling of `uio_offset` is incomplete (and I am the one to blame for that). However, I am not convinced that the offset is just an *out* argument! My idea of the purpose of the offset field is such that it should be possible to call `uiomove` multiple times in sequence, which should process subsequent data (this may be useful e.g. when some target buffer is smaller than the amount of data we want to transfer). The offset field would be updated on each call to `uiomove` so that the next call will begin transferring data from where it previously ended. That would make the offset an *in/out* argument.

However, NetBSD's implementation of `uiomove` indeed ignores the input value of the offset - as you probably have noticed. I suppose it is this way because it's a generic implementation of `uiomove`, and it is supposed to be relevant even for devices which do not support the concept of `position`. Please have a look at [`uiomove_frombuf`](http://bxr.su/NetBSD/sys/kern/subr_copy.c#154), which actually reads the offset from `uio`, and passes it to `uiomove` with `pos + offset, size - offset` arguments (!!!). Apparently our implementation of `uiomove` has the semantics of NetBSD's `uiomove_frombuf`!

The example you've linked to is actually a proof *against* your argument. Notice the call to `uiomove` appears **after** the `uio_offset` gets validated. The `tmpfs_dir_getdents` function is supposed to process a data transfer as requested by the `uio` argument - and to achieve that it starts by checking whether the `uio_offset` argument matches one of specific special cases (and if it does, a different function gets to process  the request). That confirms that the `uio_offset` is an *input* argument - it just gets processed early ([see line 840](http://bxr.su/NetBSD/sys/fs/tmpfs/tmpfs_subr.c#840)) to determine *which buffer* we'll be reading from, and therefore afterwards the actual call to `uiomove` can ignore the offset.

So it seems that you are partially right. The `uio_offset` is both input and output argument. It should be updated by `uiomove`, and sometimes it may be useful to check it after performing uiomove. But it should be also possible to use it as an input argument, to specify the position in target buffer where the data transfer should begin.

This is indeed a complex mechanism, so it's good you've put some extra efforts into investigating it, and it would be great if you could explain these issues on our next meeting so that everyone can have a clear idea on how to use `uio`. The part where you fix `uiomove` so that it increments `uio_offset` is an important fix. But don't make it ignore the input value of `uio_offset` - although it might be a good idea to split `uiomove` into `uiomove` and `uiomove_frombuf`, with only the former taking the offset into account. You may want to investigate how `uiomove_frombuf` is used in practice.

coodie:
As far as I understand this, the offset argument simply represents how much data we have currently transferred using the uio we already have. Because If we pass our uio many times around program transferring data between different buffers we might lose track of how much data we transferred. This data is incredibly useful in many places. One good example might be executing read on some file descriptor many times and trying to keep track of offset in file. If uiomove performs copying operation on `buf+offset` instead 'buf' then it becomes nearly impossible to implement lazy reading efficiently. Also, what if our data representation is not contiguous array but it is some list (as it is in the case of readdir). Normally we would probably have to transfer it to array representation and then run uiomove on that, or make some nasty shifting and passing `buf-offset` making the final implementation look terrible (I've run into that while working on readdir). Otherwise we can just locate proper element of the list (which represents offset we are supposed to read from), and starting from there run uiomove on each element of the list until we satisfy the uiomove request. 

I'm 100% sure this is correct solution and I'm not going to work on readdir until this is merged with master.

rafalcieslak:
Please read my previous comment carefully, because you are clearly missing some important details. I agree with the points you've raised. In particular, the solution I suggested (separating `uiomove` and `uiomove_frombuf`) solves the issue of lazy reading and accessing noncontiguous data. More so, it solves it *the way you want it to be solved* - while still offering the sporadic convenience of current implementation. If you really can't see my point, then I guess we'll have to discuss it in person, because I am strongly against merging this proposal *in its current form*.

coodie:
Oh well. I think I get your point. At this moment this isn't really that big issue, because I don't see where uiomove_frombuf might be used. 

I think this would be good to add `uiomove_frombuf` function anyway, so someone who isn't familiar with uio has this function already. I think adding this function to `vnode_initrd_read` will provide good example.

rafalcieslak:
My experiments suggest that `uiomove_frombuf` may be very useful for device drivers, free physmem access and buffered I/O.

Since you're recently working with many `uio`s, I thought you may want to implement #262 as well, either in this or a separate PR.

As a side note, we are currently missing an important test case, which would verify that we manage file positions correctly. Such test would have to call `read` (from userspace) multiple times and check if different parts of the file are read, it would also have to see whether `lseek` works as expected. This is somewhat related to this PR, as you slightly change the way these offsets are managed, and we have no test to confirm you didn't break anything. [However, I do have manually tested this branch with DOOM which **heavily** relies on these offsets being correct, and I've not found any problems.]

cahirwpz:
I'm pleased to see that you sorted out the issue by yourselves :-)

Having read your discussion about `uiomove` and `uiomove_frombuf`, comparing corresponding manual pages and peeking at FreeBSD and NetBSD implementations I'm convinced this PR fixes important issue and is ready to be merged.

However, as @rafalcieslak suggested, I'd like to see extra test case(s) (if possible) that would trigger the bug in the original code. As for #262 - it's probably better to implement it in another PR.

coodie:
In order to write tests you want I need to add `vnode_generic_open` which is added in other PR. This is one line change and I hope you don't mind I added this in this PR. There is big part of lseek implementation missing and I don't think this is relevant for this PR.

rafalcieslak:
Yes, current `lseek` is very limited. Please merge from `master`! We've recently started mounting the ramdisk in `/` instead of `/initrd`.

PR #264, opened 2017-04-17
Fixed file attr structure's newlib compatibility

rafalcieslak:
We've recently defined our `dev_t`, `ino_t`, `mode_t` (probably when the cpio ramdisk was introduced). Since then our `struct vattr` is no longer binary compatible with newlib's `struct stat`. This results in user programs being unable to correctly ask about file size etc! This branch fixes our vattr structure definition (in newlib compatibility mode) so that it matches newlib's layout.

I guess we'll have to eventually look into porting newlib, but until then this will be enough to get some userspace programs running correctly.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -13,13 +13,13 @@ typedef unsigned long pm_addr_t;
 typedef long off_t;
 typedef long ssize_t;
 typedef int32_t pid_t;
-typedef uint32_t dev_t;
+typedef uint16_t dev_t;
 typedef uint32_t time_t;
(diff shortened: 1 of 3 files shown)


cahirwpz:
I think it's better to align the size of `dev_t`, `ino_t`, `mode_t` and other system-wide types with syscall interface rather that particular filesystem driver. As such I'd prefer to declare `cpio_dev_t` type instead of `newlib_dev_t`. Does it make sense?

PR #265, opened 2017-04-18
Fixed occasional deadlock on `thread_join`

rafalcieslak:
A closer investigation uncovered that `sched_switch` is flawed and it's critical section ended too soon. Before `ctx_switch` re-disables interrupts, they are enabled for a short while by `critical_leave()`. If the thread gets preempted at that very moment, we end up with multiple threads marked as `RUNNING`, neither are added to run queue, etc. which caused the deadlock we occasionally observed in Travis results.

This fix extends the critical section until after `ctx_switch`, which prevents the described problem. I've restarted the Travis test 10+ times without a single failure.

I expect this is the last change needed to fix Travis false negatives.

### DIFF
--- a/sys/sched.c
+++ b/sys/sched.c
@@ -83,7 +83,7 @@ void sched_switch(thread_t *newtd) {
   newtd->td_state = TDS_RUNNING;
   critical_leave();
 
-  if (td != newtd){
+  if (td != newtd) {
     log("ss %lu | %lu", td->td_tid, newtd->td_tid);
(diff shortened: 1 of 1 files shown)


cahirwpz:
The fix makes 100% sense. It's a pity I didn't find it during a code review -_-

rafalcieslak:
For the record, this issue was introduced by #118, so a very long time ago. I suppose perfect synchronization wasn't that important back then.

PR #266, opened 2017-04-18
Decouple console functionality from UART driver code

cahirwpz:
The change aims at decoupling CBUS UART driver code from console-like device functionality. The interface is loosely based on *BSD console device interface present in [cons.h](http://bxr.su/NetBSD/sys/dev/cons.h) and described in [cons(9)](http://netbsd.gw.com/cgi-bin/man-cgi?cons+9+NetBSD-current).

To make linker sets work for files in `mips` subdirectory I had to modify top-level `Makefile` (that took me some time to figure out).

In long run I'm aiming at decoupling driver code from hardware resource with help of [bus_space(9)](http://netbsd.gw.com/cgi-bin/man-cgi?bus_space+9+NetBSD-current)-like interface which has been partially implemented in [bus-space](https://github.com/cahirwpz/mimiker/compare/bus-space) branch.

### DIFF
--- a/include/console.h
+++ b/include/console.h
@@ -1,0 +1,37 @@
+#ifndef _SYS_CONSOLE_H_
+#define _SYS_CONSOLE_H_
+
+#include <linker_set.h>
+
+typedef struct console console_t;
+
+typedef void (*cn_init_t)(console_t *);
+typedef int (*cn_getc_t)(console_t *);
+typedef void (*cn_putc_t)(console_t *, int);
+
+typedef struct console {
+  cn_init_t cn_init;
+  cn_getc_t cn_getc;
+  cn_putc_t cn_putc;
+  int cn_prio;
+} console_t;
+
+#define CONSOLE_DEFINE(name, prio) \
+  console_t name##_console = { \
+    .cn_init = name##_init, \
+    .cn_getc = name##_getc, \
+    .cn_putc = name##_putc, \
+    .cn_prio = (prio),\
+  }; \
+  SET_ENTRY(cn_table, name##_console)
+
+#define CONSOLE_DECLARE(name) \
+  extern console_t name##_console
+
+void cn_init();
+int cn_getc();
+void cn_putc(int c);
+int cn_puts(const char *s);
+int cn_write(const char *s, unsigned n);
+
+#endif
--- a/include/console.h
+++ b/include/console.h
@@ -16,17 +16,16 @@ typedef struct console {
   int cn_prio;
 } console_t;
 
-#define CONSOLE_DEFINE(name, prio) \
-  console_t name##_console = { \
-    .cn_init = name##_init, \
-    .cn_getc = name##_getc, \
-    .cn_putc = name##_putc, \
-    .cn_prio = (prio),\
-  }; \
+#define CONSOLE_DEFINE(name, prio)                                             \
+  console_t name##_console = {                                                 \
+    .cn_init = name##_init,                                                    \
+    .cn_getc = name##_getc,                                                    \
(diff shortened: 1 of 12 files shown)


cahirwpz:
@rafalcieslak, @dyniec: Please review changes.

mips/malta.c
```diff
@@ -172,7 +172,7 @@ void platform_init(int argc, char **argv, char **envp, unsigned memsize) {
 
   setup_kenv(argc, argv, envp);
 
-  uart_init();
+  cn_init();
```
rafalcieslak: I suppose some consoles may appear later in the boot process, e.g. as result of probing available devices. Do we plan to have a mechanism for re-selecting the best available console?

Also some UART devices (e.g. the two 16550s on ISA bus) can't be probed for availability, and it's up to the platform configuration to decide whether they are present or not. Supporting them is barely a matter of picking slightly different register addresses. I though it might be cool to direct kernel messages to the "ttyCBUS", while user programs would use "tty0" (@ `0x03f8`). Though that would require multiple independent consoles, and a way of opening a particular one...

mips/uart_cbus.c
```diff
@@ -29,7 +31,7 @@ void uart_init() {
   LCR = LCR_8BITS; /* 8-bit data, no parity */
 }
 
-int uart_putchar(int c) {
+static void cbus_uart_putc(console_t *dev __unused, int c) {
```
rafalcieslak: I'm interested where would buffering come into play once we decide to implement it. I suppose it would be up to the driver, so this function would store data into a synchronized cyclic buffer and return immediately, while another thread would wait until next byte can be written (presumably signaled by an interrupt), and read it from the buffer. Is my understanding correct?

include/console.h
```diff
@@ -0,0 +1,36 @@
...
+typedef void (*cn_putc_t)(console_t *, int);
+
+typedef struct console {
+  cn_init_t cn_init;
+  cn_getc_t cn_getc;
+  cn_putc_t cn_putc;
+  int cn_prio;
+} console_t;
+
+#define CONSOLE_DEFINE(name, prio)                                             \
```
cahirwpz: Right, I did not take into account code browsability factor. It's good you pointed it out, as I started using this construct in consequent changes.

mips/malta.c
cahirwpz: You're right, but we'll take care of it later, when more than one console device is handled. Corresponding function in FreeBSD is [cnadd](http://bxr.su/FreeBSD/sys/kern/kern_cons.c#168). For starter we'll leave `tty0` and `tty1` for use by user-space application.

include/mips/uart_cbus.h
```diff
@@ -1,42 +1,8 @@
...
 
-#include <stddef.h>
+#include <console.h>
 
-/*
- * Raw UART2 interface. Useful for debugging.
- * A higher-level solution will have to be implemented eventually, but
- * temporarily this should be fine.
- */
+CONSOLE_DECLARE(uart_cbus);
```
cahirwpz: Right!

mips/uart_cbus.c
cahirwpz: The driver will allocate a cyclic buffer (80 characters?) from generic memory pool available to all device drivers. `cn_putc` will put a character into the buffer, and then check if THRE (transmitter hold register empty) bit is set. If so it will initiate transfer that will be continued by interrupt handler routine if the buffer contains more than one character. If buffer is full the thread will be suspended. It will be woken up by interrupt handler routine when there's some space in the buffer to put data in.

Issue #262, opened 2017-04-17
Convenience function for preparing `uio`


rafalcieslak:
The following `uio` setup appears multiple times in kernel source:

```c
  uio.uio_op = UIO_STH;
  uio.uio_vmspace = get_some_vm_map();
  iov.iov_base = buffer;
  iov.iov_len = sizeof(buffer_data);
  uio.uio_iovcnt = 1;
  uio.uio_iov = &iov;
  uio.uio_offset = 0;
  uio.uio_resid = sizeof(buffer_data);
```

so it might be convenient to prepare a function `uio_make` or `uio_prepare` or `uio_setup` which fills in these fields in a single expression. It may need to be a macro, because it may have to create a `iov` on stack. It might be used like this `uio_make(&uio, UIO_READ, ...);`.

This was originally suggested [in this discussion](https://github.com/cahirwpz/mimiker/pull/259#discussion_r111661346).

rafalcieslak:
Resolved by merging #282.

PR #268, opened 2017-04-19
Getdirentries syscall

coodie:
This PR adds implementation of getdirentries system call.


tests/initrd.c
```diff
@@ -4,6 +4,20 @@
 #include <mount.h>
 #include <vm_map.h>
 #include <ktest.h>
+#include <dirent.h>
+
+static void prepare_uio(uio_t *uio, iovec_t *iov, char *buffer, int buf_size) {
```
rafalcieslak: If you're introducing such function, please make it globally accessible (#262) or don't use it at all. It multiple files bring their own version of uio setup routine, we'll end up with a confusing interface that is difficult to unify.

coodie: I have added prepare_uio function in another PR. After that PR gets merged with master I'll update this branch.

rafalcieslak:
I've managed to reproduce this issue with 100% success rate. The kernel works okay if I build my own `.cpio` archive, but it crashes with the one built by Travis. Note that Travis (probably) runs an older version of `cpio` than you do, so there may be some slight differences. For your convenience, here's the (zipped) `.cpio` archive as built by Travis: [travis-initrd.zip](https://github.com/cahirwpz/mimiker/files/956440/travis-initrd.zip). If you use it as the ramdisk for kernel, you should be able to investigate the nature of the crash. I don't know exactly the internals of a `cpio` archive, so you'll have to take it from here - good luck.

coodie:
@rafalcieslak @cahirwpz 

I think this branch is ready to be merged (or reviewed at least).

sys/dev_vga.c
```diff
@@ -102,10 +103,43 @@ static int dev_vga_lookup(vnode_t *v, const char *name, vnode_t **res) {
...
+static void uiomove_vga_dirent(const char *name, uio_t *uio) {
+  dirent_t *dir = NULL;
+  dir = (dirent_t *)kmalloc(M_VFS, _DIRENT_RECLEN(dir, strlen(name)), M_ZERO);
+  dir->d_namlen = strlen(name);
+  dir->d_reclen = _DIRENT_RECLEN(dir, dir->d_namlen);
+  memcpy(dir->d_name, name, dir->d_namlen + 1);
+  uiomove(dir, dir->d_reclen, uio);
+  kfree(M_VFS, dir);
+}
+
```
cahirwpz: Please declare following array here:
```c
typedef struct direntinfo {
  uint8_t d_type;
  char *d_name;
} direntinfo_t;

static direntinfo_t vga_dir[] = {
  {DT_BLK, "fb"},
  {DT_BLK, "palette"},
  {DT_CHR, "videomode"},
};
```
... and use it to create a memory block that consists of `dirent`s.

coodie: I have no idea what most of these flags mean and I don't really know how to use them, but notice that:

```
#define DT_CHR 2
#define DT_DIR 4
#define DT_BLK 6
```

Meaning that `DT_DIR & DT_BLK` will give `true` and in case of searching over directories we will want to enter fb and pallete which aren't directories.

cahirwpz: I don't think they are flags, but rather type enumerators (DT = Directory entry Type).

sys/dev_vga.c
```diff
@@ -102,10 +103,43 @@ static int dev_vga_lookup(vnode_t *v, const char *name, vnode_t **res) {
...
+      uio->uio_resid >= _DIRENT_RECLEN(dir, strlen("fb")))
+    uiomove_vga_dirent("fb", uio);
+
+  if (uio->uio_offset == palette_offset &&
+      uio->uio_resid >= _DIRENT_RECLEN(dir, strlen("palette")))
+    uiomove_vga_dirent("palette", uio);
+
+  if (uio->uio_offset == videomode_offset &&
+      uio->uio_resid >= _DIRENT_RECLEN(dir, strlen("videomode")))
+    uiomove_vga_dirent("palette", uio);
```
cahirwpz: Please don't mix code and data. You've made a classic mistake here (copy & paste). Put data into separate array and write a procedure that will transform `vga_dir` to `uio`-able buffer.

cahirwpz:
The changes I'm not going to review yet were moved to [getdirentries_tests](https://github.com/cahirwpz/mimiker/tree/getdirentries_tests) branch. We'll continue the work from there.

PR #269, opened 2017-04-21
Decouple ns16550 driver code from hw resource access method

cahirwpz:
Main purpose of this PR is to show how `bus_space` and `resource` structures help to decouple driver code from actual hardware resource(s) (in this case 16550 UART register set). Same code should be capable of handling `tty0` and `tty1` by just replacing `cbus_uart` in `in` and `out` functions to resources describing UART devices in *FDC37M81x* controller attached to ISA bus in *PIIX4* chipset.

Next step is to perform similar task (i.e. decoupling) for *GT-64120* PCI-bridge chipset by introducing `pci_read_config` and `pci_write_config` functions and corresponding function table.

### DIFF
--- a/include/bus.h
+++ b/include/bus.h
@@ -1,0 +1,52 @@
+#ifndef _SYS_BUS_H_
+#define _SYS_BUS_H_
+
+#include <common.h>
+
+typedef struct resource resource_t;
+typedef struct bus_space bus_space_t;
+
+/* `bus space` accessor routines */
+typedef uint8_t (*bus_space_read_1_t)(resource_t *handle, unsigned offset);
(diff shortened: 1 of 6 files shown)


rafalcieslak:
This looks interesting. I'm still not clear on the wider picture - like managing dynamically discovered resources, or binding a driver to a particular resource (this driver is hardcoded to use the `cbus_uart`, I expected it to be linked with the resource somewhere during platform-specific startup process) - but I see you're already preparing a `pci` branch which should demonstrate most concepts I'm missing. I don't have any comments to this implementation, it seems straight-forward.

PR #271, opened 2017-04-23
Basic signal support

rafalcieslak:
I've implemented basic signal sending, processing and delivery. This branch is pretty heavy, but I don't see a way of introducing this changes in smaller parts.

Quick overview first. What works:
 - Sending a signal from userspace (`raise`, `kill`).
 - Sending a signal from kernel space (`SIGSEGV` is sent to the offending thread on memory violation).
 - Customizing thread reaction on a signal (`signal`, `sigaction`, `SIG_IGN`, `SIG_DFL`, or a custom handler function).
 - Recursive signal handlers.

What does not work yet:
 - Signal masking.
 - Optionally switching to a different stack space for user signal handlers (I'm pretty sure this is a legacy optimization).
 - Thread/process distinction (in POSIX sense) does not exist, a signal is always delivered to one particular thread, threads never share the same PID.
 - Proper newlib integration. Newlib tries to emulate signals in userspace - by mocking `signal` and `raise` functions with some silly routines that deliver signals only to the thread that raised them. We'll need to recompile newlib with `-DSIGNAL_PROVIDED`, to disable this. We'll also need to provide newlib internals with our signal numbers - on a failed assertion newlib calls `kill(getpid(), 6)` hoping to use `SIGABRT`, but 6 is `SIGKILL` in this implementation.
 - `SIGSTOP` and `SIGCONT`.
 - Proper automatic testing of signals. Preparing an automated test will be very difficult until we have a `wait` and `waitpid`. I've prepared some demonstrations in `prog.c`, but they need to be examined manually.
 - `sigsuspend`.

My implementation is a simplified fusion of ideas from the FreeBSD design book, FreeBSD, NetBSD and Linux implementations.

The *trampoline* problem we've been discussing for many weeks turned out to be easily solvable. Very old Linux versions and pretty old FreeBSD version used a mechanism where the kernel would copy a "return from sighandler" procedure onto user process stack. Nowadays the stack is protected from execution, and this approach would not work. We wanted to overcome it by mapping a single page with executable code into process memory map... but that seems complicated, and requires the memory system to support such tricks. What Linux does is much simpler - the "return from sighandler" procedure (commonly called *trampoline*) is provided by the C library, not by the kernel.  The `sigaction` structure has a field `sa_restorer` which has to contain the address where a sighandler should return. The C library wraps `sigaction` and `signal` functions so that it fills in that field with a pointer to a trampoline function it provides. In our case, we don't need to do any cleanup after a signal handler has finished, so it's okay to use `sigreturn` syscall proxy address as the value of `sa_restorer`. I think this is a very simple and elegant solution.

The main reasons why my implementation is so much simpler than other system's is that 
1) we don't support thread groups (processes), process groups and sessions 
2) our threads do not have that many states they can be in when a signal is delivered
3) we don't support multiple processors yet
4) we don't need to stay compatible with a ton of legacy programs that exploited the way signals worked in a very old kernel version.

With these assumptions, signals are a pleasure to implement.


sys/signal.c
```diff
@@ -0,0 +1,221 @@
...
+    [SIGFPE] = "SIGFPE",   [SIGSEGV] = "SIGSEGV", [SIGKILL] = "SIGKILL",
+    [SIGTERM] = "SIGTERM", [SIGCHLD] = "SIGCHLD", [SIGUSR1] = "SIGUSR1",
+    [SIGUSR2] = "SIGUSR2",
+};
+
+void sighand_init() {
+  /* TODO: A lot of kernel components have a init function whose only purpose is
+     to perform a kmalloc_init. Maybe we could collect them with a linker set,
+     and initialize all pools together? */
+  kmalloc_init(sig_pool, 2, 2);
```
cahirwpz: I'm growing convinced that after pool allocator is merged we'll switch to single memory pool for `kmalloc`. I find passing memory pool to `kmalloc` functions cumbersome. NetBSD counterpart [malloc(9)](http://netbsd.gw.com/cgi-bin/man-cgi?malloc+9+NetBSD-current) is deprecated in favour of system-wide memory allocator: 

> These interfaces are being obsoleted and their new use is discouraged. For new code, use kmem(9) for variable-sized or one-time allocations and pool_cache(9) for frequent fixed-size allocations instead.

sys/signal.c
```diff
@@ -0,0 +1,221 @@
...
+
+static const char *signal_names[SIG_LAST] = {
+    [SIGINT] = "SIGINT",   [SIGILL] = "SIGILL",   [SIGABRT] = "SIGABRT",
+    [SIGFPE] = "SIGFPE",   [SIGSEGV] = "SIGSEGV", [SIGKILL] = "SIGKILL",
+    [SIGTERM] = "SIGTERM", [SIGCHLD] = "SIGCHLD", [SIGUSR1] = "SIGUSR1",
+    [SIGUSR2] = "SIGUSR2",
+};
+
+void sighand_init() {
+  /* TODO: A lot of kernel components have a init function whose only purpose is
```
cahirwpz: Mechanism similar to [SYSINIT](https://www.freebsd.org/doc/en/books/arch-handbook/sysinit-using.html) could be implemented to achieve it. It's not necessary to introduce it right now, altough if specified in details it could be a nice addition to task list (under category of *wish*).

sys/signal.c
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /* If the signal is ignored, don't even bother posting it. */
+  sighand_t *sh = target->td_sighand;
+  if (get_sigact(sig, sh) == SIG_IGN)
+    return 0;
+
+  bit_set(target->td_sigpend, sig);
+
+  /* TODO: If a thread is sleeping, wake it up, so that it continues execution
```
cahirwpz: What are the consequences of ignoring the kinds of sleep? Would it be the time to introduce the distinction between state `blocked` (on turnstile), `sleep` (on sleepqueue) and `interruptible sleep` (signal-aware, on sleepqueue)?

sys/signal.c
```diff
@@ -0,0 +1,221 @@
...
+
+void sighand_init() {
+  /* TODO: A lot of kernel components have a init function whose only purpose is
+     to perform a kmalloc_init. Maybe we could collect them with a linker set,
+     and initialize all pools together? */
+  kmalloc_init(sig_pool, 2, 2);
+}
+
+sighand_t *sighand_new() {
+  sighand_t *sh = kmalloc(sig_pool, sizeof(sighand_t), M_ZERO);
```
rafalcieslak: The description of signal handlers may be shared between threads (`CLONE_SIGHAND`). This is useful for multi-threaded processes, where, typically, signal handler configuration is shared by all threads.

sys/signal.c
rafalcieslak: The exact consequences of thread state on signal posting is well explained in the FreeBSD "Design and Implemnetation" book, chapter 4.7, "Posting of a signal".

Also, I am unsure what would exactly happen if we wake up a thread sleeping on a sleepqueue. It might continue execution prematurely, causing it to ignore the event it was waiting for. This can cause problems if the sleeping thread does not re-check if the condition it was waiting for is true. And if it does re-check, then it will probably go to sleep again. I suppose this requires closer examination of what do we expect to happen when waking up a sleeping thread.

sys/exception.c
```diff
@@ -9,4 +9,12 @@ void exc_before_leave(exc_frame_t *kframe) {
 
   if (td->td_flags & TDF_NEEDSWITCH)
     sched_switch(NULL);
+
+  /* First thing after switching to a thread: Process pending signals. */
+  mtx_lock(&td->td_lock);
```
rafalcieslak: Actually, `signotify` barely [marks a thread](http://bxr.su/FreeBSD/sys/kern/kern_sig.c#575) that it needs signal processing (by setting `TDF_NEEDSIGCHK` flag). Actual signal delivery processing happens within [`ast()`](http://bxr.su/FreeBSD/sys/kern/subr_trap.c#270), which is called by `MipsUserIntr` [here](http://bxr.su/FreeBSD/sys/mips/mips/exception.S#774) via [`DO_AST` macro](http://bxr.su/FreeBSD/sys/mips/include/pcb.h#87). Clearly, our equivalent of `ast` is `exc_before_leave`.

sys/signal.c
cahirwpz: [APUE](https://en.wikipedia.org/wiki/Advanced_Programming_in_the_Unix_Environment) §12.8 explains explicitely that all threads within given process share same `sighand_t` structure:

> Each thread has its own signal mask, but the signal disposition is shared by all
threads in the process. As a consequence, individual threads can block signals, but
when a thread modifies the action associated with a given signal, all threads share the
action. Thus, if one thread chooses to ignore a given signal, another thread can undo
that choice by restoring the default disposition or installing a signal handler for that
signal.

So that suggests `sighand_t` should be a member of process structure. The problem is we don't have one. Should we finally introduce the distinction? Let's discuss it during the meeting.

sys/signal.c
cahirwpz: The problem is that signal posting IMO should never wake up a sleeping thread, unless one explicitly allowed to do so, by entering `interruptible sleep` state. As I had been browsing `sleepqueue` code I recognized some of its functions are signal-aware, namely: [sleepq_wait_sig](http://bxr.su/FreeBSD/sys/kern/subr_sleepqueue.c#628) and [sleepq_timedwait_sig](http://bxr.su/FreeBSD/sys/kern/subr_sleepqueue.c#666). Indeed, by looking for [TDF_SINTR](http://bxr.su/s?refs=TDF_SINTR&project=FreeBSD) flag, seems like `sleepqueue` is the only component that cares about interruptible property.

When it comes to putting thread to sleep, apparently the most popular interface is [cv_wait_sig](http://bxr.su/s?refs=cv_wait_sig&project=FreeBSD) function and its [variants](http://bxr.su/FreeBSD/sys/sys/condvar.h#77).

rafalcieslak:
This PR is currently blocked by #278 and #276.

sys/signal.c
rafalcieslak: I've prepared a detailed description as #294.

rafalcieslak:
From now on this branch compiles newlib with `-DSIGNAL_PROVIDED`, and therefore it requries a one-time `make distclean` to force newlib to recompile.

user/newlib/overrides/include/sys/signal.h
```diff
@@ -0,0 +1,49 @@
+#ifndef _SYS_SIGNAL_H_
+#define _SYS_SIGNAL_H_
+
+#include <unistd.h>
+
+/* TODO: Should we somehow merge this file with kernel's signal.h? */
```
cahirwpz: Yes. Use `_KERNELSPACE` preprocessor definition to limit kernel interface visibility for userspace programs.

include/proc.h
```diff
@@ -24,6 +25,8 @@ struct proc {
   vm_map_t *p_uspace;      /* process' user space map */
   /* file descriptors table */
   fdtab_t *p_fdtable;
+  /* signal stuff */
+  sighand_t *p_sighand; /* Signal handlers data */
```
cahirwpz: I see no reason for introduction `sighand_t` structure. You could just replace it with:
```c
sigaction_t p_sigacts[SIG_LAST];
```
... and remove some boilerplate code. We won't share signal handlers between processes in any foreseeable future. Frankly, I don't care about `rfork` nor `clone` implementation.

include/signal.h
```diff
@@ -0,0 +1,50 @@
...
+
+typedef struct proc proc_t;
+
+typedef bitstr_t sigset_t[bitstr_size(SIG_LAST)];
+
+typedef void (*sighandler_t)(int);
+
+#define SIG_DFL (sighandler_t *)0x00
+#define SIG_IGN (sighandler_t *)0x01
+#define SIG_TERM (sighandler_t *)0x02
```
cahirwpz: I find name `SIG_TERM` somewhat confusing in the presence of `SIGTERM` symbol.

sys/signal.c
```diff
@@ -0,0 +1,200 @@
...
+
+static const char *signal_names[SIG_LAST] = {
+    [SIGINT] = "SIGINT",   [SIGILL] = "SIGILL",   [SIGABRT] = "SIGABRT",
+    [SIGFPE] = "SIGFPE",   [SIGSEGV] = "SIGSEGV", [SIGKILL] = "SIGKILL",
+    [SIGTERM] = "SIGTERM", [SIGCHLD] = "SIGCHLD", [SIGUSR1] = "SIGUSR1",
+    [SIGUSR2] = "SIGUSR2",
+};
+
+int do_kill(pid_t pid, int sig) {
+  proc_t *target = proc_find(pid);
```
cahirwpz: I can see a potential problem in here. Suppose you found a process, but after leaving `proc_find` this thread gets preempted and `target` becomes removed in the meantime. I found following code in NetBSD:
```c
/* Find the process */
mutex_enter(proc_lock);
p = proc_find(pid);
if (p == NULL) {
  mutex_exit(proc_lock);
  return ESRCH;
}
mutex_enter(p->p_lock);
mutex_exit(proc_lock);
```
... that makes me believe that `proc_find` should only check if `all_proc_lock` (should be renamed from `all_proc_list_mtx` and made global, apparently) is held when calling it.

sys/signal.c
```diff
@@ -0,0 +1,200 @@
...
+#include <errno.h>
+#include <sysent.h>
+#include <systm.h>
+#include <proc.h>
+
+static sighandler_t *signal_default_actions[SIG_LAST] = {
+    [SIGINT] = SIG_TERM,  [SIGILL] = SIG_TERM,  [SIGFPE] = SIG_TERM,
+    [SIGABRT] = SIG_TERM, [SIGSEGV] = SIG_TERM, [SIGKILL] = SIG_TERM,
+    [SIGCHLD] = SIG_IGN,  [SIGUSR1] = SIG_TERM, [SIGUSR2] = SIG_TERM,
+};
```
cahirwpz: We have at least four different default actions:
* `SA_KILL` terminate the process,
* `SA_IGNORE` ignore the signal,
* `SA_STOP` stop the process,
* `SA_CONT` continue the process if it is currently stopped.

Signal actions and properties are listed here [signalvar.h](http://bxr.su/NetBSD/sys/sys/signalvar.h#95).

sys/signal.c
```diff
@@ -0,0 +1,200 @@
...
+  if (oldact != NULL)
+    memcpy(oldact, &td->td_proc->p_sigactions[sig], sizeof(sigaction_t));
+
+  memcpy(&td->td_proc->p_sigactions[sig], act, sizeof(sigaction_t));
+
+  return 0;
+}
+
+static sighandler_t *get_sigact(int sig, proc_t *p) {
+  /* Assume p->p_lock is already owned. */
```
cahirwpz: You meant `assert(mtx_owned(p->p_lock))` ?

sys/signal.c
```diff
@@ -0,0 +1,200 @@
...
+  sighandler_t *h = p->p_sigactions[sig].sa_handler;
+  if (h == SIG_DFL)
+    h = signal_default_actions[sig];
+  return h;
+}
+
+int signal(proc_t *proc, signo_t sig) {
+
```
cahirwpz: `assert(mtx_owner(proc->p_lock))` here ?

mips/signal.c
```diff
@@ -0,0 +1,86 @@
...
+
+/* Delivers a signal to user process. */
+int platform_sendsig(int sig, sigaction_t *sa) {
+  thread_t *td = thread_self();
+  assert(mtx_owned(&td->td_lock));
+
+  /* Prepare signal context. */
+  sig_ctx_t ksc;
+
+  ksc.magic = SIG_CTX_MAGIC;
```
cahirwpz: Please use compound literals (C99 feature) for structure initialization.

mips/signal.c
```diff
@@ -0,0 +1,86 @@
...
+
+  /* Copyout signal context to user stack. */
+  sig_ctx_t *scp = (sig_ctx_t *)td->td_uctx.sp;
+  scp--;
+
+  int error = copyout(&ksc, scp, sizeof(sig_ctx_t));
+  if (error) {
+    /* This thread has a corrupted stack, we can no longer send any signals to
+       it. Forcefully terminate the thread. TODO: Maybe sending a SIGKILL is
+       safe here? */
```
cahirwpz: IMO it should be safe to deliver `SIGKILL` in this case, as it will be handled before a thread has an opportunity to use corrupted stack.

mips/signal.c
```diff
@@ -0,0 +1,86 @@
...
+    log("Thread %lu is unable to receive a signal, terminating it.",
+        td->td_tid);
+    thread_exit(-1);
+    __builtin_unreachable();
```
cahirwpz: Could you wrap it and place it in `common.h` so that `gcc` specific builtins are managed centrally?

mips/signal.c
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /* Prepare user context so that on return to usermode the handler gets
+   * executed. */
+  td->td_uctx.pc = (reg_t)sa->sa_handler;
```
cahirwpz: I don't think it does hurt not to check if `sa_handler` is a valid address (i.e. aligned and in user-space)... but you should add a comment somewhere it is safe to omit the check.

include/signal.h
rafalcieslak: I'd prefer to name it `SIGACT_TERM` which wouldn't be as ambiguous, but it must match the naming convention of `SIG_DFL` and `SIG_IGN`, and I cannot rename these two to `SIGACT_*` as these values are also used by userspace programs and their names there must stay `SIG_*`, and we are very strict about not defining the same value twice.

*Side note: I am unhappy with the current forceful merging of user and kernel headers, it implicitly mixes up usermode and kernelmode code, but I'll discuss a solution for this sometime later.*

sys/signal.c
```diff
@@ -0,0 +1,158 @@
...
+#include <klog.h>
+#include <signal.h>
+#include <thread.h>
+#include <malloc.h>
+#include <stdc.h>
+#include <errno.h>
+#include <sysent.h>
+#include <proc.h>
+
+static sighandler_t *def_sigact[NSIG] = {
```
cahirwpz: I reckon you will never store real pointers to signal handlers in this array. That's why they solved it in [slightly different way](http://bxr.su/FreeBSD/sys/kern/kern_sig.c#183), which I find less confusing.

rafalcieslak: I'll update my implementation to use this way of defining default actions, but I think it actually slightly complicates things: note that they still need to use both `SA_IGNORED` and `SIG_IGN`, [like this](http://bxr.su/FreeBSD/sys/kern/kern_sig.c#737).

include/signal.h
cahirwpz: *W.r.t. header files... I'll happily accept any other solution that will reduce likelihood of user and kernel space going out of sync with common (constants, types, etc.) declarations. AFAIU unix solution to the problem is to store kernel header files in `/usr/include/sys` and include them from corresponding files in `/usr/include/`.*

rafalcieslak: Eh, kinda. Let's look at `signal.h` in Linux. This description may be slightly inaccurate for some distros, though what I explain is true for most of them. The basic `/usr/include/signal.h` is provided by `glibc`, and it usually doesn't do much except for including `/usr/include/sys/signal.h` (sometimes it's the other way round, but anyway these files end up being equivalent). But `sys/signal.h`, and any file in `/usr/include/sys` for that matter, does not come directly from kernel sources. It is also provided by `glibc`, where it was written from scratch. But the files in `sys` (mostly) provide just the function interface, not the values of constants. These are pulled in by including files from `/usr/include/bits` directory. These files are (usually) forbidden to be included directly. These expose the *structure layout* and *constant values* for use with system interfaces. Great! So are these the files provided by kernel? No. These are also maintained by `glibc` developers.

So what *does* the kernel provide? `/usr/include/linux`. It does contain a `signal.h`, which pulls in stuff from `/usr/include/asm{,-generic}/signal*.h`. These contain mostly the same definitions as `bits/signal.h`, but these are maintained by kernel developers. Applications don't typically use interfaces from `/usr/include/linux` unless they need linux-specific features.

I recall that `uClibc` standard C library is capable of generating the `/usr/include/bits` files from `/usr/include/linux` (if we were to provide these), which seems convenient. Obviously, `newlib` is not that cool (it really doesn't care about a potentially present operating system).

cahirwpz:
Though I still have some doubts w.r.t. to locking and separtion of concerns in `sig_check` I'm going to accept this change and merge.

include/signal.h
cahirwpz: Ok. I see the point. Your observation tackles the difference between Linux and BSD worlds. *Linux* is a kernel, whereas **BSD* is an operating system. In the latter case user-space and kernel-space are tightly coupled and delivered as single product. In the former you have a choice of wildly different user-spaces (Android, ChromeOS, Debian, etc.) running on top of same kernel.

*In the long run I want Mimiker to follow BSD philosophy and be a complete OS.*

Issue #247, opened 2017-04-07
File descriptor table improvements


cahirwpz:
For simplicity file descriptor table is limited to `MAXFILE` entries and the last usable descriptor number is `NDFILE`. At some point we will be interested to have more than 20 resources opened by a single process. The task is to redesign the data structure for file descriptor table [fdtab](http://mimiker.ii.uni.wroc.pl/source/xref/mimiker/include/filedesc.h#16) so it can grow and shrink.

MoloToha:
I would suggest to create a separate structure, it's sketch looks like that 
```
typedef struct fdt_chunk {
  TAILQ_ENTRY(fdt_chunk) fdtc_chunk_list;
  file_t *fdtc_files[NDFILE];
  bitstr_t fdtc_map[bitstr_size(NDFILE)];
  int fdtc_id;
} fdt_chunk_t;
```
Such chunk contains an array of `NDFILE` pointers to file objects. A file in each chunk has `fdtc_id * NDFILE + offset` file descriptor (where `fdtc[offset]` is a pointer to file), and we can find a file with descriptor `fd` by finding a chunk with id `fd / NDFILE` and getting a pointer `fdtc_files[fd % NDFILE]`. With this approach, we substitute a fixed size array with a flexible tail queue `TAILQ_HEAD(, fdt_chunk_t) fdt_list` to which we can easily attach another chunk (it's id will be the id of last chunk in the queue plus 1), if needed, and do a cleanup of empty chunks, all at cost of performance decrease up to 52 times (if we set the limit of opened files to 1024 and `NDFILE` to 20).

rafalcieslak:
Isn't this an overly complicated solution for a simple problem? What FreeBSD does is just allocating a completely new, contiguous array for file pointers whenever a process runs out of free descriptors. See [this comment](http://bxr.su/FreeBSD/sys/kern/kern_descrip.c#134), and [this function](http://bxr.su/FreeBSD/sys/kern/kern_descrip.c#1503). Our file descriptor table is heavily based on FreeBSD's [struct filedesc](http://bxr.su/FreeBSD/sys/sys/filedesc.h#68), so you will see many similarities to our implementation.

[...]

PR #273, opened 2017-04-24
Implemented resizable file descriptor table

MoloToha:
Improved current implementation of file descriptor table so that it can grow (but isn't able to shrink yet), as requested in #247

### DIFF
--- a/sys/filedesc.c
+++ b/sys/filedesc.c
@@ -31,7 +31,7 @@ static void fd_mark_unused(fdtab_t *fdt, int fd) {
  */
 
 static void fd_growtable(fdtab_t *fdt, size_t new_size) {
-  assert(fdt->fdt_nfiles < new_size && new_size <= MAXFILES);
+  assert(fdt->fdt_nfiles < new_size && new_size <= MAXFILES && mtx_owned(&fdt->fdt_mtx));
 
   file_t **old_fdt_files = fdt->fdt_files;
   bitstr_t *old_fdt_map = fdt->fdt_map;
@@ -39,7 +39,6 @@ static void fd_growtable(fdtab_t *fdt, size_t new_size) {
   file_t **new_fdt_files =
     kmalloc(fd_pool, sizeof(file_t *) * new_size, M_ZERO);
   bitstr_t *new_fdt_map = kmalloc(fd_pool, bitstr_size(new_size), M_ZERO);
-  memset(new_fdt_map, 0, bitstr_size(fdt->fdt_nfiles * 2));
 
   memcpy(new_fdt_files, old_fdt_files, sizeof(file_t *) * fdt->fdt_nfiles);
   memcpy(new_fdt_map, old_fdt_map, bitstr_size(fdt->fdt_nfiles));
@@ -48,7 +47,7 @@ static void fd_growtable(fdtab_t *fdt, size_t new_size) {
 
   fdt->fdt_files = new_fdt_files;
   fdt->fdt_map = new_fdt_map;
-  fdt->fdt_nfiles *= 2;
+  fdt->fdt_nfiles = new_size;
 }
 
 /* Allocates a new file descriptor in a file descriptor table. Returns 0 on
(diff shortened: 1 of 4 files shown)


sys/filedesc.c
```diff
@@ -26,18 +26,51 @@ static void fd_mark_unused(fdtab_t *fdt, int fd) {
   bit_clear(fdt->fdt_map, fd);
 }
 
+/* Grows given file descriptor table to contain new_size file descriptors
+ * (up to MAXFILES)
+ */
+
+static void fd_growtable(fdtab_t *fdt, size_t new_size) {
```
rafalcieslak: This function will only work correctly if the `fdt` mutex is locked. Please either lock it recursively, or simply `assert(mtx_locked(&fdt->fdt_mtx))`.

sys/filedesc.c
```diff
@@ -26,18 +26,51 @@ static void fd_mark_unused(fdtab_t *fdt, int fd) {
...
+  memset(new_fdt_map, 0, bitstr_size(fdt->fdt_nfiles * 2));
+
+  memcpy(new_fdt_files, old_fdt_files, sizeof(file_t *) * fdt->fdt_nfiles);
+  memcpy(new_fdt_map, old_fdt_map, bitstr_size(fdt->fdt_nfiles));
+  kfree(fd_pool, old_fdt_files);
+  kfree(fd_pool, old_fdt_map);
+
+  fdt->fdt_files = new_fdt_files;
+  fdt->fdt_map = new_fdt_map;
+  fdt->fdt_nfiles *= 2;
```
rafalcieslak: This assumes that `new_size == 2 * fdt->fdt_nfiles` . This happens to be true, but you probably want to assign `new_size` to `nfiles` here.

sys/filedesc.c
```diff
@@ -72,9 +105,10 @@ void fdtab_unref(fdtab_t *fdt) {
    argument doest not make sense yet. */
 fdtab_t *fdtab_alloc() {
   fdtab_t *fdt = kmalloc(fd_pool, sizeof(fdtab_t), M_ZERO);
-  /* For now, fdt_files and fdt_map have a static size, so there is no need to
-   * separately allocate memory for them. */
   fdt->fdt_nfiles = NDFILE;
+  fdt->fdt_files = kmalloc(fd_pool, sizeof(file_t *) * NDFILE, M_ZERO);
+  fdt->fdt_map = kmalloc(fd_pool, bitstr_size(NDFILE), M_ZERO);
+  memset(fdt->fdt_map, 0, bitstr_size(NDFILE));
```
rafalcieslak: You've just called `kmalloc` with `M_ZERO`, so the allocated memory is guaranteed to be cleared, no need to `memset` it. Same goes for clearing `new_fdt_map`.

sys/filedesc.c
```diff
@@ -87,9 +121,7 @@ fdtab_t *fdtab_copy(fdtab_t *fdt) {
 
   mtx_scoped_lock(&fdt->fdt_mtx);
 
-  /* We can assume both filedescs use 20 descriptors,
-   * because we don't support other numbers. */
-  assert(fdt->fdt_nfiles == newfdt->fdt_nfiles);
+  fd_growtable(newfdt, fdt->fdt_nfiles);
```
rafalcieslak: `fd_growtable` does not work if `newsize` is smaller or equal to current size. Thus you should only call `growtable` here if it actually needs expanding.

tests/resizable_fdt.c
```diff
@@ -0,0 +1,20 @@
+#include <file.h>
+#include <filedesc.h>
+#include <ktest.h>
+
+static int test_resizable_fdt() {
+  fdtab_t *fdt_test = fdtab_alloc();
+  file_init();
```
rafalcieslak: Why do you need to call `file_init` here? It should have been already called by `kernel_init`.

sys/filedesc.c
```diff
@@ -87,9 +120,11 @@ fdtab_t *fdtab_copy(fdtab_t *fdt) {
 
   mtx_scoped_lock(&fdt->fdt_mtx);
 
-  /* We can assume both filedescs use 20 descriptors,
-   * because we don't support other numbers. */
-  assert(fdt->fdt_nfiles == newfdt->fdt_nfiles);
+  if (fdt->fdt_nfiles > newfdt->fdt_nfiles) {
+    mtx_lock(&newfdt->fdt_mtx);
```
rafalcieslak: A side note: No other threads have a reference to `newfdt` yet, so locking the mutex should be unnecessary. However, `fd_growtable` expects to have the mutex already locked, and therefore we need to lock it before growing the table. I see some redundancy here, but there is no easy way to solve this - so I suggest to leave it as it is (or maybe just a comment on why is `mtx_lock` necessary here).

PR #274, opened 2017-04-25
Add prepare_uio

coodie:
This PR adds prepare_uio functions, which prepare single-vectored uio operation structures.

### DIFF
--- a/include/uio.h
+++ b/include/uio.h
@@ -2,6 +2,7 @@
 #define _SYS_UIO_H_
 
 #include <vm.h>
+#include <common.h>
 
 typedef struct vm_map vm_map_t;
 
@@ -21,12 +22,19 @@ typedef struct uio {
   vm_map_t *uio_vmspace; /* destination address space */
 } uio_t;
 
-void prepare_uio(uio_t *uio, iovec_t *iov, uio_op_t op, vm_map_t *vm_map,
+
+/* Convenience function for initializing uio, where iov has one buffer */
+void prepare_single_uio(uio_t *uio, iovec_t *iov, uio_op_t op, vm_map_t *vm_map, size_t offset,
                  void *buffer, size_t buflen);
-void prepare_user_uio(uio_t *uio, iovec_t *iov, uio_op_t op, void *buffer,
-                      size_t buflen);
-void prepare_kernel_uio(uio_t *uio, iovec_t *iov, uio_op_t op, void *buffer,
-                        size_t buflen);
+
+#define prepare_single_user_uio(uio, op, offset, buffer, buflen) \
+    iovec_t __CONCAT(iov_,__LINE__); \
+    prepare_single_uio((uio), &__CONCAT(iov_,__LINE__), (op), get_user_vm_map(), (offset), (buffer), (buflen));
+
+#define prepare_single_kernel_uio(uio, op, offset, buffer, buflen) \
+    iovec_t __CONCAT(iov_,__LINE__); \
+    prepare_single_uio((uio), &__CONCAT(iov_,__LINE__), (op), get_kernel_vm_map(), (offset), (buffer), (buflen));
+
 
 int uiomove(void *buf, size_t n, uio_t *uio);
 int uiomove_frombuf(void *buf, size_t buflen, struct uio *uio);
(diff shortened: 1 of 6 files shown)


rafalcieslak:
I am unsure about naming these functions like `prepare_user_uio`. This name suggest this is the initialization function that must be called for any user `uio`, which is obviously not true. Maybe `prepare_basic_user_uio` or `prepare_simple_user_uio` would make it clear that this is barely a convenience function for some specific, simple kinds of data transfer?

Also note that you still need to declare `iovec_t iov` and pass it to `prepare_*_uio`. I suppose if both `prepare_*_uio` functions were macros, you could simplify things even further, something like (untested!):

```c
#define prepare_user_uio(uio, op, buffer, buflen) \
iovec_t iov_##__LINE__; \
prepare_uio(uio, &iov_##__LINE__, get_user_vm_map(), buffer, buflen);
```

rafalcieslak:
Sigh, I gotta say I liked @coodie's version slightly more (although it still needed improvements). Plan9 extensions for unnamed fields introduce needless ambiguity, which makes exact structure layout less obvious. And once we decide to enable these extensions, we have no good excuse for not using them widely, namely for all other `union` fields we use. You've wrapped all `uio`s in additional scopes, which serve no purpose besides limiting the range of a variable - which is not even explicitly declared in that new scope! I recommend either reusing `uio` variables, or giving them different names when multiple `uio`s are used by the same function (that's what the `name` argument is for, right?).

Also, in future, when rewriting someone else's proposal from scratch, please use a separate branch/PR (maybe using the original branch as merge target instead of master). You've made it difficult to compare @coodie's variant with yours, and @coodie is now unable to fix/improve his solution. I also feel that overwriting 100% of other developer's proposed changes without discussing them (or your alternative ideas) first is kinda disrespectful.

cahirwpz:
@rafalcieslak You're right. I'll revert changes and place them in another PR.

coodie:
I don't want to spend too much time discussing this change, because it's mostly done for our convenience and I'm afraid this might turn into a little bit of bikeshedding. 

I have considered idea of using macro to put iovec on stack, but I couldn't come with elegant way of doing that, because in vfs tests uio is reused so I couldn't base it on uio name (there would be name conflict), I forgot about `LINE`, while this obviously fixes this problem. I can't really see reason why not to use this, so I'll implement this.

When it comes to names, naming them `prepare` was very initial idea, and we haven't used that before and I'm ok with this. I don't really like spending too much time thinking about how to name a variable, so I'm going to take @rafalcieslak propositions and rename then accordingly (although in my opinion they are too lengthy).

cahirwpz:
@coodie I'm sorry for overwriting your changes. I guess I've got overexcited (and I'm still are) about discovery of simple subtyping for structures in GCC. Please note, you should modify `sys/exec.c` as well.

If you find #279 acceptable as an alternative you can merge it into your branch. If you decide to do so,  you'll get credit for the task as well.

include/uio.h
```diff
@@ -21,6 +22,20 @@ typedef struct uio {
   vm_map_t *uio_vmspace; /* destination address space */
 } uio_t;
 
+/* Convenience function for initializing uio, where iov has one buffer */
+void prepare_single_uio(uio_t *uio, iovec_t *iov, uio_op_t op, vm_map_t *vm_map,
+                        size_t offset, void *buffer, size_t buflen);
```
cahirwpz: I reckon it's more natural to specify `offset` (of type `off_t` !) as the last argument.

include/uio.h
```diff
@@ -21,6 +22,20 @@ typedef struct uio {
...
+
+#define prepare_single_user_uio(uio, op, offset, buffer, buflen)               \
```
cahirwpz: Could you make those names a little bit shorter?

include/uio.h
```diff
@@ -21,6 +22,20 @@ typedef struct uio {
...
+  iovec_t __CONCAT(iov_, __LINE__);                                            \
+  prepare_single_uio((uio), &__CONCAT(iov_, __LINE__), (op),                   \
+                     get_user_vm_map(), (offset), (buffer), (buflen));
+
+#define prepare_single_kernel_uio(uio, op, offset, buffer, buflen)             \
+  iovec_t __CONCAT(iov_, __LINE__);                                            \
```
cahirwpz: In use cases such as `tests/vfs.c` structure `iovec_t` is defined multiple times within the same scope. Could you check if compiler optimizes that and allocates place on stack only once?

cahirwpz:
Implemented by #282

PR #275, opened 2017-04-25
VGA driver

rafalcieslak:
Works with both OVPsim and QEMU. When launching the kernel, use `-g` flag to enable VGA output. I had to move round some PCI macros to make them accessible by the VGA driver - due to the temporary lack of a PCI bus interface.

The `vga` tests renders and displays a beautiful, photo-realistic rainbow :rainbow: in FullHD. Should look almost like [this one](http://i.imgur.com/nUzyumK.png), except our rainbow is animated!


sys/vga.c
```diff
@@ -0,0 +1,178 @@
...
+}
+
+void vga_init(vga_control_t *vga) {
+  /* Set RAM_ENABLE */
+  *VGA_REG_EXT_MOR_WRITE = 0x2;
+
+  // Set memory map select to 0
+  uint8_t memory_map_select = vga_gr_read(vga, 0x06) >> 2 & 0x3;
+  memory_map_select = 0;
+  vga_gr_write(vga, 0x06, memory_map_select << 2);
```
cahirwpz: Could you name registers numbers and magic values you use here?

cahirwpz:
Given you understood well my recent changes in driver infrastructure, would you be able to rewrite the driver to make use of new `pci_bus_driver`, BARs represented as set of `resource_t` and corresponding `bus_space`?

rafalcieslak:
I'll see what I can do.

rafalcieslak:
I am completely stuck. There are two issues that I can't find a way around:

- VGA uses several I/O [registers](http://www.osdever.net/FreeVGA/vga/portidx.htm) which I've been accessing using the physical address `0x18000000 + regaddr`. That memory region is the PCI ioports base, but apparently the VGA device does not have any I/O bars?

- I don't know how to write data to a resource. I can't directly write using `.r_start + x` address, as it only represents bus-specific offset. Should I use `bus_space_write` for all accesses? That would be a very slow way of copying an entire framebuffer byte-by-byte into device memory...

I'd appreciate some help here.

rafalcieslak:
Oh. Let me clarify my issue. I cannot access any memory behind a PCI BAR, because neither the resource address is meaningful (it's a physical address, I'd need to map it to KSEG2 in order to write to it...), and BAR's `r_bus_space` is unset. How do I use BAR resources, then?

cahirwpz:
Please have a look at [pci-bus-fixes](https://github.com/cahirwpz/mimiker/compare/pci-bus-fixes) branch. Does it solve problems that you encountered?

rafalcieslak:
I suppose so! I'll import it here so that I can continue work ASAP.

rafalcieslak:
I have some bad news. OVPsim's `CirrusLogic GD5446` has incomplete PCI support. In particular, it lacks MMIO - in fact, entire BAR1 is inaccessible (proof: `ImperasLib/source/cirrus.ovpworld.org/peripheral/GD5446/1.0/vga.c`, function `mapGPIO`, which is only referenced in `pciIint` in the same file. It explicitly declares that the entire memory region is not supported). Without this feature, it is impossible to setup VGA registers via PCI. The VGA device works correctly via ISA bus, and all registers are configurable there, which is why this device is usable at all.

Because of this, using the VGA device via PCI bus only on OVPsim is not possible. I could make my driver use ISA bus for configuring VGA registers and PCI bus for data transfer, but it would defeat the purpose of this branch.

On the other hand, QEMU has not such problem - MMIO via BAR1 works correctly (although I can't confirm whether it behaves exactly as the original GD5446), and I can use it to set up VGA registers.

cahirwpz:
Please ignore the problem for now and focus on getting the driver work under QEMU.

rafalcieslak:
Huh? The e-mail I received, as a notification of your comment, was:

> Please ignore the problem for now and focus on getting the driver work under QEMU.

Now I see you've made a mistake, and actually meant OVPsim, and not QEMU. This is very unfortunate, as I've spent this day rewriting the driver from scratch with only QEMU in mind, which simplified things **a lot**, because QEMU's `stdvga` is much simpler to configure (see for yourself, thanks to Bochs VBE extensions, the driver code is now actually readable and intuitive!).

So I've prepared a very elegant QEMU `stdvga` driver, which uses the PCI and bus interfaces. As a natural consequence of hooking the driver to `/dev/vga`, I also had to prepare a generic vga framebuffer interface. But it turns out this is exactly the opposite of what you've wanted.

I'd like you to review this code anyway. It is still some demonstration of how I use the bus interface. My suggestion is to support GD4556 with a *separate* driver - due to major differences how it is configured, although the driver would hook up to the same `vga` interface I've created, so high-level access would be entirely transparent. Writing a `cirrus` driver requires support for ISA bus first, though.

rafalcieslak:
Since display tests can't be automated anyway, I decided to remove the `tests/vga.c` rainbow test, and implemented a user-space test program instead - it computes and displays a visualization of the Mandelbrot set.

sys/stdvga.c
```diff
@@ -0,0 +1,165 @@
...
+  if (error)
+    return error;
+  /* TODO: Only update the modified area. */
+  stdvga_palette_write_buffer(stdvga, stdvga->palette_buffer);
+  return 0;
+}
+
+static int stdvga_fb_write(vga_device_t *vga, uio_t *uio) {
+  stdvga_device_t *stdvga = STDVGA_FROM_VGA(vga);
+  /* TODO: I'd love to pass the uio to the bus directly, instead of using a
```
cahirwpz: [bus_space_map](https://www.freebsd.org/cgi/man.cgi?query=bus_space_map&sektion=9) call from *BSDs allows to map `RT_MEMORY` resource into kernel (and user?) virtual address space. Some day we will implement it and use BUS_SPACE_MAP_LINEAR for frambuffer memory.

> **BUS_SPACE_MAP_LINEAR** Try to map the space so that its contents can be	accessed linearly via normal memory access methods (e.g. pointer dereferencing and structure accesses).  This is useful when software wants to do direct access to a memory device, e.g. a frame buffer. [...]

sys/stdvga.c
```diff
@@ -0,0 +1,165 @@
+#include <pci.h>
+#include <vga.h>
+#include <stdc.h>
+#include <malloc.h>
+
+#define VGA_PALETTE_SIZE (256 * 3)
+
+typedef struct stdvga_device {
+  pci_device_t *pci_device;
+  resource_t mem;
```
cahirwpz: Please keep these as pointers. Once I consolidate bus interface there will be call similar to [bus_alloc_resource_any](https://www.freebsd.org/cgi/man.cgi?query=bus_alloc_resource_any&sektion=9) to reserve a BAR.

sys/stdvga.c
```diff
@@ -0,0 +1,165 @@
...
+/* The general overview of the QEMU std vga device is available at
+   https://github.com/qemu/qemu/blob/master/docs/specs/standard-vga.txt */
+#define VGA_QEMU_STDVGA_VENDOR_ID 0x1234
+#define VGA_QEMU_STDVGA_DEVICE_ID 0x1111
+
+static void stdvga_io_write(stdvga_device_t *vga, uint16_t reg, uint8_t value) {
+  bus_space_write_1(&vga->io, reg + VGA_MMIO_OFFSET, value);
+}
+static uint8_t stdvga_io_read(stdvga_device_t *vga, uint16_t reg)
+  __attribute__((unused));
```
cahirwpz: There's `__unused` macro you can use. I guess you could use it in function definition as well.

sys/stdvga.c
```diff
@@ -0,0 +1,165 @@
...
+  int error =
+    uiomove_frombuf(stdvga->fb_buffer, stdvga->width * stdvga->height, uio);
+  if (error)
+    return error;
+  bus_space_write_region_1(&stdvga->mem, 0, stdvga->fb_buffer,
+                           stdvga->width * stdvga->height);
+  return 0;
+}
+
+MALLOC_DEFINE(stdvga_pool, "stdvga VGA driver pool");
```
cahirwpz: Ok. I think it's high time to adapt *BSD idea for predefined and globally available set of [memory pools](http://bxr.su/search?q=&defs=&refs=MALLOC_DEFINE&path=%2FNetBSD%2Fsys%2F&project=NetBSD). Obviously that should be a part of another PR.

sys/stdvga.c
```diff
@@ -0,0 +1,165 @@
...
+    return 0;
+
+  /* TODO: Initialize the pool somewhere else... This assumes "attach" will only
+     get called once. */
+  kmalloc_init(stdvga_pool, 128, 128);
+
+  stdvga_device_t *stdvga =
+    kmalloc(stdvga_pool, sizeof(stdvga_device_t), M_ZERO);
+
+  uint32_t status_command = pci_read_config(pci, 1, 4);
```
cahirwpz: I think this line is broken. You can read 4 bytes at once **only** from address aligned to 4.

Please use PCI register names, i.e.:
```c
uint32_t command = pci_read_config(pci, PCIR_COMMAND, 2);
uint32_t status = pci_read_config(pci, PCIR_STATUS, 2);
```

Issue #272, opened 2017-04-23
Adapting the standard library for user programs


rafalcieslak:
We need to finally have some control over what the standard library provides for user programs. I have devised a multi-stage plan of reconfiguring our toolchain and build system in a way that will allow us to dynamically prepare the entire sysroot when building the kernel. That would mean compiling newlib core, patching it with our custom headers and other files, adding a glue layer (`libmimiker`), and joining all these together using a fancy linker script and gcc's `--sysroot` option. I'm in the final stages of testing my proposed setup, and it appears to be working pretty well. This is a complex process, so I'll best leave details for discussion at the next meeting.

rafalcieslak:
Resolved by merging #276.

PR #276, opened 2017-04-25
Compiling newlib

rafalcieslak:
This branch introduces changes necessary for building newlib with the kernel.

**NOTE: This branch is useless with a toolchain that has newlib built-in!**

Quick overview:

- `cache` is a store for `newlib-2.5.0.tar.gz` and unpacked sources.
- `user/sysroot-newlib` is dynamically generated and is the directory where newlib gets installed after it is compiled.
- `user/sysroot-base` is versioned and contains our custom headers and other files.
- libmimiker was moved to a dedicated directory `user/libmimiker`.
- `user/sysroot` is dynamically created, and contains the full sysroot - composed of files from `sysroot-newlib`, `sysroot-base` and `libmimiker`. It may be used as `--sysroot` for compiling third-party programs. It is also copied to `/` of the ramdisk.

I also did some minor changes to unrelated Makefile rules, to improve parallelized build times, and to remove some depracated conditions. `initrd` malloc pool size had to be increased as the `cpio` archive grows larger.

We'll have to prepare a new toolchain build without newlib for this branch to make sense, I'll look into that soon. (*By the way, I've recently learned that **ALL** gcc toolchain builds are fully relocatable by design, moving the toolchain to a different directory will always work, and if you can build gcc so that it can't be moved, then gcc developers would be VERY interested in details.*). I'd also suggest to change our vendor flavor name from `unknown` to `mimiker`, so that our new toolchain name would be `mips-mimiker-elf`. This has two advantages:
1) Newlib, as well as many 3rd party programs, contain conditional code that depends on the vendor name - using a custom one like `mimiker` may simplify porting.
2) It'll give us a way to easily (temporarily) distinguish between toolchain version (kind of like we did during the switch from `mips-mti-elf` to `mipsel-unknown-elf`) and inform the developer if they are using incorrect one.

I recommend starting reading the diff from `user/Makefile`.

This branch addresses #272.


rafalcieslak:
I've prepared a newlibless toolchain, and configured Travis to use it. This inflated the diff a little bit, but overall changes are still relatively small.

cahirwpz:
I'm having some thoughts that may render the process more explicit and easier to maintain:

1. I'd like user programs to facilitate `make install DESTDIR=${sysroot}` method of installation. This way program's build and installation processes are expressed in a single `Makefile`.
2. Given the above is implemented, process of creating a ramdisk is simplified. Just take contents of `sysroot` and put it into `initrd` file. Ramdisk structure becomes explicit (`find $sysroot -ls`) and no `initrd.rules` is needed anymore. 
3. I'd be nice to keep 3rd party programs and libraries in single directory as [submodules](https://git-scm.com/docs/git-submodule). If we ever decide to fork `newlib`, just changing repository URL will do the trick.

These're my first thoughts after trying out this PR. When ideas above are sorted out, I'll have a closer look at details.

rafalcieslak:
1. I'm glad that you see the natural consequence of managing `sysroot` this way! I wanted to introduce this mechanism in a subsequent change to keep this one simpler, but since you're fine with it I can do it right here.

2. This is exactly the point that makes the idea of preparing a `sysroot` so elegant. Again, I originally wanted to investigate the removal of `initrd.rules` in a next step to keep this one simple. The only tricky part that will require to be configured with care is that (as for now) some tests come with their own collection of files, so these will also need to be installed in `sysroot`, but that's certainly not a big deal.

3. The issue with submodules is that they force us to use git. If we decide to switch to another std library that happens to use e.g. Subversion, then we would need to create and host a git mirror of the entire repository. Also ATM the size of entire `newlib-cygwin` repo is about 150MB, which is kind of heavy - and we can't do a shallow fetch for the submodule, because git servers refuse to send shallow snapshots for non-head commits (e.g. tagged releases). Naturally, both these minor issues disappear once we have our fork.

I'm also starting to feel that we're going to need to introduce changes to newlib sources (larger than just headers) much sooner rather than later. This is mainly because the internal features of newlib also use these headers, so substituting them after newlib was already compiled only works for some particular headers that are not used in std function definitions. For example `fprintf` indirectly calls `lseek` with `SEEK_SET`, and thus the proper definition of that constant must already be present when newlib gets compiled. Maybe, since we're changing the way we use newlib anyway, starting the fork right now isn't such a bad idea after all?

cahirwpz:
Just short answer to encourage you to widen the scope of PR:
1. Please do so!
2. Ditto. With both changes in place, it will be clear that this PR is a breakthrough.
3. I need some time to think about it...

rafalcieslak:
I've started by moving `./user/{sysroot,libmimiker}` to `./`, so that the `user` directory contains *only* user programs.

I see a potential difficulty with configuring the build system so that user programs install themselves into the sysroot: it would introduce a *circular dependency* between any user program and the sysroot directory. Also, tracking which files in the sysroot come from where is virtually impossible, so I'll need to wipe out the entire `sysroot` dir and re-generate it every time `make` runs, and thus the installation (not compilation) procedure for each user program will require repeating every time - in order to ensure that all dependencies are always satisfied. If I were to fix it by generating depfiles for the sysroot, then the top-level makefile would need to know the details of the installation procedure expressed in a deeply nested makefile, which would make our build scripts an incredible mess. Any wiser, extensive solution immediately starts to behave as a real *package manager*... so I'd rather not implement anything like it at the moment - I'll stay with recreating the sysroot too frequently.

cahirwpz:
IMO it's reasonable to split the process into four phases:
1. patch `newlib` with our files, install header files into `sysroot`,
2. build kernel using `sysroot` contents,
3. build `newlib` and install it into `initrd`,
4. build user programs using `initrd` contents, install them into `initrd` and create `cpio` archive.

rafalcieslak:
I suppose we may need to meet and discuss this in greater detail (we've only had 2h of meeting for the past 4 weeks and it's starting to show). The entire process is **not** as simple as you think! When designing these build stages, please keep in mind that kernel compilation is not one-off, we repeatedly run `make` during development to recompile recent changes, and therefore `make` must understand how each components depend on each other in order to correctly decide which stages must be re-done.

You imply we need newlib headers in `sysroot` even to compile the kernel. That is not true (as this branch happens to demonstrate). In fact, we need newlib to **not** be present in the sysroot, in order to ensure we don't pull any parts of userspace standard library into the kernel. Even headers are dangerous, many newlib headers contain `static inline` function implementations, and most of them are invalid if they don't get linked with newlib's `libc.a`. For kernel purposes, we use `./stdc` which provides various utilities similar to the standard library - so unless you want to replace them with kernel-side newlib, stage 1 is irrelevant.

Let me walk you throught the problem I see with mixing both newlib and user programs in the same directory you've called `initrd`. Remember that all `make` does is satisfying dependencies using target rules. Suppose we've already compiled the kernel and everything is working fine. `./initrd/` is populated with various files from various sources, many of them were copied into `initrd` as the result of invoking `make install` in a user program (or third party program) directory. All the `install` target recipe has done was `cp`-ing or `install`-ing some files into the destination directory. Now suppose we've edited some files in some of these programs, and/or the standard library. That should result in some files in `./initrd` getting updated. But how do we know a particular file in `initrd` is out-of-date? We don't track where it came from. We can't even reliably tell if it belongs to newlib, or a user program (which one?), or maybe a third party program. With `initrd.rules`, which were gathered into **centralized** installation rules, preparing dependencies for files in the `initrd` directory was simple. But now the only thing we can do to ensure that the `./initrd` is kept up to date, is re-installing **all** its contents, because, due to missing dependencies, make will never know how to perform a partial update.

Now if we need to reinstall all contents of the `./initrd` directory every time any change was done to kernel sources, then in particular we also need to reinstall `newlib`. But hey, if we do so, then immediately all user programs, which use `newlib` from `./initrd` directory require re-compiling, because the library they use has (possibly) changed! So we need to compile all of them, again, every time. After that is done, we can `make install` user programs, and they will populate `./initrd`, done.

Another possible issue concerns files removed from `initrd`. Suppose we have a program which installed `initrd/A` and `initrd/B`, but due to recent changes it no longer needs `B` and from now on it only installs `A` to `initrd`. So we should remove `initrd/B` now. But, as we don't know the origin of `initrd/B`, we can't tell what program installed it, or what it's even doing there. Is it still needed? So the only reliable way to ensure no files are left over from previous `initrd` versions, we'd need to purge all contents of the `initrd` every time before installing anything in there. But... that's not a problem at all, since we've just established that we need to reinstall everything every-time, so removing all contents beforehand doesn't do much difference.

This way we end up forced to recreate the install directory from scratch every time any user program is modified, which, in turn, required recompiling all user programs.

So, in other words: Using program-defined `make install` decentralizes dependency information. 
Without that information, top-level `Makefile`s can no longer reliably determine the minimal amount of work for updating the ramdisk according to source changes.

And naturally all of these problems become 100x harder if some of our user programs become libraries installed to the ramdisk, and other programs need these libraries to compile...

As I've said, this is not an easy process and expressing it as a series of simple "stages" hides and masks many problems. If you still believe that the exciting idea of having everything install itself into `initrd` is so simple to execute, then I ask you to either examine it more thoroughly, or tinker with a possible implementation. Reliably managing dependencies in a dynamically created installation directory is a very difficult problem. But it's a very well-known problem, and the software used to solve it is called *a package manager*.

rafalcieslak:
I've updated this branch according to the ideas we've exchanged on a recent meeting. It only slightly simplified make rules, and some extra tricks were required to work around other `make` oddities. I can't say I like the way this branch looks ATM.

I think what we're trying to do: using `make` dependencies to gather independent files into a directory, is very non-standard way of using make. *We may be using a wrong tool for the problem!* Mind you' I am completely fine with using make to compile kernel itself, and to compile each user program independently. That's following what make is designed for. What I think `make` isn't designed for is orchestrating the file installation process. Maybe we could use a nifty bash/python script for gathering files into sysroot (it would call `make install` + `fakechroot` in user program dirs) and managing dependencies between them, while still using `make` for compilation. I have some new ideas that need refining.

rafalcieslak:
Here's my conclusion after investigating these issues for many good hours. `make` is a great too for micro-managing dependencies between particular files. But it is not a good tool for managing dependencies between separate projects! This is because `make` only ever works on single files, and not entire trees. `make` also generally assumes that recipes have no side-effects, that is, their only useful output should be the recipe target itself, and certainly not stuff outside current tree, especially if it is managed by a different set of rules. Trying to force `make` to address source or installation directories results in all kinds of weird behavior, because its an ambiguous demand, and `make` was never designed to work as a package manager.

I believe it would be much simpler to build user programs and install them into `sysroot` using a dedicated python script.

Here's how I would see this. My explanation is detailed and therefore long, but the concept is actually very simple, and will immediately allow us to do all kinds of cool things!

1. Each directory in `user` would represent a single program or library. The contents of the directory are arbitrary, but it is recommended that it contains a `Makefile` with `all` and `install` targets (so exactly as they do now).
2. We would stop treating `newlib` as a special entity and place it in `./user/newlib`, since it is barely a user-mode library and nothing else. We might either have `./user/newlib` point to a third-party repository, or we might move the "download & compile" `make` script into `./user/newlib/Makefile`.
  - This way `newlib` gets uniform treatment, all user programs and libraries will be processed the same way.
3. User programs would provide a description file which contains dependency information, and, if required, compile/install script, in case a program does not use a `Makefile`. For example, a `./user/prog/rules.json` file might contain:
```json
{
  "name": "prog",
  "depends": ["stdc"]
}
```
4. The sysroot-building python script would gather all information from `user/*` directories, to have a full image of how programs and 3drparty projects depend on each other. It would then compile them in the right order.
5. The python script would maintain a cache directory with installed files. For example, `prog` wouldn't get installed to `sysroot/` directly, but to `sysroot-cache/prog/` instead. This way the `sysroot-cache` directory kind of contains *binary packages* in form of directories.
6. (**!!!**) In order to compile user program X which depends on Y and Z, the python script would create a temporary `build-sysroot` by copying (merging) the contents of `sysroot-cache/Y/` and `sysroot-cache/Z/`. The `build-sysroot` directory would be passed as `--sysroot` for compiling X. The actual compilation for X would use traditional `make`. When making X is done, the python script would use `fakechroot` to have X installed into `sysroot-cache/X/`. This way X doesn't need to be aware of the entire process, and everything is transparent!
7. Naturally, the python script would also track which projects need recompiling, but that becomes simple when we keep all *packages* as separate directory trees, it just needs to scan the `sysroot-cache/X` directory to see what is the most recently modified file, and only recompile what's necessary - using it's dependency information.
8. To prepare the actual `initd` archive, all `sysroot-cache/*/` directories are merged.

This approach would **solve all problems** I've experienced with this branch. It also has extra advantages:
 - Managing user libraries is free, they don't differ in any way from programs.
 - The process is completely transparent and user programs shouldn't need any adaptation.
 - Adding third party libraries and programs becomes super simple! Just drop a `submodule` or source directory into `user`, place a `rules.json` with dependency description, all done.
 - Compilation dependencies are airtight - only what is specified as dependencies will be present in `sysroot` used for compilation, so no unwanted stuff may be pulled in, even by mistake.
 - It is very scalable, and will stay simple to maintain even if we use a very large number of user libraries and programs.

Note that this mechanism is virtually impossible to implement using plain `make`, due to how it handles directories and complex dependencies. Actually, I see a possibly valid way of implementing this using `make` (or any build system), but it requires creating a large number of meta-files and fake dynamically generated dependencies, which would be a great pain to maintain!

I came to a strong belief that `make` is great for compilation, but not great for managing file installation and directory layout. The solution I propose is very elegant, in fact it ends up behaving as a minimalistic package manager that only moves some files around. I believe it will be also pretty simple to implement in a coherent way - so I'll be preparing a demonstration very soon. I'll be using this branch as a base, if you want to keep this PR using these weird `make` rules, please let me know and I'll move to a new branch.

Makefile
```diff
@@ -40,18 +48,22 @@ format:
...
+	cd sysroot && find -depth -print | cpio --format=crc -o -F ../$@ 2> /dev/null
 
 clean:
 	$(foreach DIR, $(SUBDIRS), $(MAKE) -C $(DIR) $@;)
 	$(RM) -f *.a *.elf *.map *.lst *~ *.log *.cpio .*.D
 	$(RM) -f tags etags cscope.out *.taghl
 
-.PHONY: format tags cscope $(SUBDIRS)
+distclean: clean
+	$(RM) -rf cache
```
cahirwpz: `sysroot` as well?

Makefile
```diff
@@ -40,18 +48,22 @@ format:
...
-include $(wildcard .*.D)
+# Detecting whether initrd.cpio requires rebuilding is tricky, because even if
+# this target was to depend on $(shell fild sysroot -type f), then make compares
+# sysroot files timestamps BEFORE recursively entering user and installing user
+# programs into sysroot. This sounds silly, but apparently make assumes no files
+# appear "without their explicit target". Thus, the only thing we can do is
+# forcing make to always rebuild the archive.
+initrd.cpio: force | user
+	@echo "Building $@..."
+	cd sysroot && find -depth -print | cpio --format=crc -o -F ../$@ 2> /dev/null
```
cahirwpz: Please use `CPIO` variable set to `cpio --format=crc`.

Makefile
```diff
@@ -1,36 +1,44 @@
...
 cscope:
 	cscope -b include/*.h ./*/*.[cS]
 
 tags:
-	find -iname '*.[ch]' -not -path "*/toolchain/*" | ctags --language-force=c -L-
-	find -iname '*.[ch]' -not -path "*/toolchain/*" | ctags --language-force=c -L- -e -f etags
-	find -iname '*.S' -not -path "*/toolchain/*" | ctags -a --language-force=asm -L-
-	find -iname '*.S' -not -path "*/toolchain/*" | ctags -a --language-force=asm -L- -e -f etags
+	@echo "Rebuilding tags..."
+	find -iname '*.[ch]' -not -path "*/toolchain/*" -and -not -path "*newlib*" | ctags --language-force=c -L-
```
cahirwpz: What about introducing additional step that generates `sources.lst` file and then reuses it to run `ctags`?

Makefile
```diff
@@ -1,36 +1,44 @@
...
+	find -iname '*.[ch]' -not -path "*/toolchain/*" -and -not -path "*newlib*" | ctags --language-force=c -L- -e -f etags
+	find -iname '*.S' -not -path "*/toolchain/*" -and -not -path "*newlib*" | ctags -a --language-force=asm -L-
+	find -iname '*.S' -not -path "*/toolchain/*" -and -not -path "*newlib*" | ctags -a --language-force=asm -L- -e -f etags
 
 # These files get destroyed by clang-format, so we exclude them from formatting
-FORMATTABLE_EXCLUDE = include/elf stdc/ include/mips/asm.h include/mips/m32c0.h
+FORMATTABLE_EXCLUDE = include/elf stdc/ include/mips/asm.h include/mips/m32c0.h sysroot newlib-
 # Search for all .c and .h files, excluding toolchain build directory and files from FORMATTABLE_EXCLUDE
 FORMATTABLE = $(shell find -type f -not -path "*/toolchain/*" -and \( -name '*.c' -or -name '*.h' \) | grep -v $(FORMATTABLE_EXCLUDE:%=-e %))
```
cahirwpz: Could you reuse `sources.lst` here as well?

Makefile.common
```diff
@@ -6,16 +6,22 @@ SOURCES = $(SOURCES_C) $(SOURCES_ASM)
...
+TRIPLET_OLD = mipsel-unknown-elf
+
+TOOLCHAIN_FOUND = $(shell which $(TRIPLET)-gcc > /dev/null; echo $$?)
+OLD_TOOLCHAIN_FOUND = $(shell which $(TRIPLET_OLD)-gcc > /dev/null; echo $$?)
+ifneq ($(TOOLCHAIN_FOUND), 0)
+  ifneq ($(OLD_TOOLCHAIN_FOUND), 0)
+    $(error "mips-mimiker-elf" toolchain not found)
+  else
+    # Old toolchain is present, but the new one is not. Instruct the developer.
+    $(error You are using "mips-unknown-elf" toolchain. Please switch to "mips-mimiker-elf")
```
cahirwpz: It's probably better to ask developer to re-read `README.md` in previous branch, rather than explicitely checking for old toolchain.

README.md
```diff
@@ -15,10 +15,10 @@ cd toolchain/mips/
 ct-ng build
 ```
 
-By default, this will build and install the `mipsel-unknown-elf` toolchnain to
-`~/local`. Update your `$PATH` so that it provides `mipsel-unknown-elf-*`,
+By default, this will build and install the `mipsel-mimiker-elf` toolchnain to
+`~/local`. Update your `$PATH` so that it provides `mipsel-mimiker-elf-*`,
 i.e. unless you've changed the install location you will need to append
-`~/local/mipsel-unknown-elf/bin` to your `PATH`.
+`~/local/mipsel-mimiker-elf/bin` to your `PATH`.
```
cahirwpz: Please add link to pre-compiled debian package with the toolchain.

user/Makefile
```diff
@@ -2,20 +2,34 @@
...
+.PHONY: clean all $(PROGDIRS)
 
-all: $(SUBDIRS)
+all: $(PROGDIRS) install
 
-$(SUBDIRS): sysroot
+install: $(PROGDIRS-inst)
+
+$(PROGDIRS):
+	@echo "Maybe compiling $@..."
```
cahirwpz: Please keep make messages format consistent, i.e. `[BUILD] $@` and `[SYSROOT] $@`.

user/Makefile
```diff
@@ -2,20 +2,34 @@
...
 	$(MAKE) -C $@
 
-sysroot: sysroot/lib/libmimiker.a
-sysroot/lib/libmimiker.a: start.o syscalls.o
+# .inst targets represent the installation of a program into sysroot.
+%.inst: %
```
cahirwpz: From my experience (`gcc`, `binutils` and generally speaking GNU programs source code) such targets are called `%-install`. It would be nice to find reasonable document describing Makefile conventions ([GNU](https://www.gnu.org/prep/standards/html_node/Makefile-Conventions.html) is not the best example) and stick to it.

cahirwpz:
Your last post is definitely worth saving as an issue in tracking system! I roughly get the idea and I like it, however there's nothing better than you explaining the whole concept in person.

PR #277, opened 2017-04-27
Hardware independent PCI bus driver

cahirwpz:
PCI bus enumeration and BAR configuration code has been decoupled from actual method of accessing PCI configuration space by introduction of `gt_pci` driver. Enumeration process creates a set of `resource_t` elements that could be allocated by driver code (after [bus_alloc_resource](https://www.freebsd.org/cgi/man.cgi?query=bus_alloc_resource&apropos=0&sektion=9&manpath=FreeBSD+11.0-RELEASE+and+Ports&arch=default&format=html) counterpart is implemented).

Proper drivers are still missing. PCI driver has a temporary structure gathering all required resources for it to work. There will be a follow up PR that will try to clean up the code in preparation for basic driver structure. I'll investigate VGA and RTC driver structure.


mips/pci.c
```diff
@@ -192,32 +170,27 @@ static void pci_bus_dump(pci_bus_t *pcibus) {
...
-static pci_bus_t pci_bus[1];
+extern pci_bus_device_t gt_pci;
 
 void pci_init() {
   kmalloc_init(mp, 1, 1);
-  pci_bus_enumerate(pci_bus);
-  pci_bus_assign_space(pci_bus, MALTA_PCI0_MEMORY_BASE, PCI_IO_SPACE_BASE);
-  pci_bus_dump(pci_bus);
+
+  pci_bus_enumerate(&gt_pci);
```
rafalcieslak: `pci_bus_enumerate` finds all devices behind this PCI bus, and creates corresponding `pci_device`s. So we'll probably want to call `enumerate` for all buses that are present in the system. Either we'll need a list of all PCI buses for that, or we'll be enumerating new buses when a PCI-PCI bridge driver is bound to a bridge (and thus a new bus is discovered).

mips/cbus.c
```diff
@@ -2,20 +2,20 @@
...
 
 static bus_space_t cbus_bus_space = {
   .read_1 = cbus_read_1, .write_1 = cbus_write_1,
 };
 
 resource_t cbus_uart[1] = {{.r_bus_space = &cbus_bus_space,
-                            .r_addr = MALTA_CBUS_UART,
-                            .r_start = 0,
-                            .r_end = 7}};
+                            .r_type = RT_MEMORY,
```
rafalcieslak: This resource represents UART registers, maybe `RT_IOPORTS` would be more appropriate?

mips/gt64120.c
```diff
@@ -0,0 +1,77 @@
...
+          ((reg) << PCI0_CFG_REG_SHIFT));
+}
+
+typedef union {
+  uint8_t byte[4];
+  uint16_t word[2];
+  uint32_t dword;
+} pci_reg_t;
+
+static uint32_t gt_pci_read_config(pci_device_t *dev, unsigned reg,
```
rafalcieslak: Would it be possible to make this function universal for all `pci_bus`es, not just the particular one called `gt`? I suppose `pci_bus` might contain information about `CFG_ADDR` and `CFG_DATA` registers, and using them would make these two functions bus-independent. Are some other PCI bus implementations configured using different mechanisms (special processor instructions come to mind)?

include/pci.h
```diff
@@ -29,36 +31,74 @@ extern const char *pci_class_code[];
...
-} pci_device_t;
-
-typedef struct {
-  unsigned ndevs;
-  pci_device_t *dev;
-} pci_bus_t;
+  resource_t bar[6];
+};
+
+typedef struct pci_bus_device {
```
cahirwpz: AFAIU eventually `pci_bus_device` will become particular instance of `device` type and `pci_bus` will become special case of `bus` type. To put it differently `pci_bus_device` holds the state of device that happens to be a PCI bus, while `pci_bus` delivers common set of functions to operate on the bus.

mips/pci.c
cahirwpz: I'm not trying yet to glue together all components with complete autoconfiguration procedure in mind. Please note that PCI bus is particular type of a device. As such it needs driver, so we need proper definitions of `device` (state) and `driver` (functions) structures. `pci_bus_enumerate` will become part of [device_attach](https://www.freebsd.org/cgi/man.cgi?query=DEVICE_ATTACH&sektion=9) counterpart.

mips/cbus.c
cahirwpz: I initially thought the same, but... Firstly, MIPS architecture has no notion of hardware I/O ports, as known from PC platform. Secondly, CBUS-UART is connected directly to the CPU by common address/data bus. Although PCI I/O space is reachable through memory as well, it's clear (for me?) that all resources mapped there should by marked with `RT_IOPORTS`.

mips/gt64120.c
cahirwpz: PCI configuration space can be accessed in various ways depending on PCI bus controller and platform architecture. For instance on classic PC architecture you have to use `in` and `out` IA-32 instructions – this method is described in corresponding [topic](http://wiki.osdev.org/PCI#Configuration_Space) on OSDev wiki.

The implementation of `gt_pci` not only touches specific GT-64120 registers, but also takes into account that MIPS processor cannot handle unaligned accesses.

include/bus.h
```diff
@@ -26,11 +26,22 @@ struct bus_space {
...
+#define RF_PREFETCHABLE 1
+
 struct resource {
   bus_space_t *r_bus_space; /* bus space accessor descriptor */
-  intptr_t r_addr;  /* address of a set of resources mapped onto the bus */
-  intptr_t r_start; /* first address of the resource */
-  intptr_t r_end;   /* last address of the resource */
+  void *r_owner;            /* pointer to device that owns this resource */
+  intptr_t r_start;         /* first physical address of the resource */
+  intptr_t r_end;           /* last physical address of the resource */
```
cahirwpz: For some reason [resource_t](http://bxr.su/FreeBSD/sys/kern/subr_rman.c#94) in FreeBSD uses the range of form [`r_start`, `r_end`] and not [`r_start`, `r_end`). If you dump contents of `/proc/ioports` and `/proc/iomem` then ranges are specified in former way, not latter.

I'd leave it as is for now. Provided it will become cumbersome I won't hesitate to change the meaning of bounds to right-open interval.

cahirwpz:
> Do you plan to introduce a generic device structure for all kinds of devices (a `pci_device` would be a specific kind of `device`)?

Yes, I do. However initially the configuration will be static to large degree. Indeed `pci_device` will become just a `device`, but I have to play with the idea of device and driver first.

cahirwpz:
Ok. I added some comments and `TODO`'s for future. Do you have any further remarks?

rafalcieslak:
Thanks for your explanations, I can now see the greater picture! These extra comments really clear things up. Consider me now in favor of merging this branch.

PR #279, opened 2017-04-28
Function helpers to create uio_t structure for single transfer

cahirwpz:
This change introduces macro `MAKE_UIO` that defines `uio_t` structure with single I/O operation. With it we capture common use case into a single line. This is an alternative to #274.

I'm planning to use `-fplan9-extensions` features in our code. I find it extremely convenient to use camouflaged subtyping for structures. This works as follows:

```c
struct A {
  int foo;
};

struct B {
  struct A;
  unsigned bar;
};

void incr(struct A *a) {
 a->foo++;
}

void func() {
 struct B b = { -1, 2 };
 incr(&b);
}
```

### DIFF
--- a/include/uio.h
+++ b/include/uio.h
@@ -21,25 +21,19 @@ typedef struct uio {
   vm_map_t *uio_vmspace; /* destination address space */
 } uio_t;
 
-/* Uses -fplan9-extensions described in:
+/* Uses struct subtyping enabled by `-fplan9-extensions` flag described in:
  * https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html */
-#define MAKE_UIO(name, op, vm_map, buf, count, offset)                         \
-  struct {                                                                     \
-    uio_t;                                                                     \
-    iovec_t iov;                                                               \
-  } name = {(uio_t){.uio_iov = &name.iov,                                      \
-                    .uio_iovcnt = 1,                                           \
-                    .uio_offset = (offset),                                    \
-                    .uio_resid = (count),                                      \
-                    .uio_op = (op),                                            \
-                    .uio_vmspace = (vm_map)},                                  \
-            (iovec_t){(buf), (count)}}
+typedef struct uio_single {
+  uio_t;
+  iovec_t iov;
+} uio_single_t;
 
-#define MAKE_UIO_USER(name, op, buf, count, offset)                            \
-  MAKE_UIO(name, op, get_user_vm_map(), buf, count, offset)
-
-#define MAKE_UIO_KERNEL(name, op, buf, count, offset)                          \
-  MAKE_UIO(name, op, get_kernel_vm_map(), buf, count, offset)
+void make_uio(uio_single_t *uio, uio_op_t op, vm_map_t *vm_map, void *buf,
+              size_t count, off_t offset);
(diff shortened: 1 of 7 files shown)


cahirwpz:
I merged my ideas with those from @coodie change. Please note `sys/exec.c` was modified to use `uio_make*` as well.

include/uio.h
```diff
@@ -21,6 +21,20 @@ typedef struct uio {
   vm_map_t *uio_vmspace; /* destination address space */
 } uio_t;
 
+/* Uses struct subtyping enabled by `-fplan9-extensions` flag described in:
+ * https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html */
+typedef struct uio_single {
```
coodie: I think adding this structure and using -plan9 extensions is pretty fancy, but it's not widely used and makes code less readable. This is pretty nice feature to be used, but it would be weird to see it only in one place and I don't think it gives us any real benefits, except that it brings confusion to the people unfamiliar with this.

tests/initrd.c
```diff
@@ -12,19 +12,9 @@ static void dump_file(const char *path) {
...
-  uio.uio_iovcnt = 1;
-  uio.uio_vmspace = get_kernel_vm_map();
-  uio.uio_iov = &iov;
-  uio.uio_offset = 0;
-  iov.iov_base = buffer;
-  iov.iov_len = sizeof(buffer);
-  uio.uio_resid = sizeof(buffer);
 
+  uio_single_t uio;
+  make_uio_kernel(&uio, UIO_READ, buffer, sizeof(buffer), 0);
```
coodie: I'll just describe in more details why I don't like this extension.

Most of people to join this project aren't going to be C experts. 
The following line:
```
    res = VOP_READ(v, &uio); 
```

Is very confusing. Someone unfamiliar with these extensions will probably think there is some magic cast going on in here. I'm 100% confident that simply naming the field in `uio_single_t` (say `uio`) and making:

``` 
make_uio_kernel(&uio.uio, UIO_READ, buffer, sizeof(buffer), 0); 
```

Is more readable.

include/uio.h
cahirwpz: Ok. You're right that this code does justify introduction of Plan9 extensions to C language. We'll revisit this idea at the next meeting. Hopefully I'll be able to show some better use cases for it.

tests/initrd.c
cahirwpz: I changed to be in line with your suggestions.

rafalcieslak:
The original example you've provided in PR description is also a great example of what bothers me most about indirect access to anonymous field. It introduces implicit conversions to otherwise very explicit language.

```c
void incr(struct A *a);

  struct B b = { -1, 2 };
  incr(&b);
```

The `incr(&b)` looks like there is some **implicit cast** going on (`struct B*` to `struct A*`, which, as a matter of fact, would cause UB), but that's *not* what actually happens here! This conversion is more like C++-style upcasting. (*I don't know the detailed plan9 compiler history, but given it's age it is likely that this extension was indeed an early attempt at richer types before C++ began to crystallize.*)

cahirwpz:
I see this GCC extension drew a lot of controversy. Let's discuss it at the next meeting.

PR #281, opened 2017-04-29
Configurable VGA video mode

rafalcieslak:
~**NOTE: This branch is built on top of #275 (VGA driver) and #280 (Imported sscanf), so the diff won't make sense until these other two branches are merged!**~

This branch introduces support for dynamically configured VGA video mode. The user-space program `mandelbrot` enables higher resolution by writing `640 480 8` to `/dev/vga/videomode`. The interface is generic, and VGA drivers may choose to support it or not.

It is also a nice demonstration of how *devfs* may be used for configuring drivers.


sys/dev_vga.c
```diff
@@ -14,10 +14,54 @@ static int dev_vga_palette_write(vnode_t *v, uio_t *uio) {
...
+
+static int dev_vga_resolution_write(vnode_t *v, uio_t *uio) {
+  vga_device_t *vga = (vga_device_t *)v->v_data;
+  uio->uio_offset = 0; /* This file does not support offsets. */
+  uint16_t xres, yres;
+  unsigned int bpp;
+  int error = vga_get_resolution(vga, &xres, &yres, &bpp);
+  if (error)
+    return error;
+  char buffer[RES_CTRL_BUFFER_SIZE];
```
cahirwpz: Because of quite limited kernel stack size, it's a good practice to keep arrays on heap (`vga_device_t` struct?).

sys/dev_vga.c
```diff
@@ -14,10 +14,54 @@ static int dev_vga_palette_write(vnode_t *v, uio_t *uio) {
   return vga_palette_write((vga_device_t *)v->v_data, uio);
 }
 
+#define RES_CTRL_BUFFER_SIZE 32
+
+static int dev_vga_resolution_write(vnode_t *v, uio_t *uio) {
+  vga_device_t *vga = (vga_device_t *)v->v_data;
+  uio->uio_offset = 0; /* This file does not support offsets. */
+  uint16_t xres, yres;
```
cahirwpz: Is it somehow justified to use `uint16_t` instead of regular `unsigned`. Later on you have to pass `%hd` to `scanf` which looks somewhat confusing.

sys/stdvga.c
```diff
@@ -88,6 +91,45 @@ static int stdvga_palette_write(vga_device_t *vga, uio_t *uio) {
...
+  *bpp = stdvga->bpp;
+  return 0;
+}
+
+static int stdvga_set_resolution(vga_device_t *vga, uint16_t xres,
+                                 uint16_t yres, unsigned int bpp) {
+  stdvga_device_t *stdvga = STDVGA_FROM_VGA(vga);
+
+  /* I have not tested larger resolutions yet. */
+  if (xres > 640 || yres > 480)
```
cahirwpz: You could either browse an array of predefined video modes or autodetect them somehow...

sys/dev_vga.c
```diff
@@ -14,10 +14,54 @@ static int dev_vga_palette_write(vnode_t *v, uio_t *uio) {
...
+  if (error)
+    return error;
+  return 0;
+}
+
+static int dev_vga_resolution_read(vnode_t *v, uio_t *uio) {
+  vga_device_t *vga = (vga_device_t *)v->v_data;
+  uint16_t xres, yres;
+  unsigned int bpp;
+  int error = vga_get_resolution(vga, &xres, &yres, &bpp);
```
cahirwpz: Hmm... I think it's better to call it *video mode* rather than *resolution* because of `bpp` component. If choose to do so then control file should probably be named `videomode`

include/vga.h
```diff
@@ -8,10 +8,16 @@ typedef struct vga_device vga_device_t;
 
 typedef int (*vga_palette_write_t)(vga_device_t *vga, uio_t *uio);
 typedef int (*vga_fb_write_t)(vga_device_t *vga, uio_t *uio);
+typedef int (*vga_get_resolution_t)(vga_device_t *vga, uint16_t *xres,
+                                    uint16_t *yres, unsigned int *bpp);
```
cahirwpz: AFAIR we use `unsigned` instead of `unsigned int` in the kernel.

sys/dev_vga.c
rafalcieslak: I initially did not include the bpp component, therefore *resolution*. I agree it should be renamed to *videomode*.

sys/dev_vga.c
rafalcieslak: I understand, but this buffer is very small, in fact 16 bytes should be more than enough. I wouldn't like to place it in `vga_device_t` struct, as this buffer is a part of the `dev_vga` interface, which is a layer on top of the `vga` interface. So for the best solution, I'd need to create yet another malloc pool for `dev_vga`... but maybe not before we have some general purpose pools in place.

sys/stdvga.c
rafalcieslak: Actually, BochsVBE extensions are very versatile, and apparently QEMU's stdvga accepts arbitrary dimensions for screen resolution - it would work correctly even with weird stuff like `12345 x 1`. I've only added this condition to impose some reasonable limit on input numbers.

This will be very different for `cirrus` driver for OVPsim (which I've also started working on), as the configuration for each video mode will be very different. Thus I intend to support only some selected video modes, and indeed checking against a list will be the only feasible solution.

sys/dev_vga.c
rafalcieslak: `vga_{get,set}_resolution` use `uint16_t` because I wanted to ensure that there are at least 12 bits for dimensions. But as it's unlikely that we'll ever want to run mimiker on an architecture with word size < 12 bits, I can probably use the generic `unsigned` instead.

PR #282, opened 2017-05-01
Initializer macro for single UIO transfer

cahirwpz:
This is alternative solution to PR #274. I think I've found the cleanest solution to the problem, that does not require using rare C extensions.

### DIFF
--- a/include/uio.h
+++ b/include/uio.h
@@ -21,6 +21,19 @@ typedef struct uio {
   vm_map_t *uio_vmspace; /* destination address space */
 } uio_t;
 
+#define UIO_SINGLE(op, vm_map, offset, buf, buflen)                            \
+  (uio_t) {                                                                    \
+    .uio_iov = (iovec_t[1]){(iovec_t){(buf), (buflen)}}, .uio_iovcnt = 1,      \
+    .uio_offset = (offset), .uio_resid = (buflen), .uio_op = (op),             \
+    .uio_vmspace = (vm_map)                                                    \
+  }
+
+#define UIO_SINGLE_KERNEL(op, offset, buf, buflen)                             \
+  UIO_SINGLE(op, get_kernel_vm_map(), offset, buf, buflen)
+
+#define UIO_SINGLE_USER(op, offset, buf, buflen)                               \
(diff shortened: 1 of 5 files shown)


cahirwpz:
@coodie @rafalcieslak Could you compare this solution against the original?

include/uio.h
```diff
@@ -21,6 +21,19 @@ typedef struct uio {
   vm_map_t *uio_vmspace; /* destination address space */
 } uio_t;
 
+#define UIO_SINGLE(op, vm_map, offset, buf, buflen)                            \
+  (uio_t) {                                                                    \
+    .uio_iov = (iovec_t[1]){(iovec_t){(buf), (buflen)}}, .uio_iovcnt = 1,      \
```
coodie: Can you briefly explain how does that work? I don't see where you get space for iov.

cahirwpz: The space for `iov` gets allocated on stack. With C99 (checked with `-std=c99 -Wpedantic`) you can do following:
```c
void foobar() {
  char *str;
  str = "foobar";
  str = (char[6]){'h', 'e', 'l', 'l', 'o', 0};
  int *tab;
  tab = (int[4]){1,2,3,4};
}
```
In both cases arrays (and string as well) will be allocated on stack (checked with optimizations turned off and `objdump -d`), unless they are constant (if so they're put into `.rodata` section). Does this explanation help?

tests/initrd.c
```diff
@@ -12,19 +12,9 @@ static void dump_file(const char *path) {
...
-  uio.uio_iovcnt = 1;
-  uio.uio_vmspace = get_kernel_vm_map();
-  uio.uio_iov = &iov;
-  uio.uio_offset = 0;
-  iov.iov_base = buffer;
-  iov.iov_len = sizeof(buffer);
-  uio.uio_resid = sizeof(buffer);
 
+  uio_t uio;
+  uio = UIO_SINGLE_KERNEL(UIO_READ, 0, buffer, sizeof(buffer));
```
rafalcieslak: Maybe:
```c
uio_t uio = UIO_SINGLE_KERNEL(UIO_READ, 0, buffer, sizeof(buffer));
```

PR #283, opened 2017-05-01
Fixed seeking for write-only files

rafalcieslak:
Fixed a tiny bug in `do_lseek`, where the file requested with a descriptor was assumed to be readable. This is not the case e.g. for `/dev/vga/fb`, and thus user programs were unable to seek back to framebuffer start.

### DIFF
--- a/sys/vfs_syscalls.c
+++ b/sys/vfs_syscalls.c
@@ -60,7 +60,7 @@ int do_lseek(thread_t *td, int fd, off_t offset, int whence) {
   /* TODO: Whence! Now we assume whence == SEEK_SET */
   /* TODO: RW file flag! For now we just file_get_read */
   file_t *f;
-  int res = fdtab_get_file(td->td_fdtable, fd, FF_READ, &f);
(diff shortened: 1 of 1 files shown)


sys/vfs_syscalls.c
```diff
@@ -60,7 +60,7 @@ int do_lseek(thread_t *td, int fd, off_t offset, int whence) {
   /* TODO: Whence! Now we assume whence == SEEK_SET */
```
cahirwpz: Could you fix other unhandled cases mentioned in this comment as well?

rafalcieslak: Actually, this will be pretty difficult until #276 (Compiling newlib) is merged - because the exact values for flags `SEEK_SET`, `SEEK_CUR` and `SEEK_END` would need to match what newlib provides in it's headers.

cahirwpz: I see... so please add corresponding issue to tracking system and merge this PR as is.

PR #284, opened 2017-05-01
Basic ps/2 8042 support

rafalcieslak:
I've implemented a very basic 8042 driver, which initial support for a ps/2 keyboard. The scancodes are gathered by a separate thread, which collects them into a buffer that user programs may read via `/dev/scancode`.

My implementation does not use interrupts (but I suppose it'll be pretty easy to enable them once we support interrupt management). Also, naturally, the `/dev/scancode` is not `poll`able, so it requires reading repeatedly (or, e.g. every frame, since events are buffered). The scancodes are not translated into keypress/keyrelease events in any way - I've only implemeted the "raw" interface.

However, even with these limitations in place, my [`headless_doom`](https://github.com/rafalcieslak/headless_doom) fork already supports the `scancode` interface, and, with #283, **DOOM is now playable!**

*EDIT: For the record, this also works under OVPsim - but the FDC37M817 ignores scan code table selection, and insists on using mode 2. Sigh. This doesn't affect this branch though, only that user-side scancode recognition is trickier.*

*EDIT2: I also have a feeling this driver may be the perfect place for demonstrating interrupt registration and handling.*


rafalcieslak:
Travis found an issue with the `pmap` test: it assumed there are no other threads in the system, and that it doesn't have to worry about preemption. But this branch introduces a "keyboard reader" thread which keeps running all the time (in a future change, it will just be woken up by an interrupt handler) - so I had to wrap both `pmap` tests into a critical section.

rafalcieslak:
I've closely examined OVPsim sources. It's absolutely clear that OVPsim's PS/2 support is a direct copy from QEMU. I've tried to find the exact QEMU version they used, and I've only managed to find out that it must have been at most 0.8... which was released way back in 2006. OVPsim's copied implementation had very little changes since then, and is quite limited. For example, it ignores requests for selecting a scan code table, and always sticks to table 2 (so I guess my hands are tied and I'll stick to table 2 as well). Many keys produce incorrect scancodes. Some keys only generate BREAK and no MAKE scancode. I can't programmatically differentiate QEMU's keyboard from OVPsim's, because they respond with the same keyboard identification code (OVPsim's uses QEMU's unmodified code).

So, generally, keyboard support on OVPsim is a real mess, and there is nothing we can do about it. I'll update this branch to deal with OVPsim as well as it's possible, but it will leave much to desire.

rafalcieslak:
This branch is currently waiting for #292 (PCI-ISA bridge) to be implemented.

PR #290, opened 2017-05-02
Moved fdtable from thread_t to proc_t

rafalcieslak:
This was certainly simpler than `uspace`. Note that until we fix `exit` to kill all process' threads, the file descriptor table never gets released (because processes are never destroyed).

### DIFF
--- a/sys/exec.c
+++ b/sys/exec.c
@@ -216,18 +216,6 @@ int do_exec(const exec_args_t *args) {
   prepare_program_stack(args, &stack_bottom);
 
   thread_t *td = thread_self();
-  /* ... file descriptor table ... */
-  /* TODO: Copy/share file descriptor table! */
-  fdtab_t *fdt = fdtab_alloc();
-  fdtab_ref(fdt);
-  td->td_fdtable = fdt;
-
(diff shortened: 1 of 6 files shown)


sys/exec.c
```diff
@@ -242,6 +230,19 @@ int do_exec(const exec_args_t *args) {
     proc_populate(proc, td);
   }
 
+  /* Prepare file descriptor table */
+  /* TODO: Copy/share file descriptor table! */
```
cahirwpz: I find this comment unclear. Unix-like `exec` does not modify descriptor table unless a descriptor was marked with `O_CLOEXEC` flag on `open`. I understand that our `exec` has unusual use case – i.e. it's used to start user-space programs from kernel, therefore we need to allocate fresh descriptor table.

rafalcieslak: You are correct. Initially `do_exec` was not meant to serve a system call (back then we did not have working syscalls), but I'm starting to move it into that direction. Thus I suppose the file descriptor table should only be created when starting the very first process (that is: `do_exec` was called by a kernel thread), otherwise the table should stay unmodified and it is `fork` who should create a copy or shared reference. I'll fix that ASAP.

PR #297, opened 2017-05-07
Early device driver infrastructure

cahirwpz:
@rafalcieslak I put this under review only because you insisted to have device drivers sooner than later. I'm pretty confident the change is going in right direction.


cahirwpz:
Please find some extra information below that probably should make it way into comments: 
* `bus_methods` structure will contain functions `bus_intr_setup`, `bus_intr_teardown`, `bus_alloc_resource` and alike,
* `device_probe` returns an integer, the higher the more suitable the driver is,
* `device_attach` returns 0 on success, and `errno.h` number on failure,
* `device_t` is not supposed to be extended as we initially discussed, all extra data should be contained within `state` and `instance`.

rafalcieslak:
About `device_t` extensions. What are your arguments against the layout we've originally discussed?

```c
struct pci_device{
    pci device data...
    device_t device;
}
```

Other device types would provide their own `xyz_device` structure. In fact *anything that contains an embedded `struct device` is a device*. I find this layout much cleaner and more consistent than hiding everything behind `void*` pointers.

My arguments for that layout:

1. Less memory allocations are performed, less pointer dereferences are required, and cache locality is improved.

2. Less typeless pointers are present in sources. 

3. pci-specific functions, like `pci_read_config` would take a `pci_device*` as an argument, instead of generic `device_t*`. That means compiler can do more type checking for us! (this is my favorite argument)

4. Source code is slightly simpler to read, because there are much less spots where you need to know what is the exact type of a device, as it can be explicitly stated in the source.

I suppose there may be more reasons to prefer the second layout - it should be easy to find them, the driver documentation from the period when Linux switched from the first to the second layout does mention some more reasons for the switch.

sys/device.c
```diff
@@ -1,3 +1,63 @@
...
+int device_probe(device_t *dev) {
+  return dev->driver->probe ? dev->driver->probe(dev) : 0;
+}
+
+int device_attach(device_t *dev) {
+  dev->state = kmalloc(M_DEV, dev->driver->size, M_ZERO);
+  return dev->driver->attach(dev);
+}
+
+int device_detach(device_t *dev) {
```
rafalcieslak: Are we going to pretend that we support device detaching? If no, let's remove this function. If yes, then I suppose this function needs to check whether `dev->driver->detach != NULL` - for example, the `stdvga` driver does not set `.detach`.

sys/device.c
```diff
@@ -1,3 +1,63 @@
...
+
+void driver_init() {
+  device_t *pcib = device_alloc();
+  pcib->driver = &gt_pci.driver;
+  device_attach(pcib);
+
+  device_t *dev;
+  SET_DECLARE(driver_table, driver_t);
+  log("Scanning %s for known devices.", pcib->driver->desc);
+  TAILQ_FOREACH (dev, &pcib->children, link) {
```
rafalcieslak: Shouldn't this loop be implemented as (final?) part of PCI attachment procedure? I suppose this code is pci-specific, and does make some assumptions about what kind of devices it operates on.

sys/device.c
```diff
@@ -1,3 +1,63 @@
...
+
+int device_detach(device_t *dev) {
+  int res = dev->driver->detach(dev);
+  if (res == 0)
+    kfree(M_DEV, dev->state);
+  return res;
+}
+
+/* Normally this file an intialization procedure would not exits! This is barely
+   a substitute for currently unimplemented device-driver matching mechanism. */
```
rafalcieslak: This comment is probably outdated.

include/pci.h
```diff
@@ -81,32 +84,28 @@ struct pci_device {
...
+static inline uint32_t pci_adjust_config(device_t *device, unsigned reg,
                                          unsigned size, uint32_t value) {
   pci_write_config(device, reg, size, value);
   return pci_read_config(device, reg, size);
 }
 
-void pci_init();
+void pci_bus_enumerate(device_t *pcib);
+void pci_bus_assign_space(device_t *pcib);
+void pci_bus_dump(device_t *pcib);
```
rafalcieslak: Don't you think this entire header would be much cleaner if all these functions took `pci_device_t*` as argument?

Also, I see that all these functions blindly assume that the `device_t` they are given is indeed a PCI device. I hope this will be always the case, but think what horrible problems will happen, if, due to a bug or a mistake, they are passed a different kind of device. If we really don't want to use `pci_device_t*` type and let the compiler check this for us, then I suggest to introduce an enum inside `device_t` which would store device type, so that we could `assert(device->type == DEV_TYPE_PCI);`.

sys/drv_stdvga.c
```diff
@@ -145,23 +144,35 @@ static int stdvga_fb_write(vga_device_t *vga, uio_t *uio) {
...
 }
 
-int stdvga_pci_attach(pci_device_t *pci) {
-  if (pci->vendor_id != VGA_QEMU_STDVGA_VENDOR_ID ||
-      pci->device_id != VGA_QEMU_STDVGA_DEVICE_ID)
+static int stdvga_probe(device_t *dev) {
+  pci_dev_data_t *pcid = dev->instance;
+
+  if (pcid == NULL)
+    return 0;
```
rafalcieslak: What is the purpose of this test? I suppose it might be turned into `assert(pcid != NULL)`, `instance` will always be set (to something) when we're probing a device.

sys/device.c
cahirwpz: Let's leave it here FTTB. There's nothing PCI-specific to this code. I guess it should become part of generic bus code in the long run.

include/pci.h
cahirwpz: Well... they don't take a `pci_device_t` but rather `pci_bus_device_t`. Those two types would not be equivalent if they existed.

sys/device.c
cahirwpz: Fixed that.

include/pci.h
rafalcieslak: Okay, so what I actually meant was this interface would ideally look like this:

```c
uint32_t pci_read_config(pci_device_t *device, unsigned reg, unsigned size);
void pci_write_config(pci_device_t *device, unsigned reg, unsigned size, uint32_t value);
uint32_t pci_adjust_config(pci_device_t *device, unsigned reg, unsigned size, uint32_t value);
void pci_bus_enumerate(pci_bus_device_t *pcib);
void pci_bus_assign_space(pci_bus_device_t *pcib);
void pci_bus_dump(pci_bus_device_t *pcib);
```

In my suggested layout I would consider both `pci_device_t` and `pci_bus_device_t` extensions of `device_t`.

rafalcieslak:
I can agree to stop whining about `device_t` structure layout if you promise me to thoroughly reconsider my idea in few weeks time, once we have a bunch of drivers in place and I provide an elegant demonstration comparing these two solutions.

cahirwpz:
Frankly... I could not come up with a clean solution implementing proposed approach. I'd be more than happy if we could work out safer and cleaner way to achieve the same.

rafalcieslak:
Fair enough. I'll keep experimenting to learn what challenges we would need to solve. Either way, let's stick to your variant for now.

include/pci.h
cahirwpz: Looks much safer to use, indeed.

cahirwpz:
Should I merge the branch as is?

rafalcieslak:
Please do. This seems to be a fine foundation, we can add next features and improvement separately.

Issue #292, opened 2017-05-05
PCI-ISA bridge driver


rafalcieslak:
We urgently need a very simple driver for the PCI-ISA bridge, so that we may access the ISA bus and start writing drivers for devices on the ISA bus. The bridge driver itself won't have much to do, at the moment all it has to do is claiming the PCI device, and preparing the ISA bus. Later it will also be responsible for dispatching probe/attach requests to ISA drivers.

cahirwpz:
Obsoleted by #380.

PR #298, opened 2017-05-08
Initial PCI-ISA bridge and RTC drivers implmementation

rafalcieslak:
There are still some missing details I'll add soon, but I need to know if I'm going in the right direction. This implementation is functional, the RTC driver exposes a `/dev` interface, and the `rtc` kernel test reads from it.

My doubts about this branch concern how I manage `resource_t` and `resource_t*` references - because of how the ISA must be dynamically created when a PCI bridge is discovered.

In my implementation: 
 - a `pci` bus has an additional `lowio` resource which covers the low I/O addresses (as reserved by PCI standard for ISA compatibility)
 - a `piix4isa` device has a transparent *reference* to the `lowio` **and** its own `isa` resource, owned by the `piix4isa` device, which represents the actual `isa` bus (it provides its own read/write operations, which relay requests to the `lowio` reference).
 - all pci devices have a transparent *reference* to the `isa` bus resource
 - a particular pci device may store in its state a *reference* to `isa` bus, in future I expect this reference to only refer to a subset of addresses within `isa` bus.

This hierarchy looks logical to me, but I am still not 100% certain I understand all the details about `resource_t`.

*Adresses #292.*


sys/drv_mc146818.c
```diff
@@ -0,0 +1,110 @@
...
+
+  if (!isad)
+    return 0;
+
+  /* TODO: Actually do some very basic test to see if the RTC is present. */
+
+  return 1;
+}
+
+static void mc146818_write_reg(mc146818_state_t *mc146818, uint8_t reg,
```
cahirwpz: That should be a part of a `bus_space` attached to resource delivered by PCI-ISA bridge. We could have exactly the same device, but accessible directly through physical memory address space, that why I'd like access methods to be abstracted away.

rafalcieslak:
Does the `mc146818->regs` resource look better now?

cahirwpz:
I missed some concepts behind ISA bus, which was pointed out in #380, which requires to approach the problem quite differently. Hence the scope of changes offered by this PR is no longer relevant for ISA bus driver development.

PR #299, opened 2017-05-08
Change compiled flag from -Og to -O0

coodie:
It is said that -Og flag is supposed not to interfere with debugging, but in fact it does. Apparently this makes register allocation, making gdb print message that value is in register instead of just printing it, and possibly some other optimizations. I find this rather annoying than useful, and obviously we don't need any kind of optimization. In future, when our kernel has more functionalities we might introduce `release` and `debug` builds.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -41,7 +41,7 @@ GCC_SYSTEM_INC = -I$(GCC_INSTALL_PATH)include/ \
                  -D_LIBC_LIMITS_H_
 
 ASFLAGS  =
-CFLAGS   = -std=gnu11 -Og -Wall -Werror -fno-builtin -nostdinc -nostdlib \
+CFLAGS   = -std=gnu11 -O0 -Wall -Werror -fno-builtin -nostdinc -nostdlib \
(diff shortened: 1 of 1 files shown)


rafalcieslak:
I don't know, if we use `-Og` by default, you can still temporarily switch to `-O0` when you need it for debugging... Is it that much of a deal? When I work on the most obscure bugs, I tend to alter my debug config a lot.

Interestingly, Travis says our test fail when compiled with `-O0` (yet they pass with `-Og`, which is a clear indication of undefined behavior). It looks like the fault has something to do with preparing the ramdisk, could you give a look at test logs, do you know where the problem might be? Could it be something you've already fixed in #268 (Getdirentries syscall)?

cahirwpz:
Please introduce extra *make* variable that will switch between `DEBUG` and `NORMAL` builds, i.e. by invoking `make DEBUG=1` user instructs *make* to produce binaries suitable for in-depth debugging.

cahirwpz:
After some investigation done by @coodie #310 issue was created.

PR #300, opened 2017-05-08
Simplified mount interface

rafalcieslak:
Initially I only wanted to remove the `if (strncmp(path, "/dev/", 5) == 0)` and associated `vfs_root_dev_vnode`, turned out some extra cleanup to the mount process was required, and I ended up with an elegant
```c
vfs_domount_named("initrd", "/");
vfs_domount_named("devfs", "/dev");
```
in `kernel_init(...)`.

### DIFF
--- a/include/mount.h
+++ b/include/mount.h
@@ -42,6 +42,11 @@ typedef struct vfsconf {
   TAILQ_ENTRY(vfsconf) vfc_list; /* Entry on the list of vfsconfs */
 } vfsconf_t;
 
+/* The list of all installed filesystem types */
+typedef TAILQ_HEAD(, vfsconf) vfsconf_list_t;
+extern vfsconf_list_t vfsconf_list;
+extern mtx_t vfsconf_list_mtx;
+
 /* This structure represents a mount point: a particular instance of a file
    system mounted somewhere in the file tree. */
 typedef struct mount {
@@ -88,9 +93,6 @@ mount_t *vfs_mount_alloc(vnode_t *v, vfsconf_t *vfc);
 /* Mount a new instance of the filesystem vfc at the vnode v. Does not support
  * remounting. TODO: Additional filesystem-specific arguments. */
 int vfs_domount(vfsconf_t *vfc, vnode_t *v);
-
-/* Mount a new instance of the filesystem named fs at the requested path. */
-int vfs_domount_named(const char *fs, const char *path);
 
 /* Finds the vnode corresponding to the given path.
  * Increases use count on returned vnode. */
--- a/include/mount.h
+++ b/include/mount.h
@@ -74,11 +74,8 @@ static inline int VFS_VGET(mount_t *m, ino_t ino, vnode_t **vp) {
   return m->mnt_vfsops->vfs_vget(m, ino, vp);
 }
 
-/* This is the / node. Since we aren't mounting anything on / just yet, there is
-   also a separate global vnode for /dev .*/
+/* This is the / node.*/
 extern vnode_t *vfs_root_vnode;
-extern vnode_t *vfs_root_dev_vnode;
-extern vnode_t *vfs_root_initrd_vnode;
 
 /* Look up a file system type by name. */
 vfsconf_t *vfs_get_by_name(const char *name);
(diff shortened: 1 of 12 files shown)


cahirwpz:
`vfs_domount_named` looks to me like a good candidate for a function that would implement [mount](http://netbsd.gw.com/cgi-bin/man-cgi?mount+2+NetBSD-current) syscall! We wouldn't use last three arguments yet, but why not introduce it right now?

rafalcieslak:
I've considered that, but I don't think a `mount` syscall would be of much value. I suppose we wouldn't use it until we have some reasonable init program.

cahirwpz:
That's right, but we could make some provisions for the future, e.g. by renaming `vfs_mount_named` to `do_mount`, at least.

sys/vfs_syscalls.c
```diff
@@ -173,3 +186,27 @@ int sys_fstat(thread_t *td, syscall_args_t *args) {
     return error;
   return 0;
 }
+
+int sys_mount(thread_t *td, syscall_args_t *args) {
+  char *user_fsysname = (char *)args->args[0];
+  char *user_pathname = (char *)args->args[1];
+
+  int error = 0;
+  char fsysname[256];
```
cahirwpz: We must stop using kernel stack so carelessly... after all we have only a page (or two?) of it! Please look for reference at FreeBSD [sys_mount](http://bxr.su/FreeBSD/sys/kern/vfs_mount.c#731) – the actually use `M_TEMP` memory pool to `copyinstr`.

include/mount.h
```diff
@@ -42,6 +42,11 @@ typedef struct vfsconf {
   TAILQ_ENTRY(vfsconf) vfc_list; /* Entry on the list of vfsconfs */
 } vfsconf_t;
 
+/* The list of all installed filesystem types */
+typedef TAILQ_HEAD(, vfsconf) vfsconf_list_t;
+extern vfsconf_list_t vfsconf_list;
```
cahirwpz: Wouldn't it make more sense for now to have linker set to store those?

rafalcieslak: Huh? We do use a linker set for kernel-provided filesystems. Please look at `vfs_init()`.

cahirwpz: Oops. I guess I'm not that familiar with this part of the kernel.

sys/vfs_syscalls.c
```diff
@@ -173,3 +187,32 @@ int sys_fstat(thread_t *td, syscall_args_t *args) {
...
+  error = copyinstr(user_pathname, pathname, sizeof(pathname), &n);
+  if (error < 0)
+    goto end;
+
+  log("mount(\"%s\", \"%s\")", pathname, fsysname);
+
+  error = do_mount(td, fsysname, pathname);
+end:
+  kfree(M_TEMP, fsysname);
+  kfree(M_TEMP, pathname);
```
rafalcieslak: BTW, @coodie `scoped_ptr` here?

sys/malloc.c
```diff
@@ -267,3 +267,5 @@ void kmem_dump(kmem_pool_t *mp) {
 /* TODO: missing implementation */
 void kmem_destroy(kmem_pool_t *mp) {
 }
+
+MALLOC_DEFINE(M_TEMP, "temporaries pool", 2, 4);
```
coodie: I like idea behind this pool. Going to be easier to implement some tests in other branches where I needed to perform some memory allocation. It wouldn't hurt to increase it's size to be even bigger (say 32 pages).

sys/vfs_syscalls.c
```diff
@@ -85,6 +88,17 @@ int do_fstat(thread_t *td, int fd, vattr_t *buf) {
   return res;
 }
 
+int do_mount(thread_t *td, const char *fs, const char *path) {
```
coodie: We would want to extend interface of mount to have extra flags argument, so we can have 'read only', 'no exec' and possibly other types of mounts. Not a big deal and can be added in future.

sys/vfs_syscalls.c
coodie: Yes, scoped_ptr would be great in this case :), but it's not part of this PR.

PR #301, opened 2017-05-08
Minor debugging experience improvements

rafalcieslak:
- I've added a `user_return` label user before `eret` in `user_exc_leave`. When debugging user programs and/or syscalls, it is very convenient to place a `tbreak` on this label, to continue debugging the user application from the moment the current system call returns.

- As our kernel `main` no longer does anything interesting, I've changed the implicit breakpoint from `main` to `kernel_init`, so that a default debugging session starts there.

- I've added a boring `panic_fail` function (similar to `assert_fail`) so that I could add an implicit breakpoint on it. This way a default debugging session will not only automatically stop at failed assertions, but also at any panic condition.

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,7 +1,8 @@
 set tcp connect-timeout 30
 target remote localhost:1234
-break main
+break kernel_init
 break assert_fail
+break panic_fail
 continue
 
 source debug/ktrace.py
--- a/.gdbinit.template
+++ b/.gdbinit.template
@@ -1,5 +1,5 @@
 set tcp connect-timeout 30
-target remote localhost:GDB_PORT
+target remote localhost:${GDB_PORT}
 break kernel_init
 break assert_fail
 break panic_fail
--- a/.gdbinit.template
+++ b/.gdbinit.template
@@ -1,5 +1,5 @@
 set tcp connect-timeout 30
-target remote localhost:1234
+target remote localhost:GDB_PORT
 break kernel_init
 break assert_fail
 break panic_fail
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ cache
 
 # Dynamically created sysroot
 sysroot
+
+# Dynamically generated gdb config
+.gdbinit
--- a/include/common.h
+++ b/include/common.h
@@ -105,12 +105,12 @@ typedef uint16_t ino_t;
 #ifndef _USERSPACE
 
 /* Terminate thread. */
-noreturn void thread_exit();
+noreturn void panic_fail();
 
 #define panic(FMT, ...)                                                        \
   __extension__({                                                              \
-    kprintf("[%s:%d] " FMT "\n", __FILE__, __LINE__, ##__VA_ARGS__);           \
-    thread_exit(-1);                                                           \
+    kprintf("[%s:%d] PANIC: " FMT "\n", __FILE__, __LINE__, ##__VA_ARGS__);    \
+    panic_fail();                                                              \
   })
 
 #ifdef DEBUG
--- a/launch
+++ b/launch
@@ -6,7 +6,7 @@ import os.path
 import signal
 import launcher
 
-UART_PORT = 8000
+UART_PORT_BASE = 8000
 
 if __name__ == '__main__':
     # Probe which tools are available.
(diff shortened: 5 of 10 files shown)


cahirwpz:
LGTM. Could you also extend launch script to redirect CBUS UART messages to `x-terminal-emulator`? I had some issues running debug session on `mimiker` server with `nc localhost 8000` opened in other terminal (OTOH each user should get other port number I guess). I think `launch` script had some issues trying to kill simulator process (in both cases of OVPsim and QEMU).

rafalcieslak:
The reason you can't use the (default) `-o xterm` on the `mimiker` server (launcher automatically falls back to stdout) is because:
```
xterm: command not found
```

cahirwpz:
Right – that makes sense! Wouldn't it be better if we changed `xterm` to `x-terminal-emulator` so that system owner can choose right tool by using `update-alternatives` utility (`urxvt` on mimiker server). I remember we had a brief discussion about it and you were somewhat reluctant to accept `x-terminal-emulator`, can you recall the argument?

I've encountered small issue – if you close `xterm` window with a button then the script fails to kill simulator (both *OVPsim* and *qemu*) correctly.

rafalcieslak:
On Debian-based systems the only guarantee about a `x-terminal-emulator` is that it accepts `-e` flag. That's not enough. For example, on my desktop, the default `x-terminal-emulator` is `gnome-terminal` which exhibits very weird behavior; it doesn't stay open as we would like it to and terminates too early, sending some odd signals to everyone around (?), which messes with the script. In order to support `x-terminal-emulator` we'd either need to test the script with a multitude of terminals (won't happen) or we would need to impose only minimum requirements for the terminal - clearly for `gnome-terminal` passing `-e` command is not enough. If you really want to use `x-terminal-emulator` and hope for the best, please enable it in a separate branch, because I expect it to introduce a lot of issues depending on terminal choice.

cahirwpz:
Ok. Then we'll stick to `xterm`.

I've uninstalled `cgdb` because it's messing up terminal completely. However the script does not check which debuggers are available and crashes trying to choose the default, which should be one of `gdbtui` or `gdb`.

When I start debugging session with `gdbtui` by invoking `launch -d test=mutex`, gdb claims the program is not being run and an exception is thrown:
```
Traceback (most recent call last):
  File "./launch", line 108, in <module>
    dbg.stop()
  File "/home/cahir/workspace/mimiker-cieslak/launcher/debuggers.py", line 23, in stop
    Launchable.stop(self)
  File "/home/cahir/workspace/mimiker-cieslak/launcher/common.py", line 54, in stop
    self.process.send_signal(signal.SIGTERM)
  File "/usr/lib/python3.4/subprocess.py", line 1681, in send_signal
    os.kill(self.pid, sig)
ProcessLookupError: [Errno 3] No such process
```

rafalcieslak:
`cgdb` is generally very inconsiderate about terminal configuration. The launcher scripts has countermeasures in place that restore a sane config after closing/killing `cgdb`.

The script does check which debuggers are available, for proof see output of `./launch --help` which lists only available options. But as the documentation states, the `-d` flag is only a shorthand for `--debugger cgdb`. It is necessary for keeping the behavior of the script unchanged. Modifying the meaning of this flag might confuse other developers who are used to it.

rafalcieslak:
I see another potential issue. The port used by GDB is currently fixed as 1234. It should differ for each user. I can configure the launcher script to request simulators to use a different port, but GDB will use whatever is present in `.gdbinit`, which is set to 1234. In order to correctly assign different GDB port numbers for each user we'd need to dynamically generate `.gdbinit` before the debugger is started.

cahirwpz:
The default behaviour of the script should be to launch a debugger that is considered to be the  common denominator (i.e. if no `gdb` is installed other choices won't work anyway), please change `-d` to `--debugger gdb`.

cahirwpz:
@rafalcieslak Please fix bug mentioned in [this post](https://github.com/cahirwpz/mimiker/pull/301#issuecomment-300751598) and the PR will be ready to merge. In described scenario `launch` script leaves simulator running in the background, so that it must be killed by hand.

rafalcieslak:
I've fixed the bug where the launch script would fail to terminate subprocesses. 

---

However, I don't know how to solve the problem that debugging does not work with OVPsim and `xterm`. The reason for it is that xterm takes a very long time to start, because X is a synchronous protocol, so if your round-trip time to mimiker server is large, it adds up for every X request. Starting xterm may take many seconds, or even a minute or two! (Yeah, X-forwarding sucks.) OVPsim has a small (few seconds), unconfigurable timeout for the incoming connection, which we perform from within xterm. If xterm doesn't start in time, the connection arrives too late, and OVPsim gets super confused.

OTOH, qemu has no such limit, and thus debugging with qemu and xterm works completely fine for me. Same goes for OVPsim + `nc`/`socat`, as these open the connection immediately.

I can't come up with a simple solution to the OVPsim timeout! When starting `xterm` on mimiker server, we'd need first to (use a script to) open a connection and buffer it's output to some file (so that the connection is opened ASAP), then we would start xterm and wait for X-forwarding to initialize, then another script running within xterm would first open the file and dump its contents, then the first script would transfer the open socket to the second script, which would then continue the connection normally. But that's a lot of work (to workaround an OVPsim problem)!

I'd appreciate some ideas on how to deal with this issue.

*Yet another reason why I prefer to run OVPsim locally - no latency.*

cahirwpz:
> I can't come up with a simple solution to the OVPsim timeout!

There's one and very simple solution - just use a terminal that doesn't take ~8 seconds to startup over SSH. `urxvt` does it in a blink of an eye. I've added another output handler.

PR #302, opened 2017-05-08
Improved devfs_install

rafalcieslak:
This branch adds an optional `DEVFS_INSTALL_FLAG_NUMBERED` flag for `devfs_install`, which allows installing multiple devices sharing the same name into `/dev` - they will be automatically numbered e.g. `/dev/vga` (== `/dev/vga0`), `/dev/vga1`, `/dev/vga2`, ... This removes the silly assumption that only the first device of a particular driver may expose a `/dev` interface.

I've also fixed a potential race condition in `devfs_install`.

### DIFF
--- a/include/devfs.h
+++ b/include/devfs.h
@@ -5,7 +5,9 @@
 
 typedef struct vnode vnode_t;
 
+#define DEVFS_INSTALL_FLAG_NUMBERED 0x1
+
 /* Installs a new device into the devfs */
-int devfs_install(const char *name, vnode_t *device);
+int devfs_install(const char *name, vnode_t *device, unsigned flags);
 
 #endif /* !_SYS_DEVFS_H_ */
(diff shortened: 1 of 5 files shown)


cahirwpz:
I've got completely different view on device naming scheme. Name and unit number of a device, according to my understanding of *BSD device drivers, comes from [devclass](http://bxr.su/FreeBSD/sys/kern/subr_bus.c#86) structure. Each device is assigned `devclass` after it's created and `device.nameunit` is set to unique identifier (`rtc0`, `video1`, `tty7`, etc.) by calling [devclass_add_device](http://bxr.su/FreeBSD/sys/kern/subr_bus.c#1627).

cahirwpz:
Obsoleted by #344. With addition of device classes (incl. `device->d_nameunit` field) and rewrite of `devfs` using (yet to be written) pseudo filesystem framework, we won't need this patch.

PR #303, opened 2017-05-09
Implementation of dup and dup2 syscalls

MoloToha:
Since I'm working on pipes, I have to implement dup and dup2 syscalls as well to be able to replace standard input and output with pipe ends.


sys/vfs_syscalls.c
```diff
@@ -173,3 +200,21 @@ int sys_fstat(thread_t *td, syscall_args_t *args) {
     return error;
   return 0;
 }
+
+int sys_dup(thread_t *td, syscall_args_t *args) {
+  int old = args->args[0];
+  int *new = (int *)(uintptr_t)args->args[1];
```
cahirwpz: Do you really trust user-space program to pass correct pointer? ;-) Please use [fuword32 & suword32](http://mimiker.ii.uni.wroc.pl/source/xref/mimiker/include/systm.h#15) calls.

sys/filedesc.c
```diff
@@ -170,6 +170,23 @@ int fdtab_install_file(fdtab_t *fdt, file_t *f, int *fd) {
...
+
+  mtx_scoped_lock(&fdt->fdt_mtx);
+
+  if (fd < 0 || fd > fdt->fdt_nfiles)
+    return -EBADF;
+
+  if (fd_is_used(fdt, fd))
+    fd_free(fdt, fd);
+
+  fdt->fdt_files[fd] = f;
```
rafalcieslak: I think you also need to `fd_mark_used`. In the original `fdtab_install_file` this was done by `fd_alloc`, but since you don't use it here, you need to manually mark that the file descriptor is now used.

rafalcieslak:
Once you have syscalls working, a great way to test this feature in a user program would be to:

 - open some file with `open` as descriptor `x`
 - `dup` descriptor 1 (stdin) to  `y`
 - `dup2` `x` into 1 (stdin)
 - `scanf` some data and verify that it matches the contents of file
 - restore original behavior by `dup2`-ing `y` into 1.

That would both test the system calls, and prove that they are capable of realizing one of their main uses.

MoloToha:
I've discovered a very strange issue while working with user space test -- scanf can be used only once, after that it does nothing (I tested that without dups, so it shouldn't be a problem with them)

user/test_dup/test_dup.c
```diff
@@ -0,0 +1,15 @@
...
+#include <fcntl.h>
+
+int main() {
+  int x = open("/tests/dup_test_file", O_RDONLY, 0);
+  int y = dup(0);
+  printf("x = %d, y = %d\n", x, y);
+  dup2(x, 0);
+  char word[100];
+  scanf("%[^\n]s", word);
+  printf("%s\n", word);
```
rafalcieslak: I'd suggest to `assert(strcmp(word,"Hello, World!")==0)` so that we can verify the test without paying attention to the output.

cahirwpz:
> I've discovered a very strange issue while working with user space test -- scanf can be used only once, after that it does nothing (I tested that without dups, so it shouldn't be a problem with them)

@MoloToha Please create an issue in the issue tracking system with `bug` label. Describe the problem and show a program that triggers unwanted behaviour.

PR #304, opened 2017-05-11
Dealing with non-ascii output when running automatic tests

rafalcieslak:
If the kernel prints out any non-ascii character, the `run_tests.py` script isn't useful as it is unable to decode binary data. This branch proposes a fix: In case ascii decoding fails for any reason, the output is still dumped, encoded as `base64`, so that the developer may investigate what exactly is wrong with it.

I don't expect this feature to be useful when working on a local machine, but will help us troubleshoot problems on Travis. It should help @czapiga investigate the problems at #253.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -4,6 +4,7 @@ import argparse
 import pexpect
 import sys
 import random
+import base64
 
 N_SIMPLE = 5
 N_THOROUGH = 100
@@ -11,6 +12,19 @@ TIMEOUT = 5
 RETRIES_MAX = 5
 REPEAT = 5
 
+# Tries to decode binary output as ASCII. On failure, prints out the entire buffer in base64.
+def safe_decode(data):
+    try:
+        return data.decode("ascii")
+    except UnicodeDecodeError as e:
+        print("Failed to decode output. Below is the output in base64.")
+        print("==== BEGIN BASE64 ====")
+        print(base64.b64encode(data).decode("ascii"))
+        print("==== END BASE64 ====")
+        print("Failed to decode output. Binary data contains non-ascii character at position %d." % e.start)
+        print("The offending byte value is %d." % data[e.start])
+        print("Above is the entire buffer encoded in base64.")
+        sys.exit(1)
 
 # Tries to start gdb in order to investigate kernel state on deadlock.
 def gdb_inspect():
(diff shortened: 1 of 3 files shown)


run_tests.py
```diff
@@ -72,7 +86,7 @@ def test_seed(seed, sim='qemu', repeat=1, retry=0):
         test_seed(seed, repeat, retry + 1)
     elif index == 3:
         print("Timeout reached.\n")
-        message = child.buffer.decode("utf-8")
+        message = safe_decode(child.buffer)
```
czapiga: As far as I understand currently message is in utf-8  and after change it will be ascii. There is no problem with that? In some places ascii was used and here it was utf-8. Is there any reason for this misalignment?

rafalcieslak: No, I consider the fact that `"utf-8"` was used here a bug.

czapiga: That is why I asked. Thanks.

cahirwpz:
I guess it'd be much better if you could replace only non-ascii character by `'<%2x>' % ord(c)`, if that can be done easily. Right now, one offending character renders output totally unreadable without decoding base64 string in external tool.

rafalcieslak:
I don't know... I don't want this output to be readable. A non-ascii character is a highly unusual situation that signifies a major bug. Ideally we'd never want to use this feature. The purpose of base64 encoding isn't to "display" the output, it's to make sure we can easily download 1:1 exact binary data from Travis. Decoding the string is simple, `base64 -d > binary.file`, and will produce the exact unmodified output, to be examined with `hexedit` or any similar tool. If I substituted some bytes with "<%2x>", it would be difficult to grep-search for similar sequence in source/binary files. Also the values of these bytes are usually irrelevant in the output, seeing the `ord(c)` in output wouldn't help to understand the issue.

Note that this feature shouldn't be of much use when working on a local machine, where you can simply run `./launch` to get the exact output. This is why I don't write the data into a binary file - the premise of this branch is to make the data retrievable from Travis, where such file would be useless.

rafalcieslak:
#306 also seems to be affected by this issue.

laky55555:
In #306 Travis says " 'utf-8' codec can't decode". Should I try to use this run_tests.py on travis to get full output?

rafalcieslak:
Using your solution, some invalid characters seem to be skipped... So binary output from kernel may be unnoticed.

PR #309, opened 2017-05-13
Vfs lookup mount fix

coodie:
In current version of vfs_lookup there was a 'bug' or more likely a misunderstanding with how lookup is supposed to work. The bug is about performing vfs_lookup where the leaf (last directory on path) has something mounted under it, in this case we want to return the mounted child.

I have looked up at FreeBSD implementation to have better understanding and possibly ease implementation, but it is incredibly difficult to read because of all the cases it covers + path caching.

Here is vfs_lookup counterpart in FreeBSD:

[http://bxr.su/FreeBSD/sys/kern/vfs_lookup.c#124](http://bxr.su/FreeBSD/sys/kern/vfs_lookup.c#124)

Note that namei procedure calls lookup in few places, while lookup performs mounting fix in this line:

[http://bxr.su/FreeBSD/sys/kern/vfs_lookup.c#789](http://bxr.su/FreeBSD/sys/kern/vfs_lookup.c#789)

There is no fixing a mounted name before we enter the lookup procedure, but if you look at it's starting argument [http://bxr.su/FreeBSD/sys/sys/namei.h#72](http://bxr.su/FreeBSD/sys/sys/namei.h#72), it is assigned in [http://bxr.su/FreeBSD/sys/kern/vfs_lookup.c#514](http://bxr.su/FreeBSD/sys/kern/vfs_lookup.c#514), meaning that the vnode we pass to lookup is already the deepest mounted subchild. In our case we have only one lookup function getting full path, so we have to fix that assumption before main lookup loop.

Our implementation still lacks a lot of features and It doesn't have to perfect yet. Please don't waste too much time discussing on this simple change, #268 depends on this.

### DIFF
--- a/include/vnode.h
+++ b/include/vnode.h
@@ -54,6 +54,10 @@ typedef struct vnode {
   mtx_t v_mtx;
 } vnode_t;
 
+static inline bool is_mountpoint(vnode_t *v) {
+  return v->v_mountedhere != NULL;
+}
+
 #if !defined(IGNORE_NEWLIB_COMPATIBILITY)
 /* This must match newlib's implementation! */
 typedef struct vattr {
--- a/sys/vfs.c
+++ b/sys/vfs.c
@@ -160,23 +160,28 @@ int vfs_domount(vfsconf_t *vfc, vnode_t *v) {
   return 0;
 }
 
-static int vfs_assign_mounted_child(vnode_t **vnode) {
+/* If `*vp` is a mountpoint, then descend into the root of mounted filesys. */
+static int vfs_maybe_descend(vnode_t **vp) {
   vnode_t *v_mntpt;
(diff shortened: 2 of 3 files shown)


rafalcieslak:
I confirm this branch fixes an important problem. Back when I provided the original implementation of out `vfs_lookup` I didn't realize that e.g. request for `/dev` should return the root of `devfs`, not the `/dev` in initrd. Therefore I consider this fix valid.

coodie:
Travis build is failing, can you see why?

rafalcieslak:
It is clear to me that the timeout hit while kernel was still printing out something. The recently merged klog tests are VERY verbose, and printing output takes a lot of time... we should stop doing that, we really don't need 5k lines of output from tests.

PR #313, opened 2017-05-14
Bus methods to setup and teardown interrupts

cahirwpz:
This PR introduces `bus_intr_setup` and `bus_intr_teardown` functions and shows how they are implemented for root device. The root device is top-level device where autoconfiguration process starts. Each platform defines its own `rootdev`. In Malta platform case, `rootdev_attach` asks `gt_pci` to discover its children.

I'm still having some doubts if clock interrupt handler should be registered through `rootdev`. I don't like the fact that core kernel functionality (i.e. clock that drives scheduler) requires driver framework to be in place.

Next step would be to implement `intr_setup` and `intr_teardown` for `gt_pci`. I'll try to find out if I can quickly add proper `ns16550` driver for `cbus_uart`.

In subsequent PRs device classes and resource allocation mechanism need to be addressed.


rafalcieslak:
I have only reviewed the generic bus-interrupt management code and not the details of hardware interaction. These changes look good to me. The interface seems just about right, and I have no remarks.

cahirwpz:
Implemented and merged – an example of usage will follow.

Issue #294, opened 2017-05-05
sysinit


rafalcieslak:
We have a lot of components in `sys/*c.` that want to prepare something during kernel startup. We've gathered such initializer functions in `kernel_init(...)`, which contains a long list of various `init` functions. The order in which these functions are called is significant, for example `vga_init` assumes the `vfs_init` has already completed. Such dependencies make it difficult to manage the contents `kernel_init()`. We would like to have an uniform interface for registering such initializer functions and specifying their dependencies. A `*.c` submode that needs a function to be called at kernel init would contain a directive similar to this (example for `vfs`):

```c
SYSINIT_ADD("vfs", vfs_init, {"vnode", "ramdisk", NULL});
```

This macro would create an instance of `sysinit_entry_t` filling it with data, and would store a pointer that instance into a linker set of all sysinit entries. No `vfs`-related function calls would need to be added to `kernel_init`.

Then, during `kernel_init()` we would read all sysinit entries gathered by the linker set, perform [topological ordering](https://en.wikipedia.org/wiki/Topological_sorting), and call the requested functions one by one. The ordering will ensure that all dependencies between subsystems are met.

Implementing this interface would help us manage the startup procedure a lot. This mechanism would be mildly similar to freebsd's [sysinit](https://www.freebsd.org/cgi/man.cgi?query=SYSINIT&sektion=9).

This task doesn't require much research or interaction with hardware/platform, and I believe it should be fun. The only potentially complex part is implementing the ordering. If any of you guys is interested, go ahead and assign yourself.

cahirwpz:
Closed by #315

PR #315, opened 2017-05-16
sysinit

dyniec:
Prototype for sysinit #294. Order is built from the end.
It's definitely not ready to merge. At this moment it works in place, but requires O(n*m) (where n-number of modules, m- number of relations) time. Also it doesn't verify if there's a cycle. It will be skipped if occurs. Soon I will move tests to separate file, and integrate inits into this.
It turns out that array literals can't be passed directly to macros.


rafalcieslak:
>  It turns out that array literals can't be passed directly to macros.

Yeah, the preprocessor does not consider `{}` brackets as parenthesis, so by writing:

```c
SYSINIT_ADD(last, dummy, {"mid1", "mid2", NULL});
```
the preprocessor would think you're passing 5 arguments to the `SYSINIT_ADD` macro. The usual workaround is to use a helper macro like this:

```c
#define DEPS(...) (char*[]){__VA_ARGS__}
```
then you can write:
```c
SYSINIT_ADD(last, dummy, DEPS("mid1", "mid2", NULL));
```
and everything is fine!

You may also move the `(char*[])` cast to inside `SYSINIT_ADD` macro, which would turn the `DEPS` macro into a universal "array initializer" helper.

dyniec:
Turns out there's program `dot` in `graphviz` package which can easily create plot graphs.
It's much easier to comprehend them than grepping through sources.
a
compared to
```grep {mips,sys}/* -e 'SYSINIT_ADD'                   
mips/clock.c:SYSINIT_ADD(mips_clock, mips_clock_init, DEPS("callout", "sched"));
mips/pmap.c:SYSINIT_ADD(pmap, pmap_init, NODEPS);
sys/callout.c:SYSINIT_ADD(callout, callout_init, NODEPS);
sys/device.c:SYSINIT_ADD(driver, driver_init, NODEPS);
sys/file.c:SYSINIT_ADD(file, file_init, DEPS("vfs", "vnode"));
sys/filedesc.c:SYSINIT_ADD(filedesc, fd_init, DEPS("file"));
sys/initrd.c:SYSINIT_ADD(ramdisk, ramdisk_init, DEPS("vm_map", "vnode"));
sys/proc.c:SYSINIT_ADD(proc, proc_init, NODEPS);
sys/sched.c:SYSINIT_ADD(sched, sched_init, DEPS("callout"));
sys/vfs.c:SYSINIT_ADD(vfs, vfs_init, DEPS("vnode"));
sys/vfs_vnode.c:SYSINIT_ADD(vnode, vnode_init, NODEPS);
sys/vm_map.c:SYSINIT_ADD(vm_map, vm_map_init, DEPS("pmap", "vm_object"));
sys/vm_object.c:SYSINIT_ADD(vm_object, vm_object_init, DEPS("pmap"));

```
I've already created python script that generates graph description from sources. If somebody considers it usefull I can upload it.

rafalcieslak:
We need every piece of documentation we can get, so it would be awesome to have such script ready to use!

By the way, note that we prefer `/* C-style comments */` over `// C++-style comments`.

As a side note, thanks to your nice graph, I noticed that *maybe* we should move contents of `ramdisk_init` into `initrd_init`? This way ramdisk initialization would happen as a filesystem init function, as a part of `vfs_init` and thus ramdisk would not need a `sysinit` entry. (Although this change wouldn't need to happen as a part of this branch.) @coodie, you have a better idea on the order ramdisk initializes things, would that be a reasonable option?

cahirwpz:
IMO the pull request is ready to be merged. Please open another PR where you will add the script (`script/show-sysinit.py`) to produce the graph visualisation of system initialization process.

Issue #314, opened 2017-05-14
Less cumbersome interface for registering devices in devfs


cahirwpz:
I'm growing dissatisfied with how devices can interact with `devfs` subsystem. Here's a proposal for the new interface. Let's discuss it at the meeting this week.

```c
static int zero_read(vnode_t *t, uio_t *uio) {
  return uiomove_frombuf(zero_page, PAGESIZE, uio);
}
static int zero_write(vnode_t *t, uio_t *uio) {
  /* ??? */
}
static vnodeops_t zero_vops = {
  .v_read = zero_read,
  .v_write = zero_write
};
static devfs_entry_t zero_devfs = {DT_BLK, &zero_devops};

static vnodeops_t vga_fb_vops = { ... };
static vnodeops_t vga_palette_vops = { ... };
static vnodeops_t vga_videomode_vops = { ... };

static devfs_entry_t vga_devfs = {
  DT_DIR, (devfs_entry_t[4]){
    {"fb", DT_BLK, &vga_fb_vops},
    {"palette", DT_BLK, &vga_palette_vops},
    {"videomode", DT_BLK, &vga_videomode_vops},
    {0}
  }
};

void devfs_install(const char *name, devfs_entry_t *entry, device_t *dev);

vnode_t *null, *vga0;

void attach() {
  null = devfs_install("null", null_devfs, NULL);
  vga0 = devfs_install("vga0", vga_devfs, stdvga0);
}

void detach() {
  devfs_remove(null);
  devfs_remove(vga0);
}
```

cahirwpz:
Obsoleted by #344.

PR #316, opened 2017-05-16
getdirentries continuation

cahirwpz:
This PR is a follow up for #268. It's main purpose is to split the change into smaller pieces and merge them individually. In the process we will identify possible improvements in various kernel interfaces. One of them was proposed in issue #314.


coodie:
I have one idea concerning direntries, I suspect that in future we will introduce few small filesystems, and in order to make them complete and usable in user space, readdir will be necessary to work with all of them. I guess it would be good idea to implement 'default' readdir (like default open vnode_open_generic, for example). Readdir is pretty complicated to write and there are many details you have to know in order to write it correctly, so I think making default readdir will save a lot of time in future, even if it will be very inefficient for most of file systems.

dev vga might be a good start to see how to implement

cahirwpz:
I conform that initial value of memory pointed by `basep` is not relevant, at least for my installation of FreeBSD 11.0. I used `find_root` program to test it.  Moreover [getdirentries(2)](https://www.freebsd.org/cgi/man.cgi?query=getdirentries&sektion=2) manual page clearly states that (in comparison to Linux manual page):

> The `getdirentries()` system call writes the position of the block read into the location pointed to by `basep`.  Alternatively, the current position pointer may be set and retrieved by `lseek(2)`.	 The current position pointer should only be set to a value returned by `lseek(2)`, a value returned in the location pointed to by `basep` or zero.

The values written into `basep` under FreeBSD make quite a lot of sense – seems that they're an offset from the beginning of `dirent` array in kernel space.

coodie:
This means that in case of directory file descriptors we could store pointer to recently read directory entry. T think that's the purpose so that you can still let the user have that flexibility of reading from given point while maintaining ease of implementation, because now the offset represents anything you want (offset in the buffer, pointer to last read element on the list, etc.). 

While there is nothing wrong with my implementation, because it's up to me how I use the offset argument. I think more generalized assumption about the offset will make implementation of initrd readdir easier.

cahirwpz:
> [...] I think making default `readdir` will save a lot of time in future, even if it will be very inefficient for most of file systems.
> `/dev/vga` might be a good start to see how to implement first default `readdir` operation.

That's a great idea! If there's an easy way to achieve it, please do so.

cahirwpz:
Obsoleted by #375.

PR #305, opened 2017-05-11
Malloc multithreaded tests

czapiga:
I think I'm just missing rand() function

PR #319, opened 2017-05-17
Import pseudo-random number generator from *BSD

cahirwpz:
@czapiga Is that would you would like to have to finish #305 ?
@rafalcieslak I guess we can replace `rand` from `ktest.c` with this PRNG. Should I do it?


rafalcieslak:
Looks good. Go ahead.

PR #320, opened 2017-05-17
High resolution timer

cahirwpz:
To increase the resolution of kernel clock we should probably migrate from `realtime_t` to `timeval_t`. Firstly hardware clock in [mips/clock.c](https://github.com/cahirwpz/mimiker/blob/master/mips/clock.c) should be rewritten to maintain `timeval_t` time, then [sys/clock.c](https://github.com/cahirwpz/mimiker/blob/master/sys/clock.c) should follow. @czapiga After that's done you can continue with #253.

### DIFF
--- a/include/clock.h
+++ b/include/clock.h
@@ -1,9 +1,7 @@
 #ifndef __CLOCK_H__
 #define __CLOCK_H__
 
-#include <stdint.h>
-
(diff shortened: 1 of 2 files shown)


cahirwpz:
Moved to #327.

PR #324, opened 2017-05-18
Python-like with statement for mutex locking

cahirwpz:
During code reviews I have noticed that `mtx_scoped_lock` blends visually with the surrounding code too much. More important issue is that it does not force a programmer to think how long a mutex is really needed to be acquired, so the mutex may be held for code that does not require such assumption. To address the problem we could write following code with `mtx_scoped_lock`:

```c
{
  mtx_scoped_lock(a_lock);
  ...
}
```

... or use `WITH_MTX_LOCK`:

```c
WITH_MTX_LOCK (a_lock) {
  ...
}
```

Please give me your feedback on usefulness of this construct.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -45,7 +45,7 @@ Cpp11BracedListStyle: true
 DerivePointerAlignment: false
 DisableFormat:   false
 ExperimentalAutoDetectBinPacking: false
-ForEachMacros:   [ TAILQ_FOREACH, SPLAY_FOREACH, RB_FOREACH ]
+ForEachMacros:   [ TAILQ_FOREACH, SPLAY_FOREACH, RB_FOREACH, WITH_MTX_LOCK ]
 IncludeCategories: 
   - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
     Priority:        2
--- a/.clang-format
+++ b/.clang-format
@@ -45,7 +45,8 @@ Cpp11BracedListStyle: true
 DerivePointerAlignment: false
 DisableFormat:   false
 ExperimentalAutoDetectBinPacking: false
-ForEachMacros:   [ TAILQ_FOREACH, SPLAY_FOREACH, RB_FOREACH, WITH_MTX_LOCK ]
+ForEachMacros:   [ TAILQ_FOREACH, SPLAY_FOREACH, RB_FOREACH, WITH_MTX_LOCK,
+                   IN_CRITICAL_SECTION, WITH_RW_LOCK ]
 IncludeCategories: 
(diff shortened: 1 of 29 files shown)


rafalcieslak:
This is awesome! I wanted to come up with such a mechanism, but I couldn't figure out how to implement it in C. The `for` trick seems to work well, although it seems to slightly abuse the `scoped_lock` macro.

I suppose we could also use a similar macro for `critical_enter/leave`.

cahirwpz:
> [...] I couldn't figure out how to implement it in C. 

I knew it was possible for some time, but I needed to consult [P99](http://p99.gforge.inria.fr/) goodness to make it happen.

> I suppose we could also use a similar macro for critical_enter/leave

The trick could be applied in many other places. Critical section is a good candidate:

```c
/* excerpt from `sys/mutex.c` */
IN_CRITICAL_SECTION() {
  while (m->m_owner != NULL)
    sleepq_wait(&m->m_owner, "mutex");
  m->m_owner = thread_self();
}
```

cahirwpz:
@rafalcieslak I implemented `WITH_MTX_LOCK`, `WITH_RW_LOCK` and `IN_CRITICAL_SECTION`. Please have a look. Be prepared to see massive diffs due to indentation changes. The most suprising thing to see is that with-statements, thanks to loop semantics, render some `goto`'s obsolete.

rafalcieslak:
While I really like these macros and how they simplify things, I think you're using them a bit too much. It's great to wrap particular pieces of code, but using them to wrap entire functions kind of defeats the purpose of cleaning up code. We have many functions that look like this:

```c
void function(){
  mtx_scoped_lock(...);
  /* lots of code */
}
```

and that's entirely fine, there is no need to introduce an extra nested scope, the function itself already provides one. Using `WITH_` in this example makes things look a bit more complex than they actually are.

Also, could we get rid of `()` from `IN_CRITICAL_SECTION() {...}`?

*I can't wait to resolving all the merge conflicts this branch will cause due to indentation changes! :grimacing: That's unavoidable, though.*

cahirwpz:
> [...] I think you're using them a bit too much.

That allowed me to evaluate different use scenarios. You're right that in majority cases scoped lock is just fine. I reworked the PR to minimize introduced indentation changes. Please note that to mark important role of scoped lock in our code, I decided to change their names to upper case to make them visually stand out. Oh, and I introduced `SCOPED_CRITICAL_SECTION()`.

cahirwpz:
@rafalcieslak Should I merge so that you can use new constructs in #329 ?

rafalcieslak:
Go ahead. Pretty much every code we write uses some kind of synchronization mechanisms, so it'll be good to have these macros ready for use ASAP.

Issue #326, opened 2017-05-18
One user program for all user-space tests


rafalcieslak:
We currently have many user-space programs that perform various of syscall interface tests. To clean up the sources, we should merge them into a single program (although possibly multiple files), which launches the test specified as a program argument. This will also help with integrating these tests with `ktest` framework, we'll merge all `tests/exec_*` into a single file providing multiple tests via variable arguments to the user test program.

Issue #311, opened 2017-05-13
Introduce long test flag for ktest


coodie:
With recent klog test introduction, many tests on travis on other branches started to fail due to random timeouts on klog test, therefore I decided to remove it from regression.

There is also a case of ramdisk_dump test, and as far as I can recall there were some problems with timeouts in synchronization.

I think we shouldn't give up on those 'long tests' because they are actually quite useful, but having a lot of long repeated tests in kernel is terrible in my opinion because this will make regression so much longer. My idea is to add flag `KTEST_LONG` for this type of tests. Tests marked as `KTEST_LONG` won't be run repeatedly, but instead they will be run just once with extended timeout.

cahirwpz:
Until we provide a reasonable solution please mark long running tests with `KTEST_FLAG_BROKEN`.

cahirwpz:
@czapiga For the record, I've just disabled long running malloc tests.

PR #328, opened 2017-05-18
Implemented utest user program

rafalcieslak:
Per #326, I've gathered all usermode tests into `utest` program. It gets integrated with kernel tests via `tests/utest.c` file. This obsoletes `misbehave`, `fd_test`, `dup_test` and most of `prog` (which I intend to turn into a sandbox program for temporary tests and experiments).


rafalcieslak:
Interestingly, running these tests with other kernel tests unveiled some memory leaks! (*Yay, we're at the point where memory leaks are a thing!* :innocent:) It's probably no surprise that `vm_map` memory pool was too small as we don't clean up `vm_map_entries` thoroughly, but it's peculiar that `file` and `filedesc` memory pools also get exhausted after some user programs start. I'll investigate.

rafalcieslak:
I won't be able to get these new tests to run correctly before #329 (Implemented exit and waitpid) is done.

rafalcieslak:
Thanks to #329 (Implemented exit and waitpid), memory leaks are fixed and automatic tests should start working now. I needed to increase overall test timeout, but that's not because the tests are long [see #311 (Introduce long test flag for ktest)], but because I've added a large number of new tests.

With new tools in place, I was able to implement a very simple usermode test that dereferences a null pointer, and ktest ensures that it got terminated with a `SIGSEGV`. Cool!

rafalcieslak:
I don't see a point in investigating Travis errors before #334 (Disable pmap in klog mask by default) is accepted. In the last failed build, 20908 lines are pmap messages, out of 30952 total, which is more than 66%, and that certainly contributes to hitting the time limit.

rafalcieslak:
I don't suppose I need to prove the usefulness of these tests, but, for the record, they've just discovered a regression in `master` introduced by merging #335 (Simplify copying data between user-space and kernel-space), see `user_fork_wait` test. I'll investigate ASAP.

cahirwpz:
@rafalcieslak are there any chances you'll find some time to fix bugs and make the PR ready for review soon? It'd be really nice if we cleaned user-space tests run procedure.

rafalcieslak:
I'll see what I can do, but it may take some time - after all, I'm debugging tens of new tests that we did not run previously. These new tests heavily rely on proper synchronization, (much more than previous tests) which makes it very tricky to troubleshoot.

cahirwpz:
Ok, I understand. Note that it's all right if we merged new framework, and then enabled tests one by one.

rafalcieslak:
I believe I have identified the problem. It occurs when thread A executes `do_waitpid` while it's child (B) executes `proc_exit` (e.g. as a result of `sig_check`, as in this case), and a context switch happens in an unfortunate moment - namely at a point where A has locked B's mutex and needs to also lock it's parents (A's) mutex as well, while, at the same time, B has locked A's mutex and wishes to lock child's (B's) mutex - which causes a bi-directional deadlock.

For reproduction, example seed is `3769242755` with `test=all` and `repeat=5` for QEMU `2.5.0` at revision `9436642`.

A possible solution would be to have `child_find_by_state` not lock child's mutex - it's unnecessary, given it only peeks at child's state, which doesn't require atomicity.

rafalcieslak:
It seems everything is okay now, and this branch is ready for review!

In meantime, two new user programs were introduced: `test_rtc` and `test_kbd`. Since these are interactive program not written with automatic testing in mind, I am not able to integrate them with `utest`, so I'll leave them untouched.

Issue #291, opened 2017-05-02
Thread exit vs process exit


rafalcieslak:
There are two ways a user program can exit: either it may request for the current thread to finish, or it may wish to end the entire process. Exitting from the last thread of a process implies ending the entire process too. Our current `exit` only finishes the current thread, and at the moment there is no way to end a process. We'll need to investigate how this can be done, and implement related features.

Off the top of my head, linux has `exit` and `exit_group` syscalls, the first finishes a single thread while the other one finishes entire process, terminating all other threads as well.

rafalcieslak:
We currently work in a model where a process may only contain a single thread. That makes this issue irrelevant for now.

rafalcieslak:
Resolved by merging #329 (Implemented exit and waitpid).

PR #329, opened 2017-05-18
Implemented exit and waitpid

### LINKED ISSUE #326: One user program for all user-space tests

rafalcieslak:
Initially, I wanted to resolve #326 (One user program for all user-space tests). It turned out that running many user programs without restarting the kernel causes memory leaks, because we never reclaim resources allocated for a process. That's because out `exit` would only terminate a thread, and not the entire process. So to fix that, I needed to implement a proper `exit`, but in order to test it in any way I had to also implement `wait/waitpid` (Note: I recommend against splitting this PR into `exit` and `wait` PRs, these two both would need to introduce the logic of process' state after death!).

The result is quite attractive, as you can see in `test_fork.c`. My `waitpid` supports pid `-1` for waiting for any child, and `WNOHANG` option for returning immediately if no children are zombies.

Naturally, I've also verified that this fixes the memory leaks.

The implementation is straight-forward, maybe except for locks in `waitpid` - but that's only because I need to use both child's and parent's locks in slightly different patterns, which may *seem* ugly at the first sight.

Since threads no longer have a return value, the tests `tests/exec_*` are no longer able to verify the exit code of the userspace thread, but I suggest to fix that *after* #328 (Implemented utest user program) is done, as it does a major refactor to these tests anyway.

*EDIT: I think this also solves #291 (Thread exit vs process exit) for now.*


rafalcieslak:
I've updated this branch to support signals. Now killing a process will also clean up it's resources, and notify whoever `wait`s for it! I've also implemented `W*` macros that decode exit code / signal number from status returned by `wait`. Both `test_fork` and `prog` make use of `wait` now.

sys/sysent.c
```diff
@@ -258,6 +259,37 @@ static int sys_dup2(thread_t *td, syscall_args_t *args) {
   return do_dup2(td, old, new);
 }
 
+static int sys_wait(thread_t *td, syscall_args_t *args) {
```
cahirwpz: I reckon you can move it to user part of `sys/wait.h` space as:
```c
#define wait(status) waitpid(-1, status, 0);
```

sys/proc.c
```diff
@@ -53,4 +63,142 @@ proc_t *proc_find(pid_t pid) {
...
+  TAILQ_FOREACH (child, &p->p_children, p_child) {
+    mtx_lock(&child->p_lock);
+    if (child->p_state == PRS_ZOMBIE)
+      return child;
+    mtx_unlock(&child->p_lock);
+  }
+  return NULL;
+}
+
+int do_waitpid(pid_t pid, int *status, int options) {
```
cahirwpz: Indeed locking in this function looks awkward. Please have another look to see if I can simplify it, after `WITH_STMT` is merged.

sys/proc.c
```diff
@@ -53,4 +63,142 @@ proc_t *proc_find(pid_t pid) {
...
+  /* Child is now a zombie. Gather its data, cleanup & free. */
+  assert(mtx_owned(&child->p_lock));
+  *status = child->p_exitstatus;
+  int retval = child->p_pid;
+
+  mtx_lock(&zombie_proc_list_mtx);
+  TAILQ_REMOVE(&zombie_proc_list, child, p_zombie);
+  mtx_unlock(&zombie_proc_list_mtx);
+
+  kfree(M_PROC, child);
```
cahirwpz: Hmm... initially I thought there could be a situation where more than one process waits for another process to finish. But apparently you can only wait for direct descendants (i.e. children) processes. Of course the question reemerges in presence of multiple threads, but that's not a thing for now.

rafalcieslak: Yes, it is very important that you can only `wait` for your children. `wait`ing doesn't just represent "blocking until child exits", but also "cleaning up after child".

Supporting multiple threads is actually quite simple.

 - No need to worry about multiple threads in child process, as the parent is only bothered when the child *process* exits, not particular threads.
 - If `wait` gets called by multiple threads of the same parent process, the situation isn't difficult to handle, as the POSIX standard (2.13) states that (I've skipped unimportant details):

*When status information becomes available to the parent process and more than one thread in the parent process is waiting for the status information [...] the matching thread or threads that obtain the status information is unspecified except that at least one of the matching threads shall obtain the status information [...]*

sys/proc.c
rafalcieslak: I think #324 (Python-like with statement for mutex locking) wouldn't help much in this function - because I need to unlock the mutex before `sleepq_wait` and then re-lock it again. But I have another idea how I could improve these locks by using two wait channels instead of one... I'll commit it soon.

rafalcieslak:
It should be now apparent what is the purpose of each lock in `waitpid`.

sys/proc.c
```diff
@@ -53,4 +60,148 @@ proc_t *proc_find(pid_t pid) {
...
+    WITH_MTX_LOCK (&p->p_lock)
+      child = child_find_by_pid(p, pid);
+
+    /* No such process, or the process is not a child. */
+    if (child == NULL)
+      return -ECHILD;
+
+    for (;;) {
+      WITH_MTX_LOCK (&child->p_lock)
+        if (child->p_state != PRS_ZOMBIE)
```
cahirwpz: Should it not be `child->p_state == PRS_ZOMBIE` ?

rafalcieslak: You are correct. I've messed it up while refactoring.

sys/proc.c
```diff
@@ -53,4 +60,148 @@ proc_t *proc_find(pid_t pid) {
...
+  WITH_MTX_LOCK (&zombie_proc_list_mtx)
+    TAILQ_REMOVE(&zombie_proc_list, child, p_zombie);
+
+  kfree(M_PROC, child);
+
+  return retval;
+}
+
+int do_waitpid(pid_t pid, int *status, int options) {
+  /* We don't have a concept of process groups yet. */
```
cahirwpz: Please create an issue and explain briefly what should be done.

sys/proc.c
```diff
@@ -53,4 +60,148 @@ proc_t *proc_find(pid_t pid) {
...
+
+  WITH_MTX_LOCK (&p->p_lock) {
+
+    assert(p->p_nthreads == 1);
+    /* If the process had other threads, we'd need to wake the sleeping ones,
+       request all of them except this one to call thread_exit from
+       exc_before_leave (using a TDF_? flag), join all of them to wait until
+       they terminate. */
+
+    /* XXX: If this process has any unwaited zombie children, assign them for
```
cahirwpz: Should not children get adopted by `init` process, without looking at `p_state` field, after process calls `exit`?

sys/proc.c
```diff
@@ -53,4 +60,148 @@ proc_t *proc_find(pid_t pid) {
...
+
+    /* Clean up process resources. */
+    vm_map_delete(p->p_uspace);
+    fdtab_release(p->p_fdtable);
+
+    /* Record some process statistics that will stay maintained in zombie
+       state. */
+    p->p_exitstatus = exitstatus;
+
+    /* Turn the process into a zombie. */
```
cahirwpz: Should not the process be turned into zombie first, and then its resources to be cleaned up and parent notified?

rafalcieslak: I don't understand. The exact order of actions in this function does not matter much, as it all happens with `p->p_lock` locked.

rafalcieslak:
Now that signal support was merged, I can add `SIGCHLD` where applicable. But... remember when I mentioned signal code is *a ton* of special cases and conditions? Well, here's one: `SIGCHLD` is ignored by default. But child process gets reaped immediately (without turning into a zombie) only if the parent **explicitly** ignores `SIGCHLD`, not by default.

So, in a way, the logic of how to perform a process exit will depend on how the parent's signal handlers are set up. Smells weird, but I suppose that's how it must be done.

sys/proc.c
cahirwpz: Logically marking a process zombie should happen first, but as you correctly noticed it does not matter. It's ok as is.

sys/proc.c
rafalcieslak: Correct. Also, until we use an init process, I need to clear the `p_parent` of all children so that they can no longer refer to their original, terminated parent - I've addressed this in a recent commit.

cahirwpz:
W.r.t. `SIGCHLD` please leave an information that we don't handle it properly. Preferably, please open an issue explaining what's missing.

rafalcieslak:
Oh... I just implemented proper support for `SIGCHLD`, and several tests in `test_fork` to prove it makes sense and fulfills the required interface. I've also added some changes that make `exit` behave correctly when the calling process was orphaned.

sys/proc.c
```diff
@@ -45,3 +54,170 @@ proc_t *proc_find(pid_t pid) {
...
+    if (p->p_parent) {
+      WITH_MTX_LOCK (&p->p_parent->p_lock) {
+        sleepq_broadcast(&p->p_parent->p_children);
+        sleepq_broadcast(&p->p_state);
+
+        /* If the parent explicitly ignores SIGCHLD, reap child immediately. */
+        if (p->p_sigactions[SIGCHLD].sa_handler == SIG_IGN) {
+          int ignore_status;
+          proc_reap(p, &ignore_status);
+          thread_exit();
```
cahirwpz: I'm afraid `noreturn` function called from within `WITH_STMT` will prevent from the lock being released :(

cahirwpz:
LGTM When final fixes are done, please merge.

PR #330, opened 2017-05-19
Simplified mutex and tailq initialization

rafalcieslak:
I was in a mood for cleanup today. I've always wondered why:

1. We rarely use `TAILQ_HEAD_INITIALIZER` macro
2. We have not implemented a `MUTEX_INITIALIZER` macro

These are very useful for initializing static/global variables. I've applied them consistently where applicable, and that rendered as much as three `init_` functions useless (!), so I've removed them entirely.

### DIFF
--- a/include/interrupt.h
+++ b/include/interrupt.h
@@ -41,9 +41,6 @@ typedef struct intr_chain {
   unsigned ic_irq;                 /* physical interrupt request line number */
 } intr_chain_t;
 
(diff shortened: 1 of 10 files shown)


czapiga:
Ship it

czapiga:
Whoops

cahirwpz:
Cleanup changes are welcome! Thanks!

Issue #327, opened 2017-05-18
High resolution timer


cahirwpz:
To increase the resolution of kernel clock we should probably migrate from `realtime_t` to `timeval_t`. Firstly hardware clock in [mips/clock.c](https://github.com/cahirwpz/mimiker/blob/master/mips/clock.c) should be rewritten to maintain `timeval_t` time, then [sys/clock.c](https://github.com/cahirwpz/mimiker/blob/master/sys/clock.c) should follow. Some preliminary work has been done in [cputimer](https://github.com/cahirwpz/mimiker/tree/cputimer) branch.

@czapiga After that's done you can continue with #253 and implement `gettimeofday` syscall.

For other time related syscalls we will need something more sophisticated. During the meeting we discussed one shot cpu high resolution (1µs) timer with following interface:

```c
typedef struct timer_event {
  TAILQ_ENTRY(timer_event) tev_link;     /* entry on list of all time events */
  timeval_t tev_alarm;                   /* absolute time when this event should be handled */
  void (*tev_func)(timer_event_t *self); /* callback function called in interrupt context */
  void *tev_data;                        /* auxiliary data */
} timer_event_t;

/* call `tev` callback after `tvdiff` time has passed */
int cpu_timer_add_event(timer_event_t *tev, timeval_t *tvdiff);
int cpu_timer_remove_event(timer_event_t *tev);

/* defined in `mips/clock.c` */
static TAILQ_HEAD(, time_event) cpu_timer_events;
```

Some remark about the way `cpu_timer` should work like:
* `cpu_timer_add_event` may work within critical section in order to insert an event into the `cpu_timer_events`,
* for each element of `cpu_timer_events` with `tev_alarm` lower than current value, `mips_clock_irq_handler` will remove an event from the list and call a callback with pointer to the event.

PR #331, opened 2017-05-20
High resolution timer

czapiga:
In this PR I will work on #327 issue. mips_clock_us will be incremented every mircosecond but for now I just wanted to check if changing types won't break anything.


czapiga:
@cahirwpz Is this how I should actually change timer resolution? It seems like after going down to 1us tests are not even executed. I have to debug why kernel freezes but first I'd like to be sure I'm not doing something stupid with mips32_set_c0

czapiga:
I noticed that I can go down with clock to ~20us on OVPsim and ~40us on qemu. Setting it to lower value makes kernel freeze in few different places

czapiga:
It was called because of  `if (ms != _ms)` in `sys/clock.c`

mips/clock.c
```diff
@@ -7,20 +7,19 @@
 #include <sched.h>
 #include <sysinit.h>
 
-/* This counter is incremented every millisecond. */
-static volatile realtime_t mips_clock_ms;
+/* System clock gets incremented every milisecond. */
+static timeval_t *sys_clock = &TIMEVAL_INIT(0, 0);
```
czapiga: How is this working? You take address of temporary object

cahirwpz:
Well... we don't want the timer to cause [interrupt storm](https://en.wikipedia.org/wiki/Interrupt_storm)!!! I wanted `clock_get` to return exact time (based on `C0_COUNT`) without making any impact on the scheduler and callouts.

mips/clock.c
cahirwpz: This is a part of C99 standard and is called *designated initializers*. It's not a *temporary*, but an *anonymous* value. On the top level it gets allocated in `.data` section, inside function it gets allocated on the stack.

czapiga:
Ok, I think I get it now..

rafalcieslak:
@czapiga Side note: ATM you are the only one in the team who uses the `not ready` label to mark unfinished PRs, others use `WIP` for that. cahirwpz, me and coodie have been using  `not ready` to mark tasks and PRs that require work in other areas, extra discussion, or additional research before work on such task can even begin.

Based on what we typically use labels for, I've created [this table](https://github.com/cahirwpz/mimiker/wiki/Github-issue-PR-labels) some time ago - it may serve as a nice overview of what do we usually mean with each label. I'm not forcing you to strictly follow these descriptions, though. It is not an officially accepted list. We may want to discuss it in future to reduce confusion and hopefully improve our labeling consistency.

cahirwpz:
@czapiga What is the progress? I'd like this task to be finished sooner than later. It will unblock #253 and enable us to add `gettimeofday` syscall.

czapiga:
I will try to end this till Thursday.

czapiga:
If it's possible I would like this to be merged then I will finish #253 and after that I will go back to `timer_event_t` interface.

PR #337, opened 2017-05-20
readdir operation for initrd with generic readdir algorithm

cahirwpz:
This change extracts code from #316 required to implement `readdir` operation for `initrd` filesystem. Preliminary work has been done by @coodie. I've only abstracted away `readdir` routine to be as independant from the filesystem as possible. Hopefully `readdir_generic` will be reused by other filesystems, in particular *pseudofs* framework (yet to be implemented) used to create `devfs`, `stdvgafs` and other device filesystems.


coodie:
Here is a list of things you have done wrong:

1. **Miscommunication**
I don't like the fact that you took my task without mentioning to me you were going to do this. Not asking, not trying to communicate it to me anyhow, just sat down and did my job. I mentioned on seminar I wasn't going to spend much time on mimiker this weekend.

2. **Taking credit for somebody's else work**
Thank you for calling over 30 hours spent working on this task 'preliminary work', I've put a lot of effort to avoid unnecessary discussions but It doesn't matter how hard I try, you never notice this. This task is simple enough and It would feel great if I could take full credit for it. 

3. **Putting effort in wrong places**
You mention quite often that you have no time to review my changes, but you have time to do my job. You are the person to teach me and other students, my changes are still unreviewed and I know they will never and ever be.  You - as an educator, leader and experienced programmer - should focus on other, possibly more difficult or vital tasks of the kernel, which no students want to do or are too difficult.

4. **No educational value**
 Remember that this is educational project not a product we will ever be able to sell or anyone ever will use for special purpose, the goal is to learn and have fun. I don't get to learn anything from that. This task is easy and should be given to student to implement, make a review.

5. **No satisfaction and fun**
This project takes a lot of time, and main benefits I can gain from this are fun and satisfaction (when will I have opportunity to work on OS in my life?), let me have those.

6. **No respect for me and my work**
You haven't noticed the work I've put into this, you have scattered entire readdir branch into few other smaller subbranches, **haven't done your job which is give feedback and make review**. Action like this carries message **nice idea, but you are a terrible programmer and I'd better do it myself**. Plus, this really rarely happens to other students, am I worse or what? Sure, I could be better and I agree I'm not a very good programmer, but this is a matter of respect and teacher-student relationship. If student doesn't get feedback and practice how can he learn?

cahirwpz:
Thank you for the effort you spend to write words of constructive criticism.

1. As we know, using GitHub to communicate has its own limitations. To make things worse, there's not enough time during weekly meeting to discuss everything, and that is why I decided to devote some extra office hours for discussions. We have no luxury to work together every day, do pair programming, talk about issues, etc. The communication is an issue and it will be – it's high time to learn to live with it. There's good will on my side to push the project forward... please do not assume I do anything wrong on purpose.

2. Formally speaking the code does not belong to me or you. Its rightful owner is University of Wroclaw. I'm trying to grade everybody's work in a fair manner – you're getting the credit for your work according to the rules I apply to every project's member. 

3. Unfortunately, I've learnt (in this project) that it frequently takes less time to rewrite sb's code than to explain how to rewrite it. During such refactoring phases I often identify underlying problems – among them misdesigned pieces of the architecture and missing abstractions & tools (in this case - pseudofs framework, apparently). If a problem seems superficial to you, it doesn't need to be - patience is a virtue. 

4. You preferred to choose an easy task, because you have a full-time job. We know that one learns the most from challenges... however if you're persistent you can find challenges even in seemingly dull tasks. As for the project, in long run it may serve as a research platform – for now nobody knows its true purpose.

5. When I was on internal interview by Google kernel team, they told me that it's not uncommon for one not to push a line of code into the kernel for a month. I rejected the offer... and now I know what they were talking about. Developing an OS kernel is not a fast paced project.

6. One of the rules I set up at the beginning of the project was (roughly): "Only elegant and clean enough solutions will be integrated into the kernel source base."
If I see value in a piece of code that cannot be immediately merged, my task is to adapt it appropriately – either by leaving remarks during code reviews or doing changes by hand. Heavy code rewrites happens to other people too – it's not uncommon, unfortunately. Nice ideas are great, but it takes some effort and talent to convey them.
We've been working on the kernel for about a year. Throughout this time you've received a lot of feedback from me, you've seen other people's code that eventually got merged. I encouraged you to read books explaining good practice (The Pragmatic Programmer, The Art of Unix Programming, etc.). The outcome of my efforts is not overwhelming, that's a pity.
Also... please note that I asked you to split the PR into smaller changes and you rejected my kind request.

*Finally... I'd much prefer if we could focus on getting your ideas hammered out into good code. Getting emotional will get us nowhere. Let's be professional.*

include/dirent.h
```diff
@@ -51,7 +53,7 @@ typedef struct dirent {
   uint16_t d_reclen; /* length of this record */
   uint16_t d_namlen; /* length of string in d_name */
   uint8_t d_type;    /* file type, see below */
-  char d_name[0];
+  char d_name[MAXNAMLEN + 1];
```
cahirwpz: It does not matter, as we use `dirent_t` not to store data, but to exchange it between kernel-space and user-space. However this is a leftover of my other idea, so I'll remove it.

include/cpio.h
```diff
@@ -44,15 +44,17 @@
...
-#define C_IFMT 0170000  /* type of file */
+#define C_FIFO 1  /* FIFO */
+#define C_CHR 2   /* Character special file */
+#define C_DIR 4   /* Directory */
+#define C_BLK 6   /* Block special file */
+#define C_REG 8   /* Regular file */
+#define C_CTG 9   /* Reserved for contiguous files */
+#define C_LNK 10  /* Reserved for symbolic links */
+#define C_SOCK 12 /* Reserved for sockets */
+
```
cahirwpz: Good point – indeed, we can.

include/cpio.h
```diff
@@ -44,15 +44,17 @@
...
+/* cpio mode to file type */
+#define CMTOFT(mode) (((mode)&0170000) >> 12)
```
cahirwpz: We don't need to expose `include/cpio.h` to the user-space, so we can modify it as we wish.

sys/vfs_readdir.c
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  for (; it; it = ops->next(it)) {
+    unsigned namlen = ops->namlen_of(it);
+    unsigned reclen = _DIRENT_RECLEN(dir, namlen);
+
+    if (uio->uio_resid < reclen)
+      break;
+
+    dir = kmalloc(M_TEMP, reclen, M_ZERO);
```
cahirwpz: I think we've discussed it during a project meeting. We're not allowed to allocate arrays and large structures on stack due to its size limitation – only 4KiB. *BSD*'s use `M_TEMP`, so do we.

include/vfs.h
```diff
@@ -19,4 +20,13 @@ int do_dup2(thread_t *td, int old, int new);
 int do_mount(thread_t *td, const char *fs, const char *path);
 int do_getdirentries(thread_t *td, int fd, uio_t *uio, off_t *basep);
 
+typedef struct readdir_ops {
+  void *(*first)(vnode_t *v);
+  void *(*next)(void *entry);
+  unsigned (*namlen_of)(void *entry);
+  void (*convert)(void *entry, dirent_t *dir);
```
cahirwpz: My assumption was that `readdir_ops_t` functions should not be concerned with memory allocation. It's up to the algorithm to manage the memory. This will simplify things if we ever keep entries in some kind of cache, if that's the way to go...

I initially thought about replacing `namlen_of` with `reclen_of`, but then (in case of `initrd`) I'd need to run `strlen()` twice.

sys/vfs_readdir.c
```diff
@@ -0,0 +1,41 @@
...
+
+  /* Locate proper directory based on offset */
+  for (; it; it = ops->next(it)) {
+    unsigned reclen = _DIRENT_RECLEN(dir, ops->namlen_of(it));
+    if (offset + reclen > uio->uio_offset) {
+      assert(it == NULL || offset == uio->uio_offset);
+      break;
+    }
+    offset += reclen;
+  }
```
cahirwpz: [VOP_READDIR(9)](http://netbsd.gw.com/cgi-bin/man-cgi?VOP_READDIR) states following:
 
>The directory contents are read into `struct dirent` structures and `uio->uio_offset` is set to the offset of the next unread directory entry.  This offset may be used in a following invocation to continue a sequential read of the directory contents.

In all contexts I know `uio_offset` value is expressed in bytes. I would prefer not change semantics of this field to anything else, just to avoid any suprises for future readers. Moreover value returned in `basep` is allowed to be passed to `lseek`, which again accepts offset in bytes, not records.

If I were to optimize directory traversing I'd probably keep `void *entry` somewhere in `file_t` and extend `VOP_READDIR` with an extra argument to pass the state of `readdir_generic`, so consecutive syscalls reuse its value. In fact there's some provision in NetBSD interface, which seems to enable some optimizations, but they do it somewhat differently:

```c
int VOP_READDIR(struct vnode *vp, struct uio *uio, kauth_cred_t cred, int *eofflag,
                off_t **cookies, int *ncookies);
```

> In case `ncookies` and `cookies` are supplied, one cookie should be returned per directory entry.  The value of the cookie for each directory entry should be the offset within the directory where the on-disk version of the following directory entry starts. That is, for each directory entry i, the corresponding cookie should refer to the offset of directory entry i + 1.

cahirwpz:
> One very non-obvious thing is that we are going to need to return `.` and `..` entries in generic readdir, I can't find any documentation on this but this is returned by `getdirentries` on linux so I assume this is necessary, they even have different inode numbers for different directories. 

Indeed I completely omitted that problem. I'm not going to add options to `readdir_generic` instead of it I'll normalize ramdisk contents so that it matches common expectations (i.e. lack of files representing `.` and `..`).

sys/vfs_readdir.c
coodie:
```
     The getdirentries() system	call writes the	position of the	block read
     into the location pointed to by basep.  Alternatively, the	current	posi-
     tion pointer may be set and retrieved by lseek(2).	 The current position
     pointer should only be set	to a value returned by lseek(2), a value
     returned in the location pointed to by basep (getdirentries() only) or
     zero.
```

From my point of view this means we can use this argument the way we want. I was also confused because basep value on my linux machine has big integers looking to be quite random numbers, so I assumed they use it to point to data structures.

I've just read the `tmpfs_readdir` code in NetBSD (FreeBSD version is terrible to read) and it seems to me that they use uio_offset to store a sequence number, and based on that find directory entry. 
[tmpfs_readdir](http://bxr.su/NetBSD/sys/fs/tmpfs/tmpfs_subr.c#807)

They keep last read directory entry in cache:
[tn_readdir_lastp](http://bxr.su/NetBSD/sys/fs/tmpfs/tmpfs.h#145)

We could introduce caching to our interface somehow.

I'm surprised that NetBSD keeps cache inside tmpfs structure instead of file_t, I can't come up with reason they do that.

sys/vfs_readdir.c
coodie: [FreeBSD allocates dirent_t on stack](http://bxr.su/FreeBSD/sys/fs/tmpfs/tmpfs_subr.c#1193), but now there is no point in doing this since we have zero length array at the end of dirent.

cahirwpz: Interesting... but on the other hand, their code base is so huge that they have even more problem at forcing some policies. Nonetheless, we should not forget to use `M_TEMP`.

cahirwpz:
> How does enumeration of inodes relate to `generic_readdir` ?

I have added a `c_parent` field to `cpio_node_t` as well. FTM their usefulness may not be apparent. I'm still trying to add `.` and `..` in a graceful manner. When I'm done with modifications I'll remove `WIP` branch so we can go back to the review process.

cahirwpz:
Blocked by #348.

cahirwpz:
Please continue the review. I added `.` and `..` handling and simplified the interface a bit.

I skipped the problem of using `seek` on directories, and did not use `file->f_data` for optimization. When the PR is merged I'll add an issue describing features that one could add to current implementation.

cahirwpz:
Please review so that we can return to #316.

cahirwpz:
I'm going to merge this branch today's evening. If you have any remarks or questions please say 
/ ask them now.

PR #339, opened 2017-05-23
Modified klog test to comply with implementation, fix bug with costum…

laky55555:
Resolving bugs in klog test introduced with changes in klog implementation.
Fixing bug in definition of klog with custom mask.


laky55555:
Maybe we should introduce change mask in scope for klog, in the same 
way/reason as with mutexes.

cahirwpz:
> Maybe we should introduce change mask in scope for klog, in the same way/reason as with mutexes.

Please feel free to add it to this PR.

PR #340, opened 2017-05-23
Added kdump klog functionality, fixed bug in segment ptable

laky55555:
Added new printing in gdb for klog.
Currently everything in klog is printed.
Maybe kdump klog should also implement some kind flags,
-v for modifying what we want to print on the screen.
[Usage] kdump klog


cahirwpz:
FYI `kl_timestamp` type was changed to `timeval_t`. Please update you code to reflect that change.

debug/physmem.py
```diff
@@ -67,18 +67,21 @@ def get_all_segments(self):
...
         pages = tailq.collect_values(freeq, 'freeq')
-        return [[idx, size, as_hex(page['paddr']), as_hex(page['vaddr'])]
-                for page in pages]
+        return [[str(idx), str(size), as_hex(page['paddr']),
+                 as_hex(page['vaddr'])] for page in pages]
 
     def dump_segment_free_pages(self, idx, segment):
-        return [self.dump_segment_freeq(idx, segment['freeq'][q], 4 << q)
-                for q in range(16)]
+        helper = []
```
laky55555: A bit ugly fix but fix with least changes to original code.
Problem was because when function self.dump_segment_freeq is called with for loop it make one more nested array

debug/physmem.py
```diff
@@ -51,8 +51,8 @@ def dump_segments(self):
         segments = self.get_all_segments()
         rows = [['Segment', 'Start', 'End', 'Pages no']]
         for idx, seg in enumerate(segments):
-            rows.append([idx, as_hex(seg['start']), as_hex(seg['end']),
-                         seg['npages']])
+            rows.append([str(idx), as_hex(seg['start']), as_hex(seg['end']),
```
laky55555: ptable that we use for output in table format can't work with numbers

debug/physmem.py
```diff
@@ -121,3 +124,61 @@ def dump_tlb(self):
...
+    def invoke(self):
+        messages, klog = self.load_klog()
+        self.dump_general_info(len(messages), klog)
+        self.dump_kernel_logs(messages)
+
+    def format_message(self, data):
+        message = data['kl_format'].string()
+        arguments = data['kl_params']
+        # If there is % escaped it is not parameter.
+        number_of_parameters = message.count('%') - 2 * message.count('%%')
```
laky55555: For using function gdb.execute print we need to give it correct number of parameters (line 147)
and I use it for so I don't need to check if parameter is address or value

laky55555:
[NOTE] After mips_timer module is initialized timer starts from beginning again.

debug/physmem.py
cahirwpz: You could make use of [itertools.chain](https://docs.python.org/2/library/itertools.html#itertools.chain) here.

debug/physmem.py
```diff
@@ -121,3 +124,61 @@ def dump_tlb(self):
...
+        params = ', '.join(str(arguments[i])
+                           for i in range(number_of_parameters))
+        try:
+            message = message.replace('"', '\\"')
+            # TODO: why do we need %zu???
```
cahirwpz: We should probably get rid of it and use `%lu` instead.

debug/physmem.py
```diff
@@ -121,3 +124,61 @@ def dump_tlb(self):
...
+            message = message.replace('%zu', '%u')
+            # Using gdb printf so we don't need to dereference addresses.
+            formated = gdb.execute(
+                'printf "' + message + ' ", ' + params, to_string=True)
+        except Exception as e:
+            print('''Error in formating message "{}" with parameters "{}"\n
+            Message skipped!!'''.format(message, params))
+            formated = message + params
+        time = str(data['kl_timestamp']['tv_sec'] * 1e6 +
```
cahirwpz: Please use `"%d.%06d" % (tv_sec, tv_usec)` string to format time.

debug/physmem.py
```diff
@@ -121,3 +124,61 @@ def dump_tlb(self):
...
+    def dump_general_info(self, number_of_logs, klog):
+        rows_general = [["Mask", "Verbose", "Log number"]]
+        mask = as_hex(klog['mask'])
+        verbose = bool(klog['verbose'])
+        rows_general.append([mask, str(verbose), str(number_of_logs)])
+        ptable(rows_general, header=True, fmt='l')
+
+    def dump_kernel_logs(self, messages):
+        rows_data = [["Time", "Line", "File", "Origin", "Message"]]
+        for message in messages:
```
cahirwpz: `rows_data.extend(messages)` ?

debug/README.md
```diff
@@ -0,0 +1,25 @@
...
+After you start kernel in debugging mode, launch script with -d flag
+[(detailed how to)](https://github.com/cahirwpz/mimiker#running) you can use follwing custom commands.
+
+* `kdump` - Print on screen current status of kernel.
+* `ktrace` - Add/remove breakpoint on selected function.
+
+kdump aguments prints:
+
+* `free_pages` - list of free pages per segment with virtual and physical address
+* `segments` -
```
cahirwpz: Please describe `segments` and `tlb`.

debug/physmem.py
laky55555: I'm not sure how to incorporate chain() to make it work with ptable. If you think it can be done feel free to change it.

Issue #342, opened 2017-05-24
Adding -Wextra  and possibly other compilation flags


czapiga:
According to my observations, because we don't use `-Wextra`, following warnings are not reported:

* `-Wmissing-field-initializers`
* `-Wsign-compare`
* `-Wtype-limits`
* `-Wstrict-prototypes` (*new!*)
* `-Wunused-parameter`

I'm not sure we should care about the last one, but others should be fixed.

PR #343, opened 2017-05-24
Adding -Wextra and -Wstrict-prototypes compile flags

czapiga:
PR for #342

### DIFF
--- a/mips/intr.c
+++ b/mips/intr.c
@@ -27,7 +27,7 @@ static intr_chain_t mips_intr_chain[8] = {
     [4] = MIPS_INTR_CHAIN(4, "hwint(2)"), [5] = MIPS_INTR_CHAIN(5, "hwint(3)"),
     [6] = MIPS_INTR_CHAIN(6, "hwint(4)"), [7] = MIPS_INTR_CHAIN(7, "hwint(5)")};
 
-void mips_intr_init() {
+void mips_intr_init(void) {
   /*
    * Enable Vectored Interrupt Mode as described in „MIPS32® 24KETM Processor
    * Core Family Software User’s Manual”, chapter 6.3.1.2.
--- a/mips/intr.c
+++ b/mips/intr.c
@@ -138,6 +138,7 @@ finalize:
 }
 
 void fpe_handler(exc_frame_t *frame) {
+  (void)frame;
   thread_t *td = thread_self();
   if (td->td_proc) {
     sig_send(td->td_proc, SIGFPE);
--- a/Makefile.common
+++ b/Makefile.common
@@ -41,7 +41,7 @@ GCC_SYSTEM_INC = -I$(GCC_INSTALL_PATH)include/ \
                  -D_LIBC_LIMITS_H_
 
 ASFLAGS  =
-CFLAGS   = -std=gnu11 -Og -Wall -Wextra -Werror -fno-builtin -nostdinc -nostdlib \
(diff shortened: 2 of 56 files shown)


rafalcieslak:
I suppose it might be better to use [`__unused`](https://github.com/cahirwpz/mimiker/blob/c1b5d2a51f0b9832f4873a35ca87f88dbe15eabe/include/common.h#L35) where applicable instead of `(void)x`.

czapiga:
It's like function signature with clean list of arguments in one line vs function signature with few lines of params and their attributes. I would like to hear a good argument before I do this work one more time (it melts my brain).

mips/intr.c
```diff
@@ -138,6 +138,7 @@ void syscall_handler(exc_frame_t *frame) {
 }
 
 void fpe_handler(exc_frame_t *frame) {
+  (void)frame;
```
coodie: I'm not quite sure, but in case of unused parameters it should be enough to change function header as such:

```c
 void fpe_handler(exc_frame_t *) {
```

rafalcieslak: I thought this is not valid in C, but only in C++?

czapiga:
Well, I have bigger problem now after changing `critical_enter()` to `critical_enter(void)` because as far as I understand `SCOPED_STMT` and `WITH_STMT` is passing `NULL` to that function(and to `void critical_leave(void);`):

> In file included from /home/czapiga/mimiker/include/klog.h:4:0,
                 from intr.c:2:
intr.c: In function 'mips_intr_setup':
/home/czapiga/mimiker/include/sync.h:22:42: error: too many arguments to function 'critical_enter'
 #define CRITICAL_SECTION WITH_STMT(void, critical_enter, critical_leave, NULL)
                                          ^
/home/czapiga/mimiker/include/common.h:116:5: note: in definition of macro 'SCOPED_STMT'
     ACQUIRE(VAL, ##__VA_ARGS__);                                               \
     ^
/home/czapiga/mimiker/include/sync.h:22:26: note: in expansion of macro 'WITH_STMT'
 #define CRITICAL_SECTION WITH_STMT(void, critical_enter, critical_leave, NULL)
                          ^
intr.c:47:3: note: in expansion of macro 'CRITICAL_SECTION'
   CRITICAL_SECTION {


Do you know how to fix that? I was trying to create `SCOPED_STMT0` and `WITH_STMT0` with now `VAL, ##__VA_ARGS__` but I failed because `RELEASE` was still getting some input param

cahirwpz:
Hmm... I think you're trying to do too much for one go. Can we start with adding only `-Wstrict-prototypes` and then with consecutive PRs add one warning option (and fix all offending code) at a time?

czapiga:
Please no.. I think it's almost ready (blocked by critical interface) and these are just some trivial changes. If you insist on splitting it then we can push `-Wextra` part first since it was already working (fixed all issues in previous commits).

cahirwpz:
@czapiga I can help you with that. It should be relatively easy to extract `-Wstrict-prototypes` changes thanks to [cherry picking](https://git-scm.com/docs/git-cherry-pick).

czapiga:
@cahirwpz Ok, I will revert some changes and leave only -Wextra part in this PR. Then I will create PR for -Wstrict-prototypes.

cahirwpz:
Greate... and please change `Makefile.common` to:
```
WFLAGS   = -Wall -Wstrict-prototypes -Werror
CFLAGS   = -std=gnu11 -Og $(WFLAGS) -fno-builtin -nostdinc -nostdlib \
            $(GCC_SYSTEM_INC) -ffreestanding
CPPFLAGS = $(WFLAGS) -DDEBUG -I$(TOPDIR)/include
```

Issue #345, opened 2017-05-24
Integrate PEP8 into Travis build process


cahirwpz:
[pep8](https://pypi.python.org/pypi/pep8) is a tool to check your Python code against some of the style conventions in [PEP 8](https://www.python.org/dev/peps/pep-0008/).

I'd like it to be intergrated with Travis build process in similar way as `verify-format.sh` is, so that build fails if our Python code is not compliant with PEP 8 rules.

cahirwpz:
#347 has been merged. Please quickly prepare another PR that fixes PEP8 warnings in our code base.

cahirwpz:
Ok. I can wait no longer. I took care of it. PR will follow soon.

cahirwpz:
Closed with #349.

PR #347, opened 2017-05-24
Pep8

dyniec:
I had to use different approach than `verify-format.sh`. I just locate `*.py` files which aren't hidden (`.ycm_extra_conf.py`) or in cache (newlib/doc) and run pep8 on every one.
Implementation of  #345


cahirwpz:
I'll merge it as is, but please quickly prepare another PR that fixes PEP8 warnings in our code base.

PR #348, opened 2017-05-25
Add VOP_CLOSE, extend VOP_READDIR with extra argument

cahirwpz:
Short description:
* add `VOP_CLOSE` operation that takes `file_t` object,
* `VOP_READDIR` takes `f_data` field of `file_t` object,
* move some function declarations from `mount.h` to `vfs.h` which they belong to,
* some minor cleanups.

I decided that `initrd_vnode_open` for `V_DIR` will create state for `readdir_generic`, and will keep it in `f_data` field of `file_t` structure. Thus we need `initrd_vnode_close` to be capable of removing that state. The state itself will be passed as an extra argument to `VOP_READDIR`.


include/vnode.h
```diff
@@ -83,28 +85,34 @@ typedef struct vattr {
 } vattr_t;
 #endif
 
+#define VOP_CALL(op, vp, ...) ((vp)->v_ops->v_##op)((vp), __VA_ARGS__)
```
coodie: I don't like idea behind this macro. This doesn't introduce any value to the code except macro. Unfamiliar programmer will spend few extra seconds to see what it means, thus making it slightly more confusing. And I'm really used to old way of VOP functions.

sys/initrd.c
```diff
@@ -243,16 +239,19 @@ static int initrd_root(mount_t *m, vnode_t **v) {
...
 static int initrd_init(vfsconf_t *vfc) {
   unsigned rd_size = ramdisk_get_size();
 
   if (!rd_size)
     return ENXIO;
 
-  initrd_ops.v_lookup = initrd_vnode_lookup;
-  initrd_ops.v_read = initrd_vnode_read;
-  initrd_ops.v_open = vnode_open_generic;
-  initrd_ops.v_getattr = initrd_vnode_getattr;
```
coodie: Idea behind putting these functions here is to disable any functionality on initrd in case we didn't load it.

cahirwpz: If we can't load the ramdisk then `ENXIO` (*no such device or address*) is returned, and mount operation fails.

PR #353, opened 2017-05-27
Timer events

czapiga:
Implements `timer_event_t` part of #327


czapiga:
This code works on OVPsim but it hangs on QEMU

cahirwpz:
Well... it's not synchronized... in this case with `CRITICAL_SECTION`. I would not expect it to work ;-)

czapiga:
What should I actually synchronize? I've tried to add `CRITICAL_SECTION` only to intr handler and then even to every function in that module and result is the same.. Also it wasn't synchronized before that's why I didn't add any `CRITICAL_SECTION` in first place

cahirwpz:
1. Following functions do need to by synchronized when calling outside of interrupt context:
```c
void cpu_timer_add_event(timer_event_t *tev);
void cpu_timer_remove_event(timer_event_t *tev);
```

I think you may safely use `CRITICAL_SECTION` within an interrupt as well.

2. Following function should never be called outside interrupt handler, so it shouldn't be even exposed in `timer.h`. It doesn't need to be synchronized.
```c
void cpu_timer_intr(void *arg)
```

3. The interrupt should only be registered (i.e. `mips_intr_setup`), when there's at least one `timer_event` available for processing, otherwise the interrupt should be disabled. Please don't add superficial event in `cpu_timer_init`.

4. You may use `klog` facility within interrupt context. However `KL_INTR` should be disabled by default, so that message printing do not interfere with interrupt processing, which is time critical. You should be able to print `klog` conveniently from gdb after kernel is stopped, as soon as @laky55555 finishes work on #340 (it's still marked as *work in progress*).

*If you're sure your code should work, but the problem persists, we'll have to investigate `mips_intr_*` routines to make sure, there's no bug there.*

czapiga:
@cahirwpz 
1. Now these functions are called only while processing interrupt
2. I made that static and removed from interface
3. I think there will be always at least one timer_event - `void mips_clock(timer_event_t *tev)` which calls `void clock(systime_t _ms)` and I add it in `static void cpu_timer_init(void)`. I can create `static void mips_clock_init(void)` in `mips/clock.c` and copy that initialization code there if thats what you want

cahirwpz:
3. That's true, but it should be handled as any other timer event (need not be a special case).

mips/timer.c
```diff
@@ -0,0 +1,72 @@
...
+      return;
+
+    struct timer_event *event;
+    TAILQ_FOREACH (event, &events, tev_link) {
+      if (tv2tk(event->tev_when) > compare) {
+        mips32_set_c0(C0_COMPARE, tv2tk(event->tev_when));
+        break;
+      }
+      TAILQ_REMOVE(&events, event, tev_link);
+      event->tev_func(event);
```
cahirwpz: If someone forgets to set `tev_func` the debugging process will be painful.

czapiga: I thought we're not doing defensive programing - we don't check input pointers against `NULL` in other places

cahirwpz: That's correct, this is just a sidenote. However, I found difficult to debug a crash caused by a jump to `NULL` pointer. Maybe it will help you.

cahirwpz:
Ok. I'm taking it over in order to debug it.

mips/timer.c
```diff
@@ -0,0 +1,76 @@
...
+    if (ticks(event) > compare)
+      break;
+    TAILQ_REMOVE(&events, event, tev_link);
+    if (TAILQ_EMPTY(&events))
+      mips_intr_teardown(cpu_timer_intr_handler);
+    assert(event->tev_func != NULL);
+    event->tev_func(event);
+  }
+
+  mips32_set_c0(C0_COMPARE, ticks(TAILQ_FIRST(&events)));
```
czapiga: Why there is no `if (TAILQ_EMPTY(&events))` ?

cahirwpz: Because I did not trigger this bug ;-) Fixed! Thanks!

PR #354, opened 2017-05-28
vnode ops extend

coodie:
While working on tmpfs prototype I realised I'll need some extra operations on vnodes and extra system calls. I have recently pushed some of these to master, but I think it needs a little review and discussion. 

I'm not so sure what vattr argument is for, but I hope this will clarify.

Tests for these functionalities will come with tmpfs implementation.


coodie:
@cahirwpz 
Can you please take a look at it?

cahirwpz:
Unfortunately, we're not going to make any use of this code.

PR #356, opened 2017-05-28
Add tmpfs

coodie:
This is WIP branch providing tmpfs. 

Depends on #354


cahirwpz:
The value of this PR diminished to the point I find it to have no value to Mimiker anymore.

Issue #359, opened 2017-05-30
/bin/mandelbrot SIGSEGV


dyniec:
Somewhere between commits 86a7b3ed57a3a718dedc9d8b8336f07976174595 and 9de1fbe4e672741e722782b824edff9a36fab2e2 `mandelbrot` program stopped working.
Between those commits `user/mandelbrot/mandelbrot.c` hasn't changed.
I start it with `./launch -g -S qemu  init=/bin/mandelbrot`.
I will take a look into changes in devices and memory.
Log on `86a7b3ed57a3a718dedc9d8b8336f07976174595`:
```[user] Entering e_entry NOW
[syscall] sbrk(448)
[syscall] sbrk(3648)
[syscall] open("/dev/vga/videomode", 2, 438)
[syscall] sys_fstat(3, 0x6ffffbc0)
[syscall] sys_read(3, 0x80001b8, 1024)
[syscall] sys_fstat(1, 0x6ffffd70)
[syscall] sys_write(1, 0x80005c0, 35)
Current resolution: 320x200, 8 BPP
[syscall] sys_write(3, 0x80001b8, 9)
[syscall] close(3)
[syscall] open("/dev/vga/palette", 1, 0)
[syscall] sys_write(3, 0x41b048, 768)
[syscall] close(3)
[syscall] open("/dev/vga/fb", 1, 0)
[syscall] sys_write(3, 0x41b348, 307200)
[syscall] close(3)
[syscall] exit(0)
[thread] Thread 'main' {0x805bd228} has finished.
```
Log on `HEAD`:
```[init] Kernel initialized!
[thread] Thread 'main' {0x8059b250} has been created.
[user] Starting program "/bin/mandelbrot"
[user] Loading user ELF: /bin/mandelbrot
[user] User ELF size: 135032
[user] Entry point will be at 0x004008e8.
[user] ELF has 6 program headers
[user] Processing a PT_LOAD segment: VirtAddr = 0x00400000, Offset = 0x00001000, FileSiz = 0x00015058, MemSiz = 0x00015058, Flags = 5
[user] Processing a PT_LOAD segment: VirtAddr = 0x00416000, Offset = 0x00017000, FileSiz = 0x000009b0, MemSiz = 0x000009b0, Flags = 6
[user] Processing a PT_LOAD segment: VirtAddr = 0x00417000, Offset = 0x00018000, FileSiz = 0x00001244, MemSiz = 0x00001244, Flags = 4
[...]

PR #368, opened 2017-06-06
Mandelbrot bug research

dyniec:
#359


cahirwpz:
After merge with master, it works on mimiker server. However, image looks kinda funky. Definitely it needs some investigation.

cahirwpz:
After running `make distclean` in master branch `mandelbrot` works like a charm!

PR #370, opened 2017-06-06
Fix all problems

rafalcieslak:
This branch enables QEMU instruction counting and makes it the default simulator.

### DIFF
--- a/launcher/simulators.py
+++ b/launcher/simulators.py
@@ -67,6 +67,7 @@ class QEMU(Launchable):
                         '-device', 'VGA',
                         '-machine', 'malta',
                         '-cpu', '24Kf',
+                        '-icount', 'shift=7',
                         '-kernel', kwargs['kernel'],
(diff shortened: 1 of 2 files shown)


cahirwpz:
Oohhh... I'm thrilled to see this change in action. Does it work for future problems as well?

rafalcieslak:
Wow! Timeout reached - and I am already able to reproduce the problem! (I had to switch to qemu 2.0  though.)

rafalcieslak:
Okay, the issue turned out to be pretty simple - nothing seems to be wrong, the timeout was reached prematurely. This kind of makes sense, as QEMU no longer executes in max-speed mode, but synchronizes everything . This, however, has a nice side effect that execution speed should be the same on all machines (except when the host machine is simply too slow), so we'll probably no longer disagree on timeout values. I suppose we might even tune the speed to more or less match Malta's processor...

cahirwpz:
LGTM

PR #373, opened 2017-06-08
NS16550 UART driver

cahirwpz:
Testing ground for NS16550 driver that should handle all devices of using same chip interface, i.e. UART0 (COM1), UART1 (COM2), UART2(cbus).


drv/ns16550.c
```diff
@@ -0,0 +1,174 @@
...
+  .v_read = ns16550_read,
+  .v_close = ns16550_close,
+};
+
+static intr_filter_t ns16550_intr(void *data) {
+  ns16550_state_t *ns16550 = data;
+  resource_t *uart = ns16550->regs;
+  intr_filter_t res = IF_STRAY;
+
+  WITH_SPIN_LOCK (&ns16550->lock) {
```
jpszczolowski: Why do we acquire a spin lock in the bottom-half?

cahirwpz: Only for symmetry with top-half. It should not break anything as it increments & decrements `td_idnest` counter.

PR #375, opened 2017-06-09
New devfs API

cahirwpz:
This change introduces API for registering device subtrees in device filesystem. The most apparent improvement is visible on `sys/dev_vga.c` file. Same functionality is expressed with ~20 lines less, but with addition of directory listing (tested with `/bin/ls_rec`).

Registering device files (`fb`, `palette` and `videomode`) in a subdirectory (`vga`) boils down to:
```c
devfs_makedir(NULL, "vga", &vga_root);
devfs_makedev(vga_root, "fb", &framebuffer_vnodeops, vga); /* /dev/vga/fb */
devfs_makedev(vga_root, "palette", &palette_vnodeops, vga); /* /dev/vga/palette */
devfs_makedev(vga_root, "videomode", &videomode_vnodeops, vga); /* /dev/vga/videomode */
```

### DIFF
--- a/sys/devfs.c
+++ b/sys/devfs.c
@@ -7,96 +7,120 @@
 #include <malloc.h>
 #include <linker_set.h>
 
-/* Structure for storage in mnt_data */
-typedef struct devfs_mount { vnode_t *root_vnode; } devfs_mount_t;
+typedef struct devfs_node devfs_node_t;
+typedef TAILQ_HEAD(, devfs_node) devfs_node_list_t;
 
-static inline devfs_mount_t *devfs_of(mount_t *m) {
-  return (devfs_mount_t *)m->mnt_data;
-}
+struct devfs_node {
+  TAILQ_ENTRY(devfs_node) dn_link;
+  char *dn_name;
+  vnode_t *dn_vnode;
+  devfs_node_list_t dn_children;
+};
 
-/* Structure for storing installed devices */
-typedef struct devfs_device {
-  char name[DEVFS_NAME_MAX];
-  vnode_t *dev;
-  TAILQ_ENTRY(devfs_device) list;
-} devfs_device_t;
+/* device filesystem state structure */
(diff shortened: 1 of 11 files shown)


cahirwpz:
@rafalcieslak Please have a quick look. I hope I didn't miss anything important.

sys/devfs.c
```diff
@@ -6,94 +6,184 @@
...
-  idev->dev = device;
+  devfs_node_t *dn = kmalloc(M_VFS, sizeof(devfs_node_t), M_ZERO);
+  dn->dn_name = kstrndup(M_VFS, name, DEVFS_NAME_MAX);
+  dn->dn_vnode = vnode_new(V_DEV, vops, data);
 
-  WITH_MTX_LOCK (&devfs_device_list_mtx)
-    TAILQ_INSERT_TAIL(&devfs_device_list, idev, list);
+  WITH_MTX_LOCK (&devfs.lock) {
+    dn->dn_ino = ++devfs.next_ino;
+    if (devfs_find_child(parent, name))
```
pwit81: I wonder if this "if" statement should be guarded by a lock. The "if" condition is already guarded. Moreover, if the condition is true, the code above mistakenly increments devfs.next_ino.

cahirwpz: There's unlikely race condition here, IIUYC. Let's assume we skip `WITH_MTX_LOCK` from L63. You could have found that there's no node in L65, but before you inserted an entry onto `parent->dn_children` in L67 you've got preempted and someone else inserted exactly the same entry.

`devfs.lock` is a global lock for all entries in devfs node tree. The locking is quite coarse in here, but it hardly matters when it comes to performance (for now). Single lock makes things simpler.

pwit81: Actually my sugestion was not to skip `WITH_MTX_LOCK`, but to move lines L65-L66 out of `WITH_MTX_LOCK` block.

PR #376, opened 2017-06-09
Revert a major regression in vnode_open_generic

rafalcieslak:
I suggest to revert this change which was introduced as a part of this [major patch](https://github.com/cahirwpz/mimiker/commit/0bb05a098b27936df30fbbcadc38f9387174dc49). I see no point of it, and because of this change **usermode programs are no longer able to write to stdout**. Note that:
```
O_RDONLY = 0
O_WRONLY = 1
O_RDWR = 2
```
These are numbers, not bitmasks. Thus masking mode with `O_RDWR` causes files that are requested in `O_WRONLY` mode (like stdout) to be opened in `FF_READ` mode, which is incorrect.

I also urge you, again, to **manually test some user programs**, especially during such extensive changes as the one that introduced this problem, as it would immediately detect this issue and save us both time! Although... I think I shouldn't care anymore.

### DIFF
--- a/sys/vfs_vnode.c
+++ b/sys/vfs_vnode.c
@@ -179,7 +179,7 @@ int vnode_open_generic(vnode_t *v, int mode, file_t *fp) {
   fp->f_ops = &default_vnode_fileops;
   fp->f_type = FT_VNODE;
   fp->f_vnode = v;
-  switch (mode & O_RDWR) {
+  switch (mode) {
     case O_RDONLY:
       fp->f_flags = FF_READ;
       break;


rafalcieslak:
FWIW, that same change broke one more thing: `PATH_MAX` for `sys_open` was increased from 256 to 1024, but the corresponding test in `fd_test` wasn't updated, and thus `fd_test` fails ATM. I'm going to fix that in `utest` (#328) instead.

cahirwpz:
I'm running some programs manually... I've already identified the issue and fixed that in [here](https://github.com/cahirwpz/mimiker/pull/375/files#diff-b7151535da8489bb609e322ffb8d91c6L40), although I think that reverting that line is plain better. Thanks!

> Although... I think I shouldn't care anymore.

You definitely should! You did a great job on testing infrastructure. I hope your #328 effort will greatly improve testing procedure in user-space.

PR #377, opened 2017-06-09
CircleCI support

rafalcieslak:
This branch enables CircleCI support (in addition to TravisCI). It will grant us much greater control over our CI environment.

After some research, I've gathered which CircleCI advantages over TravisCI will be of use to us:

- CircleCI containers are based on Docker, and in fact they allow to run **any** docker image! I've chosen Ubuntu 16.04 for a start, as it has all packages we need in decently new versions (QEMU 2.5 from repositories is totally okay for us, though switching to an even newer version in the future won't be a problem).
  - In fact, at some point we may want to create our own docker image, which would come with all utilities, toolchains and configuration pre-installed. That would be a great way to provide a standardized development environment, and would simplify the CircleCI config a lot - it would be just a matter of pulling our hypothetical docker image.
- CircleCI runs so much faster. Not only in terms of CPU, but also network bandwidth and storage I/O. ATM system setup takes 1.5 minutes, compilation takes a minute, and running all tests ~3 minutes.
  - Yes, 3 minutes for 100 seeds. That's barely 2 seconds per seed! We may be able to reduce the main tests timeout by a great deal.
  - With a custom docker image system setup would go down to as little as 5 seconds, meaning we would get formatting & compilation results ready within a minute from pushing to repository - WOW!
- Open source projects get 4 simultaneously available containers for free. If we enable them, we'll instantly get 4x as many seeds tested within the same time. (Or we might decrease the number of seeds tested per run 4 times, so that we get the same results 4 times faster.)
- SSH connection to the testing container is available OOTB, no tricks required - just click "Restart build with SSH enabled" and you get an IP+port to connect to. Super useful for testing what went wrong, reproducing issues, or even running GDB **directly** on the testing environment!
- Build runs may upload so-called artifacts, that is - arbitrary files with build and/or test results. We missed that feature on Travis when we wanted to examine what does the `initrd.cpio` look like when build by Travis. This build script uses this feature to make `mimiker.elf` and `initrd.cpio` ready for a direct download - just go to 'Artifacts' tab of build results, and you can download them with a single click. We may want to consider storing kernel output in a similar manner, so that it may be easily downloaded.
- CircleCI allows to specify different jobs in a single test run (this feature has minor bugs ATM, so I did not enable it yet). That will allow us to split the test process into multiple stages: compilation, formatting verification, pep8 verification and kernel tests. Each of these will run separately and will appear as a separate check-mark on github. This way we'll know the results of formatting check within few seconds from pushing to repo, while other test modes will be still running. Splitting these into separate checkmarks will also simplify developer's work - it will be immediately apparent when a build failed due to formatting problems or due to other reason.
- CircleCI has a very simple file cache available, which allows to store files between builds. Should we download more than just newlib during compilation, we may want to cache these downloads so that they don't need to be repeated on each build run.

Naturally, nothing will build before we enable CircleCI integration for this repository, which is as simple as heading to https://circleci.com/dashboard and flipping the right switch ON. I suppose repository owner should be the one to enable it, although I think I am also able to activate it.

Note that this config is for **CircleCI 2.0**! CircleCI 2.0 is currently in public beta, and is only enabled when requested via a link at https://circleci.com/beta-access/

For testing & preview, I've enabled CircleCI for my mimiker fork. You can see an example test build here: https://circleci.com/gh/rafalcieslak/mimiker/63


cahirwpz:
I've enabled CircleCI for my account and mimiker repository. It works like a charm and indeed is much faster than TravisCI. I found configuration script to be self-explanatory.  The set of advantages and possible improvements (over TravisCI) offered by CircleCI is impressive. I reckon the explanation you provided is worth restating and posting as *CircleCI improvements* wish-task. All in all I'm in favour of switching to CircleCI immediately. Good job!

PR #383, opened 2017-06-14
Circleci workflows

rafalcieslak:
Today early morning we were accepted into private beta of CircleCI workflows feature. Today evening workflows left beta, and are now available to all CircleCI 2.0 users.

This branch is barely a cleanup for CircleCI config I promised to do, and it shouldn't change general behavior. Job triggers are removed and `workflow` block describes dependencies between jobs, also `defaults` are used to share parts of configuration between jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,10 +1,9 @@
-version: 2
-
 defaults: &defaults
   working_directory: ~/mimiker
   docker:
     - image: rafalcieslak/mimiker-build-base:1.1
 
+version: 2
 jobs:
   verify-formatting:
     <<: *defaults
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,15 +42,3 @@ jobs:
       - attach_workspace:
           at: ~/mimiker
       - run: './run_tests.py --thorough'
-
-workflows:
-  version: 2
-  build_and_test:
-    jobs:
-      - verify-formatting
-      - verify-pep8
-      - compile
-      - kernel-tests:
-        requires:
-          - compile
-
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,44 +1,24 @@
+defaults: &defaults
+  working_directory: ~/mimiker
+  docker:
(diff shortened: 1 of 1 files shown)


rafalcieslak:
This took me a while to troubleshoot, because, at the moment, syntax errors in `workflows` section trigger no errors and just cause CircleCI to ignore build request. The workflow is now working as intended, head to https://circleci.com/gh/cahirwpz/workflows/mimiker for a nice overview.

Look at how clean & concise `.circleci/config.yml` is now!

.circleci/config.yml
```diff
@@ -1,44 +1,25 @@
...
-            }          
-            trigger_job verify-formatting
-            trigger_job verify-pep8
-            trigger_job compile
 
-  verify-formatting:
-    working_directory: ~/mimiker
-    docker:
-      - image: rafalcieslak/mimiker-build-base:1.1
+defaults: &defaults
```
cahirwpz: What does `&defaults` do here?

.circleci/config.yml
```diff
@@ -1,44 +1,25 @@
...
     steps:
       - checkout
       - run: './verify-format.sh'
 
-  verify-pep8:
-    working_directory: ~/mimiker
-    docker:
-      - image: rafalcieslak/mimiker-build-base:1.1
+  verify_pep8:
+    <<: *defaults
```
cahirwpz: What does this line mean?

.circleci/config.yml
rafalcieslak: `defaults:` creates a set of arbitrary values to be applied later. `&XYZ` specifies that the name of this defaults set shall be `XYZ`.

.circleci/config.yml
rafalcieslak: `<<: *XYZ` is used to pull in all values from previously defined set of defaults named `XYZ`. This way `verify_pep8` is populated with values for `working_directory` and `docker`.

cahirwpz:
Thanks for the explanation!

PR #384, opened 2017-06-15
Fixed small bug related to relative path

laky55555:
Because we don't implement relative paths expected error wasn't name too long but ENOENT so test didn't succeed.

### DIFF
--- a/user/fd_test/fd_test.c
+++ b/user/fd_test/fd_test.c
@@ -184,6 +184,7 @@ void test_open_path() {
   char too_long[500];
   memset(too_long, 'c', sizeof(too_long));
   too_long[sizeof(too_long) - 1] = 0;
+  too_long[0] = '/';
   /* This is very unfortunate! In our errno.h: ENAMETOOLONG is 63, but errno.h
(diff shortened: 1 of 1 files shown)


user/fd_test/fd_test.c
```diff
@@ -184,6 +184,7 @@ void test_open_path() {
   char too_long[500];
   memset(too_long, 'c', sizeof(too_long));
   too_long[sizeof(too_long) - 1] = 0;
+  too_long[0] = '/';
```
laky55555: Added forward slash so path would be relative (we are trying to test fail because name is too long)

rafalcieslak:
Nice catch! This test is supposed to provide an insanely long file name to open, but recently our limit for file name lenght was raised from 256 to 1024. This test uses 500 characters, which fits into 1024 limit. I believe a proper solution is to make this file name longer than 1024. #328, which I hope will be merged soon, completely reorganizes the way we do user-mode tests, and removes (among others) the `fd_test` program. In #328 the test procedure is moved to `utest`, and I've fixed it there by increasing the size of `too_long` to 2000.

PR #389, opened 2017-06-30
A sys_execve wrapper for do_exec

pwit81:
Some remarks:

- no support for environments
- exec_args_t.argv type is incorrect. It is const char**, should be char *const,
- available docs don't specify meaning of ARG_MAX error code in unambigous way   
- tests are perhaps pretty weak and should be improved.


cahirwpz:
Further work on this feature is continued in #422

PR #403, opened 2017-07-17
Fix crash during brk segment creation on execve in forked processes

pwit81:
A small fixup concerning preservation of old process memory image during execve attempt.

### DIFF
--- a/sys/exec.c
+++ b/sys/exec.c
@@ -105,12 +105,21 @@ int do_exec(const exec_args_t *args) {
    * We can not destroy the current vm map, because exec can still fail,
    * and in that case we must be able to return to the original address space.
    */
-  vm_map_t *vmap = vm_map_new();
-  vm_map_t *old_vmap = p->p_uspace;
+  struct {
+    vm_map_t *uspace;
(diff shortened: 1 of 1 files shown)


sys/exec.c
```diff
@@ -105,12 +105,21 @@ int do_exec(const exec_args_t *args) {
...
+  } old = {p->p_uspace, p->p_sbrk, p->p_sbrk_end};
 
   /* We are the only live thread in this process. We can safely give it a new
    * uspace. */
+  vm_map_t *vmap = vm_map_new();
+
   p->p_uspace = vmap;
+  p->p_sbrk = NULL;
+  p->p_sbrk_end = 0;
+  sbrk_attach(p); /* Attach fresh brk segment. */
```
cahirwpz: Please note that this PR is named in somewhat misleading way.

The change is mostly concerned about attaching fresh brk segment to new process image.

cahirwpz: Ok. My bad. This fix is concerned about some conditions in `sbrk_attach` not met, namely:
```c
assert(p->p_uspace && (p->p_sbrk == NULL));
```

PR #404, opened 2017-07-17
Syslimits

pwit81:
We need a header file for storing system-wide limits, like ARG_MAX (max. bytes in execve argv and envp arrays), MAX_PATH (max bytes in a path)  or IOV_MAX (max elements in an I/O vector). We borrowed one from newlib.


user/newlib/overrides/include/sys/syslimits.h
```diff
@@ -0,0 +1 @@
+/home/pwit/workspace/mimiker/include/syslimits.h
```
cahirwpz: Why do you use absolute path referring to your home directory?

pwit81: My fail. Corrected in next commit.

PR #405, opened 2017-07-18
Definition of PATH_MAX removed from include/mount.h. VFS_PATH_MAX and…

pwit81:
… PATH_MAX unified. Source files updated accordingly.

### DIFF
--- a/sys/vfs.c
+++ b/sys/vfs.c
@@ -203,7 +203,7 @@ int vfs_lookup(const char *path, vnode_t **vp) {
   vnode_lock(v);
 
   if ((error = vfs_maybe_descend(&v)))
-    goto free_mem_and_return;
+    goto end;
 
   while ((component = strsep(&pathbuf, "/")) != NULL) {
     if (component[0] == '\0')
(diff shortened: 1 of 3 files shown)


sys/vfs.c
```diff
@@ -191,10 +191,10 @@ int vfs_lookup(const char *path, vnode_t **vp) {
 
   /* Copy path into a local buffer, so that we may process it. */
   size_t n = strlen(path);
-  if (n >= VFS_PATH_MAX)
+  if (n >= PATH_MAX)
     return -ENAMETOOLONG;
-  char pathcopy[VFS_PATH_MAX];
-  strlcpy(pathcopy, path, VFS_PATH_MAX);
+  char pathcopy[PATH_MAX];
```
cahirwpz: Because `PATH_MAX` is rather large compared to kernel thread stack size (4096 bytes) please move the allocation to heap by using `M_TEMP` memory pool.

pwit81: Corrected.

PR #408, opened 2018-04-06
Increased timeouts for GDB

wmoczulsky:
Sometimes `gdb` takes longer to startup and print diagnostic messages. With following patch we give it more time. Hopefully it will resolve an issue where debugger is killed prematurely after a failed test run.


cahirwpz:
I filled out PR's main comment. LGTM.

PR #409, opened 2018-04-06
Handle FPU in a lazy way in context switching

wmoczulsky:
So far, FPU data was not preserved between context switches. Now processes that are using FPU are marked, and FPU context is saved only for them.


mips/intr.c
```diff
@@ -169,6 +169,23 @@ static void fpe_handler(exc_frame_t *frame) {
   }
 }
 
+static void cp_unusable_handler(exc_frame_t *frame) {
+  int cp_id = (frame->cause & CR_CEMASK) >> CR_CESHIFT;
+  bool kernel_mode = (frame->sr & SR_KSU_MASK) == 0;
+
+  if (cp_id != 1) {
+    panic(
```
dingorth: > A COP0 or Cache instruction was executed while the processor was running in a mode other than Debug Mode
or Kernel Mode, and the CU0 bit in the Status register was a zero.

Every userspace process executing COP0 instruction causes kernel to crash.
Maybe we should kill a userspace process which caused that instead calling panic?

mips/exc.S
```diff
@@ -143,6 +204,74 @@ user_exc_leave:
         LOAD_REG(t0, PC, k0)
         mtc0    t0, C0_EPC   
 
+
+
+        # Load FPU context process makes use of it
```
dingorth: Please fix this comment.

mips/intr.c
cahirwpz: An attempt to execute `COP0` instruction in user-space should end up by sending `SIGILL` to the offending processes. However I believe illegal instruction (aka reserved) should be services by `EXC_RI` handler.

mips/exc.S
```diff
@@ -94,6 +94,67 @@ user_exc_enter:
         addi    t0, 1
         sw      t0, TD_IDNEST(s0)
 
+
```
cahirwpz: This change is visually distinct from the surrounding code. Please align formatting!

mips/exc.S
```diff
@@ -94,6 +94,67 @@ user_exc_enter:
...
+
+        # Save FPU registers
+        # Skip if process is not using FPU
+        LOAD_PCPU(t0)
```
cahirwpz: You don't need to load `curthread` in here - it's already stored in `s0` register.

include/thread.h
```diff
@@ -103,6 +103,8 @@ typedef struct thread {
   /* signal handling */
   sigset_t td_sigpend; /* Pending signals for this thread. */
   /* TODO: Signal mask, sigsuspend. */
+
+  bool td_uses_fpu;
```
cahirwpz: Please use `td_flags` instead - that's preferred way to store thread flags.

You should read surrounding code more carefully for your changes to integrate well.

mips/exc.S
```diff
@@ -143,6 +204,74 @@ user_exc_leave:
...
+
+        # Load FPU context process makes use of it
+        # by defaul assume that process does not, and disable FPU
+        mfc0 t0, C0_STATUS
+        li t1, ~(SR_CU1)
+        and t0, t1
+        mtc0 t0, C0_STATUS
+
+        # Skip if process is not using FPU
+        LOAD_PCPU(t0)
```
cahirwpz: Please load `curthread` once into `k1` register in [exc.S:122](https://github.com/cahirwpz/mimiker/blob/master/mips/exc.S#L122) and then use it throughout following code.

wmoczulsky:
Now it should be ok.

PR #412, opened 2018-04-09
Embed initial ramdisk inside the kernel binary

j-piecuch:
* Remove associated emulator options from the launcher script
* Minor refactoring -- see commit messages


Makefile
```diff
@@ -75,16 +83,6 @@ format:
 test: mimiker.elf
 	./run_tests.py
 
-# Detecting whether initrd.cpio requires rebuilding is tricky, because even if
```
cahirwpz: What was the rationale behind moving `initrd` stuff to subdirectory? Isn't that an unnecessary change?

mips/malta.c
```diff
@@ -125,30 +125,15 @@ extern intptr_t parse_rd_start(const char *s);
...
-   */
-  assert(is_aligned(rd_start, PAGESIZE));
-  assert(is_aligned(rd_start + rd_size, PAGESIZE));
-  assert(rd_start == 0 || (intptr_t)__kernel_end <= rd_start);
-
-  intptr_t real_kernel_end =
-    (rd_start == 0) ? (intptr_t)(__kernel_end) : (intptr_t)(rd_start + rd_size);
-
-  pm_seg_t *seg = (pm_seg_t *)real_kernel_end;
+  pm_seg_t *seg = (pm_seg_t *)__kernel_end;
```
cahirwpz: I understand you'll change that to use `kernel_sbrk` with next PR. Is that correct?

j-piecuch: That's correct.

Makefile
j-piecuch: Since the source files are already separated into different directories (i.e. `mips`, `sys`, etc.), I thought I'd follow this pattern. I also didn't want to create intermediate files (`initrd.o`, `initrd.cpio`) in the root directory of the project.

cahirwpz: Hmm... `initrd` directory stores only `Makefile` at the moment and it's unlikely it will contain anything else in foreseeable future. I'll revert that change. We should discuss such decisions next time.

cahirwpz:
I'm aware of problems with `signal_segfault` test spuriously failing. Nevertheless I'll integrate this change with master.

PR #413, opened 2018-04-11
Use `kernel_sbrk` to allocate physical memory description segment.

j-piecuch:
* Use `kernel_sbrk` to allocate physical memory segment in `pm_bootstrap`.
* Replaced uses of `__kernel_start` and `__kernel_end` in `malta.c` with accessor functions `kernel_get_start` and `kernel_get_end`.
* Calls to `kernel_sbrk` can now change the value returned by `kernel_get_end`. It can be frozen by calling `kernel_sbrk_freeze`.


cahirwpz:
@j-piecuch Please review PR after my changes.

PR #414, opened 2018-04-11
Add missing `#include <mips/mips.h>` in `mips/malta.c`

j-piecuch:
The `MIPS_KSEG0_TO_PHYS` macro defined in `include/mips/mips.h`
is used in `mips/malta.c`, but the necessary header is not included.


cahirwpz:
Hmm... `include/mips/malta.h` depends on `include/mips/mips.h`.

PR #415, opened 2018-04-11
Swap include order so that `SYSROOT` is defined.

j-piecuch:
`build/flags.user.mk` adds `--sysroot=$(SYSROOT)` to `CFLAGS`, but
SYSROOT is defined in `build/common.mk`, which is included by
`build/build.mk`. Therefore, `build/build.mk` needs to be included
before `build/flags.user.mk` in order for the flag to be properly
defined.


cahirwpz:
LGTM

PR #416, opened 2018-04-15
Added priority propagation functions to scheduler.

melouber:
Two new functions available to call in scheduler:

- `void sched_lend_prio(thread_t *td, td_prio_t prio);`

- `void sched_unlend_prio(thread_t *td, td_prio_t prio);`

They will be used in turnstile implementation.
Also added `td_base_prio` to thread structure and `TDF_BORROWING` to thread flags.


cahirwpz:
LGTM

PR #417, opened 2018-04-19
Use i8254 PIT for system clock

cahirwpz:
* Add infrastructure for timers that trigger callbacks either periodically or in one-shot mode.
 * Introduce bintime_t structure for high-fidelity time measurement.
 * Add i8254 PIT driver.
 * Use timer infrastructure to trigger system clock every 1ms.
 * Repurpose MIPS CPU timer for time measurement.

### DIFF
--- a/sys/timer.c
+++ b/sys/timer.c
@@ -74,11 +74,6 @@ int tm_init(timer_t *tm, tm_event_cb_t event, void *arg) {
   return 0;
 }
 
-bintime_t tm_gettime(timer_t *tm) {
-  assert(is_initialized(tm));
-  return tm->tm_gettime(tm);
-}
-
 int tm_start(timer_t *tm, unsigned flags, const bintime_t value) {
   assert(is_initialized(tm));
 
--- a/sys/timer.c
+++ b/sys/timer.c
@@ -91,7 +91,7 @@ int tm_release(timer_t *tm) {
 
   WITH_MTX_LOCK (&timers_mtx) {
     TAILQ_INSERT_TAIL(&timers, tm, tm_link);
-    tm->tm_flags &= ~(TM_INITIALIZED | TM_RESERVED);
+    tm->tm_flags &= ~(TMF_INITIALIZED | TMF_RESERVED);
   }
 
   return 0;
--- a/sys/timer.c
+++ b/sys/timer.c
@@ -29,7 +29,6 @@ int tm_register(timer_t *tm) {
   }
 
   klog("Registered '%s' timer.", tm->tm_name);
-
   return 0;
 }
 
@@ -45,7 +44,6 @@ int tm_deregister(timer_t *tm) {
   }
 
   klog("Unregistered '%s' timer.", tm->tm_name);
-
   return 0;
 }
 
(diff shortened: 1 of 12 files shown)


sys/timer.c
```diff
@@ -0,0 +1,113 @@
...
+#include <interrupt.h>
+
+static mtx_t timers_mtx = MTX_INITIALIZER(MTX_DEF);
+static timer_list_t timers = TAILQ_HEAD_INITIALIZER(timers);
+
+#define TMF_ACTIVE 0x1000
+#define TMF_INITIALIZED 0x2000
+#define TMF_ALLOCATED 0x4000
+#define TMF_REGISTERED 0x8000
+
```
pwit81: Please explain in a comment that above defined flags are for internal use and need not be seen outside the file.

sys/timer.c
```diff
@@ -0,0 +1,113 @@
...
+    TAILQ_REMOVE(&timers, tm, tm_link);
+    tm->tm_flags &= TMF_TYPEMASK;
+  }
+
+  klog("Unregistered '%s' timer.", tm->tm_name);
+  return 0;
+}
+
+timer_t *tm_alloc(const char *name, unsigned flags) {
+  timer_t *tm = NULL;
```
pwit81: `name` argument of `tm_alloc` is unused.  Also, alloc is a bit misleading word here. Consider renaming, e.g. to `tm_lock`.

sys/timer.c
```diff
@@ -0,0 +1,113 @@
...
+
+#define TMF_ACTIVE 0x1000
+#define TMF_INITIALIZED 0x2000
+#define TMF_ALLOCATED 0x4000
+#define TMF_REGISTERED 0x8000
+
+#define is_active(tm) ((tm)->tm_flags & TMF_ACTIVE)
+#define is_initialized(tm) ((tm)->tm_flags & TMF_INITIALIZED)
+#define is_allocated(tm) ((tm)->tm_flags & TMF_ALLOCATED)
+#define is_registered(tm) ((tm)->tm_flags & TMF_REGISTERED)
```
pwit81: Please describe in a comment how the state of a timer_t object changes in time, i.e first it is registered, next allocated then initialized. Finally it can transit forth and back between initialized and active states.

sys/timer.c
cahirwpz: Done!

sys/timer.c
cahirwpz: I renamed it to `tm_reserve`. Please note that you missed the fact that `tm_release` (previously `tm_free`) is not implemented.

sys/timer.c
cahirwpz: I added inline [dot](https://www.graphviz.org/doc/info/lang.html) graph to explain it.

sys/timer.c
cahirwpz: Ok. I fixed timer reserve by name.

cahirwpz:
@pwit81 I addressed all your remarks. Is there anything else I should explain / fix?

cahirwpz:
Ok. We agreed (off-line) that the PR may be merged with master.

PR #418, opened 2018-04-20
Turnstile and mutex implementation

melouber:


tests/mutex.c
```diff
@@ -23,11 +26,11 @@ static void mtx_test_main(void *arg) {
...
   value = 0;
 
-  td[0] = thread_create("td1", mtx_test_main, NULL);
-  td[1] = thread_create("td2", mtx_test_main, NULL);
-  td[2] = thread_create("td3", mtx_test_main, NULL);
-  td[3] = thread_create("td4", mtx_test_main, NULL);
-  td[4] = thread_create("td5", mtx_test_main, NULL);
+  for (int i = 0; i < T; i++) {
+    char name[20];
+    snprintf(name, sizeof(name), "td%d", i);
```
melouber: @ilikeheaps please move changes in this file to a new branch (and a separate pull request).

sys/mutex.c
```diff
@@ -40,6 +45,13 @@ void mtx_unlock(mtx_t *m) {
   WITH_NO_PREEMPTION {
     m->m_owner = NULL;
     m->m_lockpt = NULL;
-    sleepq_signal(&m->m_owner);
+    turnstile_chain_lock(m);
+    turnstile_t *ts = turnstile_lookup(m);
+    if (ts != NULL) {
+      // TODO signal instead of broadcast (not implemented yet)
+      turnstile_broadcast(ts);
```
melouber: @ilikeheaps 
In "The Design and Implementation of the FreeBSD Operating System" they say:

> The lower cost for releasing an uncontested lock explains the algorithm used to awaken waiters on a mutex. Historically, the mutex code would only awaken a single waiter when a contested lock was released, which left the lock in a contested state if there were more than one waiter. However, leaving a contested lock ensured that the new lock holder would have to perform a more expensive unlock operation. Indeed, all but the last waiter would have an expensive unlock operation. In the current FreeBSD system, all the waiters are awakened when the lock is released. Usually they end up being scheduled sequentially, which results in them all getting to do cheaper unlock operations. If they do all end up running concurrently, they will then use adaptive spinning and will finish the chain of lock requests sooner since the context switches to awaken the threads are performed in parallel rather than sequentially. This change in behavior was motivated by documentation of these effects noted in Solaris Internals [McDougall & Mauro, 2006].

So maybe we don't want signal instead of broadcast. But I'm not sure.

sys/mutex.c
```diff
@@ -21,8 +22,12 @@ void _mtx_lock(mtx_t *m, const void *waitpt) {
   }
 
   WITH_NO_PREEMPTION {
-    while (m->m_owner != NULL)
-      sleepq_wait(&m->m_owner, waitpt);
+    while (m->m_owner != NULL) {
+      turnstile_t *ts = turnstile_trywait(m);
```
melouber: @ilikeheaps, you're using `m` as `wchan` here, whereas in current mutex implementation they're using `&m->m_owner`. @cahirwpz, is it OK?

tests/mutex_simple.c
```diff
@@ -0,0 +1,83 @@
...
+  thread_join(t1);
+  assert(s1 == 3);
+  // NOTE this assert is pretty implementation-specific
+  // some changes might make it invalid
+  assert(mtx1.m_owner == NULL);
+
+  return KTEST_SUCCESS;
+}
+
+KTEST_ADD(mutex_simple, mtx_test_simple, 0);
```
melouber: @ilikeheaps I think that you should move tests/mutex_simple.c to tests/mutex.c

sys/mutex.c
ilikeheaps: I agree. I'll change the comment appropriately.

melouber: I've added location of this explanation in the book to your comment.

tests/mutex.c
melouber: @ilikeheaps reminder

tests/turnstile_propagate_many.c
```diff
@@ -58,8 +62,9 @@ static void starter_routine(void *_arg) {
         assert(thread_self()->td_prio == propagator_prio(i - 1));
       }
 
-      /* propagator[i] waits for mtx[i-1] (owned by propagator[i-1]) */
-      // TODO add asserts
+      assert(td_is_locked(propagator[i]));
+      assert(td_is_locked_on_mtx(propagator[i], &mtx[i - 1]));
+      assert(mtx_owner(&mtx[i - 1]) == propagator[i - 1]);
```
ilikeheaps: I think that `td_is_locked_on_mtx` should itself check that the thread is indeed locked and so `assert(td_is_locked(propagator[i]));` would be unnecessary.

melouber: You're right!

include/turnstile.h
```diff
@@ -0,0 +1,40 @@
...
+
+/* Scheduler should call this after changing priority of a thread that is
+ * blocked on some turnstile.
+ *
+ * It will re-sort list of blocked threads, on which `td` is and possibly
+ * propagate `td`'s priority if it was increased.
+ *
+ * If `td` was lending priority and it was decreased, we don't change the
+ * lent priority. The borrowing thread should finish soon anyway and we
+ * don't care enough.
```
melouber: @ilikeheaps maybe:
```
If `td` was lending priority and it was decreased, we don't unlend priority as unlending is expensive and the borrowing thread should finish soon.
```
? I think that `we don't care enough` is not appropriate.

ilikeheaps: Done.

PR #419, opened 2018-04-21
Testing Coprocessor Unusable and Reserved Instruction exception in userspace.

dingorth:
* testing Coprocessor Unusable exception in userspace.
* testing Reserved Instruction exception in userspace.

### DIFF
--- a/user/utest/misbehave.c
+++ b/user/utest/misbehave.c
@@ -28,26 +28,25 @@ int test_misbehave() {
   return 0;
 }
 
-static inline void exec_cp0_instr(void){
+static inline void exec_cp0_instr(void) {
   int value;
   asm volatile("mfc0 %0, $12, 0" : "=r"(value));
 }
 
 static inline void exec_reserved_instr(void) {
-  // how to raise Reserved Instruction exception?
-  // int value;
-  // asm volatile("mfc0 %0, $12, 0" : "=r"(value));
+  int value;
+  asm volatile("mfc2 %0, $12, 0" : "=r"(value));
 }
 
 static int spawn_process(void (*proc_handler)(void)) {
   int pid;
   switch ((pid = fork())) {
-    case -1 : 
+    case -1:
       exit(EXIT_FAILURE);
-    case 0 : 
+    case 0:
       proc_handler();
       exit(EXIT_SUCCESS);
-    default : 
+    default:
       return pid;
   }
 }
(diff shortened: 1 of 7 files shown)


dingorth:
How to raise Reserved Instruction exception in userspace? Does it really make sense to do that?

dingorth:
Before merge we need to have Reserved Instruction exception handler implemented.

user/utest/misbehave.c
```diff
@@ -24,3 +27,44 @@ int test_misbehave() {
...
 }
+
+static inline void exec_cp0_instr(void) {
+  int value;
+  asm volatile("mfc0 %0, $12, 0" : "=r"(value));
+}
+
+static inline void exec_reserved_instr(void) {
+  int value;
+  asm volatile("mfc2 %0, $12, 0" : "=r"(value));
```
cahirwpz: Some MIPS CPUs implement CP2 as mentioned on wikipedia:

> For example, in the PlayStation video game console, CP2 is the Geometry Transformation Engine (GTE), which accelerates the processing of geometry in 3D computer graphics.

Reserved instructions are according to my knowledge all privileged instructions such as `eret`, `tlbp` and so on. Refer to **Table 3.23 Privileged Instructions** from *MIPS® Architecture For Programmers Volume II-A: The MIPS32® Instruction Set*.

user/utest/misbehave.c
```diff
@@ -24,3 +27,44 @@ int test_misbehave() {
...
+      exit(EXIT_FAILURE);
+    case 0:
+      proc_handler();
+      exit(EXIT_SUCCESS);
+    default:
+      return pid;
+  }
+}
+
+int test_userspace_cop_unusable(void) {
```
cahirwpz: Hence code for both functions is almost the same... please replace it with call to:
```c
static void check_exception(void (*code)(void), int signum)
```
... and move relevant code there.

user/utest/misbehave.c
dingorth: Suprisingly all of the instructions from Table 3.23 cause Coprocessor Unusable exception. At least on ours Qemu setup.

user/utest/misbehave.c
dingorth: Sure, I'll fix it.

user/utest/misbehave.c
cahirwpz: Right! Another idea – refer to chapter A.2 of same document. Choose encoding that has `*` in it. According to *Table A.1 Symbols Used in the Instruction Encoding Tables*:

> Operation or field codes marked with this symbol are reserved for future use. Executing such an instruction must cause a Reserved Instruction Exception.

dingorth:
Created issue #424

PR #420, opened 2018-04-22
Added non-static set priority function to scheduler

melouber:
As for now in scheduler, we have static function to change thread's active priority. In this pull request we are adding non-static function for other parts of kernel to change thread's base priority (and adjust active accordingly, which involves turnstiles).

### DIFF
--- a/sys/sched.c
+++ b/sys/sched.c
@@ -79,20 +79,7 @@ void sched_set_prio(thread_t *td, prio_t prio) {
   if (td->td_flags & TDF_BORROWING && td->td_prio > prio)
     return;
 
-  prio_t oldprio = td->td_prio;
   sched_set_active_prio(td, prio);
-
-  /* If thread is locked on a turnstile, let the turnstile adjust
-   * thread's position on turnstile's \a ts_blocked list. */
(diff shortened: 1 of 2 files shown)


sys/sched.c
```diff
@@ -70,20 +70,45 @@ static void sched_set_priority(thread_t *td, td_prio_t prio) {
...
+
+  td_prio_t oldprio = td->td_prio;
+  sched_set_active_prio(td, prio);
+
+  /* If thread is locked on a turnstile, let the turnstile adjust
+   * thread's position on turnstile's \a ts_blocked list. */
+
+  /* These changes are waiting for turnstile pull request because
+   * now we don't have TDS_LOCKED and turnstile_adjust. */
+  // TO BE UNCOMMENTED
```
ilikeheaps: @melouber @cahirwpz, perhaps it would make more sense if we introduced minimal changes required for the commented code (TDS_LOCKED flag and empty procedure turnstile_adjust)?

cahirwpz: Please remove all code (and comments) that are not relevant to this change.

PR #421, opened 2018-04-23
[WIP] Major changes to `tlb_refill` and `tlb_exception_handler`

j-piecuch:
* TLB misses should not happen in `tlb_refill`.
* `tlb_exception_handler` is now significantly simpler.

### DIFF
--- a/mips/ebase.S
+++ b/mips/ebase.S
@@ -1,8 +1,16 @@
 #include <mips/ctx.h>
 #include <mips/m32c0.h>
 #include <mips/regdef.h>
-
+#include <mips/tlb.h>
 #include "assym.h"
+
+#define PDE_SHIFT 22
+#define PT_SIZE (1 << 22)
+#define PD_BASE (PT_BASE + PT_SIZE)
+#define PT_BASE_UPPER (PT_BASE >> 16)
+#define PD_BASE_UPPER (PD_BASE >> 16)
+#define KSEG2_FIRST_PDE_IDX (MIPS_KSEG2_START >> PDE_SHIFT)
+
 
         # Don't allow the assembler to reorder instructions.
         .set	noreorder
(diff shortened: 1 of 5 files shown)


mips/pmap.c
```diff
@@ -347,12 +354,13 @@ void tlb_exception_handler(exc_frame_t *frame) {
   vm_map_t *map = get_active_vm_map_by_addr(vaddr);
   if (!map) {
     klog("No virtual address space defined for %08lx!", vaddr);
-  } else {
-    vm_prot_t access = (code == EXC_TLBL) ? VM_PROT_READ : VM_PROT_WRITE;
```
j-piecuch: Why has this change been reverted? The semantics are preserved and the code is more readable.
Should this change go in a separate PR?

mips/ebase.S
```diff
@@ -29,59 +26,59 @@ _ebase:
...
+tlb_refill:
+        # Consider special case when an access to page table region was from
+        # the kernel (namely pmap module). As UPD & KPD are not mapped within
+        # page table we have to deal with it separately :(
 
         mfc0    k0, C0_BADVADDR
-        srl     k1, k0, 22
-        addi    k1, k1, -(PT_BASE >> 22)
+        srl     k1, k0, PDE_SHIFT
+        addi    k1, k1, -(PT_BASE >> PDE_SHIFT)
```
j-piecuch: I don't think the use of `PDE_SHIFT` is appropriate here. Sure, it **happens to have** the same value as what we need here, but we're shifting by 22 because 22 == `log2(PT_SIZE)`, not because 22 == `PDE_SHIFT`, and `log2(PT_SIZE)` may very well be different from `PDE_SHIFT`, depending on PTE size, address space size and page size. I agree that a hard-wired constant out of nowhere doesn't look good, so defining something like `PT_SIZE_SHIFT` or `LOG2_PT_SIZE` in `include/mips/pmap.h` and using that here seems more appropriate.

mips/ebase.S
```diff
@@ -29,59 +26,59 @@ _ebase:
...
-        # We found an entry with a matching VPN2, but that doesn't mean that it's valid!
-        # We need to check it.
+        # We found an entry with a matching VPN2, but that doesn't mean that
+        # it's valid! We need to check it.
 
-        # Read the entry with matching VPN2 (note: tlbr overwrites EntryHi!)
-        tlbr
-        mtc0    k1, C0_ENTRYHI                  # Restore original EntryHi.
+        tlbr                    # Read the entry with matching VPN2
+        ehb
```
j-piecuch: Programming the MIPS32® 24KETM Core Family, 5.1 Hazard barrier instructions:
> Since some implementations work by holding up execution of all instructions after the barrier, it’s preferable to place the barrier just before the consumer, not just after the producer.

In this case the producer is `tlbr`, the consumers are `mfc0` from `C0_ENTRYLO0/1`. There is no hazard between `tlbr` and `mtc0` or `mfc0` from `C0_CONTEXT`.

mips/pmap.c
cahirwpz: If you want to clean up the code please move such changes to separate PR.

I understand there's a temptation to clean up the code as we go, but that distracts the reviewer. Thinking of separation of concerns applies to preparing patches as well.

mips/ebase.S
cahirwpz: I agree – done!

mips/ebase.S
cahirwpz: Thanks for digging it up! This note is so imporant I'll put in into a comment.

mips/ebase.S
```diff
@@ -63,7 +68,6 @@ maybe_hit:
         # it's valid! We need to check it.
 
         tlbr                    # Read the entry with matching VPN2
-        ehb
```
j-piecuch: The `ehb` instruction still needs to be there, preferably just before the consumers, that is before the `mfc0 k1, C0_ENTRYLO1`. It shouldn't be removed completely.

cahirwpz: Right! I restored `ehb` at the point. Let's not move it around for clarity's sake. Two cycles are not worth it for now.

j-piecuch:
Yes. I have no further comments.

PR #422, opened 2018-04-27
A sys_execve wrapper for do_exec [Continuation]

JakubSzczerbinski:
Update and cleanup of changes introduced [here](https://github.com/cahirwpz/mimiker/pull/389).


sys/sysent.c
```diff
@@ -359,6 +360,86 @@ static int sys_rmdir(thread_t *td, syscall_args_t *args) {
   return result;
 }
 
+static int sys_execve(thread_t *td, syscall_args_t *args) {
+  const char *user_path = (const char *)args->args[0];
+  const char **user_argv = (const char **)args->args[1];
+
+  /* Unused as environments are not yet implemented */
+  /* const char **user_envp = (const char**)args->args[2]; */
```
pwit81: Remove commented-out code.

user/dump_args/dump_args.c
```diff
@@ -0,0 +1,40 @@
...
+
+  /*   i = 0; */
+  /*   while (envp[i] != NULL) { */
+  /*     printf("Argument No. %d is: %s\n", i, argv[i]); */
+  /*     i++; */
+  /*   } */
+
+  /* } else */
+  /*   printf("Environment envp is NULL\n"); */
+
```
pwit81: Remove commented-out code.

JakubSzczerbinski:
> 'sys_execve' should do some signal setup. According to 'http://netbsd.gw.com/cgi-bin/man-cgi?
execve':

> "Signals set to be ignored in the calling process are set to be ignored in the new process. Signals which are set to be caught in the calling process image are set to default action in the new process image. Blocked signals remain blocked regardless of changes to the signal action. The signal stack is reset to be undefined (see sigaction(2) for more information)."

This is issue regarding do_exec. These changes only introduce do_exec wrapper and do_exec issues are not relevant for these changes (not directly at least). I think this issue and concerns about the type of exec_args.argv should be addressed indepently from reviewed changes.

include/syslimits.h
```diff
@@ -37,7 +37,7 @@
 #ifndef _SYS_SYSLIMITS_H_
 #define _SYS_SYSLIMITS_H_
 
-#define ARG_MAX 65536 /* max bytes for an exec function */
+#define ARG_MAX 256 /* max bytes for an exec function */
```
cahirwpz: Value of this constant is very important because it's used by user-space programs as well.

I think you completely missed the fact that this value denotes size of a buffer that stores:
 * `argv` pointer table,
 * argument strings.

**It's not an upper limit for the number of string pointers in `argv` table!**

cahirwpz:
@JakubSzczerbinski Any progress here?

cahirwpz:
Further `sys_execve` development was moved to #496

Issue #241, opened 2017-04-06
`critical_exit` should switch to another thread if indirectly requested


cahirwpz:
Intuitively [critical_leave](http://mimiker.ii.uni.wroc.pl/source/xref/mimiker/sys/sync.c#12) should switch to another thread if, by execution of critical section, a thread was added to run queue and the thread has a higher priority or currently running thread's time slice was exhausted. But we don't do it.

By looking at FreeBSD's [critical_exit](http://bxr.su/FreeBSD/sys/kern/kern_switch.c#197) it's obvious that they handle it correctly by utilizing `td_owepreempt` field in `thread`'s structure.

```C
/* k*- only accessed by curthread and from an interrupt */
volatile u_char td_owepreempt;  /* (k*) Preempt on last critical_exit */
```

The flag is set by [maybe_preempt](http://bxr.su/FreeBSD/sys/kern/sched_4bsd.c#maybe_preempt) which is called by [sched_add](http://bxr.su/FreeBSD/sys/kern/sched_4bsd.c#1396). The latter is called in many places, but one is of particular interest i.e. [turnstile_unpend](http://bxr.su/FreeBSD/sys/kern/subr_turnstile.c#939), which wakes up threads that were suspended on a turnstile.

melouber:
In Mimiker we already set `TDF_NEEDSWITCH` when time slice is exhausted or when some more important thread is added to run queue. I'd say that solution to this issue is adding this code:
```
if (td->td_pdnest == 0 && td->td_flags & TDF_NEEDSWITCH) {
  WITH_SPINLOCK(td->td_spin) {
    td->td_state = TDS_READY;
    sched_switch();
  }
}
```
at the end of `preempt_enable` function. What do you think?

cahirwpz:
Looks reasonable to me. I'll accept the change if you create PR with it.

PR #423, opened 2018-04-28
`preempt_enable` should switch to another thread if indirectly requested

melouber:
In Mimiker we set `TDF_NEEDSWITCH` when thread's time slice is exhausted or when some another thread with higher priority is added to run queue.
To solve issue #241 I add code to `preempt_enable` which preempts current thread if `TDF_NEEDSWITCH` is set.
Note that `critical_exit` in FreeBSD is counterpart of our `preempt_enable`.

### DIFF
--- a/sys/sched.c
+++ b/sys/sched.c
@@ -189,10 +189,13 @@ void preempt_disable(void) {
 void preempt_enable(void) {
   thread_t *td = thread_self();
   assert(td->td_pdnest > 0);
-  td->td_pdnest--;
 
-  if (td->td_pdnest == 0 && td->td_flags & TDF_NEEDSWITCH) {
-    WITH_SPINLOCK(td->td_spin) {
+  td->td_pdnest--;
+  if (td->td_pdnest > 0)
+    return;
+
+  WITH_SPINLOCK(td->td_spin) {
+    if (td->td_flags & TDF_NEEDSWITCH) {
       td->td_state = TDS_READY;
       sched_switch();
     }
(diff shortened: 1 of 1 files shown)


sys/sched.c
```diff
@@ -190,6 +190,13 @@ void preempt_enable(void) {
   thread_t *td = thread_self();
   assert(td->td_pdnest > 0);
   td->td_pdnest--;
+
+  if (td->td_pdnest == 0 && td->td_flags & TDF_NEEDSWITCH) {
```
cahirwpz: If `td->td_pdnest == 0` and `td->td_flags & TDF_NEEDSWITCH` are true, we enter the *if statement* and get preempted by interrupt, and after return we yield though `TDF_NEEDSWITCH` is not marked anymore. Isn't that a race condition?

melouber: Yes. I think so. I've moved `if` into spinlock.
BTW At first glance in FreeBSD they have similar situation if an interrupt happened between lines 209 and 210. http://bxr.su/FreeBSD/sys/kern/kern_switch.c#198

cahirwpz:
LGTM

sys/sched.c
melouber: @cahirwpz as I said in FreeBSD they had similar issue. I thought I would fix it, so I would have second commit in FreeBSD and therefore get more fame and glory. But someone 28 days ago noticed it as well and forestalled me... :disappointed: 
https://github.com/freebsd/freebsd/commit/e4785b15b237e7ed6ace7ace14cb2acb1287af29

PR #426, opened 2018-05-03
Little cleanup before turnstile PR

melouber:
- Move `td_prio_t` definition from `include/thread.h` to `include/common.h`
- Rename `td_prio_t` to `prio_t`
- Unify `prio_t` from `include/interrupt.h` with `prio_t` from `include/common.h` (now interrupt priority is `uint8_t` instead if `int`!)
- Add `#include <common.h>` to `include/pool.h`


cahirwpz:
LGTM

PR #427, opened 2018-05-04
Improve/fix mutex test

ilikeheaps:
Current mutex test doesn't fail (or fails rarely) with empty mutex implementation.
* yield CPU while inside the mutex to create some opportunity for failure
* make T parameter (number of threads) easily configurable
* add prefixes to variables to separate tests
* add a very basic test

The test parameters could be now significantly decreased (empty mutex implementation fails at N=1 T=2) for faster testing.


tests/mutex.c
```diff
@@ -4,39 +4,82 @@
...
+static thread_t *simple_td0;
+static int simple_status = 0;
+
+static void simple_routine(void *arg) {
+  WITH_NO_PREEMPTION {
+    simple_status = 1;
+    mtx_lock(&simple_mtx);
+  }
+  WITH_NO_PREEMPTION {
+    simple_status = 2;
```
melouber: Why do you need this `preempt_disable` and `simple_status = 2` here?

tests/mutex.c
```diff
@@ -4,39 +4,82 @@
...
-    thread_join(td[i]);
+  assert(simple_status == 1);
+  mtx_unlock(&simple_mtx);
 
-  assert(value == N * T);
+  thread_join(simple_td0);
+  assert(simple_status == 3);
+  // NOTE This assert is implementation-specific
+  // Some changes might make it invalid
+  // We could assert =!mtx_owned(...)= but that would be less restrictive
```
melouber: Please use Doxygen style instead of Org-mode style here: `=!mtx_owned(...)=`.
And please use `/* ... */` comment instead of `// ...` to be more consistent with other Mimiker source files.

tests/mutex.c
```diff
@@ -4,39 +4,82 @@
...
 #include <ktest.h>
 
-static mtx_t mtx = MTX_INITIALIZER(MTX_DEF);
-static volatile int32_t value;
+static mtx_t counter_mtx = MTX_INITIALIZER(MTX_DEF);
+static volatile int32_t counter_value;
 
-#define N 1000
-#define T 5
+#define COUNTER_N 1000
```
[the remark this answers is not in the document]
ilikeheaps: I suppose that would be outside the scope of this change.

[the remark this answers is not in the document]
melouber: @cahirwpz ?

tests/mutex.c
melouber: Then I would delete
```
WITH_NO_PREEMPTION {
  simple_status = 2;
  mtx_unlock(&simple_mtx);
}
```
and leave only
```
mtx_unlock(&simple_mtx);
```
As you don't use value `2` of `simple_status` anywhere else in this test.

tests/mutex.c
ilikeheaps: I tried changing it appropriately. Doxygen supposedly supports Markdown formatting but I'm not sure if it worked out because I use neither of these.

tests/mutex.c
ilikeheaps: I wish to leave it as it is for consistency. Actually value `2` has its meaning of not being `0`, `1` or `3`. Perhaps changing numbers to some enum:
```
enum status { INITIAL, LOCKING, UNLOCKING, DONE };
```
would make it more clear?

melouber: That would be better but still -- in my opinion setting statuses that we don't even check is useless (and confusing for me as a code reader).

tests/mutex.c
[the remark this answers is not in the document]
cahirwpz: I'll ask you to walk me through this PR during our next meeting.

cahirwpz:
LGTM

PR #428, opened 2018-05-11
Added turnstile tests

melouber:
There are three test files:
1. turnstile_simple.c
2. turnstile_complex.c
3. turnstile_adjust.c

The first one checks whether priority propagation happens.
The second one checks whether chain priority propagation happens.
The third one is added because `turnstile_simple.c` and `turnstile_complex.c` tests are using mutexes, i.e. are using functions `turnstile_wait`, `turnstile_broadcast`, `turnstile_lookup`, `turnstile_acquire`, and are not using `turnstile_adjust` function, which would be left untested.

I think that we should merge these three files into one file or put them into one folder (this would probably require changes in our build system). But now they're separate to increase readability during review.

Test `turnstile_adjust.c` was made by @ilikeheaps and I've reviewed and accepted it. I put this information here as I can't _approve changes_ to my own PR.


tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+      assert(thread_self()->td_flags & TDF_BORROWING);
+      assert_priorities(HIGH, HIGH, MED, HIGH);
+
+      assert(med_prio_mtx1_acquired == 0);
+      assert(high_prio_mtx2_acquired == 0);
+    }
+  }
+
+  /* Enabling preemption should've preempted us and switched to
+   * higher priority tasks. Our priority should've been restored. */
```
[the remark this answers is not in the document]
melouber: We don't give up CPU in `mtx_unlock`.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+       * Td3 lent td1 HIGH priority. We also got HIGH priority as chain
+       * propagation happened because td1 is blocked on mtx1, which we
+       * possess. */
+      assert(thread_self()->td_prio == HIGH);
+      assert(thread_self()->td_flags & TDF_BORROWING);
+      assert_priorities(HIGH, HIGH, MED, HIGH);
+
+      assert(med_prio_mtx1_acquired == 0);
+      assert(high_prio_mtx2_acquired == 0);
+    }
```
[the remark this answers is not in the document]
ilikeheaps: You're right, never mind.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+enum {
+  /* Priorities are multiplies of RunQueue_PriorityPerQueue
+   * so that each priority matches different run queue. */
+  LOW = 0,
+  MED = RQ_PPQ,
+  HIGH = 2 * RQ_PPQ
+};
+
+/* td3 */
+static void high_prio_task(void *arg) {
```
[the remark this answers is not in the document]
melouber: Done.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+
+/* td1 */
+static void med_prio_task1(void *arg) {
+  WITH_NO_PREEMPTION {
+    WITH_MTX_LOCK (mtx2) {
+      WITH_MTX_LOCK (mtx1) {
+        /* Td0 (low_prio_task) has released mtx1.
+         * Td3 (high_prio_task) is waiting for mtx2. */
+        med_prio_mtx1_acquired = 1;
+        assert(low_prio_mtx1_acquired == 1);
```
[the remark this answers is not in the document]
melouber: But isn't that the meaning of word acquired?

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+      assert(high_prio_mtx2_acquired == 0);
+
+      thread_yield();
+
+      /* Thread td1 acquired mtx2. Then it tried to acquire mtx1 and
+       * lent us MED priority. */
+      assert(thread_self()->td_prio == MED);
+      assert(thread_self()->td_flags & TDF_BORROWING);
+
+      /* Now we add to the game td2 and td3. */
```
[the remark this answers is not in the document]
melouber: Done.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+
+    WITH_MTX_LOCK (mtx1) {
+      low_prio_mtx1_acquired = 1;
+      assert(med_prio_mtx1_acquired == 0);
+      assert(high_prio_mtx2_acquired == 0);
+
+      thread_yield();
+
+      /* Thread td1 acquired mtx2. Then it tried to acquire mtx1 and
+       * lent us MED priority. */
```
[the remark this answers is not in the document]
melouber: Done.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+  /* Priorities are multiplies of RunQueue_PriorityPerQueue
+   * so that each priority matches different run queue. */
+  LOW = 0,
+  MED = RQ_PPQ,
+  HIGH = 2 * RQ_PPQ
+};
+
+/* td3 */
+static void high_prio_task(void *arg) {
+  WITH_MTX_LOCK (mtx2) {
```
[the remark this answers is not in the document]
melouber: Done.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+
+    /* Thread td1 (med_prio_task1) will run first after next thread_yield(). */
+    assert_priorities(LOW, MED, LOW, LOW);
+
+    WITH_MTX_LOCK (mtx1) {
+      low_prio_mtx1_acquired = 1;
+      assert(med_prio_mtx1_acquired == 0);
+      assert(high_prio_mtx2_acquired == 0);
+
+      thread_yield();
```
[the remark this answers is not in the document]
melouber: What about `assert(turnstile_lookup(mtx1) == NULL);`? It would tell us that `mtx1` was not contested when `td0` acquired it, so nothing could happen with priorities.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+      td->td_base_prio = prio;                                                 \
+    }                                                                          \
+  } while (0)
+
+static mtx_t *mtx1 = &MTX_INITIALIZER(MTX_DEF);
+static mtx_t *mtx2 = &MTX_INITIALIZER(MTX_DEF);
+static thread_t *td[T];
+static volatile bool low_prio_mtx1_acquired;
+static volatile bool med_prio_mtx1_acquired;
+static volatile bool high_prio_mtx2_acquired;
```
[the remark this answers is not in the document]
melouber: Why? I don't want to check current ownership but fact of having already acquired mutex (or not).

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+    assert(low_prio_mtx1_acquired == 1);
+    assert(med_prio_mtx1_acquired == 1);
+
+    assert_priorities(LOW, MED, MED, HIGH);
+  }
+}
+
+/* td1 */
+static void med_prio_task1(void *arg) {
+  WITH_NO_PREEMPTION {
```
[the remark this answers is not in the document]
melouber: Done.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+      lend_prio(td[2], MED);
+      lend_prio(td[3], HIGH);
+      assert_priorities(MED, MED, MED, HIGH);
+
+      thread_yield();
+
+      /* Thread td3 tried to acquire mtx2 (which is in td1's possession).
+       * Td3 lent td1 HIGH priority. We also got HIGH priority as chain
+       * propagation happened because td1 is blocked on mtx1, which we
+       * possess. */
```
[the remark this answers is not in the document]
melouber: First possession is explained by new assert in `high_prio_task`. The second one is trivial.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+  }
+}
+
+/* td1 */
+static void med_prio_task1(void *arg) {
+  WITH_NO_PREEMPTION {
+    WITH_MTX_LOCK (mtx2) {
+      WITH_MTX_LOCK (mtx1) {
+        /* Td0 (low_prio_task) has released mtx1.
+         * Td3 (high_prio_task) is waiting for mtx2. */
```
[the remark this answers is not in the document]
melouber: Thread `td0` has released `mtx1` because we are now in `mtx1`. It doesn't need any assert.
Thread `td3` is waiting for `mtx2` - what about `assert(td[3]->td_blocked->ts_wchan == mtx2);`?

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+        med_prio_mtx1_acquired = 1;
+        assert(low_prio_mtx1_acquired == 1);
+        assert(high_prio_mtx2_acquired == 0);
+
+        assert(thread_self()->td_prio == HIGH);
+        assert(thread_self()->td_flags & TDF_BORROWING);
+      }
+    }
```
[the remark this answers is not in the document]
melouber: We don't give up CPU in `mtx_unlock`.

tests/turnstile_complex.c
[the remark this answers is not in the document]
ilikeheaps: However we don't know who owned `mtx1`. Assert about `td0` and `mtx1` could go right before the attempt to lock `mtx1` - something like `assert(mtx1_owner == td0)`. Well, we still wouldn't know if somebody didn't get the mutex after `td0` unlocked it and before `td3` acquired it.
About the second one: looks okay. I'd alternatively consider `assert(td[3]->td_wchan == mtx2)`.

tests/turnstile_complex.c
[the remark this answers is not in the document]
ilikeheaps: I actually intended it as a reminder about priorities of the threads - so that the reader will immediately know which thread will run after `thread_yield();`. It is indeed written a few lines above but I think repeating it here would make it easier to read the code.

tests/turnstile_complex.c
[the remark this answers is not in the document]
ilikeheaps: These wouldn't have to fully replace the existing variables. They would be helpful for some asserts which I suggested, some of these are already implemented with `mtx->m_owner`. The problem with `mtx->m_owner` is that changing the mutex implementation details could make the test not compile or fail.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,182 @@
...
+
+/* td1 */
+static void med_prio_task1(void *arg) {
+  WITH_NO_PREEMPTION {
+    assert(mtx2->m_owner == NULL);
+    WITH_MTX_LOCK (mtx2) {
+      med_prio_mtx2_acquired = 1;
+
+      assert(mtx1->m_owner == td[0]);
+      WITH_MTX_LOCK (mtx1) {
```
ilikeheaps: Trying to acquire this mutex will cause a switch to another thread. An assert about priorities right before it (I think `assert_priorities(LOW, MED, LOW, LOW)`) would be welcome.

tests/turnstile_complex.c
```diff
@@ -0,0 +1,171 @@
...
+   * high_prio_task. */
+  assert(high_prio_mtx2_acquired == 1);
+  assert(!(thread_self()->td_flags & TDF_BORROWING));
+  assert_priorities(LOW, MED, MED, HIGH);
+}
+
+/* td2 */
+static void med_prio_task2(void *arg) {
+  /* Without turnstile mechanism this assert would fail. */
+  assert(high_prio_mtx2_acquired);
```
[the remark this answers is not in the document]
ilikeheaps: Priority propagation could work only partially though - for example only propagate to the thread owning the mutex (and not to the threads blocking that thread). The test would still fail but not necessarily in this way. I think it would randomly fail either here or on the assert at 88: `assert_priorities(HIGH, HIGH, MED, HIGH)` (`td0`, after second `thread_yield`. Look at 78: `assert_priorities(MED, MED, MED, HIGH)`: with single thread propagation this assert could be passed, `td3` would bump `td1` to `HIGH` but `td2` would still be locked and the only runnable threads would be `td0` and `td2`, both with `MED` priority.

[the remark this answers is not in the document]
melouber: My comment doesn't say what would happen if partial priority propagation was implemented.
It says that if you have no priority propagation in mutexes, `med_prio_task2` would run before `high_prio_task` acquires `mtx2`.

tests/turnstile_complex.c
melouber: Done.

tests/turnstile_complex.c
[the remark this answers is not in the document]
melouber: 1a) Now we do know who owned mtx1. `assert(mtx1->m_owner == td[0]);`
1b) My comment doesn't say that we were the first to get the mutex after `td0` released it.
2) Done.

tests/turnstile_chain.c
```diff
@@ -0,0 +1,106 @@
...
+// TODO replace asserts about priority borrowing with some accessor
+/* TODO consider where we could check restoring priorities
+ * I think that all threads should keep HIGH priority until the last propagator
+ * has acquired 2nd mutex and then the priorities would be restored to initial
+ * ones (propagator_priority(i) for propagator[i]).
+ */
+
+static mtx_t mtx[T + 1];
+// TODO consider whether this would be any useful
+// volatile int mtx_passed[T + 1];
```
melouber: I think that it wouldn't because fields in `mtx_t` are already volatile.

tests/turnstile_chain.c
```diff
@@ -0,0 +1,106 @@
...
+
+// for simpler code: propagator[0] = starter
+static thread_t *propagator[T + 1];
+static thread_t *starter;
+
+static void set_base_prio(thread_t *td, prio_t prio) {
+  WITH_SPINLOCK(td->td_spin) {
+    sched_set_prio(td, prio);
+  }
```
melouber: 1) Name `set_base_prio` is misleading as it sets both base and active prio.
2) Remember that `sched_set_prio` will be using `turnstile_adjust` in the future. Maybe we don't want to call it in our turnstile tests.

tests/turnstile_chain.c
```diff
@@ -0,0 +1,106 @@
...
+  // TODO should this remain `n + 1` and not just `n`?
+  return (n + 1) * RQ_PPQ;
+}
+
+/* n <- [1..T]
+ * nth propagator will contest nth mutex and block on (n-1)th mutex (owned
+ * by (n-1)th propagator) causing priority propagation to [0..n-1]th propagator
+ */
+static void propagator_routine(void *argn) {
+  int n = (int)argn;
```
melouber: I'd change this code to `propagator_routine(int n)` and change `thread_create` to `propagator[i] = thread_create(name, (void (*)(void *))propagator_routine, (void *)i);`.

tests/turnstile_chain.c
```diff
@@ -0,0 +1,106 @@
...
+// n <- [0..T]
+// TODO maybe rename to separate_prio or something?
+static int propagator_prio(int n) {
+  // TODO should this remain `n + 1` and not just `n`?
+  return (n + 1) * RQ_PPQ;
+}
+
+/* n <- [1..T]
+ * nth propagator will contest nth mutex and block on (n-1)th mutex (owned
+ * by (n-1)th propagator) causing priority propagation to [0..n-1]th propagator
```
melouber: I'd call these things by index `[]`, e.g. `mtx[i]` instead of `i-th mutex`, because I don't know whether or not you are counting from 0.

tests/turnstile_chain.c
```diff
@@ -0,0 +1,106 @@
...
+
+    for (int i = 1; i <= T; i++) {
+      WITH_NO_PREEMPTION {
+        set_base_prio(propagator[i], propagator_prio(i));
+        sched_add(propagator[i]);
+        assert(thread_self()->td_prio == propagator_prio(i - 1));
+      }
+
+      // check if the priorities have propagated correctly
+      for (int j = 0; j < i; j++) {
```
melouber: I'd add assert between for loops: `assert(TD_IS_LOCKED(propagator[i]));`.

tests/turnstile_chain.c
melouber: Sorry, I've misread your comment. Could you explain `mtx_passed`?

ilikeheaps: I meant it as "whether propagator[n] has entered (and possibly already exit) mtx[n]". Pretty much `mtx_acquired` in the deleted test.

tests/turnstile_chain.c
ilikeheaps: Done.

tests/turnstile_chain.c
ilikeheaps: Done.

tests/turnstile_chain.c
ilikeheaps: Done.

tests/turnstile_chain.c
ilikeheaps: Also, 2.: I don't think we should mind. We will just test a little more functions (although `turnstile_adjust` in this case will perform very simple action because `set_prio` is called only before the thread is added).

tests/turnstile_chain.c
melouber: Or add a comment that `propagator[i]` is locked on `mtx[n-1]` and when turnstile PR comes in, we'll change it to appropriate assert.

tests/turnstile_chain.c
```diff
@@ -0,0 +1,95 @@
...
+ *   priority propagation to propagator[0..n-1]
+ */
+static void propagator_routine(int n) {
+  assert(thread_self()->td_prio == propagator_prio(n));
+  assert(mtx_owner(&mtx[n]) == NULL);
+  WITH_MTX_LOCK (&mtx[n]) {
+    assert(mtx_owner(&mtx[n - 1]) == propagator[n - 1]);
+    WITH_MTX_LOCK (&mtx[n - 1]) {
+      /* Nothing interesting here. */
+    }
```
melouber: `clang-format` doesn't like these lines :disappointed: but I don't like its one-line solution:
`WITH_MTX_LOCK (&mtx[n - 1]) { /* Nothing interesting here. */ }`

tests/turnstile_chain.c
```diff
@@ -0,0 +1,95 @@
...
+
+static void starter_routine(void *_arg) {
+  assert(mtx_owner(&mtx[0]) == NULL);
+  WITH_MTX_LOCK (&mtx[0]) {
+    for (int i = 1; i <= T; i++) {
+      WITH_NO_PREEMPTION {
+        set_prio(propagator[i], propagator_prio(i));
+        sched_add(propagator[i]);
+        assert(thread_self()->td_prio == propagator_prio(i - 1));
+      }
```
melouber: I think that this last assert is unnecessary. This check happens in line 59: `assert(propagator[j]->td_prio == propagator_prio(i));` in previous `for` iteration.

tests/turnstile_chain.c
```diff
@@ -0,0 +1,95 @@
...
+        assert(td_is_borrowing(propagator[j]));
+      }
+    }
+  }
+  assert(thread_self()->td_prio == propagator_prio(0));
+  assert(!td_is_borrowing(thread_self()));
+}
+
+static int test_main(void) {
+  klog_setmask(KL_MASK(KL_TEST) | KL_MASK(KL_THREAD));
```
melouber: Why do we need it here? And in line 90?

ilikeheaps: Oh, that's to stop being flooded by random messages in klog. I'm not sure if it actually helps at all. I suppose I'll remove it since that's how it is in other tests.

tests/turnstile_chain.c
ilikeheaps: That's true. But I'd leave it as a reminder for the reader: "right now we have this priority".

tests/turnstile_chain.c
ilikeheaps: It doesn't mind the version one commit behind though:
```
WITH_MTX_LOCK(&mtx[n-1]) {
  // Nothing interesting here
}
```

tests/turnstile_chain.c
melouber: Ok.

tests/turnstile_chain.c
melouber: Ok.

PR #430, opened 2018-05-11
added reserved instruction handler and updated cp unusable handler to send signals

wmoczulsky:
Now in case of reserved instruction exception or coprocessor unusable exception (other than FPU coprocessor), SIGILL is sent to process, instead of panicking.

### DIFF
--- a/mips/intr.c
+++ b/mips/intr.c
@@ -171,7 +171,7 @@ static void fpe_handler(exc_frame_t *frame) {
 
 static void cp_unusable_handler(exc_frame_t *frame) {
   if (in_kernel_mode(frame)) {
-    panic("FPU unusable exception in kernel mode.");
+    panic("Coprocessor unusable exception in kernel mode.");
   }
 
   int cp_id = (frame->cause & CR_CEMASK) >> CR_CESHIFT;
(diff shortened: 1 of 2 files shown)


tests/utest.c
```diff
@@ -98,9 +98,10 @@ UTEST_ADD_SIMPLE(access_basic);
 UTEST_ADD_SIMPLE(stat);
 UTEST_ADD_SIMPLE(fstat);
 
-#if 0
-/* TODO Kernel does not handle such cases yet. */
 UTEST_ADD_SIMPLE(exc_cop_unusable);
```
cahirwpz: We know these tests do not work ATM. Please do not enable them.

mips/intr.c
```diff
@@ -170,23 +170,24 @@ static void fpe_handler(exc_frame_t *frame) {
...
 static void cp_unusable_handler(exc_frame_t *frame) {
-  int cp_id = (frame->cause & CR_CEMASK) >> CR_CESHIFT;
-  bool kernel_mode = in_kernel_mode(frame);
-
-  if (cp_id != 1) {
-    panic(
-      "Unexpected unusable coprocessor exception, with coprocessor id = %d\n",
-      cp_id);
+  if (in_kernel_mode(frame)) {
+    panic("FPU unusable exception in kernel mode.");
```
cahirwpz: This is not accurate. Other coprocessors may not be usable in kernel-mode as well, if they're present.

Let's `assert(cp_id < 2)` to make sure no-one tries to use other coprocessors.

PR #431, opened 2018-05-11
Testing Floating Point, Unaligned Access exceptions and wrong syscall number behaviour.

dingorth:
* refactored   `test_exc_cop_unusable()`, `test_exc_reserved_instruction()`, `test_syscall_in_bds()`
* implemented tests: floating point exception, unaligned access, ENOSYS

### DIFF
--- a/user/utest/exceptions.c
+++ b/user/utest/exceptions.c
@@ -31,9 +31,7 @@ int test_exc_integer_overflow(void) {
 int test_exc_unaligned_access(void) {
   int a[2];
   int val;
-  asm volatile("lw %0, 1(%1)"
-               : "=r"(val)
-               : "r"(a));
+  asm volatile("lw %0, 1(%1)" : "=r"(val) : "r"(a));
 
   return 0;
 }
--- a/user/utest/exceptions.c
+++ b/user/utest/exceptions.c
@@ -5,6 +5,8 @@
 #include <sys/signal.h>
 #include <stdlib.h>
 
+#include <stdio.h>
+
 int test_exc_cop_unusable(void) {
   int value;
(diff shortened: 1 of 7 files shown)


user/utest/exceptions.c
```diff
@@ -5,46 +5,38 @@
...
    * The MIPS32® Instruction Set" */
   asm volatile(".long 0x00000039"); /* objdump fails to disassemble it */
+  return 0;
 }
 
-int test_exc_cop_unusable(void) {
-  return check_exception(exec_cp0_instr, SIGILL);
+int test_exc_fpe(void) {
+  int d = __INT_MAX__;
+  asm volatile("addi %0, %0, 1" : : "r"(d));
```
wmoczulsky: Causing floating point exception using integer overflow is a little bit unusual, and would not work on most of architectures (but works on MIPS). Maybe floating point division by zero is less confusing?

cahirwpz: Current version is ok. Following document: [signal.h](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html) lists following subclasses of `SIGFPE`:

 * `FPE_INTDIV` Integer divide by zero.
 * `FPE_INTOVF` Integer overflow.
 * `FPE_FLTDIV` Floating-point divide by zero.
 * `FPE_FLTOVF` Floating-point overflow.
 * `FPE_FLTUND` Floating-point underflow.
 * `FPE_FLTRES` Floating-point inexact result.
 * `FPE_FLTINV` Invalid floating-point operation.
 * `FPE_FLTSUB` Subscript out of range.

We don't send such an information to user-space yet.

PR #432, opened 2018-05-21
Simple pipe implementation

cahirwpz:
The implementation does not support signalling (SIGPIPE) nor read / write atomicity. Our pipes are bidirectional – same as in BSD family.

It's worth to mention that write end of pipe is owner of synchronization primitives and the buffer.


cahirwpz:
@pwit81 Please review `ringbuf` first. This is natural candidate to merge it with master sooner than `pipe` branch as whole.

cahirwpz:
@pwit81 Please note that `pipe_close` might be buggy, but this is not the only reason for this PR to fail tests. There's ongoing work on fixing synchronization of `proc` structure as well.

cahirwpz:
TODO (a) make closed producer end available for reading by consumer (b) possibly remove `closed` field.

PR #437, opened 2018-05-23
Resource Management

wmoczulsky:
We need good resource manager like other BSDs. Now for example ISA and PCI devices are using the same memory addresses.


include/rman.h
```diff
@@ -0,0 +1,35 @@
...
+#define _SYS_RMAN_H_
+
+#include "bus.h"
+
+typedef size_t rman_addr;
+
+typedef struct rman rman_t;
+typedef struct rman_block rman_block_t;
+
+struct rman_block {
```
cahirwpz: I'm convinced `resource` should play the role of `rman_block`... shouldn't that be the case?

wmoczulsky:
Continued in #447

PR #439, opened 2018-05-25
pmap_setup now preserves calling thread's mapping and disables preemption.

j-piecuch:
`pmap_setup` had a bug which made a thread clear its own UPD when switched out at a certain point during initialization of a new pmap. It also didn't preserve the calling thread's UPD mapping, so  calling `pmap_new` had a side effect of actually **activating** the newly created pmap. This PR fixes both issues.


cahirwpz:
This looks ok and I'll merge this PR. However I have some doubts if nobody actively uses PD during interrupt handling. FTTB I cannot see such thing to happen.

PR #441, opened 2018-05-26
Detect kernel stack overflow in kern_exc_enter and panic when it happens.

j-piecuch:
I think that most of the intermittent test failures are caused by kernel stack overflows due to the timer interrupt interrupting its own handler. This PR introduces a check into `kern_exc_enter` so that such incidents can be caught and the kernel can safely panic with the space that's left on the stack.


j-piecuch:
You can see that the tests for this PR fail because of kstack overflow -- see the backtrace.

cahirwpz:
@j-piecuch Please merge the most recent `master` with this branch.

cahirwpz:
To reproduce the error put `break tlb_refill if $$bad == 0x80000000` into `.gdbinit.template`. With such setup kernel crashes quite consistently. However if [simulators.py:69](https://github.com/cahirwpz/mimiker/blob/master/launcher/simulators.py#L69) is changed from `shift=auto` to `shift=3` then the bug goes away under patched QEMU 2.12.0. Hence I reckon this is yet another time-measurement related issue but I might be wrong.

PR #442, opened 2018-05-26
Unit tests and another fix to ring buffer impl.

pwit81:
Corrected a bug and added some unit tests.


cahirwpz:
For the record: I backported `ringbuf` fix into `master`. Thanks for finding it!

jpszczolowski:
Why don't we have `ringbuf` tests in master branch?

jpszczolowski:
@pwit81 do we need `kmalloc` and `UIO_SINGLE_KERNEL` in these tests? I think that `ringbuf` tests should use `ringbuf` and as few other modules as possible. If we need a buffer, we can use a local array inside test function.

cahirwpz:
@pwit81 I'm somewhat confused why this PR introduces `ringbuf_alloc` and `ringbuf_destroy`.

jpszczolowski:
@pwit81 ok, I'm sorry, use of `UIO_SINGLE_KERNEL` in the tests is necessary. But I'm not sure whether we want to introduce `ringbuf_alloc`, `ringbuf_destroy` and `kmem_pool` inside `ringbuf` implementation and tests.
I've:
- changed implementation of produce and consume and added asserts (so that now these functions cannot be used wrong),
- moved each test into separate KTEST_ADD.

jpszczolowski:
@cahirwpz yes, exactly.

cahirwpz:
I think we can safely replace `ringbuf_alloc`, `ringbuf_destroy` with
```c
void ringbuf_init(ringbuf_t *rb, void *buf, size_t buflen) { ... }
```

jpszczolowski:
@cahirwpz in my opinion this PR is ready

PR #444, opened 2018-05-29
 Disable interrupts & don't generate exceptions when accessing TLB 

j-piecuch:
Depends on #443.
Diff: https://github.com/j-piecuch/mimiker/compare/special-sections...j-piecuch:tlb-intr-fix

Functions that access the TLB must not be interrupted or generate exceptions between setting necessary coprocessor 0 register and executing the instruction that consumes them.

### DIFF
--- a/debug/cpu.py
+++ b/debug/cpu.py
@@ -77,7 +77,7 @@ class TLB(object):
 
     @staticmethod
     def size():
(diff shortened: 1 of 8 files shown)


cahirwpz:
This patch seems to fix some random crashes in `master`. I'll merge it ASAP.

PR #445, opened 2018-06-02
Synchronize pmap operations using mutexes

j-piecuch:
- Add mutex to `pmap_t`
- Synchronize `pmap` operations using mutexes


cahirwpz:
LGTM

PR #446, opened 2018-06-03
vm_map: API rework & synchronization

j-piecuch:


sys/exec.c
```diff
@@ -179,10 +179,13 @@ int do_exec(const exec_args_t *args) {
         vm_addr_t end = roundup(ph->p_vaddr + ph->p_memsz, PAGESIZE);
         /* TODO: What if segments overlap? */
         /* Temporarily permissive protection. */
-        vm_map_entry_t *segment = vm_map_add_entry(
-          vmap, start, end, VM_PROT_READ | VM_PROT_WRITE | VM_PROT_EXEC);
-        /* Allocate pages backing this segment. */
-        segment->object = default_pager->pgr_alloc();
+        vm_map_entry_t *segment;
+        WITH_MTX_LOCK (&vmap->mtx) {
```
cahirwpz: Looking at the following code I'm almost convinced that `vm_map` interface is badly designed. IMHO forcing user to lock on `vm_map`'s is not necessary here. Ideally this code should look as follows:
```c
vm_object_t *obj = vm_object_alloc(VM_OBJ_DEFAULT);
vm_map_entry_t *entry = vm_map_insert(vmap, vobj, start, end,
                                      VM_PROT_READ | VM_PROT_WRITE | VM_PROT_EXEC);
```
With following procedure signatures:
```c
vm_object_t *vm_object_alloc(vm_object_type_t);
int vm_map_insert(vm_map_t *map, vm_object_t *object, 
                  vm_offset_t start, vm_offset_t end, vm_prot_t prot);
```

What do you think?

sys/mmap.c
```diff
@@ -38,7 +38,7 @@ vm_addr_t do_mmap(vm_addr_t addr, size_t length, vm_prot_t prot, int flags,
 
   vm_map_entry_t *entry;
 
-  WITH_RW_LOCK (&vmap->rwlock, RW_WRITER) {
+  WITH_MTX_LOCK (&vmap->mtx) {
```
cahirwpz: Could you devise more specialized tool like: `WITH_VMMAP_LOCKED(vmap)`?
Let's not use directly exposed fields of `vm_map_t` structure.

sys/vm_map.c
```diff
@@ -108,16 +108,23 @@ static void vm_map_remove_entry(vm_map_t *vm_map, vm_map_entry_t *entry) {
...
 
+/*
+ * Insert a vm_map_entry without a backing object into a vm_map
+ * at the specified boundaries, with specified protection attributes.
+ *
+ * Assumes the map is locked, and leaves it so.
+ */
 vm_map_entry_t *vm_map_add_entry(vm_map_t *map, vm_addr_t start, vm_addr_t end,
                                  vm_prot_t prot) {
+  assert(mtx_owned(&map->mtx));
```
cahirwpz: Maybe add `vm_map_add_entry_nolock` variant?

sys/exec.c
j-piecuch: I agree, the proposed interface is much like the one in FreeBSD. Should this change go in a separate PR?

cahirwpz: Yes, please do so.

cahirwpz: Please introduce both procedures in this PR.

cahirwpz:
@j-piecuch What is the progress with adding `vm_object_alloc` and `vm_map_insert`?

PR #447, opened 2018-06-06
bus_resource_alloc for devices & resource manager

dingorth:
Adds bus_resource_allocate_* interface to device drivers infrastructure which uses resource manager.


include/bus.h
```diff
@@ -131,6 +111,30 @@ static inline void bus_intr_teardown(device_t *dev, intr_handler_t *handler) {
   BUS_DRIVER(dev)->bus.intr_teardown(dev, handler);
 }
 
+/*! \brief Allocates resource of size \a size between \a start and
```
cahirwpz: `\brief` description must be a one-liner.

include/bus.h
```diff
@@ -131,6 +111,30 @@ static inline void bus_intr_teardown(device_t *dev, intr_handler_t *handler) {
...
+ * \param flags RF_* flags defined in rman.h
+ */
+static inline resource_t *bus_resource_alloc(device_t *dev, int type, int rid,
+                                             rman_addr_t start, rman_addr_t end,
+                                             size_t size, unsigned flags) {
+  return BUS_DRIVER(dev)->bus.resource_alloc(dev->parent, dev, type, rid, start,
+                                             end, size, flags);
+}
+
+/*! \brief Same as bus_resource_alloc but resource placement in memory is
```
cahirwpz: ditto

include/mips/malta.h
```diff
@@ -61,14 +61,21 @@ void platform_init(int argc, char **argv, char **envp, unsigned memsize);
...
 #define MALTA_PCI0_IO_BASE 0x18000000
 #define MALTA_PCI0_IO_END 0x1bdfffff
+#define MALTA_PCI0_TO_ISA_BRIDGE_BASE MALTA_PCI0_IO_BASE
+#define MALTA_PCI0_TO_ISA_BRIDGE_END (MALTA_PCI0_TO_ISA_BRIDGE_BASE + 0x100 - 1)
+#define MALTA_PCI0_EXCLUSIVE_IO_BASE (MALTA_PCI0_IO_BASE + 0x100)
+#define MALTA_PCI0_EXCLUSIVE_IO_END MALTA_PCI0_IO_END
 #define MALTA_CORECTRL_BASE 0x1be00000
+#define MALTA_CORECTRL_END (0x1e000000 - 1)
 #define MALTA_FPGA_BASE 0x1f000000
+#define MALTA_PHYS_ADDR_SPACE_END 0x7fffffff
```
cahirwpz: Should be 0x3FFFFFFF as MIPS32 processor can directly access at most 512MB

include/mips/malta.h
```diff
@@ -61,14 +61,21 @@ void platform_init(int argc, char **argv, char **envp, unsigned memsize);
...
 #define MALTA_PCI0_MEMORY_BASE 0x10000000
 #define MALTA_PCI0_MEMORY_END 0x17ffffff
 #define MALTA_PCI0_IO_BASE 0x18000000
 #define MALTA_PCI0_IO_END 0x1bdfffff
+#define MALTA_PCI0_TO_ISA_BRIDGE_BASE MALTA_PCI0_IO_BASE
+#define MALTA_PCI0_TO_ISA_BRIDGE_END (MALTA_PCI0_TO_ISA_BRIDGE_BASE + 0x100 - 1)
+#define MALTA_PCI0_EXCLUSIVE_IO_BASE (MALTA_PCI0_IO_BASE + 0x100)
+#define MALTA_PCI0_EXCLUSIVE_IO_END MALTA_PCI0_IO_END
 #define MALTA_CORECTRL_BASE 0x1be00000
+#define MALTA_CORECTRL_END (0x1e000000 - 1)
```
cahirwpz: Why it's not in computed form, while others are?

include/rman.h
```diff
@@ -0,0 +1,72 @@
...
+
+struct rman {
+  mtx_t rm_mtx;                 /* protects all fields of resource manager */
+  rman_addr_t rm_start;         /* first physical address */
+  rman_addr_t rm_end;           /* last physical adress */
+  resource_list_t rm_resources; /* all managed resources */
+};
+
+/* returns null if unable to allocate */
+resource_t *rman_allocate_resource(rman_t *rm, rman_addr_t start,
```
cahirwpz: Missing documentation.

include/rman.h
```diff
@@ -0,0 +1,72 @@
...
+                                   rman_addr_t end, size_t count, size_t align,
+                                   unsigned flags);
+
+static inline resource_t *rman_allocate_resource_anywhere(rman_t *rm,
```
cahirwpz: ditto

include/rman.h
```diff
@@ -0,0 +1,72 @@
...
+                                                          size_t count,
+                                                          size_t align,
+                                                          unsigned flags) {
+  return rman_allocate_resource(rm, 0, RMAN_ADDR_MAX, count, align, flags);
+}
+
+void rman_create(rman_t *rm, rman_addr_t start, rman_addr_t end);
```
cahirwpz: ditto

include/rman.h
```diff
@@ -0,0 +1,72 @@
...
+
+static inline void rman_create_from_resource(rman_t *rm, resource_t *res) {
+  rman_create(rm, res->r_start, res->r_end);
```
cahirwpz: Hmm... so after creating a resource manager from resource, the resource itself does not get marked as allocated? Can we create more than one rman from single resource?

mips/rootdev.c
```diff
@@ -22,28 +28,97 @@ static void rootdev_intr_teardown(device_t *dev, intr_handler_t *handler) {
...
+                                          int type, int rid, rman_addr_t start,
+                                          rman_addr_t end, size_t size,
+                                          unsigned flags) {
+
+  resource_t *r =
+    rman_allocate_resource(&rm_mem, start, end, size, size, RF_NONE);
+  if (r == NULL)
+    return NULL;
+
+  r->r_owner = child;
```
cahirwpz: That is not very different from some lines in `gt_pci_resource_alloc`. Shouldn't that be extracted to `bus_generic_resource_alloc`?

sys/drv_stdvga.c
```diff
@@ -170,9 +171,13 @@ static int stdvga_attach(device_t *dev) {
   pci_write_config(dev, PCIR_COMMAND, 2, command);
 
   stdvga_state_t *stdvga = dev->state;
-  /* TODO: This will get replaced by bus_alloc_resource* function */
-  stdvga->mem = &pcid->bar[0];
-  stdvga->io = &pcid->bar[1];
+
+  stdvga->mem = bus_resource_alloc_anywhere(
+    dev, pcid->bar_info[0].type, PCIR_BAR(pcid->bar_info[0].rid),
```
cahirwpz: Using `pcid->bar_info[0].type` here is cheating. In this case the driver should explicitely request BAR[0] as memory resource, and BAR[2] (sic!) as I/O space.

sys/rman.c
```diff
@@ -0,0 +1,114 @@
+#include <rman.h>
+#include <stdc.h>
+#include <pool.h>
+
+static pool_t P_RMAN;
+
+static resource_t *find_resource(rman_t *rm, rman_addr_t start, rman_addr_t end,
```
cahirwpz: Missing documentation.

sys/rman.c
```diff
@@ -0,0 +1,114 @@
...
+      continue;
+
+    if (len >= (rman_addr_t)count)
+      return resource;
+  }
+
+  return NULL;
+}
+
+/* !\brief ???
```
cahirwpz: Incomplete documentation.

sys/rman.c
```diff
@@ -0,0 +1,114 @@
...
+  right_res->r_resources.le_prev = &left_res;
+
+  right_res->r_end = left_res->r_end;
+  right_res->r_start = where;
+  left_res->r_end = where - 1;
+
+  return right_res;
+}
+
+/* !\brief ???
```
cahirwpz: ditto

include/rman.h
```diff
@@ -0,0 +1,72 @@
...
+  rman_addr_t r_end; /* last (inclusive) physical address of the resource */
+  unsigned r_type;   /* one of RT_* values (possibly or'ed with RT_ISA_F) */
+  unsigned r_flags;  /* or'ed RF_* values */
+  size_t r_align;    /* alignment requirements for starting physical address */
+  int r_id;          /* (optional) resource identifier */
+  LIST_ENTRY(resource) r_resources; /* link on resource manager list */
+  LIST_ENTRY(resource) r_device; /* link on resources list assigned to device */
+};
+
+struct rman {
```
cahirwpz: Should not `rman` only manage resources of a single type? If so, I'd expect the type to be encoded in this structure.

mips/rootdev.c
dingorth: Did  you mean `bus_resource_alloc` by `bus_generic_resource_alloc`, or some new function?

sys/drv_stdvga.c
dingorth: Why is it cheating? `bar_info` belongs to this device. Are you suggesting that device should pass only bar number to `bus_resource_alloc` and bus driver should get resource info from `bar_info` ?

include/mips/malta.h
dingorth: For no particular reason. I'll replace it with 0x1dffffff.

sys/drv_stdvga.c
dingorth: I agree that it is ugly. This driver should rather search `bar_info` array for non-empty bars.

cahirwpz: BAR numbers used by a driver should not depend on particular method of BAR discovery. If a device facilitate `BAR[0]` and `BAR[2]` one should be able to just pass `rid=0` and `rid=2` (respectively) to `bus_resource_alloc*`.

I'm expecting following two lines to look as follows:
```c
stdvga->mem = bus_resource_alloc_anywhere(dev, RT_MEMORY, 0, ~0, RT_PREFETCHABLE);
stdvga->io = bus_resource_alloc_anywhere(dev, RT_IOPORTS, 2, ~0, 0);
```

mips/rootdev.c
cahirwpz: A new function that contains a common piece of code used by various implementations of `bus_resource_alloc` in bus drivers. Refer to [bus_generic_probe](https://github.com/cahirwpz/mimiker/blob/master/sys/device.c#L49).

include/rman.h
wmoczulsky: > so after creating a resource manager from resource, the resource itself does not get marked as allocated?
This function is meant to be used outside rman code. Every resource returned from rman is set as allocated. RF_ALLOCATED flag is only for internal use. I added a comment.

> Can we create more than one rman from single resource?
Not using this function.
I think that this one-liner function body shows clearly what it does.

wmoczulsky: I meant, this function should not be used twice on the same resource. 

I can add a flag to resource_t telling, that it was already used for rman, in order to create an assertion. But is it really needed, or just unnecessary complication?

include/rman.h
wmoczulsky: @dingorth

sys/rman.c
wmoczulsky: I think that purpose of rman_create and find_resource are obvious, and comment would not tell much more, than signature tells. I've added comments to other functions.

include/rman.h
dingorth: Yup, I think you're right. It would be weird to manage different types of resources in a single resource manager.

sys/rman.c
cahirwpz: Ok.

PR #448, opened 2018-06-07
added flag PF_ZERO

wmoczulsky:
Having flag which tells pool allocator to returned memory with zeros, is simpler and more elegant way, than clearing data structure fields one by one in constructor.


cahirwpz:
During conversation we came to conclusion that this PR is wrongly motivated.
That was most likely caused by missing documentation. @wmoczulsky will add
doxygen-style documentation that explains purpose of this allocator and its API.

PR #451, opened 2018-06-10
Fix and enable callout tests

melouber:
* `bzero` local `callout_t` variables (there was a problem with non-zero initial `c_flags` flag within `callout_t` variables, which is checked in `_callout_setup` function) in callout tests and sleepq test
* disable interrupts instead of preemption (race conditions)
* remove one critical section as callouts already run with interrupts disabled


melouber:
These changes have been included in #451

melouber:
Closed by accident.

cahirwpz:
LGTM.

PR #452, opened 2018-06-10
Fix wrong callout creation in sleepq test

melouber:
`bzero` local `callout_t` variable as described in #451


melouber:
These changes have been included in #451

PR #455, opened 2018-06-10
Add PT access procedures & simplify tlb_refill

j-piecuch:
- Add `pmap_pte_read` and `pmap_pte_write` which don't generate TLB misses on PT access.
- Remove code handling PT accesses from `tlb_refill`

### DIFF
--- a/mips/pmap.c
+++ b/mips/pmap.c
@@ -145,6 +145,56 @@ void pmap_delete(pmap_t *pmap) {
   kfree(M_PMAP, pmap);
 }
 
+/*! \brief Inserts the TLB entry mapping \a vaddr into the TLB. */
+static inline void pmap_ensure_safe_pt_access(pmap_t *pmap, vm_addr_t vaddr) {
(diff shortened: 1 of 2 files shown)


mips/pmap.c
```diff
@@ -145,11 +145,61 @@ void pmap_delete(pmap_t *pmap) {
...
+static pte_t pmap_pte_read(pmap_t *pmap, vm_addr_t vaddr) {
+  if (!is_valid(PDE_OF(pmap, vaddr)))
+    return 0;
+  /* Interrupt handlers could generate TLB misses. */
+  SCOPED_INTR_DISABLED();
+  /*
+   * ptep can't be read from the stack after returning from
+   * pmap_ensure_safe_pt_access, as that could generate a TLB miss.
+   */
+  register pte_t *ptep = &PTE_OF(pmap, vaddr);
```
cahirwpz: AFAIK `register` qualifier can be ignored by compiler. So this shows the intent but may not match the reality.

PR #456, opened 2018-06-15
Implemented callout_drain & tweaked callout tests

melouber:
* implemented `callout_drain`
* tweaked callout tests
* added `callout_drain` test


sys/callout.c
```diff
@@ -142,4 +146,20 @@ void callout_process(systime_t time) {
   ci.last = time;
 }
 
+bool callout_drain(callout_t *handle) {
+  /* A callout may be in active state only in callout_process,
+   * which is called in bottom half (with interrupts disabled),
```
melouber: I wanted to have interrupts disabled for entire `callout_drain` function, but I don't know how to do it.
* We can't take `SCOPED_SPINLOCK(&ci.lock);` at the beginning because then we will go to sleep with spinlock acquired (another thread will try to take this spinlock and see that its owner `!= NULL` and assert `spinlock.c:24` fails).
* We can't put all the code inside `WITH_INTR_DISABLED { ... }` because it causes compiler error `control reaches end of non-void function`.

As for now, the function behaviour is correct (at least I think so) but I thought that disabling interrupts for entire function would be more clean.

cahirwpz: You've found an interesting problem. My first (and probably wrong) solution would be to extend `sleepq_wait` to accept optional `spinlock_t *` argument and then unlock it while holding `td_spin` and acquire on return. That would make it similar to `cv_wait`, but handle top-bottom half synchronization. **Do not** implement it, though. That needs more careful thinking that I can give it right now.

tests/callout.c
```diff
@@ -57,35 +53,39 @@ static int test_callout_simple(void) {
...
+   * interrupts disabled). */
   assert(intr_disabled());
-  /* This incrementation is safe as callouts run with interrupts disabled. */
-  current++;
 
-  if (current == 10)
-    sleepq_signal(callout_ordered);
+  int ord = (int)arg;
+  if (ord > 0)
+    assert(processed[ord - 1]);
```
ilikeheaps: Actually I think we could make a better assert: keep the `current` variable and `assert(ord == curr)` (+-1 depending on  the order of instructions).

melouber: Done.

sys/callout.c
```diff
@@ -130,6 +133,7 @@ void callout_process(systime_t time) {
     TAILQ_FOREACH_SAFE(elem, &detached, c_link, next) {
       TAILQ_REMOVE(&detached, elem, c_link);
       elem->c_func(elem->c_arg);
+      sleepq_broadcast(elem);
```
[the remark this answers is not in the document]
melouber: Done.

cahirwpz:
LGTM

PR #458, opened 2018-06-19
Update ISA drivers to allocate port ranges

cahirwpz:
@dingorth @wmoczulsky Please fix recently introduced code to handle use case that was introduced with this PR, i.e. I want to allocate ISA port ranges.


dingorth:
Ok, I know what causes that.
Firstly, `start`, and `end` require absolute addresses, not offsets within specific `rman`.

Secondly, `atkbdc_probe` allocates resource:
`  resource_t *regs = bus_resource_alloc(dev, RT_ISA, 0, IO_KBD + ISA_BEGIN,
                                        ISA_BEGIN + IO_KBD + IO_KBDSIZE - 1, IO_KBDSIZE, 0);`

This resource is not deallocated! It should be, but resource deallocation is not implemented.
Then when we are trying to allocate this resource for the second time in `atkbdc_attach`,

`  atkbdc->regs = bus_resource_alloc(dev, RT_ISA, 0, IO_KBD + ISA_BEGIN,
                                    ISA_BEGIN + IO_KBD + IO_KBDSIZE - 1, IO_KBDSIZE, 0);`

we can't have it because it was already allocated and it's not shareable (because shareable resources are not implemented yet). 

There is more...
All resources emerge from main resource manager which is located in `rootdev` and it manages resources of type RT_MEMORY. This implies that ALL resources in our system are of type RT_MEMORY! 

I've hotfixed it, but it looks kind of ugly.

PR #459, opened 2018-07-01
Switch out when waking a higher priority thread from sleepq

ilikeheaps:
This PR adds `sched_maybe_switch(void)` function which does the switch if a higher priority thread has been added or current thread's time slice is exhausted (=== when `TDF_NEEDSWITCH` is set). That function is used to obtain desired functionality from `sleepq_signal` and `sleepq_broadcast`.


include/sched.h
```diff
@@ -91,6 +91,13 @@ void sched_clock(void);
  */
 void sched_switch(void);
 
+/*! \brief Switch out to another thread if you shouldn't be running anymore.
+ *
+ * This function will switch if your time slice expired or a thread with higher
```
cahirwpz: `higher` twice?

cahirwpz: Perhaps you should add a note here, that this function does not detect if a thread should switch, but rather checks it and executes if needed.

sys/sched.c
```diff
@@ -204,20 +204,24 @@ void preempt_disable(void) {
   td->td_pdnest++;
 }
 
+void sched_maybe_switch(void) {
+  thread_t *td = thread_self();
+  if (!preempt_disabled())
```
cahirwpz:
```c
if (preempt_disabled())
  return;
```
General flow in well-coded C routine is (1) filter out edge cases (2) solve the problem.

include/sched.h
ilikeheaps: Done.

sys/sched.c
ilikeheaps: Done.

PR #460, opened 2018-07-01
Timed sleep

ilikeheaps:
This PR is continuation of #450.

We need timed sleep to play Doom. This PR adds timed sleep.


sys/sleepq.c
```diff
@@ -294,17 +308,44 @@ bool sleepq_broadcast(void *wchan) {
...
   bool aborted = false;
 
   if (sq != NULL) {
-    aborted = sq_wakeup(td, sc, sq, SQ_ABORT);
+    aborted = sq_wakeup(td, sc, sq, reason);
     sq_release(sq);
   }
   sc_release(sc);
 
+  // TODO maybe move this out to sleepq_abort?
```
cahirwpz: Could you elaborate? It's not obvious at first sight.

sys/sleepq.c
```diff
@@ -294,17 +308,44 @@ bool sleepq_broadcast(void *wchan) {
...
+
+bool sleepq_abort(thread_t *td) {
+  return sleepq_abort_reason(td, SQ_ABORT);
+}
+
+static void sq_timeout(thread_t *td) {
+  sleepq_abort_reason(td, SQ_TIMEOUT);
+}
+
+// TODO will timeout_ms == 0 break this?
```
cahirwpz: Maybe assert it greater than 0?

tests/sleepq_timed.c
```diff
@@ -0,0 +1,85 @@
...
+  systime_t after_sleep = getsystime();
+  systime_t diff = after_sleep - before_sleep;
+
+  if (status == SQ_TIMEOUT) {
+    timed_received++;
+    assert(diff >= SLEEP_TIME_MS);
+  } else if (status == SQ_NORMAL) {
+    signaled_received++;
+  } else {
+    assert(false);
```
cahirwpz: Do not use such things in the code. There's `panic` procedure for that, which takes a message string.

sys/sleepq.c
```diff
@@ -227,10 +238,11 @@ sq_wakeup_t _sleepq_wait(void *wchan, const void *waitpt, sq_wakeup_t sleep) {
 /* Remove a thread from the sleep queue and resume it. */
 static bool sq_wakeup(thread_t *td, sleepq_chain_t *sc, sleepq_t *sq,
                       sq_wakeup_t wakeup) {
-  /* Only sq_enter sets TDF_SLPINTR flag and it holds the same sleepq spinlock
+  /* Only sq_enter sets TDF_SLP... flags and it holds the same sleepq spinlock
    * as sq_wakeup. Hence it's safe to read flags without holding thread's
    * spinlock. */
-  if (!(td->td_flags & TDF_SLPINTR) && (wakeup == SQ_ABORT)) {
+  if ((!(td->td_flags & TDF_SLPINTR) && (wakeup == SQ_ABORT)) ||
```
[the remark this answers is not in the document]
ilikeheaps: >Could we split it into two?

Surely. I think we could also make long booleans readable by moving the `||` operator from the end of line to the beginning of the next one.

>Remember we can abort a timed sleep (signal delivery before timeout).

The `TDF_SLPINTR` flag is set when mode is at least `SQ_ABORT` so that should work.

sys/sleepq.c
```diff
@@ -218,7 +223,13 @@ sq_wakeup_t _sleepq_wait(void *wchan, const void *waitpt, sq_wakeup_t sleep) {
     if (td->td_flags & TDF_SLPINTR) {
       td->td_flags &= ~TDF_SLPINTR;
       wakeup = SQ_ABORT;
+    } else if (td->td_flags & TDF_SLPTIMED) {
+      td->td_flags &= ~TDF_SLPTIMED;
+      wakeup = SQ_TIMEOUT;
     }
+
+    assert(!(td->td_flags & TDF_SLPINTR));
```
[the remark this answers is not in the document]
ilikeheaps: Actually this checks that only one flag was set after wake-up (as we only remove one flag there). Should I remove it after all?

sys/sleepq.c
ilikeheaps: `sq_timeout` is called during `callout_process` with interrupts disabled so `sched_maybe_preempt` will immediately return. It seems a bit weird to even call it there when the callout is going to end right after this call.

sys/sleepq.c
ilikeheaps: Done.

tests/sleepq_timed.c
ilikeheaps: Done.

sys/sleepq.c
```diff
@@ -294,17 +308,44 @@ bool sleepq_broadcast(void *wchan) {
   return true;
 }
 
-bool sleepq_abort(thread_t *td) {
+// TODO rename this (and possibly `sleepq_abort`)
+static bool sleepq_abort_reason(thread_t *td, sq_wakeup_t reason) {
```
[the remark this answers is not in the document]
ilikeheaps: Done.

sys/sleepq.c
cahirwpz: Generally speaking call to `sched_maybe_preempt` is no-op while interrupts (fixed that) or preemption is disabled.

Procedures `sleep_signal`, `sleepq_broadcast` and so on, may be safely used in interrupt context. I'm quite sure eventually I'll expect callout actions to be executed in thread context, so call to `sched_maybe_preempt` seems ok here.

sys/sched.c
```diff
@@ -208,7 +208,7 @@ void preempt_disable(void) {
 }
 
 void sched_maybe_preempt(void) {
-  if (preempt_disabled())
+  if (intr_disabled() || preempt_disabled())
```
ilikeheaps: Actually `preempt_disabled()` already checks `intr_disabled()`

include/thread.h
```diff
@@ -59,6 +59,8 @@ typedef enum {
 #define TDF_BORROWING 0x00000010  /* priority propagation */
 #define TDF_SLEEPY 0x00000020     /* thread is about to go to sleep */
 #define TDF_SLPINTR 0x00000040    /* sleep is interruptible */
+#define TDF_SLPTIMED 0x00000080   /* sleep is timed (implies interruptible) */
+                                  /* (SLPINTR must be set as well) */
```
melouber: What about
```
#define TDF_SLPINTR  0x00000040
#define TDF_SLPTIMED 0x000000C0
```
?

PR #461, opened 2018-07-14
Minor improvements to `run_tests.py`

ilikeheaps:
1) print `(gdb)` prompt in the first line of interactive session
2) add `--commands [CMD ...]` argument that specifies gdb commands to run upon failure. Default commands are left unchanged (`klog`, `info registers`, `backtrace`, `list`, `kthread`).

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -116,8 +116,9 @@ if __name__ == '__main__':
     parser.add_argument('--non-interactive', action='store_true',
                         help='Do not run gdb session if tests fail.')
     parser.add_argument('--commands', nargs='*', metavar='CMD',
-                        help='Commands to automatically run in gdb session when a test fails',
-                        default=['klog', 'info registers', 'backtrace', 'list', 'kthread'])
+                        help='Commands to run in gdb session upon failure',
+                        default=['klog', 'info registers', 'backtrace',
+                                 'list', 'kthread'])
 
     try:
         args = parser.parse_args()
--- a/run_tests.py
+++ b/run_tests.py
@@ -35,7 +35,7 @@ def send_command(gdb, cmd):
 
 
 # Tries to start gdb in order to investigate kernel state on deadlock or crash.
-def gdb_inspect(interactive):
+def gdb_inspect(interactive, commands):
     gdb_port = GDB_PORT_BASE + os.getuid()
     gdb_command = 'mipsel-mimiker-elf-gdb'
     # Note: These options are different than .gdbinit.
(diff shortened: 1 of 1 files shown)


cahirwpz:
I understand the purpose of this change, however I'm not in favour of solving it this way.

After all we want interactive mode of `run_tests.py` to be friendly towards a programmer, which can be achieved by not flooding him with possibly irrelevant messages, but offering useful help.

run_tests.py
```diff
@@ -46,16 +46,15 @@ def gdb_inspect(interactive):
...
-    send_command(gdb, 'info registers')
-    send_command(gdb, 'backtrace')
-    send_command(gdb, 'list')
-    send_command(gdb, 'kthread')
+    for cmd in commands:
+        send_command(gdb, cmd)
+
     if interactive:
+        print('(gdb)', end='', flush=True)
         gdb.interact()
```
cahirwpz: I think it's much better to put all comands from removed lines 49-53 into `else` clause here.

run_tests.py
```diff
@@ -46,16 +46,15 @@ def gdb_inspect(interactive):
...
-    send_command(gdb, 'klog')
-    send_command(gdb, 'info registers')
-    send_command(gdb, 'backtrace')
-    send_command(gdb, 'list')
-    send_command(gdb, 'kthread')
+    for cmd in commands:
+        send_command(gdb, cmd)
+
     if interactive:
+        print('(gdb)', end='', flush=True)
```
cahirwpz: By default in interactive mode I'd print `backtrace` – that is always a starting point to understanding what went wrong.

But before that I'd like to see a listing of most useful commands related to debugging the kernel. I'd assume a user is familiar with _GNU gdb_, but not necessarily with our extensions to the command set like: `klog`, `cpu tlb`, `kthread`, `kdump` and so on...

run_tests.py
```diff
@@ -116,6 +115,10 @@ def test_seed(seed, interactive=True, repeat=1, retry=0):
                         help='Keep testing until some error is found.')
     parser.add_argument('--non-interactive', action='store_true',
                         help='Do not run gdb session if tests fail.')
+    parser.add_argument('--commands', nargs='*', metavar='CMD',
```
cahirwpz: After a while noone will remember another option to `run_tests.py`. It's much better to provide good defaults.

cahirwpz:
Please have a look at https://github.com/cahirwpz/mimiker/commit/44deb19756274e6f96da392dd337e17c04cfc880

ilikeheaps:
Alright. I suppose this PR doesn't bring any more value.

PR #463, opened 2018-07-28
Implementation of environment variables management.

pwit81:
### The implementation must meet following criteria:

User programs must accept another argument `envp`, so that `main` signature is:
```c
int main(int argc, char **argv, char **envp);
```
User programs must be able to access environment variables directly:
```c
extern char **environ;
```
...or through set of functions (should be implemented by _newlib_):
```c
char *getenv(const char *name);
int setenv(const char *name, const char *value, int overwrite);
int putenv(char *string);
int unsetenv(const char *name);
```

### The stack

It's shape is defined in _"System V ABI MIPS RISC Processor Supplement, 3rd edition"_, p. 30-35.
There's a useful example up there which is _"Figure 3-29: Example Process Stack"_.

### DIFF
--- a/include/exec.h
+++ b/include/exec.h
@@ -10,7 +10,6 @@ typedef struct exec_args {
    * starting process. */
   const char **argv;
   const char **envp;
-  /* TODO: Environment */
 } exec_args_t;
 
 int do_exec(const exec_args_t *prog);
--- a/include/exec.h
+++ b/include/exec.h
@@ -11,7 +11,7 @@ typedef struct exec_args {
   unsigned argc;
   const char **argv;
   unsigned envc;
-  const char **envp;  
+  const char **envp;
   /* TODO: Environment */
 } exec_args_t;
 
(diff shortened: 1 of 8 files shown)


include/exec.h
```diff
@@ -10,6 +10,8 @@ typedef struct exec_args {
    * starting process. */
   unsigned argc;
   const char **argv;
+  unsigned envc;
```
cahirwpz: FYI There's no such thing as `envc`.

cahirwpz: Ok. But please keep in mind this is an optimization. Because execve does not provide `argc` or `envc` a kernel developer may choose wrong values for those. In my view this is violation of [SPOT](https://en.wikipedia.org/wiki/Single_source_of_truth) principle. I'd prefer for a procedure to calculate the values instead of relying on user to provide good ones.

cahirwpz:
LGTM

PR #472, opened 2018-10-29
Move drivers to separate directory

mzr:
* moved drivers to `drv` directory
* adjusted `clean` directives in Makefiles

PR #474, opened 2018-11-01
Add private arguments to sysinit objects.

mzr:
Something like this would be required to specify only one generic driver init procedure.
Probably we will only need this when implementing dynamic devclasses initialization process, with every driver as a separate sysinit module.
This branch is fork of a #472


mzr:
Not needed. For now.

PR #476, opened 2018-11-06
Devclasses

mzr:


sys/bus.c
```diff
@@ -98,3 +98,55 @@ int bus_generic_probe(device_t *bus) {
   }
   return 0;
 }
+
+// identify hinted devices and resources.
+// add devices to parentbus
+// add hinted devices to devices ivars
+/*
+int bus_enumerate_hinted_children(device_t *bus){
```
Tomatosoup97: I would prefer to make a change with device hints in a separate PR - otherwise the PR contents 1) might be confusing - ideally one PR should introduce one concept, devclasses and device hints are separate ideas that does not interfere with each other *that much* 2) the less number of files PR touches the better, by joining these two ideas we might do a bunch of unrelated stuff - python scripts, FDT files along with devclasses structure

What do you think?

include/devclass.h
```diff
@@ -0,0 +1,26 @@
...
+  driver_t **drivers;
+};
+
+#define DEVCLASS_CREATE(dc_name)                                               \
+  SET_DECLARE(dc_name, driver_t);                                              \
+  devclass_t dc_name##_devclass = {.name = (#dc_name),                         \
+                                   .drivers = SET_BEGIN(dc_name)};             \
+  SET_ENTRY(devclasses, dc_name##_devclass);
+
+#define DEVCLASS_ADD(dc_name, driver_structure)                                \
```
Tomatosoup97: what about renaming it to DEVCLASS_ADD_DRV ? Having both DEVCLASS_CREATE and DEVCLASS_ADD is confusing, I imagine I would need to check in code each time to distuingush them

mzr: Good idea. I'll do that.

sys/bus.c
mzr: Let's discuss that face-to-face. It's complicated.

mzr:
Added `DEVCLASS_FOREACH` macro and implemented draft version of probe & attach proces.

mzr:
Now it is compatible with new directory structure.

include/linker_set.h
```diff
@@ -46,6 +46,9 @@
 #define SET_FOREACH(pvar, set)                                                 \
   for (pvar = SET_BEGIN(set); pvar < SET_LIMIT(set); pvar++)
 
+#define SET_FOREACH_PTR(pvar, set_b, set_e)                                    \
```
Tomatosoup97: `set_b`, `set_e` - set begin and set end ? Please make the naming more verbose so that we don't have to guess, preferably just `set_begin`, `set_end`. Also a short docstring about when I would want to use this one over `SET_FOREACH` could be helpful

mzr: done

PR #477, opened 2018-11-07
Improve docstrings for drivers infrastructure

Tomatosoup97:


include/bus.h
```diff
@@ -193,8 +196,7 @@ static inline resource_t *bus_alloc_resource_anywhere(device_t *dev,
 
 /*! \brief Allocates resource for a device.
  *
- * Basically the same as \sa bus_alloc_resource_anywhere, but resource
- * has to be identifiable by parent bus driver by \param rid.
+ * \sa bus_alloc_resource_anywhere using default resource size for the bus
```
mzr: Currently it's used only by std_vga driver. Numbers passed as `rid` are numbers of BAR registers. Parent bus has identified BARs for this device and can map them in PCI MEMORY space (seem `malta.h`), and then return as a resource to std_vga device.
`rid` is not passed to `rman_alloc_resource` it is only used by parent bus to find resource by id.

cahirwpz:
When can I expect the PR to leave WIP status?

drv/atkbdc.c
```diff
@@ -1,3 +1,5 @@
+/* atkbdc.c - AT keybord controller driver
```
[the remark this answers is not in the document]
Tomatosoup97: I saw that convention in one of the kernels, after rethinking it is dump indeed + creates additional overhead when renaming the file. Removing the names

include/bus.h
Tomatosoup97: I don't see how your comment is related to the actual change and the source code 

This is how the bus_alloc_resource_any looks like:
```c
static inline resource_t *bus_alloc_resource_any(device_t *dev, res_type_t type,
                                                 int rid, res_flags_t flags) {

  return BUS_DRIVER(dev)->bus.alloc_resource(dev->parent, dev, type, rid, 0,
                                             RMAN_ADDR_MAX, 1, flags);
}
```
It does the very same as thing as `bus_alloc_resource_anywhere` - with one difference - it sets the size to the default value equal to 1. I don't want to have a leaky abstraction here and comment that some internal rman is using the rid for certain action

include/bus.h
```diff
@@ -161,13 +161,16 @@ static inline void bus_intr_teardown(device_t *dev, intr_handler_t *handler) {
...
-/*! \brief Allocates resource of size \a size between \a start and \a end.
+/*! \brief Allocates resource for a device of size \a size between
+ * \a start and \a end.
  *
  * Should be called inside device's \fn attach function.
  *
  * \param dev device which needs resource
  * \param type resource type RT_* defined in rman.h
  * \param rid resource identifier as in \a resource_t structure
+ * \param start/end start/end addresses of the resource
```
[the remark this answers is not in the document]
Tomatosoup97: Ok, made it more clear what it means - please recheck now

include/bus.h
```diff
@@ -161,13 +161,17 @@ static inline void bus_intr_teardown(device_t *dev, intr_handler_t *handler) {
   BUS_DRIVER(dev)->bus.intr_teardown(dev, handler);
 }
 
-/*! \brief Allocates resource of size \a size between \a start and \a end.
+/*! \brief Allocates resource for a device of size \a size between
```
mzr: Is it the device of size `size` or the resource? It's confusing.

mips/gt64120.c
```diff
@@ -1,5 +1,6 @@
-/* Heavily inspired by FreeBSD / NetBSD `gt_pci.c` file. */
-
+/* GT64120 PCI bus driver
+ *
+ * Heavily inspired by FreeBSD / NetBSD `gt_pci.c` file. */
```
mzr: We could mention that this file is a merge of FreeBSDs `/sys/mips/malta/gt.c`, `/sys/mips/malta/gt_pci.c` and it is related in some way to `/sys/dev/pci/*`.

include/bus.h
mzr: Sometimes driver knows what resources it has and can ask for them using `rid`. It doesn't care what's the size of that resource. It that case the resource already exists and parentbus can look it up (using `rid`).
Probably the interface should be redesigned when we have devhints implemented.

Tomatosoup97:
After discussion with @mzr we are abandoing some of the changes in `include/bus.h` docstrings since the interface might change soon due to device hints implementation. The PR is read to review

cahirwpz:
@mzr I'm waiting for your review before I start.

include/bus.h
```diff
@@ -161,7 +161,7 @@ static inline void bus_intr_teardown(device_t *dev, intr_handler_t *handler) {
   BUS_DRIVER(dev)->bus.intr_teardown(dev, handler);
 }
 
-/*! \brief Allocates resource for a device of size \a size between
+/*! \brief Allocates for a device resource of size \a size between
```
mzr: IMO "Allocates resource of a type \a type and size \a size between \a start and \a end, for a device \a dev" sounds better.

Tomatosoup97: Agreed!

PR #481, opened 2018-11-21
Moving kmalloc from pm to vm

bgrm:


cahirwpz:
Any progress here?

cahirwpz:
@bgrm I disabled offending tests. We'll have to rethink them.

PR #483, opened 2018-11-27
Generate device hints array based on FDT conf file

Tomatosoup97:
Python script for generating static C array for device hints from FDT (Flat Device Tree) format

Obvious issues:
- FDT is exemplary and probably should be rethought


device_hints.c
```diff
@@ -0,0 +1,22 @@
...
+devhint_t hints[] = {
+{ 	.iomem_start = 0x3f8,
+	.irq = 0x4,
+	.iomem_end = 0x3ff,
+	.path = "/rootdev/pci@0/isab@0/isa@0/uart@0" },
+{ 	.iomem_start = 0x2f8,
+	.irq = 0x3,
+	.iomem_end = 0x2ff,
+	.path = "/rootdev/pci@0/isab@0/isa@0/uart@1" },
+
```
mzr: We should match the names of the devices / drivers with the `.path` string in some way.

mzr: We have to distinguish resources placed in IOMEM space and IOPORT space. See `include/mips/malta.h`.

cahirwpz:
@Tomatosoup97 I'd prefer you wrote more pythonic code, including use of list & dictionary comprehensions where possible.

Tomatosoup97:
@cahirwpz done, rewritten the code so that it uses more list comprehensions. There is one `map` left, but I think that in that case it is more natural to use it over list comprehension.

device_hints.c
Tomatosoup97: >We have to distinguish resources placed in IOMEM space and IOPORT
Valid note, done! 
>We should match the names of the devices / drivers with the .path string in some way.
Yes, we should, but I think that it is out of the scope of this PR

mzr:
Not compiling? @Tomatosoup97

Tomatosoup97:
@mzr it wasn't compiling b/c I have merged into master branch incorrectly, should work now

script/README.md
```diff
@@ -0,0 +1,26 @@
...
+    iomem - list of iomem ranges
+    ioport - list of ioport ranges
+
+For an exemplary file please take a look at `drv/device_hints.dts`.
+
+Generating generate device hints from DTS file:
+
+1. Install requirements
+
+    `pip install -r requirements.txt`
```
mzr: This command would not run on the server:
```
» which pip
pip not found
```

PR #485, opened 2018-12-05
Move callout execution to top-half

jpszczolowski:


cahirwpz:
I fixed handling exceptions in critical sections.
Now you get a nice message from the kernel:
```
KERNEL PANIC!!! 
TLB exception (load or instruction fetch) at $801095dc!
Caused by reference to $ffffe000!
HINT: Type 'info line *0x801095dc' into gdb to find faulty code line.
```
Unfortunately this is not the culprit, but rather a result of overwriting code with random data.
I belive the real culprit is in _sys/callout.c:57_.

PR #495, opened 2018-12-16
Add userspace tests for process groups

wmoczulsky:
As process groups don't exist yet, the tests were run only on linux for now
and this commit breaks compilation of mimiker.


cahirwpz:
These tests must at least compile in order to be merged with `master`.

I think the only way is to take system calls stubs from #475.

wmoczulsky:
> These tests must at least compile in order to be merged with `master`.
> 
> I think the only way is to take system calls stubs from #475.

In my opinion, we should have PR with stubs for those functions first, next PR for tests and then PR with implementation. First PR should be easy. I think, that we should work on API and internal implementation separately. Also implementation should take place after we are sure what API we want to have. In such easy feature this fragmentation is possible.

cahirwpz:
@wmoczulsky, @staffik Should I consider this PR as dead? There's completely no activity in here.

cahirwpz:
@staffik Why have you accepted a change that does not compile?

staffik:
@cahirwpz  I mean it is ok given that we have Process groups implementation merged to master too. I checked that with Process groups implementation it compiles and tests pass.

cahirwpz:
Nothing has changed here for last two months. I reckon this PR is dead.

PR #498, opened 2018-12-17
Move kernel switch context on kernel stack

cahirwpz:
Is there any benefit of doing that this way? If no, remove this branch!


cahirwpz:
Exception and interrupt context is always saved on stack. For consistency switch context should be saved on stack as well.

PR #512, opened 2019-01-03
Refactor threads' names

jpszczolowski:
My proposal of thread naming rules. Try to:

- not use `thread` word,
- prefix kernel test thread with `test`,
- prefix user test thread with `utest`,
- use format `[prefix-]module[-name[-number]]`, e.g. `callout`, `test-malloc`, `test-sleepq-waiter-2`.


jpszczolowski:
@cahirwpz done

PR #515, opened 2019-01-07
Device hints

Tomatosoup97:


cahirwpz:
I cannot see this PR moving the kernel into right direction with drivers infrastructure. Hence I decided to close it.

PR #521, opened 2019-01-15
Implement interrupt threads

jpszczolowski:


cahirwpz:
Hmm... apparently I missed the fact that C0_CAUSE.IP[7..0] bits are read-only. I suppose they get zero'ed after a read is performed, but I cannot find it in the documentation.

PR #524, opened 2019-01-21
Fix group id in Dockerfile

wmoczulsky:
`--ingroup` option allowed only one GID (https://manpages.debian.org/jessie/adduser/adduser.8.en.html)
so user was added only to `sudo` group

Previously files inside container were created with `sudo` group.

before:
```
_  mimiker git:(master) _ id
uid=1000(mimiker) gid=27(sudo) groups=27(sudo)
```
after:
```
_  mimiker git:(master) _ id
uid=1000(mimiker) gid=1000(mimiker) groups=1000(mimiker),27(sudo)

```


Tomatosoup97:
Thanks for spotting that! An alternative way would be to use `--add_extra_groups`, but your solution is also fine and avoids creating additional conf files. LGTM.

PR #525, opened 2019-01-21
Proper locking of proc_t

wmoczulsky:
In many places like exec.c:run_program() and fork.c:do_fork()  structure proc_t is not locked in accordance with annotations. I will try to fix this.


cahirwpz:
Changes introduced by this PR have little value and the PR is unlikely to be finished before semester ends.

wmoczulsky:
I changed locking strategy for proc_t. Now some fields are marked to use only in the same thread/proc, hence locking in most places is now unnecessary.

sys/fork.c
```diff
@@ -58,6 +58,8 @@ int do_fork(void) {
   memcpy(child->p_sigactions, parent->p_sigactions,
          sizeof(child->p_sigactions));
```
staffik: Is it ok that we access `child->p_uspace` / `parent->p_uspace` in https://github.com/cahirwpz/mimiker/blob/925c40d16a16e4465cd7d9c90441fce43478ffbd/sys/fork.c#L48 and `p_uspace` is marked `($) use only from the same process/thread` ?

wmoczulsky: This is why there is distinction between proc_create and proc_add. The process is invisible by other patrs of the system untill proc_add is called, so this is safe.

staffik: Is it safe to access  `parent->p_waitcv` here: https://github.com/cahirwpz/mimiker/blob/4e01cad7070f03f5571a09e0854a5b5c04d9b798/sys/proc.c#L175 ?

wmoczulsky: Yes, it is locked using global mutex in our implementation.

cahirwpz: It's always safe to call `cv_broadcast` or `cv_signal` even in interrupt context.  You must always call `cv_wait` with same mutex.

cahirwpz: In this case, it's safe as long as you're sure nobody will destroy `parent`.

PR #532, opened 2019-01-24
Define name and unit fields for a device and provide related functions

Tomatosoup97:
This change is necessary in order to search for a device in FDT. Names for the devices are based on `malta.dts` file.

* device_get_fullname
* device_construct_fullpath
* Provide tests for the above functions


cahirwpz:
I cannot see this PR moving the kernel into right direction with drivers infrastructure. Hence I decided to close it.

PR #534, opened 2019-01-28
Initial draft for device properties structure

Tomatosoup97:
Early draft of how the `devprops` (device properties) could look like. Any critique or improvement suggestion is welcome.

### DIFF
--- a/include/device.h
+++ b/include/device.h
@@ -35,35 +35,47 @@ typedef enum { DEV_BUS_NONE, DEV_BUS_PCI, DEV_BUS_ISA } device_bus_t;
 
 // TODO: There should be more key types
 typedef enum {
-    COMPATIBLE,
-    FDT_PATH,
-    VENDOR
+  CLASSCODE,
+  COMPATIBLE,
+  DEVICEID,
+  FDT_PATH,
+  VENDORID,
 } devprop_attr_key_t;
 
 // TODO: There should be more key types
 typedef enum {
-    IOPORT, // TODO: How to handle IOPORT1 and IOPORT2 ?
-    IRQ
+  BAR,
+  IOPORT, // TODO: How to handle IOPORT1 and IOPORT2 ?
+  IRQPIN,
+  IRQLINE,
 } devprop_res_key_t;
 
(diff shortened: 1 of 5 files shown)


include/device.h
```diff
@@ -41,15 +74,20 @@ struct device {
   /* Device information and state. */
   device_bus_t bus;
   driver_t *driver;
+  // TODO: most likeley we want to get rid of `instance` field
   void *instance; /* used by bus driver to store data in children */
   void *state;    /* memory requested by driver for its state*/
+  device_attr_t attributes[DEV_ATTR_MAX]; /* equivalent of FreeBSD's `ivars` */
```
mzr: Now that comment is not exactly true and I think it's more confusing than helpful.

include/device.h
```diff
@@ -30,6 +30,39 @@ struct driver {
...
+  DA_FDT_PATH,
+  /* resource types */
+  DA_IOPORT,
+  DA_MEMORY,
+  DA_IRQ,
+  /* PCI bus identifiers */
+  DA_PCI_CLASSCODE,
+  DA_PCI_DEVICEID,
+  DA_PCI_VENDORID,
+} __packed da_tag_t;
```
mzr: I'm not sure about this one, but is it possible that we may encounter unaligned access when using `__packed`?

cahirwpz: Packed unions will take the least space that can accommodate all members => in this case `char` will do. Normally it'd consume space for `int`.

tests/device.c
```diff
@@ -0,0 +1,61 @@
...
+#include <condvar.h>
+#include <ringbuf.h>
+#include <pci.h>
+#include <dev/isareg.h>
+#include <dev/ns16550reg.h>
+#include <interrupt.h>
+#include <sysinit.h>
+
+#if 0
+extern device_t *gt_pci;
```
mzr: Probably we should get reference to this device in some other, dynamic way.

cahirwpz:
This PR got heavily outdated. There's too little value in it to be worth spending time to update it to current code base. Hence, I decided to close it.

PR #526, opened 2019-01-21
Port docker to new launch

Tomatosoup97:
Related issue: https://github.com/cahirwpz/mimiker/issues/523

PR #537, opened 2019-01-30
Change Dockerfile's FROM version to :latest

jpszczolowski:
We should use `latest` version instead of `1.6` (which is out-of-date anyway). We have `latest` tag on the Docker Hubsince https://github.com/cahirwpz/mimiker/commit/11dafa31455d15aa4ad8f21a947597922377bc96.


Tomatosoup97:
@jpszczolowski please see https://github.com/cahirwpz/mimiker/pull/526, where this change is already done

jpszczolowski:
Ok, I didn't know that this change is inside #526 
But I still think that this PR is worth merging now as the tag change is not at all related to #526.

PR #540, opened 2019-02-02
Enable sbrk with negative args + munmap in simple variant

bgrm:
issue 393


cahirwpz:
Please add tests for `munmap` in `bin/utest/mmap.c`.

PR #541, opened 2019-02-08
Fix exception handling when interrupts are disabled

jpszczolowski:


mips/intr.c
```diff
@@ -275,27 +301,41 @@ void mips_exc_handler(exc_frame_t *frame) {
...
-  if (code == EXC_INTR)
-    intr_enable();
+  if (!user_mode && (!(frame->sr & SR_IE) || preempt_disabled())) {
+    /* Case 2b & 2c: we came from kernel-space, interrupts or preemption were
+     * disabled, so switching out is forbidden! */
+    PCPU_SET(no_switch, true);
+    (*handler)(frame);
+    PCPU_SET(no_switch, false);
+  } else {
+    /* Case 1 & 2: we came from user-space or kernel-space,
```
jpszczolowski:
```suggestion
    /* Case 1 & 2a: we came from user-space or kernel-space,
```

mips/intr.c
```diff
@@ -263,7 +265,31 @@ void kstack_overflow_handler(exc_frame_t *frame) {
...
-/* General exception handler is called with interrupts disabled. */
+/* Let's consider possible contexts that caused exception/interrupt:
+ *
+ * 1. We came from user-space:
+ *    interrupts and preemption must have been enabled
+ * 2. We came from kernel-space:
+ *    a. interrupts and preemption are enabled:
+ *       kernel was running in regular thread context
+ *    b. preemption is disabled, interrupts are enabled:
+ *       kernel was running in thread context and acquired sleep lock (mutex)
```
jpszczolowski: Case 2b. Being an owner of a mutex is not related to having preemption disabled IMO. The description should be something like: `kernel was running in thread context and disabled preemption`.

mips/intr.c
```diff
@@ -275,27 +301,41 @@ void mips_exc_handler(exc_frame_t *frame) {
...
-    intr_enable();
-
-  (*handler)(frame);
-
-  /* From now on till the end of this procedure interrupts are enabled. */
-  if (code == EXC_INTR)
-    intr_enable();
+  if (!user_mode && (!(frame->sr & SR_IE) || preempt_disabled())) {
+    /* Case 2b & 2c: we came from kernel-space, interrupts or preemption were
+     * disabled, so switching out is forbidden! */
```
jpszczolowski: If this is case 2b, we should enable interrupts if this is exception handling. Currently case 2b is handled always with interrupts disabled.

sys/sched.c
```diff
@@ -159,6 +159,9 @@ long sched_switch(void) {
   /* make sure we reacquire td_spin lock on return to current context */
   td->td_flags |= TDF_NEEDLOCK;
 
+  if (PCPU_GET(no_switch))
+    panic("Switching context while interrupts are disabled is forbidden!");
+
```
jpszczolowski: IMO this will never fail. Currently we set `no_switch` to true only when interrupts are disabled. And when interrupts are disabled, and `mips_exc_handler` will be called again (due to exception), there is no chance that we will context-switch as there is no `on_exc_leave` call in case 2b & 2c.

mips/intr.c
cahirwpz: Right, but that in most cases this is a consequence of acquiring a sleep lock. I'll change it to inform reader that context is somewhat wider that implied by the comment.

jpszczolowski: > Right, but that in most cases this is a consequence of acquiring a sleep lock.

How is that? We disable preemption in `sys/mutex.c` only for the time of execution of `turnstile` functions.
```
WITH_MTX_LOCK(&some_mtx) {
 // preemption is enabled here
}
```

sys/sched.c
cahirwpz: Please note that someone, by accident, may attempt to use sleep mutex in:
1) Interrupt filter code
2) Exception handler code (esp. `tlb_exception_handler`)

We don't know if it's really the case, but it's better to be safe than sorry. AFAIR I've seen similar check in _FreeBSD_ kernel.

jpszczolowski: Yes, alright, I haven't thought about calling `sched_switch` from turnstile or sleepqueue.

mips/intr.c
cahirwpz: Ok. Let's consider following code from `mtx_lock`.

```c
WITH_NO_PREEMPTION {
    while (m->m_owner != NULL)
        turnstile_wait(m, (thread_t *)m->m_owner, waitpt); 
    m->m_owner = thread_self();
    m->m_lockpt = waitpt;
}
```

It definitely may get interrupted or an exception may happen somewhere inside. If we during exception / interrupt handling we allow to switch out to another code... boom – we broke the promise. Obviously switch out will not happen as a result of calling `on_exc_leave` – that's ruled out – but IMO still a programmer may call implicitely `ctx_switch`.

mips/intr.c
cahirwpz: Right!

sys/sched.c
cahirwpz: Should I add some comment here to make the case more clear?

jpszczolowski: I don't think so. The code is clear. It's just my mind that was fixed on checking interlacing of `mips_exc_handler` calls. :wink:

mips/intr.c
cahirwpz: Right... I'll have to give it another round of thinking.

mips/intr.c
jpszczolowski: I totally agree on the cases and the actions that should be taken in each of them. I only wanted to say that this comment:
```
 *    b. preemption is disabled, interrupts are enabled:
 *       kernel was running in thread context and acquired sleep lock (mutex)
```
is misleading as:
* preemption could've been disabled anywhere in the kernel, not only in `sys/mutex.c` code,
* IMO the comment implies that "thread has acquired a mutex" means "thread has disabled preemption" -- it should be stated that the preemption is disabled only during acquir**ing** and releas**ing** of the mutex.

mips/intr.c
cahirwpz: I'll keep it as is. I added somewhat lengthy explanation of reasoning behind it.

PR #548, opened 2019-02-15
Tests for setpgid, kill and killpg.

staffik:
**Summary**
- add `bin/utest/pgrp.c` file with tests:
  - `test_setpgid` - test setpgid syscall
  - `test_kill` - test sending signal to other process
  - `test_killpg_same_group` - test sending signal to own process group
  - `test_killpg_other_group` - test sending signal to other process group
- fix issue with unwaited children in `test_fd_pipe`
- add mising `proc_unlock` to `proc_getpgid(..)`


sys/proc.c
```diff
@@ -298,8 +301,10 @@ int proc_sendsig(pid_t pid, signo_t sig) {
   }
 
   WITH_MTX_LOCK (&pgrp->pg_lock) {
-    TAILQ_FOREACH (target, &pgrp->pg_members, p_pglist)
+    TAILQ_FOREACH (target, &pgrp->pg_members, p_pglist) {
+      proc_lock(target);
```
cahirwpz: This looks horrible. I guess `sig_kill` should assume that `proc_lock` is held for the duration of the procedure.

staffik: yes

bin/utest/pgrp.c
```diff
@@ -0,0 +1,136 @@
...
+
+static void kill_tests_setup(void) {
+  sig_delivered = 0;
+  setpgid(0, 1);
+
+  sigaction_t newact;
+  sigaction_t oldact;
+
+  newact.sa_handler = sa_handler;
+  newact.sa_restorer = sigreturn;
```
cahirwpz: Not that long ago we decided to remove `sa_restorer` from `struct sigaction` since it's not a part of POSIX.1 standard.

bin/utest/fd.c
```diff
@@ -175,6 +176,9 @@ int test_fd_pipe() {
     assert_close_ok(fd[0]);
     assert_write_ok(fd[1], str, strlen(str));
     assert_close_ok(fd[1]);
+
+    int status;
+    waitpid(-1, &status, 0);
```
mzr: Shouldn't you also add `exit(EXIT_SUCCESS)` into child's part of code?

bin/utest/pgrp.c
```diff
@@ -0,0 +1,137 @@
...
+  assert(sig_delivered);
+
+  return 0;
+}
+
+/* In this tests there are five processes marked with: a, b, c, d, e.
+ * Process groups are marked with brackets: (a -> b) --> (c -> d) --> (e)
+ * Process e sends signal to the process group containing c and d.
+ * Process b sends signal to own process group containing a and b. */
+int test_killpg(void) {
```
mzr: This test looks a little bit complex, but on the other hand it is well commented. I don't have a strong opinion about it, but would be better if you could split it somehow, or come up with a test(s), testing the same functionality but "simpler".

cahirwpz: I agree with @mzr. If at least one reviewer finds the code difficult to read, it's time to rewrite it.

PR #550, opened 2019-04-16
Remove wrong interlock of interrupt thread's spinlock

jpszczolowski:
Let's look at the `sys/interrupt.c` file. Currently there are two places where the code acquires a spinlock (`delegated_lock`) and then may switch its context. Such behavior is wrong since inside `mips/switch.S` we only care about an interlock for `td_spin`. So after the context switch, `delegated_lock` is still owned (i.e. `delegated_lock->s_owner` is not `NULL`)  and other threads/interrupt contexts will fail while trying to acquire it.

Right now, even without this PR, everything seems to work fine because the interrupt thread is unused. But tests start to fail if we move UART (`drv/ns16550.c`) handling from the bottom half to the interrupt thread.

I've removed `delegated_lock` from the code. In function `intr_event_run_handlers` interrupts are already disabled, in function `intr_thread` I've disabled interrupts instead of acquiring the spinlock.


jpszczolowski:
@cahirwpz done

PR #555, opened 2019-10-06
Implement simple sigaltstack & sigprocmask & setcontext

jpszczolowski:


include/sys/context.h
```diff
@@ -1,6 +1,8 @@
 #ifndef _SYS_CONTEXT_H_
 #define _SYS_CONTEXT_H_
 
+#include <sys/ucontext.h>
```
cahirwpz: If you import this header file only for `ucontext_t *` then it's preferred to use forward opaque declaration. See below for `ctx_t` and `thread_t`.

jpszczolowski: Ok. Fixed in: https://github.com/cahirwpz/mimiker/pull/555/commits/c4202adb43ccba1cd3a40da4caf8a55b5ef0ecfc.

PR #533, opened 2019-01-25
Provide basic unit tests for FDT interface

Tomatosoup97:
Despite of the fact that the tests will assure that operations on FDT work as expected, they will provide us a basic usage guide.

PR #556, opened 2019-10-08
gdb nextuser command

mzr:
Continues execution and breaks on first user-mode command after exiting kernel-mode.
Related to #533 .


.gdbinit
```diff
@@ -14,6 +14,15 @@ break kernel_init
 break assert_fail
 break panic_fail
 
+define nextuser
```
cahirwpz: Could you provide a short description of this command that can be displayed with `help user`?

mzr: Yup.

PR #558, opened 2019-10-13
Simplify setcontext syscall by rearranging regs in exc_frame_t

jpszczolowski:


cahirwpz:
What about this comment https://github.com/cahirwpz/mimiker/blob/master/include/mips/mcontext.h#L73 ?

cahirwpz:
I don't like the idea of storing registers $k0, $k1 and $zero in `mcontext_t` – this is pointless. On the other hand it'd be nice to somehow unify [exc_frame_t](https://github.com/cahirwpz/mimiker/blob/master/include/mips/exc.h#L200) and [mcontext_t](https://github.com/cahirwpz/mimiker/blob/master/include/mips/mcontext.h#L113).

cahirwpz:
I'll merge it as is, however in future we should give all context structures another look.

PR #562, opened 2019-10-16
Remove docker development environment

mzr:
Docker development environment is no longer supported.


cahirwpz:
It was a distraction from the very beginning. Good riddance!

PR #571, opened 2019-10-28
Tmpfs

xThaid:
Add some basic functionalities of tmpfs.


cahirwpz:
Please split the PR into three smaller ones:
1) extend `vnodeops` with `VOP_RECLAIM` operation
2) add skeleton for `tmpfs` without any `vnodeops` or `vfsops`
3) implement several functions in the skeleton

Issue #553, opened 2019-10-02
break on first user-space instruction


mzr:
We need a gdb command to break on first executed instruction in the user-space.

jpszczolowski:
Implemented in #556

PR #575, opened 2019-11-02
Glue kernel-space backtrace with user-space backtrace.

mzr:
Providing unwind debug info using .cfi_* directives.
https://sourceware.org/binutils/docs/as/CFI-directives.html#CFI-directives
#553


include/mips/exc.h
```diff
@@ -7,7 +7,18 @@
...
-#define SAVE_REG(reg, offset, base) sw reg, (EXC_##offset)(base)
+#define _ADD_CFI_NO_CFI(reg, offset)
+
+#define _ADD_CFI_(reg, offset) .cfi_rel_offset reg, offset
+
+/* Resolves to _ADD_CFI_(reg, offset) when no variadic arguments given.
+ * Resolves to _ADD_CFI_NO_CFI when NO_CFI given as one, and only one
+ * variadic argument.
+ * Passing multiple variadic arguments to this macro is consiered an error.
+ */
```
jpszczolowski: `consiered` typo

include/Makefile
```diff
@@ -6,8 +6,7 @@ HEADERS = $(wildcard *.h) $(foreach dir, machine sys, $(wildcard $(dir)/*.h))
 INSTALL-FILES = $(addprefix $(SYSROOT)/usr/include/, $(HEADERS))
 
 SOURCES_H = $(foreach dir, . dev mips sys, $(wildcard $(dir)/*.h))
-FORMAT-EXCLUDE = mips/asm.h mips/m32c0.h \
-                 sys/syscall.h sys/syscallargs.h
+FORMAT-EXCLUDE = mips/m32c0.h sys/syscall.h sys/syscallargs.h
```
jpszczolowski: Maybe you could move reformatting of `include/mips/asm.h` to a different PR?

sys/mips/exc.S
```diff
@@ -21,36 +21,57 @@
...
 # [$k0] must be set to value of C0_STATUS
+//NESTED(exc_enter, 0, sp)
 exc_enter:
         mfc0    k0, C0_STATUS           
         andi    k0, SR_KSU_MASK         # Did exception occur in kernel mode?
         beqz    k0, kern_exc_enter
         nop
+        j user_exc_enter
+        nop
+
```
jpszczolowski: Why do we jump to `user_exc_enter` here instead of falling-through?

sys/mips/exc.S
```diff
@@ -21,36 +21,57 @@
...
-        lw      k0, PCPU_KSP(k0)
 
-        SAVE_CPU_CTX(sp, k0)
+        # sp contains kernel stack pointer
+        lw      sp, PCPU_KSP(k0)
+
+        .cfi_def_cfa sp, 0
+
+        // save user sp
+        sw k1, 104(sp)
```
jpszczolowski: Inconsistent comment style, maybe use `#` instead of `//`

sys/mips/exc.S
```diff
@@ -21,36 +21,57 @@
         .local skip_fpu_save
         .local skip_fpu_restore
 
+        .cfi_sections .debug_frame, .eh_frame
+
 # [$k0] must be set to value of C0_STATUS
+//NESTED(exc_enter, 0, sp)
 exc_enter:
```
jpszczolowski: What about this `//NESTED(exc_enter, 0, sp)`?

include/Makefile
cahirwpz: Did that on my own. Please clean up `mips/asm.h` header file.

include/mips/asm.h
```diff
@@ -119,33 +87,36 @@
...
+  .cfi_startproc
 
 /*
- * XLEAF
- *	declare alternate entry to leaf routine
- */
+ *  * XLEAF
+ *   *  declare alternate entry to leaf routine
+ *    */
 #define XLEAF(x)                                                               \
```
jpszczolowski: Unnecessary asterisks.

include/mips/asm.h
```diff
@@ -119,33 +87,36 @@
...
+  .cfi_startproc
 
 /*
- * LEAF_NOPROFILE
- *	No profilable leaf routine.
- */
+ *  * LEAF_NOPROFILE
+ *   *  No profilable leaf routine.
+ *    */
 #define LEAF_NOPROFILE(x)                                                      \
```
jpszczolowski: Unnecessary asterisks.

include/mips/asm.h
```diff
@@ -172,7 +176,8 @@
   .globl _C_LABEL(x);                                                          \
   .ent _C_LABEL(x), 0;                                                         \
   _C_LABEL(x) :;                                                               \
-  .frame sp, fsize, retpc
+  .frame sp, fsize, retpc;                                                     \
+  .cfi_startpoic
```
cahirwpz: Typo!

include/mips/exc.h
```diff
@@ -7,7 +7,18 @@
...
+#define _ADD_CFI_NO_CFI(reg, offset)
+
+#define _ADD_CFI_(reg, offset) .cfi_rel_offset reg, offset
+
+/* Resolves to _ADD_CFI_(reg, offset) when no variadic arguments given.
+ * Resolves to _ADD_CFI_NO_CFI when NO_CFI given as one, and only one
+ * variadic argument.
+ * Passing multiple variadic arguments to this macro is considered an error.
+ */
+#define SAVE_REG(reg, offset, base, ...)                                       \
```
cahirwpz: Could you be more direct and rename it to `SAVE_REG_CFI`, restore original `SAVE_REG` and remove `_ADD_CFI_*` clutter?

include/mips/exc.h
```diff
@@ -56,7 +72,10 @@
   SAVE_REG(t0, SR, reg);                                                       \
   SAVE_REG(t1, CAUSE, reg);                                                    \
   SAVE_REG(t2, BADVADDR, reg);                                                 \
-  SAVE_REG(t3, PC, reg)
+  /* Saving value of user-space PC just before syscall. */                     \
+  SAVE_REG(t3, PC, reg, NO_CFI);                                               \
+  .cfi_return_column t3;                                                       \
```
cahirwpz: What does `.cfi_return_column` do ?

lib/libc/sys/syscalls.S
```diff
@@ -1,5 +1,7 @@
 #include <machine/syscall.h>
 
+.cfi_sections .debug_frame, .eh_frame
```
cahirwpz: I removed those entirely, rebuild repo from scratch and kernel-user transition in backtrace still works. Are those really needed? What they were supposed to do then?

sys/mips/ebase.S
```diff
@@ -15,6 +15,8 @@
 
         .section .ebase, "ax"
 
+        .cfi_sections .debug_frame, .eh_frame
```
cahirwpz: ditto

sys/mips/exc.S
```diff
@@ -21,36 +21,48 @@
         .local skip_fpu_save
         .local skip_fpu_restore
 
+        .cfi_sections .debug_frame, .eh_frame
```
cahirwpz: ditto

sys/mips/exc.S
```diff
@@ -21,36 +21,48 @@
...
         beqz    k0, kern_exc_enter
         nop
 
+
 # IMPORTANT: While exception level is active the code may only access kernel
 # stack and local pcpu structure! FTTB These are accessed through kseg0.
 # It is UNSAFE to use k0/k1 registers when TLB miss handler may be triggered!
 
 NESTED(user_exc_enter, USER_EXC_FRAME_SIZE, ra)
+        .cfi_signal_frame
```
cahirwpz: This one is really required. Without it backtrace stops at first stack frame in user space. Good job!

sys/mips/exc.S
```diff
@@ -21,36 +21,48 @@
...
+
+
+        # Move user sp to temporary register and replace it with kernel sp.
+        move    k1, sp
+        .cfi_register sp, k1
```
cahirwpz: Could you add comments to `.cfi_register` and `.cfi_def_cfa`. I reckon nobody will remember what the directives do in 6 months.

sys/mips/exc.S
```diff
@@ -21,36 +21,48 @@
...
+        .cfi_sections .debug_frame, .eh_frame
+
 # [$k0] must be set to value of C0_STATUS
 exc_enter:
         mfc0    k0, C0_STATUS           
         andi    k0, SR_KSU_MASK         # Did exception occur in kernel mode?
         beqz    k0, kern_exc_enter
         nop
 
+
```
cahirwpz: Could you remove extranous empty lines in this file?

sys/mips/exc.S
```diff
@@ -21,36 +21,48 @@
...
+        # Exception frame will be saved on kernel stack. 
         # Fetch exception frame pointer where context will be saved.
         LOAD_PCPU_KSEG0(k0)
-        lw      k0, PCPU_KSP(k0)
 
-        SAVE_CPU_CTX(sp, k0)
+        # sp contains kernel stack pointer
+        lw      sp, PCPU_KSP(k0)
+
+        .cfi_def_cfa sp, 0
```
cahirwpz: If a directive is associated with given instruction do not insert newline between them.

build/flags.mk
```diff
@@ -1,4 +1,4 @@
 ASFLAGS  +=
 WFLAGS   += -Wall -Wextra -Wno-unused-parameter -Wstrict-prototypes -Werror
-CFLAGS   += -std=gnu11 -Og -ggdb3 $(WFLAGS)
+CFLAGS   += -std=gnu11 -Og -ggdb3 $(WFLAGS) -fno-asynchronous-unwind-tables
```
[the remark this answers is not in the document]
cahirwpz: Seems that everything works when it's removed.

lib/libc/sys/syscalls.S
mzr: According to [doc](https://sourceware.org/binutils/docs/as/CFI-directives.html#CFI-directives):

> `.cfi_sections` may be used to specify whether CFI directives should emit `.eh_frame` section and/or `.debug_frame` section. If section_list is `.eh_frame`, `.eh_frame` is emitted, if section_list is `.debug_frame`, `.debug_frame` is emitted. To emit both use `.eh_frame`, `.debug_frame`. 

But:
> The default if this directive is not used is .cfi_sections .eh_frame.

So we should be fine without these.

include/mips/exc.h
mzr: According to doc: 
> Change return column register, i.e. the return address is either directly in register or can be accessed by rules for register.

I only use it to later insert `.cfi_rel_offset t3, EXC_PC`. To my best knowledge this is the only way to specify return address to instruction just before entering kernel-space.

mzr: Without this we miss last frame before entering kernel-space. I.e. `write` which calls syscall.

PR #576, opened 2019-11-03
Reading from pipe should return only what's available

jpszczolowski:
From _Advanced Programming in the UNIX Environment (Third Edition)_, section 3.7:
> Data is read from an open file with the `read` function. [...] There are several cases in which the number of bytes actually read is less than the amount requested:
> * [...]
> * When reading from a pipe or FIFO. If the pipe contains fewer bytes than requested, read will return only what is available.

In our case `/bin/ksh` is reading from pipe using `nbytes` set to 65535 (in order to read everything). Right now Mimiker blocks this read until `65535` bytes are available.


cahirwpz:
Right! `read(...)` on pipe is allowed to return short count. But when there's nothing to read it must wait, unless the other end of pipe was closed. Returning 0 from `read(...)` always means `EOF`.

You removed:
```
      /* the buffer is empty so wait for some data to be produced */
      cv_wait(&producer->nonempty, &producer->mtx);
```
... which was a bad idea, and that forced you to busy wait in `assert_read_equal_blocking`.

PR #577, opened 2019-11-03
Implement fcntl syscall (only F_DUPFD command)

jpszczolowski:
`fcntl` syscall is used (and needed) by `/bin/ksh` while using pipes (e.g. calling `cat /bin/ls | wc`), so that's another step to have a working shell.


sys/kern/vfs_syscalls.c
```diff
@@ -125,6 +125,22 @@ int do_dup2(proc_t *p, int oldfd, int newfd) {
...
+  /* TODO: Currently only F_DUPFD command is implemented. */
+  if (cmd != F_DUPFD)
+    return ENOSYS;
+
+  file_t *f;
+  int error;
+
+  if ((error = fdtab_get_file(p->p_fdtable, fd, 0, &f)))
+    return error;
+  if ((error = fdtab_install_file_at_min(p->p_fdtable, f, arg, newfdp)))
```
cahirwpz: You missed to drop a reference to `f` here in case there was an error.

cahirwpz: I see... there's similar mistake in other functions in this file. I find current interface of `fdtab_install_file_at` a little bit awkward. I suspect it might be prone to a data race – consider a preemption between `fdtab_get_file` and `fdtab_install_file`.

I'll attempt to quickly improve the interface.

PR #579, opened 2019-11-04
jmp to __sc_error inside SYSCALL is now local

mzr:
`bnez` is I-type instruction so it can't jump too far.
That caused failure in build 9010 of pull/578 (#578) on CircleCI.

Now it jumps to local label, and then `j` instruction (J-type) is used to jump further.

### DIFF
--- a/include/mips/syscall.h
+++ b/include/mips/syscall.h
@@ -27,8 +27,12 @@
   LEAF(name);                                                                  \
   li v0, num;                                                                  \
   syscall;                                                                     \
-  bnez v1, _C_LABEL(__sc_error);                                               \
+  bnez v1, ___sc_error##name;                                                  \
+  nop;                                                                         \
   j ra;                                                                        \
+  nop;                                                                         \
+  ___sc_error##name : j _C_LABEL(__sc_error);                                  \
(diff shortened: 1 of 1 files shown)


include/mips/syscall.h
```diff
@@ -27,8 +27,12 @@
   LEAF(name);                                                                  \
   li v0, num;                                                                  \
   syscall;                                                                     \
-  bnez v1, _C_LABEL(__sc_error);                                               \
+  bnez v1, ___sc_error##name;                                                  \
+  nop;                                                                         \
   j ra;                                                                        \
+  nop;                                                                         \
+  ___sc_error##name : j _C_LABEL(__sc_error);                                  \
```
cahirwpz: You're introducing lots of global symbols this way. Preferred way to code that is to use real local labels.
```
bnez v1, 1f
...
1:  j _C_LABEL(__sc_error)
```

cahirwpz:
Please note that you need to use `.set noreorder` to control instructions inserted into delay slots. Please always check with `objdump` if generated assembly is matches your expectations. I could see double `nop` instruction scattered around.

PR #582, opened 2019-11-07
Pass arguments to userspace programs.

pj1031999:
Add option to pass arguments to init program using launch. Everything after --
will be passed as arguments.
Example command:
./launch -d init=/bin/ls -- / /usr

### DIFF
--- a/sys/kern/main.c
+++ b/sys/kern/main.c
@@ -9,7 +9,7 @@
 
 /* Borrowed from mips/malta.c */
 char *kenv_get(const char *key);
-char **kenv_get_init_args(int *n);
+char **kenv_get_init(void);
 
 int kmain(void) {
   char *init = kenv_get("init");
@@ -19,16 +19,7 @@ int kmain(void) {
   proc_add(proc_create(thread_self(), NULL));
 
   if (init) {
-    int ninit_args;
-    char **init_args = kenv_get_init_args(&ninit_args);
-    char **args = kmalloc(M_TEMP, sizeof(char *) * (ninit_args + 2), 0);
-    args[0] = init;
-    for (int i = 0; i < ninit_args; ++i) {
-      args[i + 1] = init_args[i];
-    }
-    args[ninit_args + 1] = NULL;
-    run_program(init, args, (char *[]){NULL});
-    kfree(M_TEMP, args);
+    run_program(init, kenv_get_init(), (char *[]){NULL});
   } else if (test) {
     ktest_main(test);
   } else {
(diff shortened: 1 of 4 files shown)


sys/kern/main.c
```diff
@@ -17,7 +19,20 @@ int kmain(void) {
   proc_add(proc_create(thread_self(), NULL));
 
   if (init) {
-    run_program(init, (char *[]){init, NULL}, (char *[]){NULL});
+    int ninit_args;
+    char **init_args = kenv_get_init_args(&ninit_args);
+
+    if (init_args) {
+      char **args = kmalloc(M_TEMP, sizeof(char *) * (ninit_args + 2), 0);
```
xThaid: Shouldn't this memory leak be handled?

sys/mips/malta.c
```diff
@@ -122,6 +125,19 @@ char *kenv_get(const char *key) {
...
+    char *arg = _kenv.argv[i];
+    if ((strncmp("--", arg, 2) == 0)) {
+      if (n) {
+        *n = _kenv.argc - i - 1;
+      }
+      return _kenv.argv + i + 1;
+    }
+  }
+  return NULL;
+}
```
jpszczolowski: An idea: changing `return NULL;` to `*n = 0; return NULL;` would enable you to merge two cases in file `sys/kern/main.c`: `if (init_args) { ... } else { ... }` into just one. Then you could delete:
 ```
run_program(init, (char *[]){init, NULL}, (char *[]){NULL});
```
and always call:
```
run_program(init, args, (char *[]){NULL});
```

File: sys/kern/main.c [hunk 3]
```diff
@@ -5,9 +5,11 @@
 #include <sys/proc.h>
 #include <sys/thread.h>
 #include <sys/ktest.h>
+#include <sys/malloc.h>
 
 /* Borrowed from mips/malta.c */
 char *kenv_get(const char *key);
+char **kenv_get_init_args(int *n);
```
jpszczolowski: I find it a little bit confusing that `kenv_get_init_args()` returns the result both as a regular return value and through a pointer. But maybe it's just nitpicking.

sys/kern/main.c
pj1031999: Done. (d6fd9d470780afc6d27bc4f28035418f04d7a828)

sys/mips/malta.c
pj1031999: Done. (d6fd9d470780afc6d27bc4f28035418f04d7a828)

sys/kern/main.c
pj1031999: From my point of view it is the simplest solution in this case. But if you know something more elegant (and not more complicated) feel free to share your idea.

PR #583, opened 2019-11-10
Implement vmem subsystem (create & add & alloc calls)

jpszczolowski:
This PR is based on NetBSD's vmem implementation, which is available here: https://nxr.netbsd.org/xref/src/sys/kern/subr_vmem.c

**What is vmem?** You can find more information here:

> The vmem(9) API is a general address space allocator.  It is used internally by kmem(9), pool(9), vm(9), and other kernel subsystems and device drivers to allocate regions of various kinds of address spaces.

Source: `MEMORYALLOCATORS(9)`, _NetBSD Kernel Developer's Manual_, https://netbsd.gw.com/cgi-bin/man-cgi?memoryallocators+9+NetBSD-current

> In this section we describe a new general−purpose resource allocator, vmem, which provides guaranteed constant−time performance with low fragmentation. Vmem appears to be the first resource allocator that can do this.

Source: _Magazines and Vmem: Extending the Slab Allocator to Many CPUs and Arbitrary Resources, J. Bonwick et al. (2001)_, https://www.usenix.org/legacy/event/usenix01/full_papers/bonwick/bonwick.pdf


sys/kern/vmem.c
```diff
@@ -29,45 +32,111 @@ typedef struct vmem {
...
-typedef struct vmem_bt {
-  TAILQ_ENTRY(vmem_bt) bt_seglist;
+typedef struct bt {
+  TAILQ_ENTRY(bt) bt_seglist;
   union {
-    LIST_ENTRY(vmem_bt) u_freelist;
-    LIST_ENTRY(vmem_bt) u_hashlist;
+    LIST_ENTRY(bt) u_freelist;
+    LIST_ENTRY(bt) u_hashlist;
   } bt_u;
```
cahirwpz: We can use anonymous unions in C11. You don't need macro definitions below.

include/sys/vmem.h
```diff
@@ -0,0 +1,24 @@
+#include <sys/types.h>
+
+#define VMEM_SLEEP 0x00000001
+#define VMEM_NOSLEEP 0x00000002
+#define VMEM_INSTANTFIT 0x00001000
+#define VMEM_BESTFIT 0x00002000
+
+typedef uintptr_t vmem_addr_t;
+typedef size_t vmem_size_t;
+typedef unsigned int vm_flag_t;
```
cahirwpz: This typename suggests `vm_flag_t` is somehow related to virtual memory subsystem. Please rename to `vmem_flags_t`.

jpszczolowski: Good point! In fact, I have already renamed it to `vmem_flags_t` on my local branch. :D

sys/kern/vmem.c
jpszczolowski: Oh, I see, and it would simplify the code. Thanks!

sys/kern/vmem.c
```diff
@@ -0,0 +1,140 @@
...
+  vmem_seglist_t vm_seglist;
+  vmem_freelist_t vm_freelist[VMEM_MAXORDER];
+  vmem_hashlist_t vm_hashlist[VMEM_MAXHASH];
+  LIST_ENTRY(vmem) vm_alllist;
+} vmem_t;
+
+typedef enum { BT_TYPE_FREE, BT_TYPE_BUSY, BT_TYPE_SPAN } bt_type_t;
+
+typedef struct bt {
+  TAILQ_ENTRY(bt) bt_seglist;
```
cahirwpz: If that's an entry on list, then it's called `link`. Unfortunately this convention is not used throughout whole code base.

Please rename to `bt_seglink` and perform similar name transformation for other links.

jpszczolowski: Thanks, done!

cahirwpz:
Please find extra review for this code.

PR #584, opened 2019-11-13
Fix log2 function in mimiker.h

jpszczolowski:
Currently our `log2(x)` implementation works fine iff `x` is a power of 2. This PR fixes it.
You can find a simple comparison of old-`log` vs new-`log` here: https://ideone.com/BNRNFU

### DIFF
--- a/include/sys/mimiker.h
+++ b/include/sys/mimiker.h
@@ -15,7 +15,7 @@
 #include <sys/param.h>
 #include <sys/types.h>
 
-#define log2(x) (__builtin_ffs(x) - 1)
+#define log2(x) (63 - __builtin_clzl(x))
 #define ffs(x) (size_t)(__builtin_ffs(x))
 #define clz(x) (size_t)(__builtin_clz(x))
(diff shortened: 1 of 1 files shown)


include/sys/mimiker.h
```diff
@@ -15,7 +15,7 @@
 #include <sys/param.h>
 #include <sys/types.h>
 
-#define log2(x) (__builtin_ffs(x) - 1)
+#define log2(x) (31 - __builtin_clz(x))
```
cahirwpz: I'm afraid it's not going to work on _AArch64_ :(

cahirwpz: Well in fact... it won't work with `int16_t` nor `int8_t`...

jpszczolowski: > Well in fact... it won't work with int16_t nor int8_t...

It does work with shorter data types than int since the builtin's prototype is in fact `int __builtin_clz (unsigned int x)`. It converts the argument to unsigned int.

PR #585, opened 2019-11-13
Extend vfs_lookup to return parent vnode.

xThaid:
`vfs_lookup` interface has changed. Now it accepts `nameiop_t`, which denotes lookup operation and returns `nameidata_t`, which encapsulates results: both `vnode` and its parent. When `vfs_lookup` is called with `NAMEI_CREATE` operation, returned parent `vnode` is held and locked. Otherwise parent is released.


include/sys/vfs.h
```diff
@@ -12,6 +12,31 @@ typedef struct statfs statfs_t;
...
+
+/* Path name component flags */
+#define NAMEI_ISLASTPC 0x00000001 /* this is last component of pathname */
+
+/*
+ * Encapsulation of lookup parameters.
+ */
+typedef struct {
+  nameiop_t pc_nameiop;
+  uint32_t pc_flags;
```
pj1031999: Are you planning more flags?

xThaid: Yes, there will be more flags.

PR #587, opened 2019-11-14
let gdb read kernel memory

mzr:


mzr:
Breaks #575

mzr:
#590 Solves the same problem without breaking user-backtrace.

PR #590, opened 2019-11-18
Read kernel memory while on bp in user-space.

mzr:
This patch lets gdb read kernel memory while on bp in user-space. User processes still are not allowed to read kernel memory. This code looks nasty but this is the only working solution among few others.


cahirwpz:
I merged my version of patch into `master`. If you find the patch to be buggy please rewrite it. Otherwise close the PR.

PR #593, opened 2019-11-19
vmem: comments & simplify vmem_alloc & little refactor

jpszczolowski:
* Added comments in `vmem.h` and to structures in `vmem.c`
* Changed `vmem_assert_locked()` function to `VMEM_ASSERT_LOCKED()` macro
* Make use of `align()` and `is_aligned()` from `include/sys/mimiker.h`
* Change `vm_quantum_mask` field to `vm_quantum`, as the latter is used more frequently (note that `mask` = `quantum` - 1)
* Delete `vmem_align_and_fit()` call from `vmem_alloc()` (no need to align as all existing segments are already aligned to `vm_quantum`, no need to check size as it's INSTANTFIT strategy)
* Delete "`btnew_prev`" case inside `vmem_alloc()`, as it would never happen
* Check that segments' start addresses are aligned to `vm_quantum` in `vmem_check_sanity()`
* Check that segments' sizes are at least equal `vm_quantum` in `vmem_check_sanity()`

### DIFF
--- a/sys/kern/vmem.c
+++ b/sys/kern/vmem.c
@@ -239,7 +239,7 @@ int vmem_alloc(vmem_t *vm, vmem_size_t size, vmem_addr_t *addrp) {
   for (vmem_freelist_t *list = first; list < end; list++) {
     bt = LIST_FIRST(list);
     if (bt != NULL) {
-      /* This is INSTANTFIT strategy, we know that any segment found on these
+      /* This is instant-fit strategy, we know that any segment found on these
        * lists is large enough. */
       assert(bt->bt_size >= size);
       found = true;
--- a/sys/kern/vmem.c
+++ b/sys/kern/vmem.c
@@ -16,6 +16,8 @@
 #define ORDER2SIZE(order) ((vmem_size_t)1 << (order))
 #define SIZE2ORDER(size) ((int)log2(size))
 
+#define VMEM_ASSERT_LOCKED(vm) assert(mtx_owned(&(vm)->vm_lock))
+
 typedef TAILQ_HEAD(vmem_seglist, bt) vmem_seglist_t;
 typedef LIST_HEAD(vmem_freelist, bt) vmem_freelist_t;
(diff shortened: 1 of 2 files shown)


sys/kern/vmem.c
```diff
@@ -65,7 +86,7 @@ static vmem_freelist_t *bt_freehead_tofree(vmem_t *vm, vmem_size_t size) {
  * For VMEM_INSTANTFIT, returns the list in which any blocks are large enough.
```
cahirwpz: I guess that comment is not relevant anymore?

sys/kern/vmem.c
```diff
@@ -251,56 +238,41 @@ int vmem_alloc(vmem_t *vm, vmem_size_t size, vmem_addr_t *addrp) {
   vmem_check_sanity(vm);
 
   bt_t *bt;
-  vmem_addr_t start;
   bool found = false;
   for (vmem_freelist_t *list = first; list < end; list++) {
     bt = LIST_FIRST(list);
-    if (bt != NULL && vmem_align_and_fit(bt, size, align, &start) == 0) {
+    if (bt != NULL) {
+      /* This is INSTANTFIT strategy, we know that any segment found on these
```
cahirwpz: Ditto?

sys/kern/vmem.c
jpszczolowski: Yes, I've deleted part of this comment: https://github.com/cahirwpz/mimiker/pull/593/commits/5e62fb32623934c1f8408ab62e709f7b7fa8a4e5

sys/kern/vmem.c
jpszczolowski: I'd leave a mention that this is instant-fit... But I've changed caps to regular font: https://github.com/cahirwpz/mimiker/pull/593/commits/1c0c2fff9b1a44e87ef5a3a2f11e3fb6be1b147f

PR #601, opened 2019-11-23
Implement mkdir syscall

xThaid:
Implemented [mkdir](https://netbsd.gw.com/cgi-bin/man-cgi?mkdir+2+NetBSD-current) syscall.

### DIFF
--- a/sys/kern/vfs.c
+++ b/sys/kern/vfs.c
@@ -341,6 +341,29 @@ int vfs_lookup(const char *path, vnode_t **vp) {
   return error;
 }
 
+int vnr_create(const char *path, vnode_t **dvp, componentname_t *cn) {
+  vnrstate_t vs;
+  vnrstate_init(&vs, VNR_CREATE, path);
+  int error = vfs_nameresolve(&vs);
+  if (error)
+    return error;
+
+  if (vs.vs_vp != NULL) {
+    if (vs.vs_vp != vs.vs_dvp)
+      vnode_put(vs.vs_dvp);
+    else
+      vnode_drop(vs.vs_dvp);
+
+    vnode_drop(vs.vs_vp);
+    return EEXIST;
+  }
+
+  *dvp = vs.vs_dvp;
+  memcpy(cn, &vs.vs_cn, sizeof(componentname_t));
+
+  return 0;
+}
+
 int vfs_open(file_t *f, char *pathname, int flags, int mode) {
   vnode_t *v;
   int error = 0;
--- a/sys/kern/vfs.c
+++ b/sys/kern/vfs.c
@@ -341,7 +341,7 @@ int vfs_lookup(const char *path, vnode_t **vp) {
   return error;
 }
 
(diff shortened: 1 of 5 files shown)


sys/kern/vfs.c
```diff
@@ -341,6 +341,29 @@ int vfs_lookup(const char *path, vnode_t **vp) {
...
+    return error;
+
+  if (vs.vs_vp != NULL) {
+    if (vs.vs_vp != vs.vs_dvp)
+      vnode_put(vs.vs_dvp);
+    else
+      vnode_drop(vs.vs_dvp);
+
+    vnode_drop(vs.vs_vp);
+    return EEXIST;
```
cahirwpz: Should not the code be returned by `vfs_nameresolve`?
Seems like `vfs_nameresolve` implementation details are leaking here.

sys/kern/vfs.c
```diff
@@ -341,6 +341,29 @@ int vfs_lookup(const char *path, vnode_t **vp) {
   return error;
 }
 
+int vnr_create(const char *path, vnode_t **dvp, componentname_t *cn) {
```
cahirwpz: I guess we should rename it to `vfs_namecreate` and `vfs_lookup` to `vfs_namelookup` (possibly in another PR).

sys/kern/vfs_syscalls.c
```diff
@@ -179,7 +181,36 @@ int do_unlink(proc_t *p, char *path) {
...
-  return ENOTSUP;
+  int error;
+  vattr_t va;
+  vnode_t *vn, *dvn;
+  componentname_t cn;
+
+  if ((error = vnr_create(path, &dvn, &cn)))
+    return error;
+
+  if (cn.cn_namelen > NAME_MAX) {
```
cahirwpz: Should not that case be handled by `vfs_nameresolve` or `vfs_nameresolve`?

cahirwpz:
One more thing... please add a useful description for the PR.

sys/kern/vfs_syscalls.c
xThaid: Moved to `vfs_namecreate`.

sys/kern/vfs.c
xThaid: I didn't put this code in `vfs_nameresolve`, since this case is only handled during `VNR_CREATE` operation. The question is how should vnr operation specific functions look like. Should it contain any extra logic or just be wrappers of `vfs_nameresolve` to simplify calls?

sys/kern/vfs.c
xThaid: `vfs_lookup` renamed in another PR.

sys/kern/vfs.c
cahirwpz: Let's postpone the decision. We need to develop `vfs_nameresolve` that handle more use cases. Then we'll reevaluate `vfs_namecreate`.

PR #602, opened 2019-11-25
vmem: implement free() & destroy() calls

jpszczolowski:
* Parameters `base` and `size` have been removed from `vmem_create()` call, so that fresh `vmem` instance has no initial span, and one must call `vmem_add()` explicitly
* Implemented `vmem_free()`
* Implemented `vmem_destroy()` with extensive sanity check -- the kernel check that the destroyed `vmem` instance is empty and its memory is not corrupted
* Updated `sys/tests/vmem.c` test, so that it uses `vmem_free()` and `vmem_destroy()`


cahirwpz:
Please resolve conflicts and add PR description.

PR #603, opened 2019-11-25
Implementation of getcwd, chdir syscalls

JakubSzczerbinski:
> The working directory of a process is a directory of a hierarchical file system, if any, dynamically associated with each process.
> When the process refers to a file using a simple file name or relative path (as opposed to a file designated by a full path from a root directory), the reference is interpreted relative to the working directory of the process.

Implementation of getcwd and chdir syscalls based on [NetBSD](http://bxr.su/NetBSD/sys/kern/vfs_getcwd.c)

- Stores working directory as vnode in proc structure,
- Sets working directory to root during exec,
- Copies working directory during fork,
- Chdir: sets working directory as vnode retrieved by lookup on its argument,
- Getcwd: returns path from root to working directory vnode,
- Assumes there is a one global filesystem root (doesn't support chroot),
- Only contains setting and getting working directory. (Does not contain other related changes, for example lookup of relative paths)

NetBSD man pages: [chdir](https://netbsd.gw.com/cgi-bin/man-cgi?chdir), [getcwd](https://netbsd.gw.com/cgi-bin/man-cgi?getcwd++NetBSD-current)


jpszczolowski:
Please follow Mimiker's include guard convention. You can find an example here: https://github.com/cahirwpz/mimiker/blob/884ecdb9493dafa864d44abd69781cd006fd8689/include/sys/malloc.h#L1-L2
https://github.com/cahirwpz/mimiker/blob/884ecdb9493dafa864d44abd69781cd006fd8689/include/sys/malloc.h#L44

bin/utest/Makefile
```diff
@@ -1,6 +1,7 @@
 TOPDIR = $(realpath ../..)
 
 SOURCES = \
+	getcwd.c \
```
mzr: Add new source files so they are sorted alphabetically.

include/sys/proc.h
```diff
@@ -58,6 +59,7 @@ struct proc {
   sigaction_t p_sigactions[NSIG]; /* (@) description of signal actions */
   condvar_t p_waitcv;             /* (a) processes waiting for this one */
   int p_exitstatus;               /* (@) exit code to be returned to parent */
+  vnode_t *cwd;                   /* Current working directory */
   /* program segments */
   vm_segment_t *p_sbrk; /* ($) The entry where brk segment resides in. */
```
jpszczolowski: * You should add information about `cwd`'s locking strategy (e.g. `(@)`, `(a)`, `($)`).
* You should prefix fields inside `struct proc` with `p_`.

sys/kern/vfs.c
```diff
@@ -47,6 +48,10 @@ static vfs_init_t vfs_default_init;
 /* Global root vnodes */
 vnode_t *vfs_root_vnode;
 
+vnode_t *root_vnode(void) {
+  return vfs_root_vnode;
+}
```
jpszczolowski: Global symbols in `sys/kern/vfs.c` file should be prefixed with `vfs_`.

sys/kern/exec.c
```diff
@@ -409,6 +409,9 @@ __noreturn void run_program(const char *path, char *const *argv,
   fdtab_hold(fdt);
   p->p_fdtable = fdt;
 
+  /* Set current working directory to root directory */
+  p->p_cwd = vfs_root_vnode;
```
cahirwpz: You forgot to bump up reference counter on `vfs_root_vnode`.

sys/kern/syscalls.c
```diff
@@ -310,24 +312,43 @@ static int sys_chdir(proc_t *p, chdir_args_t *args, register_t *res) {
...
 
 static int sys_getcwd(proc_t *p, getcwd_args_t *args, register_t *res) {
-  __unused char *u_buf = args->buf;
-  __unused size_t len = args->len;
+  char *u_buf = args->buf;
+  size_t len = args->len;
 
-  return ENOTSUP;
+  char *path = kmalloc(M_TEMP, PATH_MAX, 0);
+  char *path_start = path + PATH_MAX;
```
cahirwpz: Why cannot you just pass pointer to `PATH_MAX` as second argument to `do_getcwd`?

sys/kern/vfs_name_in_dir.c
```diff
@@ -0,0 +1,48 @@
+#include <sys/klog.h>
+#include <sys/dirent.h>
+#include <sys/vnode.h>
+#include <sys/vm_map.h>
+#include <sys/malloc.h>
+#include <sys/libkern.h>
+#include <sys/vfs.h>
+
+int vfs_name_in_dir(vnode_t *dir, vnode_t *node, char *bufp, char **bpp) {
```
cahirwpz: Last argument should be `size_t *len` ?

cahirwpz:
Please also provide reasonable description for the PR, as requested by @pwit81 on #docs channel.

cahirwpz:
Please resolve conflicts and write PR description.

sys/kern/vfs_lookup.c
```diff
@@ -0,0 +1,51 @@
...
+
+  int offset = 0;
+  size_t last = *lastp;
+  uio_t uio;
+
+  dirent_t *dirents = kmalloc(M_TEMP, PATH_MAX, 0);
+
+  for (;;) {
+    uio = UIO_SINGLE_KERNEL(UIO_READ, offset, dirents, PATH_MAX);
+    if ((error = VOP_READDIR(dv, &uio, NULL)))
```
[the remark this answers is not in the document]
cahirwpz: Ouch. In fact directory offset is stored right in `uio`. NVM.

sys/kern/vfs_syscalls.c
```diff
@@ -304,3 +305,53 @@ int do_ioctl(proc_t *p, int fd, u_long cmd, void *data) {
...
+
+  for (;;) {
+    while (uvp->v_mount)
+      uvp = uvp->v_mount->mnt_vnodecovered;
+
+    if (uvp == vfs_root_vnode)
+      break;
+
+    componentname_t cn = COMPONENTNAME("..");
+    if ((error = VOP_LOOKUP(uvp, &cn, &lvp)))
```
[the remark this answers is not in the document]
cahirwpz: Looks better, though I had to fix loop not to return immediately but rather perform cleanup.

Issue #594, opened 2019-11-19
Kernel panic in ksh pipes when nonexistent command is used


jpszczolowski:
1. Run `/bin/ksh`
2. Enter `nonexistent_command | wc`
3. Boom! Kernel panic
```
KERNEL PANIC!!!
TLB exception (load or instruction fetch) at $c0115774!
Caused by reference to $000000f4!
HINT: Type 'info line *0xc0115774' into gdb to find faulty code line.
```

jpszczolowski:
@JakubSzczerbinski you may be interested in this

cahirwpz:
Seems like there's an attempt to deliver a signal to a zombie process that still resides in a group created by `ksh`.

PR #607, opened 2019-11-29
Fixing #594. Do not deliver signals to zombies, ...

mzr:
* signals are not delivered to zombies
  That caused #594 .
* pids start at 1
  Pids shouldn't start at 0, [because](https://www.freebsd.org/cgi/man.cgi?query=kill&apropos=0&sektion=2&manpath=FreeBSD+12.1-RELEASE+and+Ports&arch=default&format=html):
```
If	pid is zero:
	     The sig signal is sent to all processes whose group ID is equal
	     to	the process group ID of	the sender, and	for which the process
	     has permission; this is a variant of killpg(2).
```


mzr:
```
# echo a | wc | wc
echo a | wc | wc
       1       3      25
# 
```
Another pipe is not possible due to memory exhaustion in pipe buffers.

cahirwpz:
LGTM

PR #608, opened 2019-11-29
[RFC] tmux++

mzr:
ctrl + m: enable/disable mouse.
[ enter | [ ]: vi-style copy mode

I hope that rest is self-explanatory.

Copying to system clipboard is... tricky.
In order to do that we would need xsel/xclip installed.


cahirwpz:
LGTM

.tmux.conf
```diff
@@ -1,7 +1,63 @@
+# cat << EOF > /dev/null
+# This file is a shell/tmux.conf polyglot
+# https://en.wikipedia.org/wiki/Polyglot_(computing)
+
 set exit-unattached off
 set exit-empty off
 set -g status-position top
 
 set -g default-terminal "screen-256color"
 set -g mouse on
```
cahirwpz: Is mouse mode turned on by default a good idea?

.tmux.conf
```diff
@@ -1,7 +1,63 @@
...
 setw -g xterm-keys on
+set -g status-keys vi
```
cahirwpz: Could you add an inline comment for those two? I cannot understand the difference between those two without looking into `tmux` documentation.

.tmux.conf
```diff
@@ -1,7 +1,63 @@
+# cat << EOF > /dev/null
```
cahirwpz: Neat trick :)

PR #610, opened 2019-11-29
Remove unnecessary complexity from #607

mzr:
I was fiddling with some C syntax and forgot to delete it... AFAIK standard says that uninitialized indices are initialized to 0.


cahirwpz:
Duplicate of #609

PR #611, opened 2019-11-30
Implement readdir v-node operation for tmpfs

xThaid:
- Added node identifiers in tmpfs (simple incremental IDs)
- Directories in tmpfs now remember theirs parent (and also number of links is correctly updated)
- Implemented `readdir` syscall (using our helper routine `readdir_generic`)


cahirwpz:
i-node of ".." entry in root directory of tmpfs should be set to i-node of mount point, i.e. directory of filesystem where tmpfs was mounted. Please check it on Linux with `ls -ia`.

PR #612, opened 2019-11-30
Add some vfs tests

xThaid:
Added some vfs userspace tests checking basic functionality of creating/deleting directories.


bin/utest/main.c
```diff
@@ -53,6 +53,7 @@ int main(int argc, char **argv) {
   CHECKRUN_TEST(setjmp);
   CHECKRUN_TEST(sigaction_with_setjmp);
   CHECKRUN_TEST(sigaction_handler_returns);
+  CHECKRUN_TEST(vfs);
```
cahirwpz: This test name does not indicate which part of vfs you're about to test. Please be more specific when naming tests.

PR #616, opened 2019-12-09
Add support for KASAN (Kernel Address Sanitizer)

jpszczolowski:
Work based on KASANs from Linux and NetBSD kernels.

TODO: description


sys/kern/kasan.c
```diff
@@ -166,13 +164,33 @@ void kasan_mark(const void *addr, size_t size, size_t sz_with_redz,
...
+  uint32_t *ptr;
+
+  nentries =
+    ((size_t)&__CTOR_END__ - (size_t)&__CTOR_LIST__) / sizeof(uintptr_t);
+
+  ptr = &__CTOR_LIST__;
+  for (i = 0; i < nentries; i++) {
+    void (*func)(void);
+
+    func = (void *)(*ptr);
```
cahirwpz: On `Aarch64` you cannot cast `uint32_t` to `void *`. Use `intptr_t` whenever you need an integer that has the same size as pointer.

jpszczolowski: Thanks. I've refactored this function in the following commit: https://github.com/cahirwpz/mimiker/pull/616/commits/2bcaf0ed19f94053aa048cc707f9110122594080

PR #618, opened 2019-12-13
Add support for creating file using open syscall.

xThaid:
We can create new regular files using `open` syscall with `O_CREATE` flag specified. I have also handled `O_APPEND` flag. It doesn't have any effect now, but at least it doesn't return with error, as it did before.

### DIFF
--- a/sys/kern/vfs.c
+++ b/sys/kern/vfs.c
@@ -11,6 +11,7 @@
 #include <sys/linker_set.h>
 #include <sys/sysinit.h>
 #include <sys/mimiker.h>
+#include <sys/stat.h>
 
 /* Internal state for a vnr operation. */
 typedef struct {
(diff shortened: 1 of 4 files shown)


sys/kern/vfs.c
```diff
@@ -343,24 +344,16 @@ int vfs_namelookup(const char *path, vnode_t **vp) {
   return error;
 }
 
-int vfs_namecreate(const char *path, vnode_t **dvp, componentname_t *cn) {
+int vfs_namecreate(const char *path, vnode_t **dvp, vnode_t **vp,
```
cahirwpz: I wonder... why signature of this function doesn't look like this: `int vfs_namecreate(const char *path, vnode_t **dvp, vnode_t **vp, const char **namep)`. Won't `cn` always be a copy of last component from `path`? Why are we even consider copying it? Such change would simplify a call to `VOP_CREATE` as you would not need to create `namecopy`.

cahirwpz: During recent face-to-face discussion we came to a conclusion that some `vnodeops` must be modified to accept `componentname_t *` instead of `const char *` to make this code not to use dynamic allocation. Please address that in near future.

PR #620, opened 2020-01-04
Add read/write operations for tmpfs

xThaid:
Read and write operations for tmpfs are now fully working. Addressing a file's data blocks is based on a structure adopted from [ext2](http://web.mit.edu/tytso/www/linux/ext2intro.html) filesystem. The key features are fixed logical block size, ease of data location and indirect blocks. Default level of indirection is 2, which should be enough (file's size can be up to roughly 4GiB). 
I also have written some unit tests for this code.

### DIFF
--- a/bin/utest/main.c
+++ b/bin/utest/main.c
@@ -54,6 +54,7 @@ int main(int argc, char **argv) {
   CHECKRUN_TEST(sigaction_with_setjmp);
   CHECKRUN_TEST(sigaction_handler_returns);
   CHECKRUN_TEST(vfs_dir);
+  CHECKRUN_TEST(vfs_rw);
 
   CHECKRUN_TEST(setpgid);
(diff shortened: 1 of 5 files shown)


cahirwpz:
I wonder... maybe you tried to make the code too generic by assuming someone will be tempted to change value of `INDIR_LEVELS`?

xThaid:
I know this code may be too generic, but I feel like this is more clear and compact than hard-coding two levels of indirection. In my opinion, changing file size one block at a time also simplifies whole logic. I tried to write something else, but all resulted in rather longer and repetitive code with multiple if statements.

cahirwpz:
I'm willing to accept simpler code, even if it involves making it more verbose.

xThaid:
Second version is in another PR #624.

PR #626, opened 2020-01-20
Relative nameresolve

JakubSzczerbinski:


cahirwpz:
Please provide PR description!

The change works for simple case, but it fails when `.` or `..` is at the beginning of path:
```
# cd tmp
# echo foo > bar
# cat bar
foo
# cat ./bar
No such file or directory.
# cat ../tmp/bar
No such file or directory.
```

cahirwpz:
I'll accept partial implementation FTTB

PR #629, opened 2020-01-21
Truncate syscall

xThaid:
I have implemented [truncate, ftruncate](https://netbsd.gw.com/cgi-bin/man-cgi?truncate+2+NetBSD-current) syscall. I have written also some unit tests.


bin/utest/vfs.c
```diff
@@ -65,6 +65,27 @@ int test_vfs_rw(void) {
...
+  assert(read(3, rdbuf, 10000) == 7777);
+  assert(!memcmp(wrbuf, rdbuf, 2048));
+  /* Rest of the file should be zeroed. */
+  for (int i = 2048; i < 7777; i++)
+    assert(((uint8_t *)rdbuf)[i] == 0);
+
+  truncate(TESTDIR "/file", 0);
+  assert_lseek_ok(0, 0, SEEK_SET);
+  assert(read(3, rdbuf, 2048) == 0);
+
```
cahirwpz: Please test `truncate` on device files and directories as well.

xThaid: `truncate` requires operations `VOP_ACCESS` and `VOP_SETATTR` to be implemented, so calling this syscall on device files results in `ENOTSUP` errno. I don't know if testing this behavior makes any sense.

cahirwpz: Ok. Let's take care about it later.

PR #636, opened 2020-01-27
Access kernel stack using only KSEG2 addresses

jpszczolowski:
We would like to always access kernel stack through `kseg2` addresses, so that we can enable KASAN stack instrumentation.


jpszczolowski:
Note: some CFI directives have been deleted, I'll have to fix user-backtrace

PR #646, opened 2020-01-31
Symbolic links

xThaid:
This is early stage of symbolic links. I have added [readlinkat](https://netbsd.gw.com/cgi-bin/man-cgi?readlinkat+2+NetBSD-current) syscall. For now, it is only supported by the initrd file system.


cahirwpz:
LGTM though name resolution does not handle symbolic links. I'll accept it as an intermediate step.

PR #649, opened 2020-02-12
umask syscall

xThaid:
Implemented [umask](https://netbsd.gw.com/cgi-bin/man-cgi?umask+2+NetBSD-current) system call. Access permissions taken as argument by `mkdir(2)` and `open(2)` system calls are now restricted by the `umask`.


cahirwpz:
LGTM

PR #650, opened 2020-02-20
Symbolic links resolving

xThaid:
Paths with symbolic links in it are now correctly resolved.


xThaid:
Our previous efforts to remove memory allocation for the path during name resolving was quite useless. We need extra buffer anyway as we modify path while following symbolic links.

PR #652, opened 2020-03-16
Enable tmux mouse support

jpszczolowski:
In my opinion it would be nice to have tmux mouse support enabled by default.

### DIFF
--- a/.tmux.conf
+++ b/.tmux.conf
@@ -18,7 +18,7 @@ set exit-empty off
 set -g status-position top
 
 set -g default-terminal "screen-256color"
-set -g mouse off
+set -g mouse on
 setw -g xterm-keys on # If this option is set, tmux will generate xterm(1)-style 
                       # function key sequences; these have a number included to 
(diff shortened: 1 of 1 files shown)


pj1031999:
I disagree. From my point of view it is very annoying.

jpszczolowski:
@pj1031999 FYI mouse mode has (almost) always been enabled in Mimiker but has been recently disabled in one of the final commits of https://github.com/cahirwpz/mimiker/pull/608 (I don't know why)

cahirwpz:
This is a matter of local user settings. One is free to configure `tmux` as he wishes and current configuration file should ease that process by providing extensive comments.

Hence I propose to revert to original settings (`off`) and a comment to `set -g mouse` line that explains what are the advantages and disadvantages of using this option.

jpszczolowski:
> This is a matter of local user settings. One is free to configure tmux as he wishes

@cahirwpz I've tried putting `set -g mouse on` into `.tmux.conf.local` file, but it doesn't seem to work.

cahirwpz:
Fixed by #654

PR #654, opened 2020-03-19
Fix tmux overrides.

mzr:
This seems to be the simplest solution. There is an env var `MIMIKER_REPO` set by `./launch`.
It is set to `./launch`'s cwd, and then it is used in `.tmux.conf` to source `/.tmux.conf.local`.


cahirwpz:
Tested it with `.tmux.conf.local` contents:
```
set -g mouse on
```
... and it works flawlessly.

@mzr Thanks for fixing that!

PR #657, opened 2020-03-24
Yet another attempt to clean up vfs_syscalls.c

xThaid:
* Fix function call in vfs_namelookupat routine
* Move repeatedly used code in a single functions
* Clean up vfs_open


sys/kern/vfs_syscalls.c
```diff
@@ -106,6 +107,23 @@ static int vfs_open(file_t *f, char *pathname, int flags, int mode) {
   return error;
 }
 
+/* Unlocks and decrements reference counter for both vnode and its parent */
+static void vnode_put2(vnode_t *v, vnode_t *dv) {
```
cahirwpz: Maybe `vnode_put_both` and `vnode_drop_both`?

xThaid: Done.

PR #662, opened 2020-04-01
Fix argument passing in system calls

xThaid:
032 ABI uses only 4 registers for parameter passing later arguments are passed on stack. This was not handled yet. Because of that, some system calls like `linkat` weren't working.


cahirwpz:
`kernel_tests_clang` failures are due to recent changes in `mips/ebase.S` which will be fixed soon.

PR #666, opened 2020-04-06
pmap cleanup

pj1031999:
Fixed usage of pte_t and pde_t.


cahirwpz:
Don't you want to clean up `boot.c` as well (i.e. line 77) ?

PR #673, opened 2020-04-14
Fix potential lost wakeup in cv_wait(_timed).

### LINKED ISSUE #668: Potential lost wakeup in condition variables
`cv_wait` and `cv_wait_timed` are supposed to atomically release a lock and go to sleep:
https://github.com/cahirwpz/mimiker/blob/4336f9a8ea297741370dd311959dcd8ec038dc38/sys/kern/condvar.c#L27-L34
Preemption is disabled between unlocking and going to sleep, but we might get an interrupt. Some interrupt filter routines call `cv_signal` (e.g. `atkbdc_intr`), so there is a potential lost wakeup.
One solution would be to use `WITH_INTR_DISABLED` instead of `WITH_NO_PREEMPTION`.

j-piecuch:
Fixes #668


cahirwpz:
I reckon this PR tries to workaround some fundamental issues. I'm still looking into #242 to decide.

cahirwpz:
I consider it as a temporary solution and accept it.

PR #674, opened 2020-04-14
Fix race between sig_kill and _sleepq_wait.

### LINKED ISSUE #669: Threads can go to sleep with pending signals
When sending a signal to a process that is sleeping interruptibly, we wake it up using `sleepq_abort`. However, the thread may go to sleep again before calling `on_user_exc_leave`, so the processing of the signal can be delayed indefinitely.
When going to sleep (interruptibly) in `_sleepq_wait`, we should check if there are pending signals, and if so, immediately return `EINTR`.
[Relevant NetBSD code](http://bxr.su/NetBSD/sys/kern/kern_sleepq.c#245)

j-piecuch:
Fixes #669

### DIFF
--- a/sys/kern/sleepq.c
+++ b/sys/kern/sleepq.c
@@ -216,6 +216,14 @@ static void sq_leave(thread_t *td, sleepq_chain_t *sc, sleepq_t *sq) {
 static int _sleepq_wait(void *wchan, const void *waitpt, sleep_t sleep) {
   thread_t *td = thread_self();
   int status = 0;
+#define EARLY_INTR(td, sleep)                                                  \
+  (((td)->td_flags & TDF_NEEDSIGCHK) != 0 && sleep == SLP_INTR)
+
+  /* If there are pending signals, interrupt the sleep immediately. */
+  WITH_SPIN_LOCK (&td->td_spin) {
+    if (EARLY_INTR(td, sleep))
+      return EINTR;
+  }
 
   if (waitpt == NULL)
     waitpt = __caller(0);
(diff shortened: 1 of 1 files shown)


sys/kern/sleepq.c
```diff
@@ -216,18 +216,33 @@ static void sq_leave(thread_t *td, sleepq_chain_t *sc, sleepq_t *sq) {
...
+   * We can also get a signal in here -- interrupt early if we got one.
+   * The first signal check is an optimization that saves us the call
+   * to sq_enter. */
 
   WITH_SPIN_LOCK (&td->td_spin) {
     if (td->td_flags & TDF_SLEEPY) {
       td->td_flags &= ~TDF_SLEEPY;
+      if (EARLY_INTR(td, sleep)) {
+        td->td_flags &= ~(TDF_SLPINTR | TDF_SLPTIMED);
+        return EINTR;
```
cahirwpz, line 244: At some point I'll remove all references to obscure `WITH_*` macros.
I'd much prefer to have a single exit point from the critical section.

sys/kern/sleepq.c
```diff
@@ -216,18 +216,33 @@ static void sq_leave(thread_t *td, sleepq_chain_t *sc, sleepq_t *sq) {
 static int _sleepq_wait(void *wchan, const void *waitpt, sleep_t sleep) {
   thread_t *td = thread_self();
   int status = 0;
+#define EARLY_INTR(td, sleep)                                                  \
```
cahirwpz, line 219: Generally speaking, we avoid macro definitions in procedure scope. 
You could've choosen a better name for it as well.
Maybe `TD_INTERRUPTED` ?

j-piecuch: Ok, I'll make it a static function, maybe something like `_sleepq_interrupted_early`?

cahirwpz: Seems ok.

PR #675, opened 2020-04-14
Add sched_yield syscall.

j-piecuch:


include/sys/sched.h
```diff
@@ -105,6 +105,12 @@ long sched_switch(void);
  */
 void sched_maybe_preempt(void);
 
+/*! \brief Unconditionally yield the CPU to another thread.
+ *
+ * \note Panics if interrupts or preemption are disabled.
+ */
+void sched_yield(void);
```
cahirwpz, line 112: `sched_yield` is defined in `include/sched.h` both on _Linux_ and _BSD_ systems.

sys/kern/sched.c
```diff
@@ -211,6 +211,16 @@ void sched_maybe_preempt(void) {
   }
 }
 
+void sched_yield(void) {
```
cahirwpz, line 214: In kernel space we can just call it `yield` as they do: http://bxr.su/NetBSD/sys/kern/kern_synch.c#272

include/sys/sched.h
cahirwpz: I'm in favour of hiding most declarations from this file from user-land programs with simple trick we use in other header files: http://bxr.su/NetBSD/sys/sys/sched.h#143 (`#ifdef _KERNEL`)

j-piecuch: ...my bad, I forgot to add the `include/sched.h` file to the commit.

PR #677, opened 2020-04-17
mtx_init & spin_init should set proper lock type

jpszczolowski:
Right now `spin_init` does not set spinlock's type to `LK_SPIN`, but it should. Look at `SPIN_INITIALIZER` implementation:
https://github.com/cahirwpz/mimiker/blob/019d2787d696c24e1d6e14b579cd5eaeedbbd92f/include/sys/spinlock.h#L33-L36

`spin_init` is used in `atkbdc` and `ns16550`:
https://github.com/cahirwpz/mimiker/blob/019d2787d696c24e1d6e14b579cd5eaeedbbd92f/sys/drv/ns16550.c#L163

So in fact `ns16550->lock` (of type `spin_t`) is used inside `cv_wait` **as a mutex**.
https://github.com/cahirwpz/mimiker/blob/019d2787d696c24e1d6e14b579cd5eaeedbbd92f/sys/drv/ns16550.c#L56-L60

But **we mustn't use mutexes in the lower half**!

### DIFF
--- a/sys/kern/mutex.c
+++ b/sys/kern/mutex.c
@@ -14,7 +14,7 @@ void mtx_init(mtx_t *m, lock_type_t type) {
   m->m_owner = NULL;
   m->m_count = 0;
   m->m_lockpt = NULL;
-  m->m_type = type;
+  m->m_type = type | LK_SLEEP;
 }
 
 void _mtx_lock(mtx_t *m, const void *waitpt) {
--- a/sys/kern/spinlock.c
+++ b/sys/kern/spinlock.c
@@ -14,7 +14,7 @@ void spin_init(spin_t *s, lock_type_t type) {
   s->s_owner = NULL;
   s->s_count = 0;
   s->s_lockpt = NULL;
-  s->s_type = type;
+  s->s_type = type | LK_SPIN;
 }
 
 void _spin_lock(spin_t *s, const void *waitpt) {


jpszczolowski:
I've just checked `mutex_init` and it also doesn't set lock type to `LK_SLEEP`.

sys/kern/mutex.c
```diff
@@ -14,7 +14,7 @@ void mtx_init(mtx_t *m, lock_type_t type) {
   m->m_owner = NULL;
   m->m_count = 0;
   m->m_lockpt = NULL;
-  m->m_type = type;
+  m->m_type = type | LK_SLEEP;
```
j-piecuch, line 17: `LK_SLEEP` is defined as `0`, so this changes nothing.
Are logical operations on enums idiomatic C?
I also have an issue with the name `LK_SLEEP`: we don't actually go to sleep on a sleepqueue
when blocking on a mutex. I think `LK_BLOCK` or `LK_BLOCKING` is more appropriate.
The documentation in `sys/lock.h` should be updated.
Why not have something like this:
```c
typedef int lock_type_t;
#define LK_SLEEP 0x0
#define LK_SPIN 0x1
#define LK_TYPE 0x1
#define LK_RECURSE 0x2
```
The type of a lock can be extracted using `type & LK_TYPE`, and `LK_RECURSE` is now an ordinary flag.

jpszczolowski: > `LK_SLEEP` is defined as 0, so this changes nothing.

Yes, but the same thing is done here:
https://github.com/cahirwpz/mimiker/blob/019d2787d696c24e1d6e14b579cd5eaeedbbd92f/include/sys/mutex.h#L26-L29
IMO it's more clear and errorproof in case of changing `LK_SLEEP` to some other value in the future.

> Are logical operations on enums idiomatic C?
> I also have an issue with the name `LK_SLEEP`:
> [...]

I agree that `lock_type_t` implementation may not be perfect, but this should probably be addressed in some other issue / PR? My PR is just fixing `*_init` functions so that spinlocks are not used as mutexes.

cahirwpz:
I'll merge it as is and pass the baton to @j-piecuch for further improvements of lock infrastructure.

PR #679, opened 2020-04-20
Adjustments to lock_type_t

j-piecuch:
- Add a new lock type (`LK_BLOCK`) that represents ordinary mutexes that use turnstiles
- Adjust lock type values so that it is possible to check that the caller doesn't try to set the lock type themselves in `mtx_init` and `spin_init`

### DIFF
--- a/include/sys/lock.h
+++ b/include/sys/lock.h
@@ -26,12 +26,12 @@ typedef enum {
    * When a thread tries to acquire a sleeping lock that is owned by another
    * thread, it will go to sleep on a sleepqueue. */
   LK_TYPE_SLEEP = 3,
-  /*!\var LK_RECURSE
+  /*!\var LK_RECURSIVE
    * \brief Flag indicating a recursive lock.
    *
    * The lock may be acquired by the owner multiple times, and must
    * be released exactly as many times. */
-  LK_RECURSE = 4
+  LK_RECURSIVE = 4
 } lk_attr_t;
 
 typedef struct spin spin_t;
@@ -66,7 +66,7 @@ static inline bool lk_sleep_p(lock_t l) {
 
 /* !\brief Predicates checking flags of a lock */
 static inline bool lk_recursive_p(lock_t l) {
-  return lk_attr(l) & LK_RECURSE;
+  return lk_attr(l) & LK_RECURSIVE;
 }
 
 #endif /* !_SYS_LOCK_H_ */
--- a/include/sys/lock.h
+++ b/include/sys/lock.h
@@ -1,7 +1,11 @@
 #ifndef _SYS_LOCK_H_
 #define _SYS_LOCK_H_
 
-/*! \enum Kind of lock. */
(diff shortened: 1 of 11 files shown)


include/sys/lock.h
```diff
@@ -1,27 +1,25 @@
...
-   * When a thread acquires spinning lock interrupts will be disabled.
-   */
-  LK_SPIN = 1,
-  /*!\var MTX_RECURSE
-   * \brief Type of recursive lock.
-   *
-   * The owner may lock it multiple times, but must release it as many times as
-   * she acquired it. */
-  LK_RECURSE = 2,
-} lock_type_t;
```
cahirwpz, line 24: IIRC There's one important advantage of having enums in C that may not be obvious. `gdb` understands values of enum types and print them symbolically. So instead of `3` you would get `LK_SPIN | LK_RECURSE`.

Please correct me if I'm wrong.

include/sys/lock.h
```diff
@@ -1,27 +1,25 @@
...
+ * When a thread tries to acquire a blocking lock that is owned by another
+ * thread, it will block and switch out to another thread. */
+#define LK_BLOCKING 0x1
+/*!\brief Type of spin locks.
+ *
+ * Interrupts are disabled upon acquiring a spin lock. */
+#define LK_SPIN 0x2
+/*!\brief Mask used to extract a lock's type (blocking/spin). */
+#define LK_TYPE 0x3
+/*!\brief Flag indicating a recursive lock.
```
jpszczolowski, line 18: You can consider changing `LK_TYPE` to `LK_TYPE_MASK`. IMO it would be easier to understand without looking into this comment, and more consistent with other code in the kernel.

include/sys/lock.h
j-piecuch: You're right, it's a significant advantage of using enums over ints. It seems that using enums as flags isn't as bad as I thought it was.

include/sys/lock.h
j-piecuch: Good point.

jpszczolowski:
What do you think of an idea of having a struct for lock attributes that contains two fields: enum-type and int-flags? It would be very self-descriptive.

j-piecuch:
> What do you think of an idea of having a struct for lock attributes that contains two fields: enum-type and int-flags? It would be very self-descriptive.

Yes, but it would double the size of the type... I don't know if we want that.
I added some comments to `lock_attrs_t` that hopefully clarify its usage.

jpszczolowski:
@j-piecuch I've added some getters in `sys/lock.h` file. Please have a look.

j-piecuch:
> @j-piecuch I've added some getters in `sys/lock.h` file. Please have a look.

LGTM. I also thought about changing `LK_{BLOCK,SPIN,SLEEP}` to `LK_TYPE_{BLOCK,SPIN,SLEEP}` to make it obvious that they represent the lock's type without looking at the documentation.

PR #680, opened 2020-04-21
sigprocmask implementation

j-piecuch:

PR #681, opened 2020-04-21
Add WUNTRACED and WCONTINUED support to wait4

j-piecuch:
`WUNTRACED` lets the caller wait for a child process to stop.
`WCONTINUED` lets the caller wait for a child process to continue.
This PR includes changes from #680


sys/kern/proc.c
```diff
@@ -339,48 +339,75 @@ static bool is_zombie(proc_t *p) {
...
-        *cldpidp = pid;
-        return 0;
+        if ((options & WCONTINUED) && (child->p_state == PS_NORMAL) &&
+            (child->p_flags & PF_CONTINUED)) {
+          child->p_flags &= ~PF_CONTINUED;
+          *status = MAKE_STATUS_SIG_CONT();
+          return 0;
+        }
+
+        if (pid > 0)
```
cahirwpz, line 404: Why do we need this condition?

sys/kern/proc.c
```diff
@@ -339,48 +339,75 @@ static bool is_zombie(proc_t *p) {
...
+          break;
       }
 
-      /* No zombie child was found. */
+      if (!any)
```
cahirwpz, line 408: Is there a way to check the same thing without introducing boolean variable?

sys/kern/proc.c
cahirwpz, line 404: Why do we need this condition?

j-piecuch: We don't _need_ it, it's an optimization. If `pid > 0`, we're waiting only for the child with the specified pid, so once we found it there's no point in iterating over the rest of the children.

sys/kern/proc.c
j-piecuch: I couldn't think of any obvious way to do it.

sys/kern/proc.c
```diff
@@ -339,64 +339,72 @@ static bool is_zombie(proc_t *p) {
...
+  if (pid == child->p_pid)
+    return true;
+  /* pid == -1 => any child  */
+  if (pid == -1)
+    return true;
+  /* pid == 0 => child with PGID same as ours */
+  if (pid == 0 && child->p_pgrp == pg)
+    return true;
+  /* pid < -1 => child with PGID equal to -pid */
+  if (pid < -1 && child->p_pgrp->pg_id != -pid)
```
j-piecuch, line 353: There should probably be `==` instead of `!=` here.

include/sys/signal.h
```diff
@@ -100,6 +100,7 @@ int sig_check(thread_t *td);
  * If the default action for a signal is to terminate the process and
  * corresponding signal handler is not set, the process calls `sig_exit`.
  *
+ * \note Must be called with all_proc_mtx and proc::p_mtx locked!
```
j-piecuch, line 103: I would clarify that the current process's `p_mtx` must be held.

sys/kern/signal.c
```diff
@@ -144,30 +144,28 @@ int do_sigprocmask(int how, const sigset_t *set, sigset_t *oset) {
...
-    proc->p_flags &= ~PF_STOPPED;
-    proc->p_flags |= PF_CONTINUED;
-    proc_t *parent = proc->p_parent;
-    if (parent)
-      cv_broadcast(&parent->p_waitcv);
+  if (p->p_state == PS_STOPPED && continued) {
+    p->p_state = PS_NORMAL;
+    p->p_flags &= ~PF_STOPPED;
+    p->p_flags |= PF_CONTINUED;
+    cv_broadcast(&p->p_parent->p_waitcv);
```
j-piecuch, line 168: Doesn't `init` have no parent (i.e. its `p_parent` is `NULL`)?

sys/kern/proc.c
j-piecuch: I guess we could introduce an `error` variable like the one here: http://bxr.su/NetBSD/sys/kern/kern_exit.c#1021

cahirwpz: Done!

sys/kern/proc.c
cahirwpz: Added a comment to clarify that.

sys/kern/proc.c
cahirwpz: Fixed!

include/sys/signal.h
cahirwpz: Done!

sys/kern/signal.c
cahirwpz: I've been thinking of introducing `proc0` structure as a [sentinel value](https://en.wikipedia.org/wiki/Sentinel_value) to fix similar issues. I don't want to have extra checks scattered around the code to test for that very rare case where we handle `init` process.

PR #682, opened 2020-04-21
Increase timeouts in run_tests.py

jpszczolowski:
Sometimes we cannot see why a kernel test failed because there is a timeout while waiting for gdb to spawn / finish executing our command.

This PR increases timeouts in `run_tests.py` by some arbitrary values.


jpszczolowski:
This PR does not solve the problem. Please see issue #686

PR #683, opened 2020-04-22
Move callout structures from stack to dynamically or statically allocated memory

cahirwpz:
* Move thread sleep callout from stack to thread structure.
* Review `tests/sleepq.c`
* Move one callout in `tests/callout.c` to bss section.

### DIFF
--- a/sys/kern/thread.c
+++ b/sys/kern/thread.c
@@ -111,6 +111,7 @@ void thread_delete(thread_t *td) {
 
   kmem_free(td->td_kstack.stk_base, PAGESIZE);
 
+  callout_drain(&td->td_slpcallout);
   sleepq_destroy(td->td_sleepqueue);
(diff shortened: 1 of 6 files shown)


sys/kern/thread.c
```diff
@@ -50,6 +51,7 @@ static void thread_init(thread_t *td, prio_t prio) {
   td->td_lock = MTX_INITIALIZER(0);
   cv_init(&td->td_waitcv, "thread waiters");
   LIST_INIT(&td->td_contested);
+  bzero(&td->td_slpcallout, sizeof(callout_t));
```
j-piecuch, line 54: I think we should also drain the thread's callout before freeing it in `thread_delete`, otherwise the callout function (e.g. `sq_timeout`) may refer to freed memory.

cahirwpz: Right! I've applied corresponding change.

PR #688, opened 2020-04-28
Don't unlock the process in sig_kill.

j-piecuch:
This only led to errors in sig_kill where we forgot to unlock it.


cahirwpz:
LGTM However I'd add an information here: https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/include/sys/signal.h?r=0d9fc7f1#82 that it returns with `p::p_lock` being held as well.

PR #689, opened 2020-04-28
Implement auto-reap.

j-piecuch:
If the parent of an exiting child ignores SIGCHLD, the child shouldn't leave any state for the parent to reap.


cahirwpz:
Resolve conflicts please.

cahirwpz:
LGTM

PR #690, opened 2020-04-28
Rework PID management

j-piecuch:
It shouldn't be possible to allocate a new PID such that there is a process group with PGID equal to the new PID (that would mean that processes could accidentally become leaders of unrelated process groups). Currently, no such checks are performed on PID allocation. This PR fixes this issue and makes PID lookups faster by using a table instead of a linked list.

### DIFF
--- a/sys/kern/exec.c
+++ b/sys/kern/exec.c
@@ -398,9 +398,8 @@ __noreturn void run_program(const char *path, char *const *argv,
                             char *const *envv) {
   proc_t *p = proc_self();
 
-  assert(pgrp_enter(p, p->p_pid) == 0);
-
   assert(p != NULL);
+  assert(pgrp_enter(p, p->p_pid) == 0);
 
   klog("Starting program '%s'", path);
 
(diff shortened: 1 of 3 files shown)


sys/kern/exec.c
```diff
@@ -398,9 +398,8 @@ __noreturn void run_program(const char *path, char *const *argv,
                             char *const *envv) {
   proc_t *p = proc_self();
 
-  pgrp_enter(p, 1);
-
   assert(p != NULL);
+  assert(pgrp_enter(p, p->p_pid) == 0);
```
cahirwpz, line 402: It's always a bad idea to pass expression that have side effects to macros.
What if I decide to compile kernel with `NDEBUG` defined?
https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/include/assert.h?r=fc854717#6

j-piecuch: Oh, right. Fixed.

j-piecuch:
This PR is superseded by #709, closing.

PR #693, opened 2020-04-28
KASAN & allocators: cleanup

jpszczolowski:
* Remove `kasan_` prefix from static KASAN functions
* Better KASAN test fail report by calling `ktest_failure`
* Use non-static `kasan_mark_valid` instead of static `kasan_shadow_clean` (they did the same thing)
* Delete `kasan_shadow_clean`
* Cleanup in `malloc.c` & `pool.c`
* Provide additional `vmem`'s `P_BT_BOOTPAGE` page

### DIFF
--- a/build/arch.mips.mk
+++ b/build/arch.mips.mk
@@ -7,11 +7,11 @@ ELFTYPE := elf32-littlemips
 # Set KASAN flags
 ifeq ($(KASAN), 1)
 ifeq ($(KERNEL), 1)
-	# Added to files that are sanitized
-	CFLAGS_KASAN = -fsanitize=kernel-address -fasan-shadow-offset=0xD8000000 \
-								 --param asan-globals=1 \
-								 --param asan-stack=1 
-	# Added to all files
-	CFLAGS += -DKASAN
+  # Added to files that are sanitized
+  CFLAGS_KASAN = -fsanitize=kernel-address -fasan-shadow-offset=0xD8000000 \
+                 --param asan-globals=1 \
+                 --param asan-stack=1 
+  # Added to all files
+  CFLAGS += -DKASAN
 endif
 endif
--- a/build/arch.mips.mk
+++ b/build/arch.mips.mk
@@ -8,7 +8,9 @@ ELFTYPE := elf32-littlemips
 ifeq ($(KASAN), 1)
 ifeq ($(KERNEL), 1)
 	# Added to files that are sanitized
-	CFLAGS_KASAN = -fsanitize=kernel-address --param asan-globals=1 --param asan-stack=1 -fasan-shadow-offset=0xD8000000
+	CFLAGS_KASAN = -fsanitize=kernel-address -fasan-shadow-offset=0xD8000000 \
+								 --param asan-globals=1 \
+								 --param asan-stack=1 
 	# Added to all files
 	CFLAGS += -DKASAN
 endif
--- a/include/mips/kasan.h
+++ b/include/mips/kasan.h
@@ -25,11 +25,11 @@ static_assert(KASAN_MD_SHADOW_SIZE % SUPERPAGESIZE == 0,
   (KASAN_MD_SHADOW_START -                                                     \
    (KASAN_MD_SANITIZED_START >> KASAN_SHADOW_SCALE_SHIFT))
 
-__always_inline static inline int8_t *md_addr_to_shad(uintptr_t addr) {
+__always_inline static inline int8_t *kasan_md_addr_to_shad(uintptr_t addr) {
   return (int8_t *)(KASAN_MD_OFFSET + (addr >> KASAN_SHADOW_SCALE_SHIFT));
 }
 
-__always_inline static inline bool md_addr_supported(uintptr_t addr) {
+__always_inline static inline bool kasan_md_addr_supported(uintptr_t addr) {
   return addr >= KASAN_MD_SANITIZED_START && addr < KASAN_MD_SANITIZED_END;
 }
 
--- a/include/mips/kasan.h
+++ b/include/mips/kasan.h
@@ -25,11 +25,11 @@ static_assert(KASAN_MD_SHADOW_SIZE % SUPERPAGESIZE == 0,
   (KASAN_MD_SHADOW_START -                                                     \
    (KASAN_MD_SANITIZED_START >> KASAN_SHADOW_SCALE_SHIFT))
 
-__always_inline static inline int8_t *kasan_md_addr_to_shad(uintptr_t addr) {
+__always_inline static inline int8_t *md_addr_to_shad(uintptr_t addr) {
   return (int8_t *)(KASAN_MD_OFFSET + (addr >> KASAN_SHADOW_SCALE_SHIFT));
 }
 
-__always_inline static inline bool kasan_md_addr_supported(uintptr_t addr) {
+__always_inline static inline bool md_addr_supported(uintptr_t addr) {
   return addr >= KASAN_MD_SANITIZED_START && addr < KASAN_MD_SANITIZED_END;
 }
 
--- a/sys/kern/pool.c
+++ b/sys/kern/pool.c
@@ -67,9 +67,7 @@ typedef struct pool_item {
 
 /* Pool of pool_t objects. */
 static pool_t P_POOL[1];
-
-#define POOL_BOOTPAGE_CNT 1
-static alignas(PAGESIZE) uint8_t P_POOL_BOOTPAGE[PAGESIZE * POOL_BOOTPAGE_CNT];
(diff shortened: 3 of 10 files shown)


sys/kern/pool.c
```diff
@@ -196,32 +203,31 @@ void *pool_alloc(pool_t *pool, unsigned flags) {
  * (maybe leave one) */
 void pool_free(pool_t *pool, void *ptr) {
   debug("pool_free: pool = %p, ptr = %p", pool, ptr);
-
   assert(pool->pp_state == ALIVE);
 
   /* Mark the item as invalid */
   kasan_mark(ptr, 0, pool->pp_itemsize, KASAN_CODE_POOL_USE_AFTER_FREE);
 
-  WITH_MTX_LOCK (&pool->pp_mtx) {
```
cahirwpz, line 205: Is this change really necessary?

sys/kern/pool.c
```diff
@@ -64,6 +64,12 @@ typedef struct pool_item {
   unsigned long pi_data[0] __aligned(PI_ALIGNMENT);
 } pool_item_t;
 
+/* Pool of pool_t objects. */
+static pool_t P_POOL[1];
+
+#define POOL_BOOTPAGE_CNT 1
+static alignas(PAGESIZE) uint8_t P_POOL_BOOTPAGE[POOL_BOOTPAGE_CNT][PAGESIZE];
```
cahirwpz, line 71: I'd much prefer to have it defined as:
```c
static alignas(PAGESIZE) uint8_t P_POOL_BOOTPAGE[PAGESIZE*2];
```

sys/kern/vmem.c
```diff
@@ -76,11 +76,17 @@ typedef struct bt {
...
+ * number of free tags. For more information, please see bt_alloc and bt_refill
+ * methods in NetBSD's vmem and M_NOGROW flag in Mimiker. */
+#define VMEM_BT_BOOTPAGE_CNT 2
+static alignas(PAGESIZE) uint8_t P_BT_BOOTPAGE[VMEM_BT_BOOTPAGE_CNT][PAGESIZE];
 
 void vmem_bootstrap(void) {
   pool_add_page(P_VMEM, P_VMEM_BOOTPAGE);
-  pool_add_page(P_BT, P_BT_BOOTPAGE);
+  for (int i = 0; i < VMEM_BT_BOOTPAGE_CNT; i++)
+    pool_add_page(P_BT, P_BT_BOOTPAGE[i]);
```
cahirwpz, line 89: Could you actually slightly change `pool_add_page` interface to accept size of second argument aligned to pagesize? I'd like to have:

```
static alignas(PAGESIZE) uint8_t P_BT_BOOTPAGE[PAGESIZE*2];

void vmem_bootstrap(void) {
  ...
  pool_add_page(P_BT, P_BT_BOOTPAGE, sizeof(P_BT_BOOTPAGE));
}
```

cahirwpz: This will require to modify slightly `add_slab` as well, but that's ok.

build/arch.mips.mk
```diff
@@ -8,7 +8,9 @@ ELFTYPE := elf32-littlemips
 ifeq ($(KASAN), 1)
 ifeq ($(KERNEL), 1)
 	# Added to files that are sanitized
-	CFLAGS_KASAN = -fsanitize=kernel-address --param asan-globals=1 --param asan-stack=1 -fasan-shadow-offset=0xD8000000
+	CFLAGS_KASAN = -fsanitize=kernel-address -fasan-shadow-offset=0xD8000000 \
+								 --param asan-globals=1 \
```
cahirwpz, line 12: Could use better alignment.

```
	CFLAGS_KASAN = -fsanitize=kernel-address -fasan-shadow-offset=0xD8000000 \
                       --param asan-globals=1 --param asan-stack=1 
```

include/mips/kasan.h
```diff
@@ -25,11 +25,11 @@ static_assert(KASAN_MD_SHADOW_SIZE % SUPERPAGESIZE == 0,
   (KASAN_MD_SHADOW_START -                                                     \
    (KASAN_MD_SANITIZED_START >> KASAN_SHADOW_SCALE_SHIFT))
 
-__always_inline static inline int8_t *kasan_md_addr_to_shad(uintptr_t addr) {
+__always_inline static inline int8_t *md_addr_to_shad(uintptr_t addr) {
```
cahirwpz, line 28: Mmm... everything else in this file is prefixed with `KASAN` why not leave the prefix for these functions as well. We can include this header from any file (not that it would make much sense).

sys/kern/pool.c
jpszczolowski: In quarantine PR, this function will be renamed to `_pool_free` and called from quarantine code (if KASAN is enabled) or from `pool_free` (if KASAN is disabled) **with mutex already acquired**, so this `WITH_MTX_LOCK (&pool->pp_mtx)` will be deleted anyway. I wanted to move this code change from that PR to this one, as it does not introduce any new features.

build/arch.mips.mk
jpszczolowski: As GitHub displays tab using 6 spaces, the alignment will never look good here if we indent `CFLAGS_KASAN` using tab. Right now my code change looks properly in an editor that displays tab as 2 spaces.

I've discovered that the whole block within `ifeq ... endif` can be indented with spaces -- no tab before `CFLAGS_KASAN` needed. Maybe that would be the solution?

include/mips/kasan.h
jpszczolowski: I thought that it would be consistent if non-static functions begin with `kasan_`, and static ones don't. But this file is included from `mips/boot.c`, so I agree with you. Fixed in https://github.com/cahirwpz/mimiker/pull/693/commits/4a22641144052705330fe686599150bac4f922c7

sys/kern/vmem.c
jpszczolowski: Should it create one big slab (so we also need to add field `ph_size` to `pool_slab_t`, as it'll needed in `destroy_slab`), or many single-paged slabs?

sys/kern/pool.c
jpszczolowski: In quarantine PR, `POOL_BOOTPAGE_CNT` will have to be larger if KASAN is enabled (quarantine increases `sizeof(pool_t)` significantly), that's why I introduced `POOL_BOOTPAGE_CNT`.

So, would:
```
#define POOL_BOOTPAGE_CNT 1
static alignas(PAGESIZE) uint8_t P_POOL_BOOTPAGE[PAGESIZE * POOL_BOOTPAGE_CNT];
```
be OK?

And:
```
#ifdef KASAN
#define POOL_BOOTPAGE_CNT some-bigger-number
#else
#define POOL_BOOTPAGE_CNT 1
#endif /* !KASAN */
static alignas(PAGESIZE) uint8_t P_POOL_BOOTPAGE[PAGESIZE * POOL_BOOTPAGE_CNT];
```
in the future.

build/arch.mips.mk
cahirwpz: > I've discovered that the whole block [...]
Yes, please do so.

sys/kern/pool.c
cahirwpz: It's no use of trying to give the structure to a workaround. I'll finally solve the problem with `M_NOGROW` at some point.

sys/kern/pool.c
cahirwpz: Ok.

sys/kern/vmem.c
cahirwpz: Yes. One big slab is preferred solution.

sys/kern/pool.c
jpszczolowski: So what do you expect me to do here?

PS I think that size of `P_POOL_BOOTPAGE` isn't related to `M_NOGROW`. During `pool_bootstrap` we are calling `INVOKE_CTORS(pool_ctor_table)` which allocates various memory pools out of "master pool" `P_POOL`. As this happens before `kmem_bootstrap`, "master pool" has to use static memory.

build/arch.mips.mk
jpszczolowski: Fixed in https://github.com/cahirwpz/mimiker/pull/693/commits/7bc832c02d22ee09c29473c71830ed225b61ab16

sys/kern/vmem.c
jpszczolowski: Implemented in https://github.com/cahirwpz/mimiker/pull/693/commits/b42c0feec66a339a33a2bed40373d183962e56fb

sys/kern/pool.c
jpszczolowski: Fixed in https://github.com/cahirwpz/mimiker/pull/693/commits/e4eca308c7799b49baf32adb53ddf638f6c5d127

PR #697, opened 2020-04-30
KASAN: instrument allocators & memory-related functions

jpszczolowski:
* Add `alloca` instrumentation
* Add `copyin`, `copyinstr`, `copyout` instrumentation
* Add `pool` & `kmalloc` instrumentation that detects overflow and use-after-free bugs


cahirwpz:
Please note that you should use `#if KASAN` instead of `#ifdef KASAN` since `KASAN` is always defined but it has (boolean) value now.

PR #700, opened 2020-05-08
Credentials syscalls

franciscozdo:


include/sys/cred.h
```diff
@@ -0,0 +1,30 @@
...
+#include <sys/types.h>
+#include <sys/proc.h>
+#include <stdint.h>
+
+#define NGROUPS_MAX 16
+
+/*
+ * Kernel view of credencials
+ */
+struct cred {
```
xThaid, line 13: It would be better to use `typedef` in declaration of struct, e.g.:
```c
typedef struct cred {
...
} cred_t;
``` 
And then use type `cred_t` instead of `struct cred`.

include/sys/proc.h
```diff
@@ -49,6 +50,7 @@ struct proc {
   TAILQ_ENTRY(proc) p_child;  /* (a) link on parent's children list */
   thread_t *p_thread;         /* (@) the only thread running in this process */
   pid_t p_pid;                /* (!) Process ID */
+  struct cred *p_cred;        /* Process credentials */
   char *p_elfpath;            /* (!) path of loaded elf file */
```
jpszczolowski, line 54: Please add an information about `p_cred`'s corresponding lock
For instance: if `p_cred` is protected by `p_lock`, change the comment to: `/* (@) Process credentials */`

sys/kern/cred_syscalls.c
```diff
@@ -0,0 +1,87 @@
+#include <sys/cred.h>
+#include <sys/proc.h>
+#include <sys/errno.h>
+
+int do_getresuid(proc_t *p, uid_t *ruid, uid_t *euid, uid_t *suid) {
+  cred_t cr = p->p_cred;
+  if (ruid != NULL)
```
jpszczolowski, line 7: It'd be more effective to use a pointer to `cred_t` instead of creating a local copy of the structure on the stack.
```
cred_t *cr = &p->p_cred;
if (ruid != NULL)
  *ruid = cr->cr_ruid;
[...]
```

sys/kern/cred_syscalls.c
```diff
@@ -0,0 +1,87 @@
...
+    *euid = cr.cr_euid;
+  if (suid != NULL)
+    *suid = cr.cr_suid;
+  return 0;
+}
+
+int do_getresgid(proc_t *p, gid_t *rgid, gid_t *egid, gid_t *sgid) {
+  cred_t cr = p->p_cred;
+  if (rgid != NULL)
+    *rgid = cr.cr_rgid;
```
jpszczolowski, line 19: Ditto.

sys/kern/cred_syscalls.c
```diff
@@ -0,0 +1,87 @@
...
+    *sgid = cr.cr_sgid;
+  return 0;
+}
+
+int do_setresuid(proc_t *p, uid_t ruid, uid_t euid, uid_t suid) {
+  proc_lock(p);
+
+  cred_t *oldcred = &p->p_cred;
+  cred_t newcred; /* XXX: maybe not on stack */
+  cred_copy(&newcred, oldcred);
```
pj1031999, line 32: I think that you are wrong here.

`void cred_copy(cred_t *src, cred_t *dest)`
So you copy _random_ value from the stack to your old credentials and lost them forever.

franciscozdo: Thanks. I changed `cred_copy` to look like `memcpy`. Now it will be less confusing. 
```void cred_copy(cred_t *dest, cred_t src)```

PR #702, opened 2020-05-12
Implement statvfs & fstatvfs system calls

xThaid:
[statvfs & fstatvfs](https://netbsd.gw.com/cgi-bin/man-cgi?statvfs+2+NetBSD-current) are now implemented, but at the moment they are quite nonfunctional since none of our file systems supports it.


sys/kern/syscalls.c
```diff
@@ -861,3 +863,40 @@ static int sys_sched_yield(proc_t *p, void *args, register_t *res) {
   thread_yield();
   return 0;
 }
+
+static int sys_statvfs(proc_t *p, statvfs_args_t *args, register_t *res) {
+  int error;
+  const char *u_path = args->path;
+  statvfs_t *u_buf = args->buf;
+  statvfs_t *buf = kmem_alloc(sizeof(struct statvfs), M_WAITOK);
```
cahirwpz, line 871: I know `statvfs` is quite large, but come on – our stack is almost empty upon kernel entry ;-)

Beside `kmem_alloc` is low-level allocator that should be used only for multiple page sized regions. It's not ok to use it for structure allocations.

cahirwpz:
LGTM

PR #703, opened 2020-05-12
Gettimeofday syscall.

wiklam:
Adding gettimeofday syscall with user test.


include/sys/time.h
```diff
@@ -51,6 +59,21 @@ typedef struct bintime {
     .sec = 0, .frac = ((1ULL << 63) / (hz)) << 1                               \
   }
 
+static inline time_t tm2sec(tm_t t) {
+  time_t res = 0;
+  int month_int_days[13] = {0,   0,   31,  59,  90,  120, 151,
```
pj1031999, line 64: Why not `static const`?

include/sys/time.h
```diff
@@ -51,6 +59,21 @@ typedef struct bintime {
     .sec = 0, .frac = ((1ULL << 63) / (hz)) << 1                               \
   }
 
+static inline time_t tm2sec(tm_t t) {
```
pj1031999, line 62: `static` in header file?

include/sys/time.h
wiklam: My bad. Thank you!

jpszczolowski: @pj1031999 `const` OK, but why `static` here?

include/sys/time.h
```diff
@@ -51,6 +59,20 @@ typedef struct bintime {
...
+  time_t res = 0;
```
jpszczolowski, line 63: Why not `tm_t *` instead of `tm_t` function argument? This structure is quite big and now it's copied by value.

include/sys/time.h
```diff
@@ -186,11 +234,15 @@ struct itimerval {
 
 #ifdef _KERNEL
 
-/* XXX: Do not use this function, it'll get removed. */
-timeval_t get_uptime(void);
+/* Time measured from the  start of system. */
+bintime_t binuptime(void);
```
jpszczolowski, line 238: Double space in the comment

include/sys/timer.h
```diff
@@ -42,6 +42,10 @@ typedef struct timer {
   void *tm_priv;              /*!< private data (usually device_t *) */
 } timer_t;
 
+/* Used by rtc to initalize system boottime using rtc */
+void boottime_init(tm_t *t);
+/* Used to set/change system boottime */
+void tm_setclock(const timespec_t *ts);
 /*! \brief Used by a driver to make timer device available to the system. */
```
jpszczolowski, line 49: You can consider changing the comments to `/*! \brief ... */` format to be consistent with the rest of the file

sys/kern/time.c
```diff
@@ -2,7 +2,19 @@
 #include <sys/time.h>
 
 int do_clock_gettime(clockid_t clk, timespec_t *tp) {
-  return ENOTSUP;
+
+  switch (clk) {
```
jpszczolowski, line 6: Unnecessary newline

include/sys/time.h
wiklam: Stayed by the convention that convert function takes the argument in such a way, but for this structure, it will be a better idea to use a pointer.

bin/utest/time.c
```diff
@@ -0,0 +1,13 @@
+#include "utest.h"
+
+#include <stdio.h>
+#include <sys/time.h>
+
+int test_gettimeofday(void) {
+  timeval_t time;
+  gettimeofday(&time, NULL);
+
+  printf("Timeofday: %lld second and %d microsecond\n", time.tv_sec,
```
cahirwpz, line 10: This test doesn't test anything.
The test should run without any manual inspection.
Only when a test fails manual inspection is required.
Please rethink how are you going to test `gettimeofday`.

include/sys/time.h
```diff
@@ -51,6 +59,22 @@ typedef struct bintime {
     .sec = 0, .frac = ((1ULL << 63) / (hz)) << 1                               \
   }
 
+static inline time_t tm2sec(tm_t *t) {
```
cahirwpz, line 62: This procedure is way too big to be exposed as `static inline`!

include/sys/time.h
```diff
@@ -5,13 +5,21 @@
 
 #define CLK_TCK 100 /* system clock ticks per second */
 
+/* 1ns = (2^64) / 1000000000 */
+#define BINTIME_SCALE_NS ((uint64_t)18446744073ULL)
```
cahirwpz, line 9: I don't like the idea of polluting public namespace with macrodefinitions that are used to implement a single procedure used by the kernel.

include/sys/timer.h
```diff
@@ -42,6 +42,8 @@ typedef struct timer {
   void *tm_priv;              /*!< private data (usually device_t *) */
 } timer_t;
 
+/*! \brief  Used to set/change the system boottime */
+void tm_setclock(const bintime_t *ts);
```
cahirwpz, line 46: s/ts/bt/

sys/kern/timer.c
```diff
@@ -9,6 +9,7 @@
 static mtx_t timers_mtx = MTX_INITIALIZER(0);
 static timer_list_t timers = TAILQ_HEAD_INITIALIZER(timers);
 static timer_t *time_source = NULL;
+static bintime_t timebasebin = BINTIME(0);
```
cahirwpz, line 12: Strange choice for naming. Shouldn't it be `basebintime`?

wiklam: Agree, but in my opinion, `boottime` will be the best option.

PR #706, opened 2020-05-18
Remove unused bus_spac_ map flags

pj1031999:
According to the results of the investigation, the following flags are not used
in NetBSD and FreeBSD. Let's remove them.

- BUS_SPACE_MAP_CACHEABLE
- BUS_SPACE_MAP_LINEAR
- BUS_SPACE_MAP_PREFETCHABLE


cahirwpz:
LGTM

PR #708, opened 2020-05-19
sigsuspend() and signal handler mask implementation

j-piecuch:
This fixes an issue with `ksh` where the shell would wait forever for the completion of a command.
I encountered a couple of difficulties:
- I'm unable to call `sleepq_abort()` when holding `td_spin` of the thread I'm trying to wake up.
- `_sleepq_wait()` returns immediately if `TDF_NEEDSIGCHK` is set, but I think this flag is only a safe approximation of whether any signal is actually pending. I think we should call `sig_pending()` in `_sleepq_interrupted_early()`, but that requires acquiring a process lock.


sys/mips/signal.c
```diff
@@ -68,32 +69,40 @@ int sig_send(signo_t sig, sigaction_t *sa) {
 }
 
 int sig_return(void) {
+  int error = 0;
   thread_t *td = thread_self();
-  SCOPED_MTX_LOCK(&td->td_lock);
   sig_ctx_t ksc;
-  exc_frame_t *uframe = td->td_uframe;
```
cahirwpz, line 74: Please strive to minimize volume of code changes – that'll make reviews go faster. Especially avoid unnecessary code movement or indentation. I think current `sig_return` could made `static` and renamed to `copy_sigcontext`. While new `sig_return` would contain new code and call to `copy_sigcontext`.

sys/mips/signal.c
```diff
@@ -26,15 +27,15 @@ static void stack_unusable(thread_t *td, register_t sp) {
   __unreachable();
 }
 
-int sig_send(signo_t sig, sigaction_t *sa) {
+int sig_send(signo_t sig, sigset_t *return_mask, sigaction_t *sa) {
```
cahirwpz, line 30: Though we diverged from BSD naming for signal functions, I'd still keep signatures as close as possible to their BSD counterparts. So in this case:
``` 
int sig_send(signo_t sig, sigaction_t *sa, sigset_t *mask) ...
```
Please note that we prefer to use short names in function signatures, as BSDs do.

sys/kern/signal.c
```diff
@@ -129,14 +131,47 @@ int do_sigprocmask(int how, const sigset_t *set, sigset_t *oset) {
...
+    WITH_SPIN_LOCK (&td->td_spin) {
+      if (sig_pending(td))
+        td->td_flags |= TDF_NEEDSIGCHK;
+      else
+        td->td_flags &= ~TDF_NEEDSIGCHK;
+    }
+  }
+
+  int error;
+  while ((error = sleepq_wait_intr(td, "sigsuspend()")) == 0)
```
cahirwpz, line 167: When does spurious wakeup happen?

sys/kern/signal.c
```diff
@@ -129,14 +131,47 @@ int do_sigprocmask(int how, const sigset_t *set, sigset_t *oset) {
...
+    klog("Spurious wakeup in sigsuspend()!");
+
+  assert(error = EINTR);
```
cahirwpz, line 170: Umm... `==` instead of `=` ?

include/sys/thread.h
```diff
@@ -104,8 +108,9 @@ typedef struct thread {
   char *td_name;   /*!< (@) name of thread */
   tid_t td_tid;    /*!< (@) thread identifier */
   /* thread state */
-  thread_state_t td_state;    /*!< (!) thread state */
-  volatile uint32_t td_flags; /*!< (!) TDF_* flags */
+  thread_state_t td_state;        /*!< (!) thread state */
+  volatile uint32_t td_flags;     /*!< (!) TDF_* flags */
+  volatile tdp_flags_t td_pflags; /*!< (p) TDP_* flags */
```
cahirwpz, line 113: `td_pflags` counterpart in _FreeBSD_ is private (that's what `p` in `pflags` stands for) to thread.
It looks somewhat suspicious that you need to acquire process' lock to use this field.
Can you explain?

sys/mips/signal.c
j-piecuch: ... but the signature is different anyway (I'm assuming you're referring to FreeBSD's `sendsig`):
- `sendsig`'s first argument is the handler, not the signal number
- `sendsig`'s second argument is a `ksiginfo` structure, which contains the signal number
So, if we really wanted to "mimic" `sendsig`'s signature, `sigaction_t` should be first and `signo_t` should be second.
Besides, OpenBSD's signature looks like this:
```c
void sendsig(sig_t _catcher, int _sig, sigset_t _mask, const siginfo_t *_si);
```
...so according to that signature, `sig_send`'s signature should look like:
```c
int sig_send(sigaction_t *sa, signo_t sig, sigset_t *mask)
```
or
```c
int sig_send(sigaction_t *sa, sigset_t *mask, signo_t sig)
```
depending on whether you decide to substiture `sig` for the signal number or the `siginfo_t` in the OpenBSD signature.

sys/mips/signal.c
j-piecuch: I find that such refactoring only makes it harder to follow the code when reading it -- I can understand it when it leads to code reuse. To minimize noise in the diff, you can ignore whitespace changes by checking "hide whitespace changes" in the settings at the top of the page. Unfortunately, you need to set it manually every time, there's no way to make it the default.

sys/mips/signal.c
```diff
@@ -68,32 +69,40 @@ int sig_send(signo_t sig, sigaction_t *sa) {
...
-  sig_ctx_t *scp = (sig_ctx_t *)((intptr_t *)uframe->sp + 1);
-  int error = copyin_s(scp, ksc);
-  if (error)
-    return error;
-  if (ksc.magic != SIG_CTX_MAGIC) {
-    klog("User context at %p corrupted!", scp);
-    sig_trap(uframe, SIGILL);
-    return EINVAL;
+
+  SCOPED_MTX_LOCK(&td->td_proc->p_lock);
```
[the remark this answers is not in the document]
j-piecuch: Good point, only `do_sigprocmask` needs it.

sys/kern/signal.c
j-piecuch: Fixed.

sys/kern/signal.c
j-piecuch: It's not supposed to happen, but It's what FreeBSD does [here](http://bxr.su/FreeBSD/sys/kern/kern_sig.c#1550). Now that I think about it, an assertion would be better.
Maybe something along the lines of [`kpause`](http://bxr.su/NetBSD/sys/kern/kern_synch.c#230) would be handy here.

include/sys/thread.h
j-piecuch: NetBSD's implementation uses a `l_sigrestore` field in `struct lwp`, which is protected by the thread's process lock: http://bxr.su/NetBSD/sys/kern/sys_sig.c#619
I want to set `TDP_OLDMASK` and do the mask manipulation atomically, so protecting it with `p_lock` seems reasonable, since `td_sigmask` is protected by it as well.

sys/mips/signal.c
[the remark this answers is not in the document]
j-piecuch: Fixed.

include/sys/thread.h
j-piecuch: Also the `p` in `pflags` here stands for "process", I wasn't aware that it stood for "private" in FreeBSD.

sys/mips/signal.c
cahirwpz: Ok. I took me a few seconds to find the option.

sys/kern/signal.c
```diff
@@ -200,7 +235,10 @@ void sig_kill(proc_t *p, signo_t sig) {
       /* If the thread is sleeping interruptibly (!), wake it up, so that it
        * continues execution and the signal gets delivered soon. */
       if (td_is_interruptible(td)) {
-        sleepq_abort(td);
+        /* XXX Maybe TDF_NEEDSIGCHK should be protected by a different lock? */
+        spin_unlock(&td->td_spin);
```
cahirwpz, line 239: Ok. I see... there's a problem with transferring lock ownership. That won't be fixed till summer :(

include/sys/thread.h
cahirwpz: IMHO it feels much more natural for `TDP_OLDMASK` bit to be a part of thread's private state. That's something that I initally expected and got curious why other choice was made.

sys/mips/signal.c
j-piecuch: Done.

include/sys/thread.h
```diff
@@ -135,6 +140,7 @@ typedef struct thread {
   /* signal handling */
   sigset_t td_sigpend; /*!< (p) Pending signals for this thread. */
   sigset_t td_sigmask; /*!< (p) Signal mask */
+  sigset_t td_oldmask; /*!< (p) Signal mask from before sigsuspend() */
```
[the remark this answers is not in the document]
j-piecuch: Done.

include/sys/thread.h
j-piecuch: On second thought, `td_oldsigmask` and `TDP_OLDSIGMASK` are only accessed by the owning thread, so making them thread-private is a good idea. I have made the corresponding changes.

sys/kern/signal.c
j-piecuch: I'm not sure if it would be better, but I don't mind it if you prefer it that way. I have made the corresponding change.

cahirwpz:
I like how you attach a meaningful description to each commit in the branch. It's easy afterwards to prepare single large commit for `master` branch.

PR #709, opened 2020-05-20
Manage PIDs and PGIDs using hashtables.

j-piecuch:
This PR supersedes #690


sys/kern/proc.c
```diff
@@ -29,26 +38,43 @@ static proc_list_t proc_list = TAILQ_HEAD_INITIALIZER(proc_list);
 static proc_list_t zombie_list = TAILQ_HEAD_INITIALIZER(zombie_list);
 static pgrp_list_t pgrp_list = TAILQ_HEAD_INITIALIZER(pgrp_list);
 
-/* Pid 0 is never available, because of its special treatment by some
- * syscalls e.g. kill. */
-static bitstr_t pid_used[bitstr_size(NPROC)] = {1};
+static pgrp_t *pgrp_lookup(pgid_t pgid);
+static proc_t *proc_find_raw(pid_t pgid);
```
cahirwpz, line 42: `s/pgid/pid`

What `raw` does stand for? Maybe `proc_find_nolock` instead?

sys/kern/proc.c
```diff
@@ -29,26 +38,43 @@ static proc_list_t proc_list = TAILQ_HEAD_INITIALIZER(proc_list);
...
-static pid_t pid_alloc(void) {
-  assert(mtx_owned(all_proc_mtx));
 
-  pid_t pid;
-  bit_ffc(pid_used, NPROC, &pid);
-  if (pid < 0)
-    panic("Out of PIDs!");
-  bit_set(pid_used, pid);
-  return pid;
+static void proc_init_hashtbl(void) {
```
cahirwpz, line 46: Maybe rename to `proc_init`? There's no need for extra function right now.

sys/kern/proc.c
```diff
@@ -29,26 +38,43 @@ static proc_list_t proc_list = TAILQ_HEAD_INITIALIZER(proc_list);
...
+    TAILQ_INIT(&proc_hashtbl[i]);
+    TAILQ_INIT(&pgrp_hashtbl[i]);
+  }
+}
+
+static bool pid_is_taken(pid_t pid) {
+  /* PID 0 is reserved. */
+  if (pid == 0)
+    return true;
+  if (proc_find_raw(pid) != NULL)
```
cahirwpz, line 57: Is `!= NULL` necessary here and below?

sys/kern/proc.c
```diff
@@ -58,7 +84,7 @@ static pgrp_t *pgrp_lookup(pgid_t pgid) {
   assert(mtx_owned(all_proc_mtx));
 
   pgrp_t *pgrp;
-  TAILQ_FOREACH (pgrp, &pgrp_list, pg_link)
+  TAILQ_FOREACH (pgrp, PGRP_HASH_CHAIN(pgid), pg_hash)
```
cahirwpz, line 87: Since `TAILQ_FOREACH` macro expands second argument multiple times it's advisable to introduce extra variable to keep result of `PGRP_HASH_CHAIN`. This is not absolutely necessary here, since `PGRP_HASH_CHAIN(pgid)` doesn't have side effect, but it's a good habit nonetheless.

This remark applies to every such use throught this change of course.

sys/kern/proc.c
```diff
@@ -152,30 +178,39 @@ void proc_add(proc_t *p) {
...
       TAILQ_INSERT_TAIL(CHILDREN(p->p_parent), p, p_child);
   }
 
   klog("Process PID(%d) {%p} has been created", p->p_pid, p);
 }
 
-proc_t *proc_find(pid_t pid) {
+/* Lookup a process in the PID hash table.
+ * The returned process, if any, is NOT locked. */
+proc_t *proc_find_raw(pid_t pid) {
```
cahirwpz, line 191: You defined it `static` earlier? How come the symbol leaks outside?!

sys/kern/proc.c
```diff
@@ -253,21 +288,19 @@ __noreturn void proc_exit(int exitstatus) {
...
+    proc_t *init = proc_find_raw(1);
+    assert(init != NULL);
     proc_reparent(p, init);
 
     TAILQ_REMOVE(&proc_list, p, p_all);
     TAILQ_INSERT_TAIL(&zombie_list, p, p_zombie);
 
     /* When the process is dead we can finally signal the parent. */
     proc_t *parent = p->p_parent;
-    if (!parent)
```
cahirwpz, line 269: To remove all checks for `p->p_parent` being `NULL` we should finally introduce `proc0` as they do in [FreeBSD](http://bxr.su/FreeBSD/sys/kern/init_main.c#proc0_init) and [NetBSD](http://bxr.su/NetBSD/sys/kern/kern_proc.c#proc0_init). What do you think?

sys/kern/proc.c
j-piecuch: `raw` means the returned process is not locked and there's no `proc_is_alive()` check.

sys/kern/proc.c
j-piecuch: Semantically, `p` and `p != NULL` are equivalent here. It's a matter of taste and clarity, I suppose.

sys/kern/proc.c
j-piecuch: Ok... I'll make the change, but then I guess you should also change the code here: https://github.com/cahirwpz/mimiker/blob/583f2b82e94a58b33b9ba9c21fde9996dc9d9eb1/sys/kern/proc.c#L380
...and here: https://github.com/cahirwpz/mimiker/blob/583f2b82e94a58b33b9ba9c21fde9996dc9d9eb1/sys/kern/proc.c#L215

sys/kern/proc.c
j-piecuch: I think it's a good idea, but someone needs to implement it.

sys/kern/proc.c
cahirwpz: You're right! Would you be so kind and do it in this PR ?

sys/kern/proc.c
j-piecuch: An omission on my part, the function is meant to be static.

sys/kern/proc.c
cahirwpz: You're right, it's better now to focus on adding sessions to the system.

sys/kern/proc.c
```diff
@@ -14,11 +14,20 @@
...
 #include <sys/malloc.h>
 #include <sys/vfs.h>
+#include <sys/sysinit.h>
 #include <bitstring.h>
 
-#define NPROC 64 /* maximum number of processes */
+/* Allocate PIDs from a reasonable range, can be changed as needed. */
+#define PID_MAX 255
+#define NBUCKETS ((PID_MAX + 1) / 4)
+#define PIDHASH(pid) (pid % NBUCKETS)
```
cahirwpz, line 23: It's good idea to surround `pid` with parenthesis. Just in case someone passed `pid + 1` to this macro ;-)

sys/kern/proc.c
j-piecuch: Not to mention other list macros like `TAILQ_INSERT_TAIL`, which expands its first argument multiple times and is used e.g. here: https://github.com/cahirwpz/mimiker/blob/583f2b82e94a58b33b9ba9c21fde9996dc9d9eb1/sys/kern/vm_physmem.c#L116

cahirwpz: You're right. It has been used in many context that way. Sometimes it gave us a headache. Maybe we should switch to `static inline` (for `CHILDREN`, `FREELIST` and so on) in long-run?

If you wish you can ignore my initial remark.

sys/kern/proc.c
cahirwpz: Since it's so similar to `pgrp_lookup` please rename it to `proc_lookup`.

sys/kern/proc.c
j-piecuch: We could also make the macros evaluate their arguments exactly once using temporary variables, which can be generated using [`__COUNTER__`](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html). I haven't seen it done like that anywhere before, but it should work in theory... I even wrote a working proof of concept:
```c
#define CAT(a, b) CAT2(a, b) // force expand
#define CAT2(a, b) a##b      // actually concatenate
#define UNIQUE(var) CAT(var, __COUNTER__)

#define WITH_2_TEMP_VARS(e1, e2, BODY, ...)                                    \
  WITH_2_TEMP_VARS_(UNIQUE(__tmpvar), e1, UNIQUE(__tmpvar), e2, BODY,          \
                    __VA_ARGS__)

#define WITH_2_TEMP_VARS_(v1, e1, v2, e2, BODY, ...)                           \
  do {                                                                         \
    typeof(e1) v1 = (e1);                                                      \
    typeof(e2) v2 = (e2);                                                      \
    BODY(v1, v2, __VA_ARGS__);                                                 \
  } while (/*CONSTCOND*/ 0)  

#define TAILQ_INSERT_TAIL(head, elm, field)                                    \
  WITH_2_TEMP_VARS(head, elm, TAILQ_INSERT_TAIL_, field)

#define TAILQ_INSERT_TAIL_(head, elm, field)                                   \
  do {                                                                         \
    (elm)->field.tqe_next = TAILQ_END(head);                                   \
    (elm)->field.tqe_prev = head->tqh_last;                                    \
    *head->tqh_last = (elm);                                                   \
    head->tqh_last = &(elm)->field.tqe_next;                                   \
  } while (/*CONSTCOND*/ 0)
```

sys/kern/proc.c
j-piecuch: I think that will lead to confusion between `proc_find` and `proc_lookup`. I would prefer renaming `pgrp_lookup` to `pgrp_find`

sys/kern/proc.c
j-piecuch: Done.

sys/kern/proc.c
j-piecuch: Done.

PR #713, opened 2020-05-26
Separating kernelspace from userspace for time related functions/structs.

wiklam:


wiklam:
The proposal was used in #716 and #717.

PR #716, opened 2020-05-31
Replacing timeval with bintime in kernel.

wiklam:
Changing the representation of how we store time in kernel to more accurate, which give higher precision (2^-64 ≈ 5 * 10^-20 (bintime) instead of 10^-6 (timeval)).

### DIFF
--- a/include/sys/time.h
+++ b/include/sys/time.h
@@ -46,11 +46,6 @@ typedef struct bintime {
     .frac = (uint64_t)((fp - __builtin_floor(fp)) * (1ULL << 63) * 2)          \
   }
 
-#define BINTIME2(s, fp)                                                        \
-  (bintime_t) {                                                                \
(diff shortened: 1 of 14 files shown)


include/sys/time.h
```diff
@@ -46,6 +46,11 @@ typedef struct bintime {
     .frac = (uint64_t)((fp - __builtin_floor(fp)) * (1ULL << 63) * 2)          \
   }
 
+#define BINTIME2(s, fp)                                                        \
```
pj1031999, line 49: This name doesn't look good for me.

If I understand correctly you introduce `BINTIME2` only because clang doesn't support static declaration combined with `__builtin_floor` in tests. If you doesn't plan to use it in other places maybe it will be good idea to move this macro into dedicated tests or change something other in tests?

PR #717, opened 2020-06-02
Removing unused macros and function duplications related with time.

wiklam:
After replacing timeval by bintime some functions/structures will not be used anymore. Functions related only with timeval are duplicated e.g. timeval_add, timeval_clear.


cahirwpz:
LGTM

PR #718, opened 2020-06-04
Replacing timeval by bintime for rusage in kernel.

wiklam:
This change brings us one step closer to remove timeval from kernel. Struct krusage could be defined just for kernel space but in such situation, we have to use some implicit properties of the structures in _wait4 and probably in some future syscalls related to rusage.


wiklam:
The cost of removing timeval from kernel is higher than the benefits (the complexity of the code rise). We should try to limit the scope for timeval and timespec (e.g. to syscall interface).

PR #722, opened 2020-06-14
Fixing clk_tck definition & moving tm2sec to time.c

wiklam:
Changing value  **clk_tck** to 1000, cause our system tick take 1ms.
Moving **tm2sec** to more appropriate file for this type of function.


cahirwpz:
LGTM

PR #723, opened 2020-06-15
Map cbus_uart into virtual address space

pj1031999:
* Move `cn_init` to `kernel_init` after initialization of memory subsystems
* Map `cbus_uart` into virtual address space

### DIFF
--- a/sys/kern/console.c
+++ b/sys/kern/console.c
@@ -2,7 +2,6 @@
 #include <sys/console.h>
 #include <sys/linker_set.h>
 
-
 static void dummy_init(console_t *dev __unused) {
 }
 static void dummy_putc(console_t *dev __unused, int c __unused) {
--- a/sys/kern/console.c
+++ b/sys/kern/console.c
@@ -21,6 +21,10 @@ static console_t dummy_console = {
 
 CONSOLE_ADD(dummy_console);
 
+void cn_early_init(void) {
+  cn = &dummy_console;
+}
+
 void cn_init(void) {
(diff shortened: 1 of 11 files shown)


sys/mips/uart_cbus.c
```diff
@@ -28,6 +31,12 @@ static bool is_set(unsigned offset, uint8_t mask) {
 }
 
 static void cbus_uart_init(console_t *dev __unused) {
+  vaddr_t handle = kva_alloc(PAGESIZE);
```
cahirwpz, line 34: I'd expect that to be implemented by `cbus_bus_space` interface.

sys/mips/uart_cbus.c
```diff
@@ -28,6 +31,12 @@ static bool is_set(unsigned offset, uint8_t mask) {
...
+  /* MALTA_CBUS_UART is MALTA_FPGA_BASE + 0x900 */
```
cahirwpz, line 35: Maybe `MALTA_CBUS_UART` should be set as relative to `MALTA_FPGA_BASE` to avoid 0x900 below?

sys/kern/console.c
```diff
@@ -21,6 +21,10 @@ static console_t dummy_console = {
 
 CONSOLE_ADD(dummy_console);
 
+void cn_early_init(void) {
```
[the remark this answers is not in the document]
pj1031999: done

sys/mips/uart_cbus.c
pj1031999: done

PR #725, opened 2020-06-22
Use devclasses for device initialization

pj1031999:
* fix `bus_generic_probe` -- now it initializes only devices in correct _devclass_
* use _devclass_ interface for device initialization
* introduce new member of `device_t` for access to _devclass_
* remove `gt_pci_bus` and `gt_pci` global references
* stop using `SYSINIT_ADD` for device initialization
* use `bus_generic_probe` everywhere where it is possible


cahirwpz:
LGTM

PR #732, opened 2020-07-17
Build on Aarch64

pj1031999:
Changes required to make Mimiker build for AArch64.

From now we can build Mimiker for two platform:
- `make BOARD=malta` for Malta board (default)
- `make BOARD=rpi3` for Raspberry Pi 3

Introduce new `--board` argument for `run_tests.py` and `launch`.
By default we want to emulate Malta board `--board malta` but we can emulate rpi3 `--board rpi3`.

Add build tests with CircleCi.


cahirwpz:
LGTM

PR #740, opened 2020-08-14
Directly-mapped page table for AArch64

pj1031999:
Build directly-mapped page table.

`DMAP` will be located at `DMAP_BASE` which is `0xffffff8000000000`.


cahirwpz:
There's subtle problem with current solution. When a page table grows we'll have to write to `l*_*` tables as well, in order to map new page directories. That forces us to map those pages in virtual addresses.

Let's start to implement `pmap` - that will force us to understand problem better. I hope we'll avoid direct map after all.

PR #741, opened 2020-08-14
AArch64 exceptions.

pj1031999:
According to sys/arm64/arm64/exception.S from FreeBSD we only need to handle:
- Synchronous EL1h
- IRQ EL1h
- Synchronous 64-bit EL0
- IRQ 64-bit EL0


cahirwpz:
Although exception handling is not yet finished I'll accept this PR as intermediate step.

PR #752, opened 2020-09-11
Implement tests for pmap.

pj1031999:
Tests added in #743.

- `test_pmap_kenter` for `pmap_kenter` and `pmap_kremove`
- `test_pmap_page` for `pmap_copy_page` and `pmap_zero_page`
- `test_pmap_extract` for `pmap_extract`

### DIFF
--- a/sys/tests/pmap.c
+++ b/sys/tests/pmap.c
@@ -4,6 +4,7 @@
 #include <sys/vm_physmem.h>
 #include <sys/ktest.h>
 #include <sys/sched.h>
+#include <sys/kmem.h>
 
 #define PAGES 16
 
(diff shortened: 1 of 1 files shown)


cahirwpz:
Please use `kasan_mark_valid` and `kasan_mark_invalid` instead disabling tests for KASAN. The procedures are quite well documented. You can basically copy some lines from `kmem.c`.

Please add `pmap_kextract` procedure to `pmap` interface.

BTW. Kernel pmap is always active. You cannot make it inactive for obvious reasons. I'm not sure there's need to call `pmap_activate` in these tests.

PS. Could you have a look at `test_kernel_pmap` and make it work again (with requisite changes).

PR #758, opened 2020-09-19
Revert almost the greatest idea in MIPS pmap

cahirwpz:
* Treat kseg0 as direct map for free.
 * Remove workarounds.


pj1031999 (approved):
This change makes me very sad :(

LGTM

PR #762, opened 2020-10-04
Simplify ctx_switch.

pj1031999:
Simplify ctx_switch.


cahirwpz:
Remember that first thing you need to do after `.ctx_resume` label is to switch stack, so that `vm_map_switch` executes on target thread stack.

PR #764, opened 2020-10-07
Introduce pmap_growkernel

pj1031999:
Implement [`pmap_growkernel`](https://www.freebsd.org/cgi/man.cgi?query=pmap_growkernel&apropos=0&sektion=9&manpath=FreeBSD+12.1-RELEASE+and+Ports&arch=default&format=html) based on _NetBSD_ version for _AArch64_.

Implementation for:
- _AArch64_
- _MIPS_


include/sys/vmem.h
```diff
@@ -19,7 +20,7 @@ void init_vmem(void);
 
 /*! \brief Create a new vmem arena.
  * You need to specify quantum, the smallest unit of allocation. */
-vmem_t *vmem_create(const char *name, vmem_size_t quantum);
+vmem_t *vmem_create(const char *name, vmem_size_t quantum, bool grow);
```
j-piecuch, line 23: I have a feeling @cahirwpz won't like this and would prefer the third parameter to be `vmem_flags_t` instead.

sys/aarch64/pmap.c
```diff
@@ -602,3 +604,25 @@ void pmap_delete(pmap_t *pmap) {
...
+/*
+ * Increase usable kernel virtual address space to at least maxkvaddr.
+ * Allocate page table (level 1) if needed.
+ */
+vaddr_t pmap_growkernel(vaddr_t maxkvaddr) {
+  pmap_t *pmap = pmap_kernel();
+  vaddr_t va;
+
+  /* For 8 pages used by kernel we need 1 page for KASAN. */
+  maxkvaddr = roundup(maxkvaddr, 8 * L2_SIZE);
```
j-piecuch, line 617: The `roundup()` here is so that the shadow map is grown by a multiple of `L2_SIZE`, is my understanding correct? If so, I'd like the comment to say it more directly.

sys/kern/vmem.c
```diff
@@ -242,24 +253,30 @@ int vmem_add(vmem_t *vm, vmem_addr_t addr, vmem_size_t size) {
...
        vm->vm_name);
 
   return 0;
 }
 
+int vmem_add(vmem_t *vm, vmem_addr_t addr, vmem_size_t size) {
+  int error;
+  WITH_MTX_LOCK (&vm->vm_lock)
+    error = vmem_add_nolock(vm, addr, size);
+  return error;
```
j-piecuch, line 272: Why not:
```c
SCOPED_MTX_LOCK(&vm->vm_lock);
return vmem_add_nolock(vm, addr, size);
```

sys/aarch64/pmap.c
```diff
@@ -602,3 +604,25 @@ void pmap_delete(pmap_t *pmap) {
   free_asid(pmap->asid);
   pool_free(P_PMAP, pmap);
 }
+
+/*
+ * Increase usable kernel virtual address space to at least maxkvaddr.
+ * Allocate page table (level 1) if needed.
```
j-piecuch, line 610: Why does the function allocate level 1 pages, but not level 2 or 3?

sys/kern/vmem.c
```diff
@@ -270,12 +287,25 @@ int vmem_alloc(vmem_t *vm, vmem_size_t size, vmem_addr_t *addrp,
...
+      if (grow_called || !(vm->vm_flags & VM_GROW))
+        goto fail;
+
+      vaddr_t old = maxkvaddr;
+      maxkvaddr = pmap_growkernel(maxkvaddr + size);
+      klog("%s: increase kernel end %08lx -> %08lx", __func__, old, maxkvaddr);
+      grow_called = true;
+      if (vmem_add_nolock(vm, old, maxkvaddr - old) != 0)
+        goto fail;
+      goto retry;
```
j-piecuch, line 305: `vmem` is supposed to be a general-purpose allocator that can be used not only for kernel virtual address space allocation. This change tightly couples the `VM_GROW` flag with kernel virtual address space allocation, i.e. no one else can use the `VM_GROW` flag. I don't like this. Maybe the KVA-specific code could be moved to `kva_alloc()`?

sys/kern/vmem.c
```diff
@@ -270,12 +287,25 @@ int vmem_alloc(vmem_t *vm, vmem_size_t size, vmem_addr_t *addrp,
...
     vmem_check_sanity(vm);
 
+  retry:
     bt = bt_find_freeseg(vm, size);
 
     if (bt == NULL) {
-      pool_free(P_BT, btnew);
       klog("%s: block of %lu bytes not found in '%s'", __func__, size,
            vm->vm_name);
+      if (grow_called || !(vm->vm_flags & VM_GROW))
```
j-piecuch, line 296: I noticed that `kva_alloc()` passes the `M_NOGROW` flag to `vmem_alloc()`. First, I'm not sure `kva_alloc()` should to that. Second, we should do something with that flag here.

sys/mips/pmap.c
```diff
@@ -509,6 +510,7 @@ static void pmap_setup(pmap_t *pmap) {
 void init_pmap(void) {
   pmap_setup(&kernel_pmap);
   kernel_pmap.pde = _kernel_pmap_pde;
+  pmap_maxkvaddr = rounddown((vaddr_t)__kernel_end, PAGESIZE);
```
j-piecuch, line 513: Why are you rounding down and not up?

j-piecuch (changes requested):

include/sys/vmem.h
pj1031999: done

sys/kern/vmem.c
pj1031999: No reason for that. I changed it according to your suggestion.

sys/mips/pmap.c
pj1031999: I think that it doesn't matter as long as last used page will be mapped.
But `roundup` is better here - we don't want to have anything after end - I changed that.

sys/aarch64/pmap.c
pj1031999: My mistake in comment.

sys/kern/vmem.c
pj1031999: I'm not sure what is a purpose of `M_NOGROW`. We don't use that flag in `vmem` code.
@cahirwpz could you explain origin of that flag?

sys/kern/vmem.c
pj1031999: Ok, but we use `vmem` only as a backend for `kmem`. If I remember correctly @cahirwpz said that it is a correct place for calling `pmap_growkernel`.
I can move that code into `kva_alloc` but then from my point of view it will be more complicated.

j-piecuch: Hmmm... ok, I'm not going to argue with him.

pj1031999: I will mark this conversation as unresolved and I will wait for his opinion - maybe I didn't understand him correctly.

pj1031999: I have moved everything to `kmem`.

sys/kern/vmem.c
pj1031999: I have moved everything to `kmem`.

sys/aarch64/pmap.c
pj1031999: done

sys/kern/kmem.c
```diff
@@ -28,8 +28,24 @@ static void kick_swapper(void) {
...
+
+  bool restarted = false;
+
+retry:
+  if (vmem_alloc(kvspace, size, &start, M_NOGROW)) {
+    if (restarted)
+      return 0;
+
+    vaddr_t old = maxkvaddr;
+    maxkvaddr = pmap_growkernel(maxkvaddr + size);
```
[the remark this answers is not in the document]
pj1031999: fixed

sys/aarch64/pmap.c
```diff
@@ -82,6 +82,7 @@ static const pte_t vm_prot_map[] = {
 
 static pmap_t kernel_pmap;
 paddr_t _kernel_pmap_pde;
+static vaddr_t pmap_maxkvaddr;
```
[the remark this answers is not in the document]
pj1031999: There is no difference between them now. That implementation is based on NetBSD version where `pmap_maxkvaddr` is private for pmap and uvm manages their own kernel end using return value of `pmap_growkernel`. On the other hand FreeBSD shares the same variable with other subsystems globally.

So now I'm not sure what is your expectation.

sys/kern/kmem.c
```diff
@@ -28,8 +30,26 @@ static void kick_swapper(void) {
...
+      maxkvaddr = pmap_growkernel(maxkvaddr + size);
+      klog("%s: increase kernel end %08lx -> %08lx", __func__, old, maxkvaddr);
+
+      if (vmem_add(kvspace, old, maxkvaddr - old))
+        return 0;
+    }
+
+    restarted = true;
+    goto retry;
+  }
```
j-piecuch, line 52: Consider two threads: A and B. Both threads concurrently call `kva_alloc()` and the first call to `vmem_alloc()` fails in both threads. Then, thread A grows `kvspace` and after that, thread B grows `kvspace`. We end up with `kvspace` grown by too much, as growing it only once would be sufficient for both threads.
I think we could fix this by checking if the value of `maxkvaddr` changed between the first call to `vmem_alloc()` and acquiring `maxkvaddr_lock`:
```c
retry:
  vaddr_t old = maxkvaddr;
  if (vmem_alloc(kvspace, size, &start, M_NOGROW)) {
    if (restarted)
      return 0;

    WITH_MTX_LOCK (&maxkvaddr_lock) {
      if (maxkvaddr > old) {
        mtx_unlock(&maxkvaddr_lock);
        goto retry;
      }
      ...
    }

    restarted = true;
    goto retry;
  }

```

sys/kern/kmem.c
```diff
@@ -28,8 +30,26 @@ static void kick_swapper(void) {
...
   vmem_addr_t start;
-  if (vmem_alloc(kvspace, size, &start, M_NOGROW))
-    return 0;
+
+  bool restarted = false;
+
+retry:
+  if (vmem_alloc(kvspace, size, &start, M_NOGROW)) {
+    if (restarted)
+      return 0;
```
j-piecuch, line 39: I think this can lead to `kva_alloc()` returning 0 even if `kvspace` can be grown.
Consider what happens if another thread calls `kva_alloc()` right after we call `vmem_add()`, but before we retry the call to `vmem_alloc()`. The other thread can take all the space we just added, so our second attempt to `vmem_alloc()` will fail. I think we should just remove the `restarted` variable altogether and keep trying until we succeed.

j-piecuch (changes requested):

sys/aarch64/pmap.c
```diff
@@ -602,3 +604,28 @@ void pmap_delete(pmap_t *pmap) {
   free_asid(pmap->asid);
   pool_free(P_PMAP, pmap);
 }
+
+/*
+ * Increase usable kernel virtual address space to at least maxkvaddr.
+ * Allocate page table (levels 1, 2, 3) if needed.
+ */
+vaddr_t pmap_growkernel(vaddr_t maxkvaddr) {
```
xThaid, line 612: What about making an assertion `assert(maxkvaddr > pmap_maxkvaddr)` ?

sys/kern/kmem.c
pj1031999: done

sys/kern/kmem.c
pj1031999: done

sys/aarch64/pmap.c
pj1031999: done

sys/kern/kmem.c
```diff
@@ -30,26 +30,29 @@ static void kick_swapper(void) {
...
   vmem_addr_t start;
+  vaddr_t old;
 
-  bool restarted = false;
-
-retry:
-  if (vmem_alloc(kvspace, size, &start, M_NOGROW)) {
-    if (restarted)
-      return 0;
+  WITH_MTX_LOCK (&maxkvaddr_lock)
```
j-piecuch, line 35: Taking the lock is not necessary if we assume that word-sized memory reads and writes are atomic. FreeBSD makes that assumption, so I think we could make it too.

xThaid: This would be detected by the KCSAN as a data race, because there is a possibility of two accesses to that variable and one of them is write.

pj1031999: done

sys/aarch64/pmap.c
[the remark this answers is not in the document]
j-piecuch: I wasn't aware that NetBSD had two different variables for `mavkvaddr`. I assumed that they have different meanings, otherwise there would be no need to have two different variables. You could add a description here like "`pmap_maxkvaddr` is used to track the range of virtual kernel addresses for which page table pages have been allocated".

sys/mips/pmap.c
```diff
@@ -547,3 +549,34 @@ void pmap_delete(pmap_t *pmap) {
...
   pool_free(P_PMAP, pmap);
 }
+
+#define L1_SPACE_SIZE (PAGESIZE * PAGESIZE / sizeof(pte_t))
+
+/*
+ * Increase usable kernel virtual address space to at least maxkvaddr.
+ * Allocate page table (level 1) if needed.
+ */
+vaddr_t pmap_growkernel(vaddr_t maxkvaddr) {
```
j-piecuch, line 559: This function modifies the global variable `pmap_maxkvaddr` without any synchronization. We're holding `maxkvaddr_lock` while calling it, but it's not required. Could you add an assertion here that checks that we're holding `maxkvaddr_lock`?

j-piecuch (changes requested):

sys/mips/pmap.c
j-piecuch: My bad, we're holding `pmap_kernel()->mtx` when modifying `pmap_maxkvaddr`, but some accesses are unlocked, so we need to make them atomic to avoid tripping up KCSAN.

pj1031999: done

j-piecuch (approved):

franciscozdo (approved):

j-piecuch (approved):

xThaid (approved):

sys/mips/pmap.c
```diff
@@ -552,33 +548,27 @@ void pmap_delete(pmap_t *pmap) {
...
-  /*
-   * For 8 pages used by kernel we need 1 page for KASAN.
-   * shadow map is grown by a multiple of L1_SPACE_SIZE.
-   */
-  maxkvaddr = roundup(maxkvaddr, 8 * L1_SPACE_SIZE);
+  maxkvaddr = roundup(maxkvaddr, PAGESIZE);
 
   WITH_MTX_LOCK (&pmap->mtx) {
-    for (va = pmap_maxkvaddr; va <= maxkvaddr; va += L1_SPACE_SIZE) {
+    for (va = vm_kernel_end; va < maxkvaddr; va += PAGESIZE) {
```
j-piecuch, line 565: Isn't incrementing `va` by `PAGESIZE` a bit inefficient? The step size used to be `L1_SPACE_SIZE` here. Same for AArch64.

sys/mips/pmap.c
```diff
@@ -552,33 +548,27 @@ void pmap_delete(pmap_t *pmap) {
...
       if (!is_valid_pde(PDE_OF(pmap, va)))
         pmap_add_pde(pmap, va);
     }
-    pmap_maxkvaddr = va;
   }
 
-  /* TODO(pj) add new region into shadow map */
+  vm_kernel_end = va;
```
j-piecuch, line 571: I think `vm_kernel_end = maxkvaddr` is correct. Same for AArch64. It makes no difference as long as the step size in the loop above is the same as the alignment of `maxkvaddr`.

sys/mips/pmap.c
pj1031999: For `PAGESIZE` `pmap_growkernel` is called three times.
For `L1_SPACE_SIZE` two times.

sys/mips/pmap.c
pj1031999: As you said it makes no difference. From my point of view step in loop should be always a divider of alignment of `maxkvaddr`. But I have changed that for readability.

sys/mips/pmap.c
j-piecuch: I'm not concerned about how many times `pmap_growkernel()` is called, I'm concerned about how many times `pmap_add_pde()` is called. For instance, if the current PTE index of `vm_kernel_end` is 0 (i.e. it's aligned to `L1_SPACE_SIZE`), and we call `pmap_growkernel(vm_kernel_end + L1_SPACE_SIZE)`, Then we will call `pmap_add_pde()` `L1_SPACE_SIZE / PAGESIZE` times, even though calling it once would suffice. Is my thinking correct?

pj1031999: Oh, now I see what you mean.
We will call `pmap_add_pde()` exactly the same number of times in both cases.
But now we will walk through page table `L1_SPACE_SIZE / PAGESIZE` times in `PDE_OF` - similarly for AArch64.
Do you want to revert that change and go back to align `maxkvaddr` to `L1_SPACE_SIZE`?

j-piecuch: Yes, that's what I meant - unnecessary page table walks.
I think aligning `maxkvaddr` to `L1_SPACE_SIZE` and using steps of `L1_SPACE_SIZE` is the best way to solve this.

pj1031999: done

j-piecuch (approved):

sys/aarch64/pmap.c
```diff
@@ -602,3 +602,26 @@ void pmap_delete(pmap_t *pmap) {
...
+ * Allocate page table (levels 1, 2, 3) if needed.
+ */
+void pmap_growkernel(vaddr_t maxkvaddr) {
+  assert(mtx_owned(&vm_kernel_end_lock));
+  assert(maxkvaddr > vm_kernel_end);
+
+  pmap_t *pmap = pmap_kernel();
+  vaddr_t va;
+
+  maxkvaddr = roundup(maxkvaddr, L2_SIZE);
```
j-piecuch, line 617:
```suggestion
  maxkvaddr = roundup2(maxkvaddr, L2_SIZE);
```

sys/mips/pmap.c
```diff
@@ -547,3 +547,30 @@ void pmap_delete(pmap_t *pmap) {
...
+ * Allocate page table (level 1) if needed.
+ */
+void pmap_growkernel(vaddr_t maxkvaddr) {
+  assert(mtx_owned(&vm_kernel_end_lock));
+  assert(maxkvaddr > vm_kernel_end);
+
+  pmap_t *pmap = pmap_kernel();
+  vaddr_t va;
+
+  maxkvaddr = roundup(maxkvaddr, L1_SPACE_SIZE);
```
j-piecuch, line 564:
```suggestion
  maxkvaddr = roundup2(maxkvaddr, L1_SPACE_SIZE);
```

j-piecuch (changes requested):

j-piecuch (approved):

cahirwpz (approved):

PR #765, opened 2020-10-09
Port `su` program

franciscozdo:
Depends on changes in #766 and #767.


etc/master.passwd
```diff
@@ -1,2 +1,2 @@
 root::0:0::0:0:root:/root:/bin/ksh
-
+paco::1000:1000::0:0:paco:/root:/bin/ksh
```
pj1031999, line 2: Is _paco_ a special username? Why not something like _user_ or _mimiker_?
I'm wonder if `/root` is a good home directory for any user other than _root_.

lib/libc/gen/initgroups.c
```diff
@@ -0,0 +1,68 @@
...
+#include <errno.h>
+#include <grp.h>
+
+int initgroups(const char *uname, gid_t agroup) {
+  gid_t groups_list[NGROUPS];
+  int ngroups;
+  gid_t *groups = groups_list;
+  int rval;
+
+  //_DIAGASSERT(uname != NULL);
```
pj1031999, line 48: Why this line is commented? Do you plan to use it in the future?

usr.bin/su/Makefile
```diff
@@ -0,0 +1,11 @@
+TOPDIR = $(realpath ../..)
+
+PROGRAM = su
+SOURCES = su.c suutil.c
+FORMAT-EXCLUDE = $(SOURCES)
+
+LDLIBS = -lutil
+
+#CPPFLAGS += -DALLOW_GROUP_CHANGE -DALLOW_EMPTY_USER
```
pj1031999, line 9: Why this line is commented? Do you plan to use these flags later?

usr.bin/su/suutil.h
```diff
@@ -0,0 +1,35 @@
...
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+__BEGIN_DECLS
+int chshell(const char *);
+char *ontty(void);
+__END_DECLS
```
pj1031999, line 35: `__BEGIN_DECLS` and `__END_DECLS` are defined in `sys/cdefs.h`. How can we see them here?

pj1031999:
I see a lot of `#if 0 ... #endif` in code. Is it ok?

usr.bin/su/suutil.h
franciscozdo: done.

include/sys/param.h
```diff
@@ -111,6 +111,8 @@
 
 #define powerof2(x) ((((x)-1) & (x)) == 0)
 
+#define NGROUPS NGROUPS_MAX
```
cahirwpz, line 114: This is already defined but earlier.

lib/libc/gen/initgroups.c
franciscozdo: removed

usr.bin/su/Makefile
franciscozdo: Yes. Flag `DALLOW_GROUP_CHANGE` could be used in the future. `DALLOW_EMPTY_USER` is now used.

include/sys/param.h
franciscozdo: Removed.

etc/master.passwd
franciscozdo: For now changed to users 'admin' and 'user'. Home directory for all users is now /

PR #766, opened 2020-10-10
Implementation of execv

franciscozdo:


bin/utest/exec.c
```diff
@@ -0,0 +1,45 @@
...
+    printf("parent: child exited with status %d\n", status);
+    assert(status == 0);
+    return 0;
+  }
+  /* child */
+  char *args[] = {"echo", "test", NULL};
+  char *env[] = {"test", "environment", NULL};
+  int r = execve("/bin/echo", args, env);
+  printf("child: execv failed with %d (%s)\n", r, strerror(errno));
+  _exit(1);
```
pj1031999, line 26: Is there any reason you are not using `exit` here?

bin/utest/exec.c
```diff
@@ -0,0 +1,45 @@
...
+    assert(r == pid);
+    printf("parent: child exited with status %d\n", status);
+    assert(status == 0);
+    return 0;
+  }
+  /* child */
+  char *args[] = {"echo", "test", NULL};
+  int r = execv("/bin/echo", args);
+  printf("child: execv failed with %d (%s)\n", r, strerror(errno));
+  _exit(1);
```
pj1031999, line 44: Is there any reason you are not using `exit` here?

bin/utest/exec.c
```diff
@@ -0,0 +1,45 @@
...
+#include <sys/wait.h>
+#include <assert.h>
+
+int test_execve() {
+  int pid = fork();
+  if (pid) {
+    /* parent */
+    int status;
+    int r = wait(&status);
+    assert(r == pid);
```
pj1031999, line 16: It is not clear for me how are you catching `fork` failures here.

Let's assume that `fork` is broken. Then `pid` is -1. `wait` should failed and `r` will be -1. `status` is uninitialized here so it can be equal to 0. Both assertion will pass and you return 0 which means test success but you never call `execve`.

franciscozdo: Thanks. I've added error handling.

bin/utest/exec.c
franciscozdo: Changed to `exit`.

bin/utest/exec.c
```diff
@@ -0,0 +1,54 @@
...
+
+#include <stdio.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+#include <sys/wait.h>
+#include <assert.h>
+#include <stdlib.h>
+
+int test_execve() {
```
pj1031999, line 11: The signature of this function is different in the header file `int test_execve(void)`.
In C `foo(void)` is different than `foo()`. Second variant could take any number of parameters of unknown types.

The same comment applies to `test_execv`.

bin/utest/exec.c
```diff
@@ -0,0 +1,54 @@
...
+  int pid = fork();
+  if (pid < 0) {
+    fprintf(stderr, "fork failed: %s\n", strerror(errno));
+    return 1;
+  }
```
pj1031999, line 16: Why not `assert(pid != -1);`?

pj1031999 (changes requested):
Please fix the signatures of these functions.

bin/utest/exec.c
franciscozdo: I want to display error message :)

pj1031999 (approved):
LGTM

cahirwpz:
@franciscozdo Undoubtedly `exec` function is an important addition to the set of supported `libc` functions. However tests for `execve` seem to me to be a little redundant as we use this functionality all the time.

In long run I'd much prefer to "borrow" testing framework and test from NetBSD. I don't think implementing system call tests is effective use of our time. Especially when we consider the fact that we took libc and other user-space libraries and programs from NetBSD.

Please remove the tests and leave only the wrapper for `execve`.

PR #772, opened 2020-10-10
Handling the lack of a timer in tm_reserve

pj1031999:
Do not dereference NULL address when timer is not found.


franciscozdo (approved):
LGTM

PR #775, opened 2020-10-13
Import seq from NetBSD

komik0:
This cs introduce seq (well known command from NetBSD/Linux)


komik0:
pushed from wrong branch

PR #779, opened 2020-10-17
Regenerate assym.h when a dependency of genassym.cf changes

### LINKED ISSUE #516: Adding member to `struct proc` breaks tests
Tests failed when added new member `LIST_ENTRY(proc) dummy` to `struct proc` and runned `./run_tests.py`.

cahirwpz:
Should fix #516 as well


pj1031999 (commented):
Thanks for that fix! It works very well.
One thing that worries me is a lot of "sed: can't read genassym.cf: No such file or directory" messages. 

What do you think about:
`assym.h: genassym.cf $(shell $(GENASSYM-DEPS) genassym.cf 2> /dev/null)`
In case of lack files required by genassym.cf or assym.h make still generates correct error message.

PR #780, opened 2020-10-18
Port su program from NetBSD

franciscozdo:
I removed some code from BSD implementation (macro `BSD4_4`). I don't think we will use it.

Added syscalls which do nothing but `su` needs them.
* `getpriority()`
* `setpriority()`

Added some dummy functions that do nothing just return 0 or "" to avoid commenting out code in su. 
* `getusershell()`
* `setusershell()`
* `ttyname()`
* `crypt()`
* `getpass()`


usr.bin/su/su.c
```diff
@@ -0,0 +1,557 @@
...
+	enum { UNSET, YES, NO } iscsh = UNSET;
+	const char *user, *shell, *avshell;
+	char *username, **np;
+#ifdef SU_ROOTAUTH
+	char *userpass;
+#endif
+	char *class;
+	char shellbuf[MAXPATHLEN], avshellbuf[MAXPATHLEN];
+#ifdef LOGIN_CAP
+	time_t pw_warntime = _PASSWORD_WARNDAYS * SECSPERDAY;
```
franciscozdo, line 105: This line was moved here (between macro) because it is only used when `LOGIN_CAP` is used.

pj1031999 (approved):
LGTM

cahirwpz:
Could we adapt:
 * `getusershell` & `setusershell` from http://bxr.su/OpenBSD/lib/libc/gen/getusershell.c
 * `ttyname` from http://bxr.su/OpenBSD/lib/libc/gen/ttyname.c (let's just comment out lines 90 - 107)
?

franciscozdo:
> Could we adapt:
> 
> * `getusershell` & `setusershell` from http://bxr.su/OpenBSD/lib/libc/gen/getusershell.c
> * `ttyname` from http://bxr.su/OpenBSD/lib/libc/gen/ttyname.c (let's just comment out lines 90 - 107)
>   ?

I have adapted `getusershell` and `setusershell` but `ttyname` left. IMO there is too much to add. It must be done better than just copying from BSD.

PR #783, opened 2020-10-22
Cred syscalls v2

franciscozdo:


sys/kern/cred_syscalls.c
```diff
@@ -131,3 +139,110 @@ int do_setgroups(proc_t *p, int ngroups, const gid_t *gidset) {
...
+   * when we don't set ruid and euid we also don't set suid
+   * (the last condition in or)
+   */
+  if (ruid == (uid_t)-1 && (euid == cur_ruid || euid == cur_euid ||
+                            euid == cur_suid || euid == (uid_t)-1))
+    cur_euid = -1;
+
+  error = change_resuid(&p->p_cred, ruid, euid, cur_euid);
+
+fail:
```
j-piecuch, line 191: `end` feels like a more appropriate name for this label, since we arrive here even if we succeed.

franciscozdo: Yes. `end` will be beter.

cahirwpz:
LGTM. Please resolve conflict before we proceed with merge.

PR #784, opened 2020-10-23
Kill QEMU after simulation

xThaid:
Added executing `pkill -9 qemu` after tmux session is closed. Just to make sure it won't hang forever.

I also removed `session.kill_session()`, because it only caused the error and as far as I know `server.kill_server()` destroys all sessions too.

### DIFF
--- a/launch
+++ b/launch
@@ -5,6 +5,7 @@ import os
 import os.path
 import subprocess
 import sys
+import time
 
 from launcher import *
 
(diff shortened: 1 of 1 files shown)


launch
```diff
@@ -51,9 +52,12 @@ def DevelRun(sim, dbg, uarts):
         session.select_window(dbg.name if dbg else '/dev/cons')
         session.attach_session()
     finally:
-        session.kill_session()
         server.kill_server()
 
+        # Give QEMU a chance to exit gracefully
+        time.sleep(0.2)
```
cahirwpz, line 58: Could we call `sim.stop()` here instead?

xThaid: `sim.stop()` kills the process created by the `subprocess` in the `sim.run()` procedure.
The problem is that QEMU is managed by tmux, so we can't stop it using `subprocess`.

PR #785, opened 2020-10-27
Make it possible to signal a process group without holding all_proc_mtx

j-piecuch:
This PR changes our locking rules to make it possible to signal a process group without holding `all_proc_mtx`.
Summary of changes:
- Add a mutex to `pgrp_t` that protects the list of members
- `proc_t::p_parent`, `proc_t::p_pglist`, `proc_t::p_pgrp` and `pgrp_t::pg_members` are now protected by a pair of locks: holding either one suffices for reading, but both must be held for writing.
- `proc_t::p_flags` is now protected by `proc_t::p_lock` instead of `all_proc_mtx`
- Add a rule that permits acquiring parent process's lock while holding the child process's lock

As a result of making these changes, some stuff broke, which this PR also fixes:
- There was a potential lost wakeup in `do_waitpid`: fixed by introducing the `PF_CHILD_STATE_CHANGED` flag.
- Unlocking a mutex while holding the current thread's lock is incorrect, since `mtx_unlock` can acquire it. This bug was already present before I made the changes, the changes merely made the bug actually cause a panic. I fixed it by introducing the `TDF_STOPPING` flag.


sys/kern/proc.c
```diff
@@ -648,11 +683,21 @@ int do_waitpid(pid_t pid, int *status, int options, pid_t *cldpidp) {
...
 
-      /* Wait until one of children changes a state. */
-      klog("PID(%d) waits for children (pid = %d)", p->p_pid, pid);
-      cv_wait(&p->p_waitcv, all_proc_mtx);
+      proc_lock(p);
+      if (p->p_flags & PF_CHILD_STATE_CHANGED) {
+        /* A child changed state while we were checking: recheck. */
+        p->p_flags &= ~PF_CHILD_STATE_CHANGED;
+        proc_unlock(p);
+        goto check_children;
```
franciscozdo, line 691: Can we make here a normal loop without `goto`?

franciscozdo (commented):
Please update information about locking in `sig_post` function description in `include/sys/signal.h`

sys/kern/proc.c
```diff
@@ -555,20 +590,20 @@ int proc_sendsig(pid_t pid, signo_t sig) {
...
 
-  TAILQ_FOREACH (target, &pgrp->pg_members, p_pglist) {
-    WITH_MTX_LOCK (&target->p_lock)
-      sig_kill(target, sig);
+    if (pid < -1) {
+      pgrp = pgrp_lookup(-pid);
+      if (!pgrp)
+        return EINVAL;
+    }
+    mtx_lock(&pgrp->pg_lock);
```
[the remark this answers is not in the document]
j-piecuch: Yes. If we dropped `all_proc_mtx` before acquiring `pgrp->pg_lock`, `pgrp` could disappear between the two operations.

sys/kern/proc.c
j-piecuch: We could, but a loop seems out of place here, since there's no actual looping that's happening: we would need to insert a `break` at the end of the loop, and `goto check_children` would be replaced by `continue`. I think `goto` is the better choice in this case.

franciscozdo: Okay.

sys/kern/signal.c
```diff
@@ -173,6 +173,14 @@ int do_sigsuspend(proc_t *p, const sigset_t *mask) {
   return EINTR;
 }
 
+/* Call with td->td_lock held! */
+static void sig_continue_thread(thread_t *td) {
+  if (td->td_flags & TDF_STOPPING)
```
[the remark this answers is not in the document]
j-piecuch: I guess it won't hurt :)

franciscozdo (approved):

sys/kern/proc.c
```diff
@@ -259,46 +261,68 @@ static pgrp_t *pgrp_lookup(pgid_t pgid) {
...
 }
 
 static int _pgrp_enter(proc_t *p, pgrp_t *target) {
+  pgrp_t *old_pgrp = p->p_pgrp;
+  assert(old_pgrp);
+
+  if (old_pgrp == target)
+    return 0;
+
   pgrp_increase_jobc(p, target);
```
cahirwpz, line 297: I got confused looking at those two lines. Without diving into `pgpr_*_jobc` implementation this looks almost like no-op. Did we choose good names for these procedures?

sys/kern/proc.c
```diff
@@ -259,46 +261,68 @@ static pgrp_t *pgrp_lookup(pgid_t pgid) {
...
+      TAILQ_REMOVE(&p->p_pgrp->pg_members, p, p_pglist);
+      p->p_pgrp = NULL;
+    }
   }
+
+  if (TAILQ_EMPTY(&pgrp->pg_members))
+    pgrp_remove(pgrp);
 }
 
 static int _pgrp_enter(proc_t *p, pgrp_t *target) {
```
cahirwpz, line 290: A short comment describing this function's purpose could help a reader...

cahirwpz (commented):
Is there a chance you could split out fixes to another PR, so that I can review them separately first ?

I feel a bit overwhelmed by the scope of this PR.

sys/kern/proc.c
j-piecuch: Maybe `pgrp_jobc_enter` and `pgrp_jobc_leave` are better names? Honestly, I don't think this is a big deal. One can easily find the implementation of these procedures and read the comment that explains their purpose.

j-piecuch: We could also add `old_pgrp` as a parameter to `pgrp_decrease_jobc`. It would look less like a no-op then.

j-piecuch:
Superseded by #803

sys/kern/proc.c
cahirwpz: I'll change it as suggested.

PR #786, opened 2020-10-28
TTY: Implement line editing in canonical mode

j-piecuch:


sys/kern/tty.c
```diff
@@ -173,6 +216,123 @@ static void tty_outq_putc(tty_t *tty, uint8_t c) {
...
+ */
+static void tty_echo(tty_t *tty, uint8_t c) {
+  assert(mtx_owned(&tty->t_lock));
+
+  int lflag = tty->t_lflag;
+  /* If ECHO is not set, only echo newlines provided ECHONL is set. */
+  if (!(lflag & ECHO) && (!(lflag & ECHONL) || c != '\n'))
+    return;
+  /* If ECHOCTL is set, echo control characters as ^A, ^B etc. */
+  if ((lflag & ECHO) && ((c <= 0x1f && c != '\t' && c != '\n') || c == 0x7f)) {
```
cahirwpz, line 231: Two magic constants... is there any macro that defines them?

sys/kern/tty.c
```diff
@@ -173,6 +216,123 @@ static void tty_outq_putc(tty_t *tty, uint8_t c) {
...
+    tty_output(tty, '^');
+    if (c == 0x7f)
```
cahirwpz, line 233: ditto

sys/kern/tty.c
```diff
@@ -191,16 +353,74 @@ void tty_input(tty_t *tty, uint8_t c) {
...
-      tty_output(tty, CTRL('g'));
-      tty_notify_out(tty);
+  if (lflag & ICANON) {
+    /* Canonical mode character processing takes place here. */
+    if (CCEQ(cc[VERASE], c)) {
+      /* Erase/backspace */
+      uint8_t erased;
+      if (tty_line_unputc(tty, &erased))
+        tty_erase(tty, erased);
+      goto notify;
```
cahirwpz, line 363: Try to refrain yourself from using `goto`'s in cases where control flow is already complex.

cahirwpz (commented):
Introduced control flow is very complex and difficult to read. Please do everything you can to keep it simple to read and understand.

sys/kern/tty.c
j-piecuch: There's no meaning to this constant apart from "the character 0x7f". For this reason, I don't think using a macro here makes sense. If I were to use a macro here, what would it be called, `CHAR_7F`?

sys/kern/tty.c
j-piecuch: I have added one.

sys/kern/tty.c
j-piecuch: Done.

sys/kern/tty.c
cahirwpz: Isn't that a code for delete character?

j-piecuch: It is. I have added the macro.

cahirwpz (approved):
LGTM

PR #768, opened 2020-10-10
AArch64 timer.

pj1031999:
Enable timer on AArch64.

It depends on changes from #769, #770, #771, #772, #787, #788.

PR #788, opened 2020-10-28
AArch64 interrupts

pj1031999:
Add interrupts related functions from #768.
It should be merged after #787.


xThaid:
As far as I'm concerned, looks good.

xThaid (approved):

sys/aarch64/rootdev.c
```diff
@@ -5,10 +5,42 @@
...
+ * 32 local interrupts -- one per CPU but now we only support 1 CPU
+ * It is different than bcm2835reg.h layout where we have:
+ *  - 4x32 base interrupts
+ *  - 32 GPU0 interrupts -- 128 offset in table
+ *  - 32 GPU1 interrupts -- 160 offset in table
+ *  - 32 base interrupts -- 192 offset in table
+ */
+#define N_IRQ (BCM2835_NIRQ + BCM2836_NIRQPERCPU)
+
+/* These should be should be exposed for the children of rootdev. */
```
ThomasStachovsky, line 27: A tiny typo in the comment

sys/aarch64/rootdev.c
```diff
@@ -45,11 +77,49 @@ static bus_space_t *rootdev_bus_space = &(bus_space_t){
...
+static void rootdev_intr_handler(device_t *dev, void *arg) {
+  assert(dev != NULL);
+  rootdev_t *rd = dev->state;
+
+  /* Handle local interrupts. */
+  intr_handle(rootdev_local_handle + BCM2836_LOCAL_INTC_IRQPENDINGN(0),
+              &rd->intr_event[LOCAL_OFFSET]);
+
+  /* Handle base interrupts. */
+  intr_handle(rd->arm_base + (BCM2835_INTC_IRQBPENDING & OFFSET_MASK),
```
ThomasStachovsky, line 113: Can You explain why we use OFFSET_MASK here and in the lines below? I can't seem to find a case where it makes a difference i.e. BCM2835_INTC_IRQBPENDING is already a relative address (because BCM2835_INTC_BASE is).

ThomasStachovsky (approved):
Looks good to me.
I asked a question regarding one of the additions.

sys/aarch64/rootdev.c
pj1031999: done

sys/aarch64/rootdev.c
pj1031999: Thanks! It's my mistake.
It should be `BCM2835_ARMICU_BASE` + `BCM2835_INTC_*`.
It should make more sense now.

ThomasStachovsky (approved):
Everything looks good now!

mohrcore (approved):

mohrcore:
LGTM

sys/aarch64/rootdev.c
```diff
@@ -45,11 +77,52 @@ static bus_space_t *rootdev_bus_space = &(bus_space_t){
...
+  intr_event_add_handler(event, handler);
 }
 
 static void rootdev_intr_teardown(device_t *dev, intr_handler_t *handler) {
-  /* TODO(pj) */
+  intr_event_remove_handler(handler);
+}
+
+/* Read 32 bit pending register located at va and run handlers. */
+static inline void intr_handle(vaddr_t va, intr_event_t *events) {
```
cahirwpz, line 92: Does compiler honour `inline` in case of function as large as this one?

sys/aarch64/rootdev.c
```diff
@@ -45,11 +77,52 @@ static bus_space_t *rootdev_bus_space = &(bus_space_t){
...
+  assert(va != 0);
+  assert(events != NULL);
+
+  uint32_t pending = *(uint32_t *)va;
+
+  for (uint32_t irq = 0; irq < 32; irq++) {
```
cahirwpz, line 98: Umh... in most cases pending will be either 0 or will have a single bit. Could you optimize for such cases. Please use `ffs`.

sys/aarch64/rootdev.c
cahirwpz: Could you pass a pointer that has a meaningful name instead of just `va` ?

sys/aarch64/rootdev.c
```diff
@@ -45,11 +77,52 @@ static bus_space_t *rootdev_bus_space = &(bus_space_t){
...
+static void rootdev_intr_handler(device_t *dev, void *arg) {
+  assert(dev != NULL);
+  rootdev_t *rd = dev->state;
+
+  /* Handle local interrupts. */
+  intr_handle(rootdev_local_handle + BCM2836_LOCAL_INTC_IRQPENDINGN(0),
+              &rd->intr_event[LOCAL_OFFSET]);
+
+  /* Handle base interrupts. */
+  intr_handle(rd->arm_base + ((BCM2835_ARMICU_BASE + BCM2835_INTC_IRQBPENDING) &
```
cahirwpz, line 113: Since many `BCM2835_ARM_BASE` is page aligned, maybe we should rename `BCM2835_ARMICU_BASE` to `BCM2835_ARMICU_OFFSET` and give it a value relative to `BCM2835_ARM_BASE`. You would no need `OFFSET_MASK` then.

cahirwpz (changes requested):

sys/aarch64/rootdev.c
cahirwpz: Do you have to manually apply interrupt mask?

sys/aarch64/rootdev.c
pj1031999: Yes, compiler honour `inline` in that case.

PR #790, opened 2020-10-30
Allow sleeping with a locked vnode.

j-piecuch:
This is a workaround that allows us to sleep in implementations of `VOP_READ` and `VOP_WRITE`. In the future, we should replace it with a proper lock that allows sleeping.


cahirwpz:
LGTM

PR #791, opened 2020-10-30
Allow processes to set the process group of their children.

j-piecuch:
`ksh` needs it for proper job control.


bin/utest/pgrp.c
```diff
@@ -45,6 +39,98 @@ static void sa_handler(int signo) {
...
+int test_setpgid_leader(void) {
+  signal(SIGUSR1, sa_handler);
+
+  pid_t cpid = fork();
+  if (cpid == 0) {
+    /* Become session leader. */
+    assert(setsid() == getpid());
+    /* Can't change pgrp of session leader. */
+    assert(setpgid(0, 0));
+    while (!sig_delivered)
```
cahirwpz, line 51: Since we have `sigprocmask` and `sigsuspend` at our disposal, could you replace `sched_yield` with proper synchronisation?

bin/utest/pgrp.c
```diff
@@ -45,6 +39,98 @@ static void sa_handler(int signo) {
...
+  /* Wait until child becomes session leader. */
+  while (getsid(cpid) != cpid)
+    sched_yield();
+
+  /* Can't change pgrp of session leader. */
+  assert(setpgid(cpid, getpgid(0)));
+
+  kill(cpid, SIGUSR1);
+
+  int status;
```
cahirwpz, line 65: This code repeats over and over... could we replace it with a function, for instance `assert(child_exitcode(pid, 0))`?

bin/utest/pgrp.c
```diff
@@ -45,6 +39,98 @@ static void sa_handler(int signo) {
...
+    /* Can't change pgrp of session leader. */
+    assert(setpgid(0, 0));
+    while (!sig_delivered)
+      sched_yield();
+    return 0;
+  }
+
+  /* Wait until child becomes session leader. */
+  while (getsid(cpid) != cpid)
+    sched_yield();
```
cahirwpz, line 58: ditto

cahirwpz (commented):
Kernel side looks good to me. However userspace code could use some care.

Also... I realized we have a small bug in `do_sigsuspend` implementation https://github.com/cahirwpz/mimiker/blame/1db7bcbee9f455e5eeaf4e44e25ea4b521d3c597/sys/kern/signal.c#L171 - could you fix that?

PR #792, opened 2020-10-30
Tiny changes.

MichalBlk:
Fix a typo and remove a needless header.


cahirwpz:
Could you split all of those changes into three separate PRs. In case of rman I'd like to request additional changes if possible. Perhaps interrupt handling also calls for some improvements. I'd like to discuss these issues separately.

PR #793, opened 2020-10-31
Improve hardware resource management

MichalBlk:
Resource manager should manage multiple regions.


cahirwpz:
For the record: we need a single rman to manage multiple disjoint ranges of memory / ioports / irqs.

PR #794, opened 2020-10-31
Improve interrupt handling framework

MichalBlk:
Improvements:
- fix of the intr_event_run_handlers():
     - don't lose interrupts on the same interrupt event (I know that it doesn't affect us for now), 
     - handle undefined filter functions (since filter isn't mandatory),
     - properly count handlers on an interrupt event

### DIFF
--- a/include/sys/interrupt.h
+++ b/include/sys/interrupt.h
@@ -51,7 +51,7 @@ typedef struct intr_handler intr_handler_t;
 typedef enum {
   IF_STRAY = 0,    /* this device did not trigger the interrupt */
   IF_FILTERED = 1, /* the interrupt has been handled and can be EOId */
-  IF_DELEGATE = 2, /* the handler should be run in private thread */
+  IF_DELEGATE = 2  /* the handler should be run in private thread */
 } intr_filter_t;
 
 /*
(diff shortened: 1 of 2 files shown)


cahirwpz:
For the record: we need to dynamically create and bind interrupt threads with `intr_event` and provide fast method of waking them up in case a task was delegated from filter routine.

jpszczolowski:
After your change if several `ih_filter`s return `IF_DELEGATE`, we'll call `ie_disable` **once** here:
https://github.com/cahirwpz/mimiker/blob/692d86121a22eded3ca903541afadc168441a96d/sys/kern/interrupt.c#L117-L119
but call `ie_enable` **more than once** (i.e. for each `ih`) in `intr_thread` here:
https://github.com/cahirwpz/mimiker/blob/692d86121a22eded3ca903541afadc168441a96d/sys/kern/interrupt.c#L143-L144

Is that intended?

jpszczolowski:
I'm not sure, but maybe we want to call `ie_disable` once and `ie_enable` once (after last `ih`, that comes from a given `ie`, has been serviced by the `intr_thread`).

In FreeBSD they do something similar by calling `ie->ie_pre_ithread` once:
http://bxr.su/FreeBSD/sys/kern/kern_intr.c#1409
and then `ie->ie_post_ithread` once:
http://bxr.su/FreeBSD/sys/kern/kern_intr.c#1212
BTW their infrastructure is a little bit different -- they have an `intr_thread` per each `intr_event`.

cahirwpz:
@jpszczolowski FYI our short-term goal is to add single `thread` per `intr_event` on demand – i.e. when first handler is registered that requires execution of `ih_handler` in interrupt thread context.

BTW it's good to see you here! Given your experience with _Mimiker_, it would be great if you could help us with code reviews from time to time!

PR #796, opened 2020-10-31
Bugfixes from #785

j-piecuch:


include/sys/proc.h
```diff
@@ -60,8 +60,9 @@ typedef enum { PS_NORMAL, PS_STOPPED, PS_DYING, PS_ZOMBIE } proc_state_t;
 
 typedef enum {
   /* Cleared when continued or reported by wait4. */
-  PF_STOPPED = 0x1,   /* Set on stopping */
-  PF_CONTINUED = 0x2, /* Set when continued */
+  PF_STOPPED = 0x1,             /* Set on stopping */
```
cahirwpz, line 63: Can we actually merge `PF_STOPPED` and `PF_CONTINUED` into `PF_STATE_CHANGED` ?

cahirwpz (commented):
I've got a question that is not directly related to this change, but still I'd like you to answer it.

include/sys/proc.h
j-piecuch: Yes, I think it's possible.

cahirwpz: I have a pending change that does that I'll commit it.

sys/kern/signal.c
```diff
@@ -300,23 +311,23 @@ void sig_post(signo_t sig) {
...
     klog("Stopping thread %lu in process PID(%d)", td->td_tid, p->p_pid);
     p->p_state = PS_STOPPED;
-    p->p_flags &= ~PF_CONTINUED;
-    p->p_flags |= PF_STOPPED;
-    cv_broadcast(&p->p_parent->p_waitcv);
-    proc_unlock(p);
-    WITH_MTX_LOCK (&p->p_parent->p_lock)
+    p->p_flags |= PF_STATE_CHANGED;
+    WITH_PROC_LOCK(p->p_parent) {
+      proc_wakeup_parent(p->p_parent);
```
cahirwpz, line 316: Why waking up process is not a part of `sig_kill` responsibility?

j-piecuch: Because sending a signal doesn't change the state of the process. For instance, I can send `SIGSTOP` and `SIGKILL`, and if `SIGKILL` is delivered first, then the process won't stop at any time.

j-piecuch: `SIGCONT` is an exception: the act of sending it changes the state of the target process, so we notify the parent in `sig_kill()` in that case.

sys/kern/proc.c
```diff
@@ -448,8 +448,18 @@ static void proc_reparent(proc_t *old_parent, proc_t *new_parent) {
 
   /* The new parent might be waiting for its children to change state,
    * so notify the parent so that they check again. */
-  if (new_parent)
-    cv_broadcast(&new_parent->p_waitcv);
+  if (new_parent) {
```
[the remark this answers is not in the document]
j-piecuch: Right now, `proc_reparent()` always gets called with non-null `new_parent`, so it's a sensible change.

sys/kern/signal.c
cahirwpz: Should not `sig_kill(p, SIGCHLD)` attempt to wake up `p` if `p` is sleeping on `p_waitcv` ?

j-piecuch: Sending a signal is one thing, and notifying the parent process of a child's state change is another. An ordinary process can send `SIGCHLD` to another process, and that doesn't mean that a child has changed state.

cahirwpz: You're right! Thanks!

cahirwpz (approved):
LGTM

PR #798, opened 2020-11-01
Get rid of the TDF_SLEEPY flag.

j-piecuch:
Now we hold a spinlock all the way from entering a thread into the sleepq until `sched_switch()`, so we can't receive a signal
or be woken up in the middle.
This also eliminates the need for 2 signal checks and fixes a bug where we didn't leave the sleepq if the second signal check finds that we should interrupt the sleep:  https://github.com/cahirwpz/mimiker/blob/5073f58cd2cf623713f7f0e9614bb66e7208b563/sys/kern/sleepq.c#L171-L174


cahirwpz:
Though we're holding `td_lock` for a little bit longer, I recognize the value of code simplification. If this fixes a bug, then it's even better. LGTM.

PR #801, opened 2020-11-02
Fix locking during fork

franciscozdo:
We were copying credentials of parent process with no lock held. Unless we have only one thread per process it won't break. But it will be safer and less error prone in the future.


pj1031999 (approved):
LGTM

cahirwpz (approved):
LGTM

PR #802, opened 2020-11-02
Signal privilege checks

franciscozdo:
Introduce checking privilege to signal other process.


sys/kern/cred.c
```diff
@@ -1,9 +1,21 @@
...
   assert(mtx_owned(&from->p_lock));
   memcpy(&to->p_cred, &from->p_cred, sizeof(cred_t));
 }
+
+int cred_cansignal(proc_t *p, cred_t *cred) {
+  assert(mtx_owned(&p->p_lock));
+
+  if (cred->cr_ruid != p->p_cred.cr_ruid &&
+      cred->cr_ruid != p->p_cred.cr_suid &&
+      cred->cr_euid != p->p_cred.cr_ruid && cred->cr_euid != p->p_cred.cr_suid)
```
j-piecuch, line 17: I assume the semantics of this check are based on some standard or an existing implementation in some other kernel -- can you point me to it so that I can compare the two?

j-piecuch (changes requested):
We need to sort out the locking issues before merging this.

sys/kern/cred.c
franciscozdo: Yes. It is described in standard: [kill()](https://pubs.opengroup.org/onlinepubs/009695399/). And [example](http://bxr.su/FreeBSD/sys/kern/kern_prot.c#1552) in code of FreeBSD.

sys/kern/proc.c
```diff
@@ -697,3 +697,14 @@ int do_waitpid(pid_t pid, int *status, int options, pid_t *cldpidp) {
   }
   __unreachable();
 }
+
+int proc_cansignal(proc_t *receiver, proc_t *sender, signo_t sig) {
+  assert(mtx_owned(&receiver->p_lock));
+  assert(mtx_owned(&sender->p_lock));
```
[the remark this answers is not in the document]
franciscozdo: I have it in mind. It was my first try, and I plan to do it in some kind of first solution you proposed.

sys/kern/cred.c
j-piecuch: Thanks, looks good!

sys/kern/proc.c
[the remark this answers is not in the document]
franciscozdo: Please check if I have done it right now.

j-piecuch (changes requested):
In addition to my comments in the code:
- I don't like that all these signal-related functions are in `proc.c`. I think they should go into `signal.c`.
- I would also like you to add a `*` flag to `p_cred` in `proc.h`, which means it's safe to read `p_cred` of the current process without locking it.
@cahirwpz what do you think about migrating signal-related `proc_*` functions to `signal.c`? Or maybe another file like `proc_signal.c`?

sys/kern/proc.c
```diff
@@ -569,42 +569,81 @@ __noreturn void proc_exit(int exitstatus) {
...
+  int error = 0;
+  proc_t *target;
+  cred_t *cred = kmalloc(M_TEMP, sizeof(cred_t), 0);
+
+  if (!cred)
+    return ENOMEM;
+
+  WITH_PROC_LOCK(p) {
+    cred_copy(cred, p);
+  }
```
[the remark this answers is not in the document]
franciscozdo: Would it be true if we will have multiple threads in the future? Then another thread can change credentials in the same time. I am not sure if I can leave it without locking (it would be hard to find when aligning code to multiple threads).

sys/kern/proc.c
```diff
@@ -697,3 +736,17 @@ int do_waitpid(pid_t pid, int *status, int options, pid_t *cldpidp) {
...
+
+int proc_cansignal(proc_t *p, cred_t *cred, proc_t *target, signo_t sig) {
+  assert(mtx_owned(&target->p_lock));
+
+  /* process can signal itself */
+  if (p == target)
+    return 0;
+
+  /* process can send SIGCONT to every process in the same session */
+  if (sig == SIGCONT && p->p_pgrp->pg_session == target->p_pgrp->pg_session)
```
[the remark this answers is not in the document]
franciscozdo: Changed. I store value of `p->p_pgrp->pg_session` to check against it without problems.

include/sys/proc.h
```diff
@@ -143,7 +143,7 @@ proc_t *proc_find(pid_t pid);
  * (pid > 0) sends signal to the process with the ID specified by pid.
  * (pid = 0) sends signal to processes in process group of the calling process.
  * (pid <-1) sends signal to processes in process group with ID equal (-pid). */
-int proc_sendsig(pid_t pid, signo_t sig);
+int proc_sendsig(proc_t *p, pid_t pid, signo_t sig);
```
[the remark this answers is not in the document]
franciscozdo: Done.

sys/kern/proc.c
```diff
@@ -569,42 +569,81 @@ __noreturn void proc_exit(int exitstatus) {
...
+}
+
+int proc_sendsig(proc_t *p, pid_t pid, signo_t sig) {
+
+  if (sig >= NSIG)
+    return EINVAL;
+
+  int error = 0;
+  proc_t *target;
+  cred_t *cred = kmalloc(M_TEMP, sizeof(cred_t), 0);
```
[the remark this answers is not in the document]
franciscozdo: Okay. Now I use credentials directly from process structure.

sys/kern/proc.c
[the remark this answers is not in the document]
j-piecuch: I wouldn't worry about multithreaded processes for now.

sys/kern/cred.c
```diff
@@ -7,3 +7,8 @@ void cred_fork(proc_t *to, proc_t *from) {
   assert(mtx_owned(&from->p_lock));
   memcpy(&to->p_cred, &from->p_cred, sizeof(cred_t));
 }
+
+void cred_copy(cred_t *cr, proc_t *p) {
+  assert(mtx_owned(&p->p_lock));
+  memcpy(cr, &p->p_cred, sizeof(cred_t));
+}
```
j-piecuch, line 14: I don't think we need this anymore, do we?

sys/kern/cred_checks.c
```diff
@@ -0,0 +1,32 @@
...
+      cred->cr_ruid != target->p_cred.cr_suid &&
+      cred->cr_euid != target->p_cred.cr_ruid &&
+      cred->cr_euid != target->p_cred.cr_suid)
+    return EPERM;
+
+  return 0;
+}
+
+int proc_cansignal(cred_t *cred, session_t *s, proc_t *target, signo_t sig) {
+  assert(mtx_owned(&target->p_lock));
```
j-piecuch, line 22: I would like the signature of this function to look like this:
```c
/* Checks whether the current process has permission to send `sig` to `target` process. */
int proc_cansignal(proc_t *target, signo_t sig)
```
We can require `all_proc_mtx` to be held in addition to `&target->p_lock` -- then we will be able to safely access the current process's session.

sys/kern/proc.c
```diff
@@ -569,42 +569,76 @@ __noreturn void proc_exit(int exitstatus) {
   thread_exit();
 }
 
+static int proc_pgsignal(cred_t *cred, session_t *s, pgid_t pgid, signo_t sig) {
+  pgrp_t *pgrp = NULL;
+
+  WITH_MTX_LOCK (all_proc_mtx) {
```
j-piecuch, line 575: We can hold `all_proc_mtx` for the whole duration of the function. This can hurt performance a little bit, but it simplifies the code. It's also required after you make the changes I suggested in `proc_cansignal()`.

sys/kern/proc.c
```diff
@@ -569,42 +569,76 @@ __noreturn void proc_exit(int exitstatus) {
...
+
+  int error;
+  proc_t *p = proc_self();
   proc_t *target;
+  session_t *session;
+
+  WITH_PROC_LOCK(p) {
+    /* Take a snapshot of session ptr. We will use it in privilege check. */
+    session = p->p_pgrp->pg_session;
+  }
```
j-piecuch, line 616: Not needed if we call `proc_cansignal()` with `all_proc_mtx` held, which I suggest we do.

sys/kern/proc.c
```diff
@@ -569,42 +569,76 @@ __noreturn void proc_exit(int exitstatus) {
...
 
   if (pid > 0) {
     WITH_MTX_LOCK (all_proc_mtx)
```
j-piecuch, line 619: We could make this a `SCOPED_MTX_LOCK`.

j-piecuch (changes requested):
I'm in favor of simplifying the code at the expense of more coarse-grained locking.

sys/kern/cred.c
franciscozdo: Yes. Removed.

sys/kern/cred_checks.c
franciscozdo: Applied your suggestion.

sys/kern/cred.c
j-piecuch: It hasn't been removed here. Have you pushed your branch?

j-piecuch (approved):

cahirwpz (approved):
LGTM

PR #803, opened 2020-11-02
Make it possible to signal a process group without holding all_proc_mtx.

j-piecuch:
This is the non-bugfix part of #785.


cahirwpz (approved):
LGTM

PR #806, opened 2020-11-02
Add error handling while sending signals

franciscozdo:
This changes will prepare signal sending mechanism for future changes in checking privileges to sending signals.


include/sys/signal.h
```diff
@@ -81,14 +81,18 @@ typedef struct ctx ctx_t;
  *
  * \sa sig_post
  * \note Must be called with p::p_lock held. Returns with p::p_lock held.
+ *
+ * \returns error code if can't signal process
  */
-void sig_kill(proc_t *p, signo_t sig);
+int sig_kill(proc_t *p, signo_t sig);
```
j-piecuch, line 87: I don't think we want to add permission checking to `sig_kill()`: its responsibility should be to send a signal to a process. We can create another procedure that checks permissions before calling `sig_kill()` or simply check permissions manually before calling it. This is not unreasonable, since we need permission checks only when sending a signal on behalf of a user, i.e. in `sys_kill()`. NetBSD does manual permission checks as needed before calling [`kpsignal2()`](http://bxr.su/NetBSD/sys/kern/kern_sig.c#kpsignal2): see [`kill1()`](http://bxr.su/NetBSD/sys/kern/sys_sig.c#kill1).

include/sys/signal.h
```diff
@@ -81,14 +81,18 @@ typedef struct ctx ctx_t;
...
 
 /*! \brief Signal all processes in a process group.
  *
  * \note Must be called with pg::pg_lock held. Returns with pg::pg_lock held.
+ *
+ * \returns error code if can't signal any of the processes
  */
-void sig_pgkill(pgrp_t *pg, signo_t sig);
+int sig_pgkill(pgrp_t *pg, signo_t sig);
```
j-piecuch, line 95: Likewise, I think the use of `sig_pgkill()` should be similar to use of [`kpgsignal()`](http://bxr.su/NetBSD/sys/kern/kern_sig.c#885): permission checks should be carried out elsewhere. We might need another procedure for signaling a process group on behalf of a user, similar to [`killpg1()`](http://bxr.su/NetBSD/sys/kern/kern_sig.c#809).

j-piecuch (changes requested):
I disagree with adding permission checks directly to `sig_kill()` and `sig_pgkill()`. These procedures should responsible only for sending a signal, so that they can be called on behalf of the kernel (e.g. from the TTY subsystem), as well as on behalf of users, after checking permissions.

franciscozdo:
According to comments this changes are not needed.

PR #807, opened 2020-11-03
Clarify behavior of sig_post().

j-piecuch:
It wasn't clear to me that calling `sig_post()` multiple times won't make us lose pending signals. It turns out it's safe, but it's far from obvious. I have added some comments to help future readers of the code.


franciscozdo (approved):
LGTM

PR #808, opened 2020-11-04
Discard pending signals when ignoring them via sigaction().

j-piecuch:
This gives us the nice property that if `sig_pending()` returns a signal, then that signal isn't ignored. I will be using this property in the future.


franciscozdo (approved):
LGTM

cahirwpz (approved):
LGTM

PR #809, opened 2020-11-05
Move TTY I/O functions into fileops. 

j-piecuch:
This allows us to bypass the vnode layer during I/O, which we need to do, since the vnode layer acquires the vnode lock for the duration of `VOP_READ` and `VOP_WRITE`.


franciscozdo (approved):
LGTM

cahirwpz (approved):
LGTM

PR #811, opened 2020-11-06
Properly catch stop signals in sleepq_wait().

j-piecuch:
Sending a stop signal to a sleeping thread shouldn't interrupt its sleep (in the sense of returning `EINTR`).
It should make the thread stop, and after it's resumed, return from `sleepq_wait()` without an error, so it's somewhat like a spurious wakeup. We expect callers to handle such wakeups correctly.


franciscozdo (approved):
LGTM
I don't see better way to wait for child's action too.

cahirwpz (commented):
Given `sleepqueue` and `turnstile` are low-level tools for implementing synchronisation mechanisms I'd like them not to be tied to high-level abstractions like signals and processes in any way. Could you move `sq_check_signals` functionality outside of `sleepq.c` ?

j-piecuch:
We decided that a better approach is for `SIGSTOP` to interrupt the sleep, stop at the userspace boundary, and restart the interrupted syscall if no signals with handlers were caught.

PR #817, opened 2020-11-09
Sequential bar numbers.

MichalBlk:
Make device bars appear sequentially in the PCI device structure
so that a driver does not have to check which bar is vaild and which is not.

### DIFF
--- a/sys/drv/stdvga.c
+++ b/sys/drv/stdvga.c
@@ -173,7 +173,7 @@ static int stdvga_attach(device_t *dev) {
 
   stdvga->mem =
     bus_alloc_resource_any(dev, RT_MEMORY, 0, RF_ACTIVE | RF_PREFETCHABLE);
-  stdvga->io = bus_alloc_resource_any(dev, RT_MEMORY, 2, RF_ACTIVE);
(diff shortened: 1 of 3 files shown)


sys/drv/stdvga.c
```diff
@@ -173,7 +173,7 @@ static int stdvga_attach(device_t *dev) {
 
   stdvga->mem =
     bus_alloc_resource_any(dev, RT_MEMORY, 0, RF_ACTIVE | RF_PREFETCHABLE);
-  stdvga->io = bus_alloc_resource_any(dev, RT_MEMORY, 2, RF_ACTIVE);
```
pj1031999, line 176: Why we need to change `rid` here?

pj1031999 (approved):
LGTM

sys/drv/stdvga.c
MichalBlk: Now all valid bars of a device appear in sequence. 
Before this fix, the VGA driver had to be aware of the fact that we sense the VGA bars as 
valid, invalid, valid, invalid, invalid, and invalid.
Since VGA only uses two bars, after the fix, bar 2 (and the rest) are invalid and therefore cannot be used.

pj1031999: Thanks for your explanation.

PR #821, opened 2020-11-13
Implement setting uid and gid according to set-user/group-ID bits

franciscozdo:
When set-user-ID bit is set in the new process file mode then effective user id is set to owner of file and saved user id is set to old effective user id. The same is done when set-group-ID bit is set.


sys/kern/cred.c
```diff
@@ -7,3 +7,14 @@ void cred_fork(proc_t *to, proc_t *from) {
   assert(mtx_owned(&from->p_lock));
   memcpy(&to->p_cred, &from->p_cred, sizeof(cred_t));
 }
+
+void cred_exec_change_id(proc_t *p, uid_t uid, gid_t gid) {
+  if (uid != (uid_t)-1) {
```
wiklam, line 12: Assert for holding a lock can prevent future bugs ;)

wiklam (approved):
LGTM

sys/kern/exec.c
```diff
@@ -302,13 +303,32 @@ static void destroy_vmspace(exec_vmspace_t *saved) {
...
+  *uid = *gid = -1;
+
+  if (VOP_GETATTR(vn, &attr))
+    return false;
+
+  if (attr.va_mode & S_ISUID)
+    *uid = attr.va_uid;
+
+  if (attr.va_mode & S_ISGID)
+    *uid = attr.va_gid;
```
xThaid, line 317: Shouldn't be here `*gid` instead of `*uid`?

franciscozdo: Thanks. It should.

xThaid (approved):

sys/kern/exec.c
```diff
@@ -339,6 +356,10 @@ static int _do_execve(exec_args_t *args) {
     use_interpreter = true;
   }
 
+  uid_t uid;
+  gid_t gid;
+  bool setid = check_setid(vn, &uid, &gid);
```
cahirwpz, line 361: I think this implementation is prone to TOCTOU related vulnerabilities. How do we ensure nobody changed the binary in the meantime? Suppose we have a binary with `suid` or `sgid` that got replaced between lines 361 and 373 - is that possible? If so, then I'd be a good idea to leave a comment here to future readers to make them aware of it.

cahirwpz (approved):
LGTM but I'd like you to answer a question and possibly add a comment.

sys/kern/exec.c
franciscozdo: Yes, it is possible. I left a comment.

j-piecuch: Why don't we lock the vnode then?

franciscozdo: It will be resolved by adding `ETXTBSY` error here and in filesystem syscalls.

PR #820, opened 2020-11-13
Manage IRQ resources.

cahirwpz:
An attempt to manage IRQ as if they were resources. To use an interrupt you have to first allocate if from device's bus.

PR #826, opened 2020-11-15
Somewhat more reasonable resource manager

MichalBlk:
The main goals:
  - Implement multiple regions within a single resource manager.
  - ~~Add proper alignment mechanism (via `res_flags_t`).~~ (alignment is now separated from `res_flags_t` and is only part of internal driver interface).


cahirwpz:
Please resolve conflicts after #820 merge.

MichalBlk:
I've limited sharing to irqs and updated PR due to #820 .

sys/aarch64/timer.c
```diff
@@ -80,7 +80,7 @@ static int arm_timer_attach(device_t *dev) {
   };
 
   state->irq_res =
-    bus_alloc_irq(dev, 0, BCM2836_INT_CNTPNSIRQ_CPUN(0), RF_ACTIVE);
+    bus_alloc_irq(dev, 0, BCM2836_INT_CNTPNSIRQ_CPUN(0), RF_ACTIVE | RF_SHAREABLE);
```
pj1031999, line 83: Why we need that flag here?

sys/mips/cbus.c
```diff
@@ -16,10 +16,5 @@ static bus_space_t cbus_bus_space = {
...
 
-resource_t cbus_uart[1] = {{.r_bus_tag = &cbus_bus_space,
-                            .r_bus_handle = 0,
-                            /* RT_MEMORY since CBUS UART is a device mapped in
-                             * common physical address space of MIPS CPU. */
-                            .r_type = RT_MEMORY,
-                            .r_start = 0,
-                            .r_end = 7}};
+resource_t cbus_uart[1] = {
+  {.r_bus_tag = &cbus_bus_space, .r_bus_handle = 0, .r_start = 0, .r_end = 7}};
```
pj1031999, line 20: Could you format it?

include/sys/rman.h
```diff
@@ -6,82 +6,101 @@
...
- * power of 2.
+ * passed via `flags` argument (see `rman_make_alignment_flags`).
  *
  * \returns NULL if could not allocate a resource
  */
-resource_t *rman_alloc_resource(rman_t *rm, rman_addr_t start, rman_addr_t end,
-                                size_t count, size_t bound, res_flags_t flags);
+resource_t *rman_reserve_resource(rman_t *rm, rman_addr_t start,
+                                  rman_addr_t end, size_t count,
+                                  res_flags_t flags);
```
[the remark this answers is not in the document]
MichalBlk: We don't need the `dev` pointer because our resource doesn't have the `r_dev` (or `r_owner`). It may change soon regarding resource lists.

sys/aarch64/timer.c
MichalBlk: Should this interrupt be exclusive to the timer?

sys/mips/cbus.c
MichalBlk: This is the outcome of the `make format` command. Should I format it manually to make it look like it does now?

include/sys/rman.h
```diff
@@ -6,82 +6,101 @@
...
+  RF_FIRSTSHARE = 4,
+  RF_ACTIVE = 8,
 } res_flags_t;
 
-/* XXX: since a resource should be created only as a result of a call to
- * `rman_alloc_resource` perhaps we could make the definition private? */
+#define RF_ALIGNMENT_SHIFT 10 /* alignment size bit starts bit 10 */
+#define RF_ALIGNMENT_MASK (0x003F << RF_ALIGNMENT_SHIFT)
+#define RF_ALIGNMENT_LOG2(x) ((x) << RF_ALIGNMENT_SHIFT)
+#define RF_ALIGNMENT(x) (((x)&RF_ALIGNMENT_MASK) >> RF_ALIGNMENT_SHIFT)
```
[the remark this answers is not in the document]
MichalBlk: Comments added. Since these macros are used exclusively by rman so I moved them to rman.c.

include/sys/rman.h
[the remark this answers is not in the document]
pj1031999: Thanks for your explanation.

sys/mips/cbus.c
pj1031999: Look for drivers declaration in `/sys/drv`. They use `/* clang-format off */` and `/* clang-format on */` for that purpose.
It's not _very_ important but from my point of view it's really helpful.

include/sys/rman.h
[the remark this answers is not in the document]
pj1031999: Thanks!

sys/aarch64/timer.c
pj1031999: I thought about "what is a difference now". Previously it just worked so what is a implication of that flag?

MichalBlk: It doesn't really make a difference for now. It allows future sharing of the `BCM2836 INT_CNTP SIRQ CPU(0)` interrupt. I added it in order to remove any confusions associated with any future attempt to share this interrupt. Do you think that we should delay this change to some future PR?

sys/mips/cbus.c
MichalBlk: Done.

sys/aarch64/timer.c
pj1031999: As long as it works I don't care but from my point of view the best what we can do now is discuss that offline with @cahirwpz and then make a decision.

pj1031999: After offline discussion – let it be.

mohrcore (changes requested):
I made my initial review. It might be incomplete, because I'm yet to read through `rman_region_reserve_resource` and `rman_reserve_resource`, however I have a couple of comments to share already and I believe some changes should be made.

I will post an update after I'm done analysing `rman_region_reserve_resource` and `rman_reserve_resource`.

sys/kern/rman.c
```diff
@@ -1,83 +1,252 @@
...
-
-    /* Does region of `count` size fit in the gap at `bound` aligned address? */
-    if (first + count - 1 <= last) {
-      *start_p = align(first, bound);
-      *res_p = curr_r;
+    rstart = RMAN_ALIGN(max(start, gap_start), amask);
+    /* Check for overflow. */
+    if ((rstart + count - 1) < rstart)
+      break;
+    rend = min(gap_end, max(rstart + count - 1, end));
```
mohrcore, line 197: Evalution of `max(rstart + count)` seems to be unnecessary here. `rend` is only used at line _199_:
```C
if (rstart <= rend && rend - rstart >= count - 1) {
```
and after substituting `rend` with `rstart + count - 1` the condition of the _if statement_ becomes
```C
(rstart + count - 1) - rstart >= count - 1
```
This is obviously **always true**.

I suggest replacing `max(rstart + count - 1, end)` with `end`.

mohrcore (changes requested):
Ok, I finished reading through the rest of the code. I left one more comment with suggested change.

sys/kern/rman.c
mohrcore: Uh, never mind. That was a stupid remark.

However I'm still confused by this line.
I don't understand tho why do you compare `rstart + count - 1` with `end`. Is there really a case when `rstart + count - 1` is greater than `end`? Wouldn't that mean that there's not enough space for the resource?
I think that lines _186-188_ and _162-163_ already handle such scenarios.

mohrcore (changes requested):
I found one of my comments to be nonsense. I stroke it through. But I still have some doubts about the line I commented at.

sys/mips/gt64120.c
```diff
@@ -359,22 +363,18 @@ static resource_t *gt_pci_alloc_resource(device_t *dev, res_type_t type,
...
     }
 
-    if (type == RT_MEMORY) {
-      int error = bus_activate_resource(dev, type, rid, r);
-      assert(error == 0);
-    } else {
-      rman_activate_resource(r);
-    }
+    int error = bus_activate_resource(dev, type, rid, r);
+    assert(error == 0);
```
pj1031999, line 377: Should we fail here? Why not to free resource and return NULL?

sys/kern/rman.c
```diff
@@ -87,29 +256,48 @@ resource_t *rman_alloc_resource(rman_t *rm, rman_addr_t first, rman_addr_t last,
...
-void rman_activate_resource(resource_t *r) {
-  rman_t *rm = r->r_rman;
+void rman_release_resource(resource_t *r) {
+  rman_t *rm = RESOURCE_GET_RMAN(r);
+  rman_region_t *rg = r->r_rg;
 
-  WITH_MTX_LOCK (&rm->rm_lock)
-    r->r_flags |= RF_ACTIVE;
+  WITH_MTX_LOCK (&rm->rm_lock) {
+    /* XXX: maybe we should ensure that (r->r_flags & RF_ACTIVE) == 0? */
```
pj1031999, line 274: So we should or not to ensure that?

sys/kern/rman.c
```diff
@@ -1,83 +1,252 @@
...
+void rman_init(rman_t *rm, const char *name) {
+  rm->rm_name = name;
+  mtx_init(&rm->rm_lock, 0);
+  TAILQ_INIT(&rm->rm_regions);
+}
+
+void rman_manage_region(rman_t *rm, rman_addr_t start, rman_addr_t end) {
+  rman_region_t *rg, *s, *t;
+
+  rg = kmalloc(M_RES, sizeof(rman_region_t), M_ZERO);
```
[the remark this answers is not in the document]
MichalBlk: Please notice that we to the allocation here because we want to avoid calling `kmalloc` (without `M_NOWAIT`) inside `WITH_MTX_LOCK (&rm->rm_lock) { ...}`.

sys/kern/rman.c
```diff
@@ -1,83 +1,252 @@
...
-    *start_p = align(start, bound);
+  amask = RMAN_ALIGNMENT_GET(flags);
+  start = RMAN_ALIGN(max(rg->rg_start, start), amask);
+  end = min(rg->rg_end, end);
+  if (end - count + 1 < start)
+    return false;
+
+  /* Does the resource fit before the first resource in the region? */
+  s = TAILQ_FIRST(&rg->rg_resources);
+  if (!s || (s && s->r_start - count >= start)) {
```
pj1031999, line 167: Sometimes you are writing `!expr` sometimes `expr == NULL`. Please be consistent.

sys/kern/rman.c
```diff
@@ -1,83 +1,252 @@
...
-    return false;
+/* Alignment mask to extract the alignment bits. */
+#define RMAN_ALIGNMENT_MASK (0x003F << RMAN_ALIGNMENT_SHIFT)
+
+/* Extracts alignment bits. */
+#define RMAN_ALIGNMENT_EXTRACT(x)                                              \
+  (((x)&RMAN_ALIGNMENT_MASK) >> RMAN_ALIGNMENT_SHIFT)
+
+/* Decodes alignment bits from the `flags` argument. */
+#define RMAN_ALIGNMENT_GET(x) ((1ull << RMAN_ALIGNMENT_EXTRACT(x)) - 1)
```
pj1031999, line 39: You use that in assignments for `rman_addr_t`. Could you cast that here?

sys/kern/rman.c
```diff
@@ -1,83 +1,252 @@
...
+
+#define RMAN_ALIGN(addr, amask) (((addr) + (amask)) & ~(amask))
```
pj1031999, line 41: Maybe static inline functions will be better here?

sys/kern/rman.c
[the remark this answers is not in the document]
mohrcore: Ok, fair enough. Allocating memory for resource in before, just in case it's going to be used seems to be a better choice than calling kmalloc and potentially going to sleep while holding a mutex.

mohrcore (approved):
LGTM

sys/kern/rman.c
```diff
@@ -1,83 +1,252 @@
...
 
-static bool rman_find_gap(rman_t *rm, rman_addr_t *start_p, rman_addr_t end,
-                          size_t count, size_t bound, resource_t **res_p) {
-  assert(mtx_owned(&rm->rm_lock));
+/* Alignment bits start at bit 10. */
+#define RMAN_ALIGNMENT_SHIFT 10
 
-  rman_addr_t start = *start_p;
+/* Positions the alignment bits for the `flags` argument. */
+#define RMAN_ALIGNMENT_ENCODE(x) ((x) << RMAN_ALIGNMENT_SHIFT)
```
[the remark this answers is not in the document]
mohrcore: Resolved with 0b44da5.

sys/kern/rman.c
mohrcore: Resolved in e3b65cc.

PR #828, opened 2020-11-16
Reset caught signals on exec.

j-piecuch:


franciscozdo (approved):
LGTM

sys/kern/signal.c
```diff
@@ -397,6 +397,34 @@ void sig_pgkill(pgrp_t *pg, ksiginfo_t *ksi) {
...
+  assert(mtx_owned(&p->p_lock));
+  thread_t *td = p->p_thread;
+
+  /* The signal mask, pending and ignored signals remain unchanged.
+   * Caught signals have their action reset to SIG_DFL.
+   * If a pending signal becomes ignored due to its default action,
+   * it is discarded. */
+  for (signo_t sig = 1; sig < NSIG; sig++) {
+    if (sig_ignored(p->p_sigactions, sig)) {
+      /* Invariant check. */
```
cahirwpz, line 410: It's clear that next line is invariant check, while it may be not obvious why you check it. Please improve the comment.

include/sys/signal.h
```diff
@@ -143,6 +143,11 @@ int sig_send(signo_t sig, sigset_t *mask, sigaction_t *sa, ksiginfo_t *ksi);
  * \note This is machine dependent code! */
 int sig_return(void);
 
+/*! \brief Perform signal-related tasks on process exec.
```
cahirwpz, line 146: I think `sig_onexec` name is already self-descriptive and there's no need to repeat it here. Perhaps you could write what this procedure actually does (i.e. resets handlers).

cahirwpz (commented):
Code structure wise => LGTM. Comments could be improved.

sys/kern/signal.c
j-piecuch: I don't _need_ to check it, but it's a good opportunity to do so. I have changed the comment to say exactly what the invariant is.

cahirwpz (approved):
LGTM

PR #829, opened 2020-11-16
PIT timer with 64 bit counter

wiklam:
Simplifying the Programmable Interval Timer infrastructure and removing some races during time update.

We do not guarantee that we won't miss a whole period before we read the counter. We could use RTC interrupts to read the counter more often then PIT to prevent tick loss, but after a discussion, it will be too complicated handling it that way.


sys/drv/pit.c
```diff
@@ -12,48 +12,67 @@
...
-  value |= inb(TIMER_CNTR0);
-  value |= inb(TIMER_CNTR0) << 8;
-  return pit->period_cntr - value;
+  count |= inb(TIMER_CNTR0);
+  count |= inb(TIMER_CNTR0) << 8;
+  return count;
+}
+
+static uint64_t pit_get_counter64(pit_state_t *pit) {
+  static uint16_t counter16_last = 0;
```
cahirwpz, line 50: Using global state that is not tied to `pit_state_t` prevents this driver to be instantiated twice. While this is not a problem, it's considered a bad practice while implementing device drivers. Please move `counter16_last` and `counter64_last` to `pit_state_t`.

sys/drv/pit.c
```diff
@@ -12,48 +12,67 @@
...
+  static counter_t counter64_last = (counter_t){.val = 0};
+  uint16_t counter16_now, ticks;
+  uint32_t oldlow;
+  SCOPED_INTR_DISABLED();
```
cahirwpz, line 54: Using interrupt disable in device driver code is a bad practice. Spin locks do the same and are future-proof against SMP systems.

sys/drv/pit.c
```diff
@@ -12,48 +12,67 @@
...
+  if (counter16_last >= counter16_now)
+    ticks = counter16_last - counter16_now;
+  else
+    ticks = counter16_last + (pit->period_cntr - counter16_now);
+
+  counter16_last = counter16_now;
+
+  oldlow = counter64_last.lo;
+
+  if (oldlow > (counter64_last.lo = oldlow + ticks))
```
cahirwpz, line 66: Assignment in comparison... whoa... this is atrocious to readers. Please don't do it.

cahirwpz (commented):
I spotted some things that need improvements. This is not a complete review.

sys/drv/pit.c
wiklam: Done.

sys/drv/pit.c
wiklam: Done.

sys/drv/pit.c
wiklam: Done.

sys/drv/pit.c
```diff
@@ -12,48 +12,78 @@
...
+
+static uint64_t pit_get_counter64(pit_state_t *pit) {
+  uint16_t counter16_now, ticks;
+  uint32_t oldlow;
+
+  counter16_now = pit_get_counter16(pit);
+  if (pit->counter16_last >= counter16_now)
+    ticks = pit->counter16_last - counter16_now;
+  else
+    ticks = pit->counter16_last + (pit->period_cntr - counter16_now);
```
pj1031999, line 60: It's confusing for me. Could you add some comments?

sys/drv/pit.c
```diff
@@ -12,48 +12,78 @@
...
 #define PIIX4_IDE_DEVICE_ID 0x7111
 
+typedef union {
+  /* assumes little endian order */
+  struct {
+    uint32_t lo;
+    uint32_t hi;
+  };
+  uint64_t val;
+} counter_t;
```
pj1031999, line 22: It looks like some universal structure.
Could you add some explanation why we need that structure and why it's specific for PIT timer?

wiklam: Good idea!

sys/drv/pit.c
wiklam: Sure!

sys/drv/pit.c
```diff
@@ -17,43 +17,66 @@ typedef struct pit_state {
...
+  uint16_t counter16_now, ticks;
+  uint32_t oldlow;
+
+  counter16_now = pit_get_counter16(pit);
+  /* PIT counter counts from n to 1 and when we get to 1 an interrupt
+     is send and the counter starts from the beginning (n = pit->period_cntr)*/
+  if (pit->counter16_last >= counter16_now)
+    ticks = pit->counter16_last - counter16_now;
+  else
+    ticks = pit->counter16_last + (pit->period_cntr - counter16_now);
```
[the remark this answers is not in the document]
wiklam: We can not guarantee it. We could use RTC interrupts to read the counter more often then PIT will do to prevent tick loss, but after a discussion, it will be too complicated handling it that way.

ThomasStachovsky (approved):

mohrcore (approved):
LGTM

include/sys/time.h
```diff
@@ -160,6 +160,17 @@ struct itimerval {
...
+/* It allows to extend the capacity
+   for counters with 16/32 bits, . */
+typedef union {
+  /* assumes little endian order */
+  struct {
+    uint32_t lo;
+    uint32_t hi;
+  };
+  uint64_t val;
+} counter_t;
```
cahirwpz, line 172: This type is globally visible in the kernel after inclusion of `sys/time.h`. For that reason I think that the typename is too ambiguous. That would be ok if defined in `_.c` file, but not in a header file.

sys/drv/pit.c
```diff
@@ -14,43 +14,67 @@ typedef struct pit_state {
   spin_t lock;
   resource_t *irq_res;
   timer_t timer;
-  uint64_t period_frac;        /* period as a fraction of a second */
-  uint16_t period_cntr;        /* period as PIT counter value */
-  volatile uint16_t last_cntr; /* last seen counter value */
-  volatile bool overflow;      /* counter overflow detected? */
-  bintime_t time;              /* last time measured by the timer */
+  uint16_t period_cntr;     /* period as PIT counter value */
+  uint16_t counter16_last;  /* last seen counter value */
```
cahirwpz, line 18: Could you change `counter16_last` and `counter64_last` into somewhat more self-descriptive identifiers?

sys/drv/pit.c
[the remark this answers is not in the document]
cahirwpz: Readers in the future may ask the same question. Please leave a comment.

sys/drv/pit.c
```diff
@@ -14,43 +14,67 @@ typedef struct pit_state {
...
+}
+
+static uint64_t pit_get_counter64(pit_state_t *pit) {
+  uint16_t counter16_now, ticks;
+  uint32_t oldlow;
+
+  counter16_now = pit_get_counter16(pit);
+  /* PIT counter counts from n to 1 and when we get to 1 an interrupt
+     is send and the counter starts from the beginning (n = pit->period_cntr)*/
+  if (pit->counter16_last >= counter16_now)
```
cahirwpz, line 47: Please simplify following three lines by extracting common subexpression if possible.

sys/drv/pit.c
```diff
@@ -14,43 +14,67 @@ typedef struct pit_state {
...
+    ticks = pit->counter16_last - counter16_now;
+  else
+    ticks = pit->counter16_last + (pit->period_cntr - counter16_now);
+
+  pit->counter16_last = counter16_now;
+
+  oldlow = pit->counter64_last.lo;
```
cahirwpz, line 54: Could you explain to a reader, why following lines are needed?

cahirwpz (changes requested):
The code requires some changes before it's merged.

sys/drv/pit.c
wiklam: Done.

sys/drv/pit.c
wiklam: Done.

include/sys/time.h
wiklam: Done.

pj1031999 (approved):
LGTM

sys/drv/pit.c
wiklam: Done.

PR #831, opened 2020-11-16
Shared memory in fork

bkjg:
Add sharing shared memory instead of copying in fork syscall


include/sys/vm_object.h
```diff
@@ -3,14 +3,18 @@
...
 /* At the moment assume object is owned by only one vm_map */
 typedef struct vm_object {
+  mtx_t mtx;
   vm_pagelist_t list;
   vm_pagetree_t tree;
   size_t size;
   size_t npages;
   vm_pager_t *pager;
+  refcnt_t ref_counter;
 } vm_object_t;
```
pj1031999, line 18: Which fields are protected by `mtx`? Please add comment.

sys/kern/vm_map.c
```diff
@@ -328,8 +330,15 @@ vm_map_t *vm_map_clone(vm_map_t *map) {
...
     TAILQ_FOREACH (it, &map->entries, link) {
-      vm_object_t *obj = vm_object_clone(it->object);
-      vm_segment_t *seg = vm_segment_alloc(obj, it->start, it->end, it->prot);
+      vm_object_t *obj;
+      vm_segment_t *seg;
+      if (it->flags & VM_SHARED) {
+        refcnt_acquire(&it->object->ref_counter);
+        seg = it;
+      } else {
+        obj = vm_object_clone(it->object);
```
pj1031999, line 339: It's not clear for me what should be done with `ref_counter` in `vm_object_clone`.
Is it possible to have `ref_counter` different than `1` here?
Could you explain?

sys/kern/vm_object.c
```diff
@@ -21,21 +21,32 @@ vm_object_t *vm_object_alloc(vm_pgr_type_t type) {
...
 void vm_object_free(vm_object_t *obj) {
-  while (!TAILQ_EMPTY(&obj->list)) {
-    vm_page_t *pg = TAILQ_FIRST(&obj->list);
-    TAILQ_REMOVE(&obj->list, pg, obj.list);
-    vm_page_free(pg);
+  if (!refcnt_release(&obj->ref_counter)) {
+    return;
+  }
+
+  assert(obj->ref_counter == 0);
```
pj1031999, line 35: Do we need this assert?

sys/kern/vm_object.c
```diff
@@ -21,21 +21,32 @@ vm_object_t *vm_object_alloc(vm_pgr_type_t type) {
...
+
+  WITH_MTX_LOCK (&obj->mtx) {
+    while (!TAILQ_EMPTY(&obj->list)) {
+      vm_page_t *pg = TAILQ_FIRST(&obj->list);
+      TAILQ_REMOVE(&obj->list, pg, obj.list);
+      vm_page_free(pg);
+    }
   }
   pool_free(P_VMOBJ, obj);
 }
```
pj1031999, line 45: Are we sure that nothing will change the value of `ref_counter` between `refcnt_release` and `WITH_MTX_LOCK`?

pj1031999 (changes requested):

include/sys/vm_object.h
cahirwpz: Refer to `include/sys/proc.h` for examples.

j-piecuch (changes requested):
`vm_object_remove_page()` is a public function, but it requires callers to hold `obj->mtx`. This is unlike any other public function in this file. I suggest adding a static function `vm_object_remove_page_locked()` and using that inside `vm_object_remove_range()` and `vm_object_remove_page()` (which should acquire `obj->mtx` before calling it).

sys/kern/vm_object.c
bkjg: We aren't. I will change it

sys/kern/vm_object.c
bkjg: This assert was creating for checking correctness purposes. I will remove it.

pj1031999: Okay but if your condition is not met then you should return from that function in previous if statement.

sys/kern/vm_map.c
bkjg: vm_object_clone clones only the data of the vm_object. That means ref_counter will be equal to 1, so it's not possible to have ref_counter different than 1 here

include/sys/vm_object.h
bkjg: Done

bkjg:
@j-piecuch Thank you for your comments! You're absolutely right! I fixed it.

bin/utest/vm_map.c
```diff
@@ -0,0 +1,79 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+int test_sharing_memory_simple(void) {
+  char *map =
+    mmap(NULL, 4096, PROT_READ | PROT_WRITE, MAP_ANON | MAP_SHARED, 0, 0);
```
pj1031999, line 15: > MAP_ANON
> Map anonymous memory not associated with any specific file. The file descriptor is not used for creating MAP_ANON regions, and must be specified as -1. The mapped memory will be zero filled.

mmap(2) - NetBSD Manual Pages

Should we stick to it?

bin/utest/vm_map.c
```diff
@@ -0,0 +1,79 @@
...
+
+  assert(strcmp(map, "Hello, World!") == 0);
+
+  assert(munmap(map, 4096) == 0);
+  return 0;
+}
+
+int test_sharing_memory_child_and_grandchild(void) {
+  char *map =
+    mmap(NULL, 4096, PROT_READ | PROT_WRITE, MAP_ANON | MAP_SHARED, 0, 0);
```
pj1031999, line 41: Same as above.

bin/utest/vm_map.c
```diff
@@ -0,0 +1,79 @@
...
+    if (pid == 0) {
+      /* grandchild */
+      strcpy(map, "Hello from grandchild!");
+      exit(0);
+    } else {
+      /* child */
+      int status;
+      waitpid(-1, &status, 0);
+      assert(WIFEXITED(status));
+      assert(strcmp(map, "Hello from grandchild!") == 0);
```
pj1031999, line 64: Probably `WIFEXITED` doesn't check anything on `waitpid` failure.

sys/kern/sbrk.c
```diff
@@ -22,8 +22,8 @@ void sbrk_attach(proc_t *p) {
   /* Initially allocate one page for brk segment. */
   vaddr_t addr = SBRK_START;
   vm_object_t *obj = vm_object_alloc(VM_ANONYMOUS);
-  vm_segment_t *seg =
-    vm_segment_alloc(obj, addr, addr + PAGESIZE, VM_PROT_READ | VM_PROT_WRITE);
+  vm_segment_t *seg = vm_segment_alloc(obj, addr, addr + PAGESIZE,
+                                       VM_PROT_READ | VM_PROT_WRITE, 0);
```
pj1031999, line 26: `vm_flags_t` is a enum. `VM_FILE` is easier to read than `0`.

sys/kern/vm_map.c
```diff
@@ -328,8 +331,17 @@ vm_map_t *vm_map_clone(vm_map_t *map) {
   WITH_MTX_LOCK (&map->mtx) {
     vm_segment_t *it;
     TAILQ_FOREACH (it, &map->entries, link) {
-      vm_object_t *obj = vm_object_clone(it->object);
-      vm_segment_t *seg = vm_segment_alloc(obj, it->start, it->end, it->prot);
+      vm_object_t *obj;
+      vm_segment_t *seg;
+      if (it->flags & VM_SHARED) {
+        refcnt_acquire(&it->object->ref_counter);
+        obj = it->object;
```
pj1031999, line 338: Do you set `VM_SHARED` anywhere in code?

sys/kern/sbrk.c
bkjg: Okay, I will change it

bin/utest/vm_map.c
bkjg: Maybe you're right. I added assert to check if waitpid succeeded.

sys/kern/vm_map.c
bkjg: Previously mimiker didn't support sharing the memory. That's why doesn't exist code in kernel that set this particular flag. However now there is a possibility to have shared memory and you can accomplish that by mapping the memory with MAP_SHARED flag and then one of functions for creating vm_segment_t structure will set it. (you can see the example in tests for this functionality)

pj1031999: Ok. You call `mmap` with `MAP_SHARED` flag and then you copy it into flags in `sys_mmap` and it works because definitions from `sys/mman.h` fit into `vm_flags_t`.
Please add comments about that assumption.

bin/utest/vm_map.c
bkjg: I changed it

sys/kern/vm_map.c
bkjg: Okay, done

pj1031999: I though about adding comment into `vm.h`
> If you do something here then look into `mman.h`
And the same in `mman.h`.

cahirwpz: Please refer to `sys/kern/mmap.c`. We check correspondence between user and kernel constants there.

sys/kern/exec.c
```diff
@@ -284,7 +284,7 @@ static void enter_new_vmspace(proc_t *p, exec_vmspace_t *saved,
   vm_object_t *stack_obj = vm_object_alloc(VM_ANONYMOUS);
   vm_segment_t *stack_seg =
     vm_segment_alloc(stack_obj, USER_STACK_TOP - USER_STACK_SIZE,
-                     USER_STACK_TOP, VM_PROT_READ | VM_PROT_WRITE);
+                     USER_STACK_TOP, VM_PROT_READ | VM_PROT_WRITE, VM_FILE);
```
cahirwpz, line 287: `VM_FILE` here? We don't have file mappings! Why not `VM_ANON`? Similarly in many other places...

include/sys/vm_object.h
```diff
@@ -3,14 +3,23 @@
...
-  vm_pagetree_t tree;
-  size_t size;
-  size_t npages;
-  vm_pager_t *pager;
+  mtx_t mtx;
+  vm_pagelist_t list;   /* (@) List of pages */
+  vm_pagetree_t tree;   /* (@) Tree of pages */
+  size_t npages;        /* (@) Number of pages */
+  vm_pager_t *pager;    /* Pager type and page fault function for object */
+  refcnt_t ref_counter; /* (a) How many objects refer to this object? */
```
franciscozdo, line 22: @cahirwpz Should we consider adding prefix to every field in this structure? (like `vo_*` or `vmo_*`) We follow this rule in other structures, and it is easier to read code when we access fields with descriptive names. If we want to follow this rule here it might be better to change this until our structure gets much bigger.

franciscozdo (changes requested):

sys/kern/mmap.c
```diff
@@ -9,12 +9,28 @@
...
+static_assert(VM_PROT_READ == PROT_READ, "VM_PROT_READ != PROT_READ");
+static_assert(VM_PROT_WRITE == PROT_WRITE, "VM_PROT_WRITE != PROT_WRITE");
+static_assert(VM_PROT_EXEC == PROT_EXEC, "VM_PROT_EXEC != PROT_EXEC");
+
+static_assert(VM_FILE == MAP_FILE, "VM_FILE != MAP_FILE");
+static_assert(VM_ANON == MAP_ANON, "VM_ANON != MAP_ANON");
+static_assert(VM_SHARED == MAP_SHARED, "VM_SHARED != MAP_SHARED");
+static_assert(VM_PRIVATE == MAP_PRIVATE, "VM_PRIVATE != MAP_PRIVATE");
+static_assert(VM_FIXED == MAP_FIXED, "VM_FIXED != MAP_FIXED");
+static_assert(VM_STACK == MAP_STACK, "VM_STACK != MAP_STACK");
```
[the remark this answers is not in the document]
pj1031999: For me `static_assert` is fine here.

pj1031999 (commented):
Besides what Franek said, I'm interested in why the kernel has slowed down noticeably.
Could you explain that?

include/sys/vm_map.h
```diff
@@ -9,6 +9,11 @@
 typedef struct vm_map vm_map_t;
 typedef struct vm_segment vm_segment_t;
 
+typedef enum {
+  VM_SEG_PRIVATE = 0, /* private memory (default) */
+  VM_SEG_SHARED = 1,  /* shared memory */
+} vm_seg_flags_t;
+
```
j-piecuch, line 16: I don't think we need `VM_SEG_PRIVATE`, since we need only a single bit to convey the private/shared status of a segment: if `VM_SEG_SHARED` is set, the segment is shared, and if it's not set, it is private. Also, having an enum member with value 0 for a flag type is weird (how do we test for it)? I therefore suggest removing `VM_SEG_PRIVATE`.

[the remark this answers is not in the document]
j-piecuch: They are different types with different uses, e.g. it makes little sense to have `VM_FILE` or `VM_FIXED` as a possible flag for a segment (at least that's how I see it).

sys/kern/vm_map.c
```diff
@@ -233,13 +234,21 @@ int vm_map_insert(vm_map_t *map, vm_segment_t *seg, vm_flags_t flags) {
...
     return error;
   if ((flags & VM_FIXED) && (start != seg->start))
     return ENOMEM;
+
+  if (flags & VM_SHARED)
+    map_flags |= VM_SEG_SHARED;
+
   seg->start = start;
   seg->end = start + length;
+  seg->flags = map_flags;
```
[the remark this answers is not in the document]
j-piecuch: I agree... I feel like the caller should set the flags in the segment and this function should not touch them.
Maybe it's a good idea to add a `vm_seg_flags_t flags` argument to `vm_segment_alloc()`?

sys/kern/vm_map.c
```diff
@@ -328,8 +337,19 @@ vm_map_t *vm_map_clone(vm_map_t *map) {
...
+      if (it->flags & VM_SEG_SHARED) {
+        refcnt_acquire(&it->object->ref_counter);
+        obj = it->object;
+      } else {
+        /* vm_object_clone will clone the data from the vm_object_t
+         * and will return the new object with ref_counter equal to one */
+        obj = vm_object_clone(it->object);
+      }
+      seg = vm_segment_alloc(obj, it->start, it->end, it->prot);
+      seg->flags = it->flags;
```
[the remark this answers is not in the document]
j-piecuch: +1

bin/utest/vm_map.c
```diff
@@ -0,0 +1,73 @@
...
+  if (pid == 0) {
+    /* child */
+    strcpy(map, "Hello, World!");
+    exit(0);
+  }
+
+  /* parent */
+  int status;
+  assert(waitpid(-1, &status, 0) == pid);
+  assert(WIFEXITED(status));
```
j-piecuch, line 31: You can use the `wait_for_child_exit()` helper function from `util.h`.

bin/utest/vm_map.c
```diff
@@ -0,0 +1,73 @@
...
+    assert(WIFEXITED(status));
+    assert(strcmp(map, "Hello from grandchild!") == 0);
+    strcpy(map, "Hello from child!");
+    exit(0);
+  }
+
+  /* parent */
+  int status;
+  assert(waitpid(-1, &status, 0) == pid);
+  assert(WIFEXITED(status));
```
j-piecuch, line 69: ditto here

j-piecuch (changes requested):

j-piecuch:
> Besides what Franek said, I'm interested in why the kernel has slowed down noticeably.
> Could you explain that?

@pj1031999 I ran `./run_tests.py --thorough` on this branch and `master`, and here are my results:
`master`:
```
real	19m26.254s
user	16m19.070s
sys	1m2.625s
```
`bkjg:shared_memory`:
```
real	19m38.627s
user	16m30.054s
sys	1m0.461s
```
Given that the `shared_memory` branch has 2 more tests, I wouldn't say the kernel has noticeably slowed down.

pj1031999:
@j-piecuch for me it was `16m` vs `19m` on clang.

include/sys/vm_map.h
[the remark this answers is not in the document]
bkjg: Yes, that's how the @j-piecuch said: these flags are used for different purposes. Segment can represent shared memory, private or even copy-on-write segments but for example VM_FILE doesn't have any sense here.

bkjg: I assumed that maybe in the future we would like to check if a given segment is private or shared. And if you look at vm_flags_t enum, you could see the similar behavior there: there is a default VM_FILE flag, the value for this member is equal to zero. So similar question would be here: why wouldn't we remove this flag too? I think it's important to keep consistency in code and use the same rules.

bin/utest/vm_map.c
bkjg: I didn't know this function. Many thanks! I will change it! :smiley:

bin/utest/vm_map.c
bkjg: Done :smile:

sys/kern/vm_map.c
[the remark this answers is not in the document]
bkjg: You're both right. I added segment flags as the argument to vm_segment_alloc function.

sys/kern/vm_map.c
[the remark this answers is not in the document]
bkjg: Thanks! I did that!

include/sys/vm_map.h
j-piecuch: `vm_flags_t` is supposed to reflect the `MAP_*` flags, so `VM_FILE` is 0 because `MAP_FILE` is 0, and we can't change `MAP_FILE` because it's exposed to userspace and it's the default value. Besides, having a 0 valued flag enum member is dangerous when a user wants to check a flag: when I want to check if a segment is shared, do I check `(flags & VM_SEG_SHARED)` or `!(flags & VM_SEG_PRIVATE)`? Clearly the latter is wrong, but it's not obvious at all without reading the enum definition.

include/sys/vm_object.h
cahirwpz: @franciscozdo Yes, but in another PR.

@bkjg Please put it on your TODO list.

include/sys/vm_map.h
cahirwpz: Done.

PR #832, opened 2020-11-16
Copy-on-write inside fork syscall

bkjg:
Add copy-on-write to fork.


pj1031999:
@bkjg what is a status of that PR? Should I start reviewing the code?

bkjg:
> @bkjg what is a status of that PR? Should I start reviewing the code?

@pj1031999 To successful review of this PR we should wait until #831 is merged, because this PR depends on changes from that PR, so they are visible here too.

cahirwpz:
Please resolve conflicts after #831 merge.

j-piecuch (changes requested):

sys/kern/vm_physmem.c
```diff
@@ -250,7 +252,7 @@ void vm_page_free(vm_page_t *page) {
 
   SCOPED_MTX_LOCK(physmem_lock);
 
-  klog("%s: free %lx of size %ld", __func__, page->paddr, page->size);
+  klog("%s: free %lx of size %ld\n", __func__, page->paddr, page->size);
```
pj1031999, line 255: Why `\n`?

include/sys/vm.h
```diff
@@ -60,6 +61,7 @@ struct vm_page {
   paddr_t paddr;                  /* physical address of page */
   pg_flags_t flags;               /* page flags (used by physmem as well) */
   uint32_t size;                  /* size of page in PAGESIZE units */
+  refcnt_t ref_counter;           /* number of references to this page */
```
pj1031999, line 64: Maybe some prefix like `vmp_` will be a good idea here.

include/sys/vm_object.h
```diff
@@ -14,12 +16,13 @@
...
 
 typedef struct vm_object {
-  mtx_t mtx;
+  rwlock_t mtx;
   vm_pagelist_t list;   /* (@) List of pages */
   vm_pagetree_t tree;   /* (@) Tree of pages */
   size_t npages;        /* (@) Number of pages */
   vm_pager_t *pager;    /* Pager type and page fault function for object */
   refcnt_t ref_counter; /* (a) How many objects refer to this object? */
+  vm_object_t *shadow_object; /* (@) Pointer to backing object */
```
pj1031999, line 25: Maybe some prefix like `vmo_` will be a good idea here.

lib/libc/gen/getpagesize.c
```diff
@@ -1,4 +1,4 @@
-#include <sys/vm.h>
+#include <machine/vm_param.h>
```
pj1031999, line 1: `machine/vm_param.h` is already included in `sys/vm.h`.

sys/kern/vm_map.c
```diff
@@ -340,17 +340,35 @@ vm_map_t *vm_map_clone(vm_map_t *map) {
...
+          obj = vm_object_alloc(VM_SHADOW);
+          shadow = it->object->shadow_object;
+          obj->shadow_object = shadow;
+        } else {
+          shadow = it->object;
+          obj = vm_object_alloc(VM_SHADOW);
+          obj->shadow_object = shadow;
+          it->object = vm_object_alloc(VM_SHADOW);
+          it->object->shadow_object = shadow;
+          vm_object_set_readonly(shadow);
```
pj1031999, line 362: Could you explain what should be done here with pages marked as `VM_PROT_READ | VM_PROT_EXEC`?

sys/kern/vm_pager.c
```diff
@@ -17,7 +17,61 @@ static vm_page_t *anon_pager_fault(vm_object_t *obj, off_t offset) {
...
+        prev->shadow_object = it->shadow_object;
+        prev->pager = it->pager;
+        if (it->shadow_object != NULL) {
+          refcnt_acquire(&it->shadow_object->ref_counter);
+        }
+
+        vm_object_free(it);
+      }
+    }
+  }
```
pj1031999, line 66: It's hard to read that code.
At the end of that function you always call `vm_object_add_page` so you could write

```
  if (...) {
    ...
    goto out;
  }
  
  ...

out:
  vm_object_add_page(obj, offset, new_pg);
  return new_pg;
```
It is simpler for me.

I also don't understand logic here.
You call 2x `refcnt_acquire` and after that `refcnt_release` – it looks strange without explanation.

cahirwpz:
Please merge `master` into your branch and resolve conflicts.

include/sys/vm_object.h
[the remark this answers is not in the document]
bkjg: Great point! Thanks! Changed.

sys/aarch64/pmap.c
```diff
@@ -519,3 +519,10 @@ void pmap_delete(pmap_t *pmap) {
   free_asid(pmap->asid);
   pool_free(P_PMAP, pmap);
 }
+
+void pmap_set_page_readonly(vm_page_t *pg) {
```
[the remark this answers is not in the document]
bkjg: I didn't think about that before. Thanks! Fixed

sys/kern/vm_physmem.c
bkjg: My mistake. Sorry. I reversed it.

bin/utest/vm_map.c
```diff
@@ -66,3 +68,33 @@ int test_sharing_memory_child_and_grandchild(void) {
   assert(munmap(map, pgsz) == 0);
   return 0;
 }
+
+int test_cow_private_simple(void) {
```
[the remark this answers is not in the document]
bkjg: Sounds good. I changed it

include/sys/vm_object.h
```diff
@@ -14,12 +16,13 @@
  */
 
 typedef struct vm_object {
-  mtx_t mtx;
+  rwlock_t mtx;
```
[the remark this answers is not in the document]
bkjg: I thought that in case when we are using the cow, reading access will be more common than the writing access so predictably rwlocks would be faster than mutexes.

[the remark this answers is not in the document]
j-piecuch: It is true that with cow reads from `vm_objects` will become more frequent, but in order to benefit performance, read accesses must be _far_ more frequent than write accesses, and read accesses must often occur concurrently. As for the first requirement, that may be the case, although I haven't looked through the code. As for the second, I don't think concurrent read accesses are likely, since we don't support SMP, so concurrent access can occur only when a thread is preempted while holding the lock.
I'm also a little concerned about the fact that `rwlock` uses `sleepq` functions, so there may be starvation and priority inversion issues, since you sometimes acquire a write lock while holding a mutex, e.g. in `vm_page_fault()`.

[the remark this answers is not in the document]
bkjg: Okay, I see what you mean. I agree with you. That's right. I also am worried a little about the fact that we don't use rwlocks in mimiker frequently, so they aren't tested well, so we don't know if they work properly for sure. I will try to split this PR to several smaller ones, so rwlocks would be separate change that maybe will be skipped.

sys/kern/vm_map.c
```diff
@@ -340,17 +340,35 @@ vm_map_t *vm_map_clone(vm_map_t *map) {
...
+      vm_object_t *shadow;
 
       if (it->flags & VM_SEG_SHARED) {
         refcnt_acquire(&it->object->ref_counter);
         obj = it->object;
       } else {
-        /* vm_object_clone will clone the data from the vm_object_t
-         * and will return the new object with ref_counter equal to one */
-        obj = vm_object_clone(it->object);
+        if ((it->flags & VM_SEG_NEED_COPY) != 0 && it->object->npages == 0) {
```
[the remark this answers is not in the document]
bkjg: Fixed

sys/kern/vm_pager.c
bkjg: This functionality was moved to another PR.

include/sys/vm_object.h
bkjg: Do you mean adding to each field?

sys/kern/vm_map.c
bkjg: I think in case when sth is executable, then we also should set is as readonly and just disregard the fact that this is VM_PROT_EXEC, don't do anything special in that case

lib/libc/gen/getpagesize.c
bkjg: I remember I had some problems with this header, but I restored the old version

include/sys/vm_object.h
pj1031999: Yes.

sys/kern/vm_map.c
pj1031999: Maybe I don't understand something but if you mark page as RO you can't execute any instruction from that page so how it work's in that case? Do you kill new process in that case or make copy of .text?

sys/mips/pmap.c
```diff
@@ -347,9 +347,10 @@ void pmap_remove(pmap_t *pmap, vaddr_t start, vaddr_t end) {
       paddr_t pa;
       if (pmap_extract_nolock(pmap, va, &pa)) {
         vm_page_t *pg = vm_page_find(pa);
-        WITH_MTX_LOCK (pv_list_lock)
+        WITH_MTX_LOCK (pv_list_lock) {
           pv_remove(pmap, va, pg);
-        pmap_pte_write(pmap, va, empty_pte(pmap), 0);
+          pmap_pte_write(pmap, va, empty_pte(pmap), 0);
```
j-piecuch, line 352: Do we need to hold `pv_list_lock` to call `pmap_pte_write()`?

j-piecuch (changes requested):
There's a significant locking issue.

sys/kern/vm_map.c
```diff
@@ -340,17 +340,41 @@ vm_map_t *vm_map_clone(vm_map_t *map) {
   WITH_MTX_LOCK (&map->mtx) {
     vm_segment_t *it;
     TAILQ_FOREACH (it, &map->entries, link) {
+      vm_seg_flags_t flags = it->flags;
```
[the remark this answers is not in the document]
bkjg: oh you're right. I didn't notice it. I will fix it

include/sys/vm_map.h
```diff
@@ -10,8 +10,9 @@ typedef struct vm_map vm_map_t;
 typedef struct vm_segment vm_segment_t;
 
 typedef enum {
-  VM_SEG_SHARED = 1,  /* shared memory */
-  VM_SEG_PRIVATE = 2, /* private memory (default) */
+  VM_SEG_SHARED = 1,    /* shared memory */
+  VM_SEG_PRIVATE = 2,   /* private memory (default) */
+  VM_SEG_NEED_COPY = 4, /* copy on write segment, needs copying of pages */
```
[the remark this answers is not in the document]
bkjg: hmm I like the idea. I will change it

sys/kern/vm_object.c
```diff
@@ -83,13 +83,18 @@ void vm_object_remove_range(vm_object_t *object, off_t offset, size_t length) {
 }
 
 void vm_object_free(vm_object_t *obj) {
-  if (!refcnt_release(&obj->ref_counter))
-    return;
-
   WITH_MTX_LOCK (&obj->mtx) {
```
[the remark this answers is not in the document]
bkjg: Yes, you're right. I will change it

sys/mips/pmap.c
bkjg: Hmm no. I will fix it.

sys/aarch64/pmap.c
```diff
@@ -511,24 +511,34 @@ pmap_t *pmap_new(void) {
...
   pool_free(P_PMAP, pmap);
 }
+
+void pmap_set_page_prot(vm_page_t *pg, vm_prot_t prot) {
+  SCOPED_MTX_LOCK(pv_list_lock);
+  pv_entry_t *pv;
+  TAILQ_FOREACH (pv, &pg->pv_list, page_link) {
+    pmap_protect(pv->pmap, pv->va, pv->va + PAGESIZE, prot);
+  }
+}
```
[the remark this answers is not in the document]
bkjg: Okay, I will create a separate PR with this fix.

bin/utest/vm_map.c
```diff
@@ -15,6 +15,7 @@ int test_sharing_memory_simple(void) {
   size_t pgsz = getpagesize();
   char *map =
     mmap(NULL, pgsz, PROT_READ | PROT_WRITE, MAP_ANON | MAP_SHARED, -1, 0);
+
   assert(map != (char *)MAP_FAILED);
```
pj1031999, line 19: Please don't add empty lines in random pieces of code.

bin/utest/vm_map.c
```diff
@@ -66,3 +68,33 @@ int test_sharing_memory_child_and_grandchild(void) {
...
+  } else {
+    /* parent */
+    wait_for_child_exit(pid, 0);
+
+    assert(strcmp(map, "Hello, World!") == 0);
+  }
+
+  assert(munmap(map, pgsz) == 0);
+  return 0;
+}
```
pj1031999, line 100: That test doesn't work on AArch64 - you get infinity loop  in pmap. Please fix that or explain what is a source of that problem.

pj1031999 (changes requested):

bin/utest/vm_map.c
bkjg: I cannot reproduce this problem. Could you provide some example of that case? Or what can I do to observe this situation?

bin/utest/vm_map.c
bkjg: Fixed.

bin/utest/vm_map.c
pj1031999:
```
git clone https://github.com/bkjg/mimiker
cd mimiker
git checkout copy-on-write
make -j4 BOARD=rpi3
./launch --board rpi3 -d test=user_anon_private_simple
```

bkjg: That was because of the bug with permissions. It's fixed now.

include/sys/vm_object.h
```diff
@@ -19,6 +20,8 @@ typedef struct vm_object {
   size_t npages;        /* (@) Number of pages */
   vm_pager_t *pager;    /* Pager type and page fault function for object */
   refcnt_t ref_counter; /* (a) How many objects refer to this object? */
+  vm_object_t *backing_object; /* (@) Pointer to backing object */
+  refcnt_t shadow_counter;     /* (a) For how many objects this is a shadow? */
```
j-piecuch, line 24: If I understand the code correctly, `shadow_counter` is the number of shadow objects for which this object is the __backing__ object. In other words, it's the number of objects immediately (i.e. without other objects in between) shadowing this object.

sys/kern/vm_map.c
```diff
@@ -371,15 +372,42 @@ vm_map_t *vm_map_clone(vm_map_t *map) {
...
+          obj->backing_object = backing;
+          it->object = vm_object_alloc(VM_SHADOW);
+          it->object->backing_object = backing;
+          /* we set a backing object twice, so we have to acquire
+           * backing->shadow_counter twice too */
+          refcnt_acquire(&backing->shadow_counter);
+          /* all pages in backing object are now read-only,
+           * that refers also to pages which previously had VM_PROT_EXEC set
+           */
+          vm_object_set_prot(backing, VM_PROT_READ);
```
j-piecuch, line 402: I'm concerned about how this would interact with pages mapped without `VM_PROT_READ`... For instance pages mapped as execute-only (I don't know if it's a thing on `aarch64`), or guard pages with `VM_PROT_NONE`: when cloning such a segment, you would enable read access in the page table, and read access wouldn't cause a page fault.

j-piecuch (changes requested):

sys/aarch64/pmap.c
```diff
@@ -496,25 +496,26 @@ void pmap_set_modified(vm_page_t *pg) {
...
+    return EACCES;
 
-    if ((prot & VM_PROT_EXEC) && (pte & ATTR_SW_NOEXEC))
-      return EACCES;
-  }
-
-  vm_page_t *pg = vm_page_find(pa);
-  assert(pg != NULL);
+  if ((prot & VM_PROT_EXEC) && (pte & ATTR_SW_NOEXEC))
+    return EACCES;
```
pj1031999, line 519: I don't understand that change. Why did you move `vm_page_find` before access checking? I don't see any advantages.
Please don't set `pte` and `pg` at the beginning of that function - it's confusing. You always use only values returned by `pmap_lookup_pte` and `vm_page_find` so _default_ 0 is not good here.

pj1031999 (changes requested):

sys/kern/vm_object.c
```diff
@@ -98,24 +103,35 @@ void vm_object_free(vm_object_t *obj) {
...
+void vm_object_set_prot(vm_object_t *obj, vm_prot_t prot) {
+  assert(mtx_owned(&obj->mtx));
+
+  vm_page_t *pg;
+  TAILQ_FOREACH (pg, &obj->list, obj.list) { pmap_set_page_prot(pg, prot); }
+}
+
+bool vm_object_is_backing(vm_object_t *obj) {
+  if (obj == NULL)
+    return false;
```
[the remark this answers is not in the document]
bkjg: Done.

sys/mips/pmap.c
```diff
@@ -462,25 +463,29 @@ void pmap_set_modified(vm_page_t *pg) {
...
-    pte_t pte = pmap_pte_read(pmap, va);
+    pte = pmap_pte_read(pmap, va);
+  }
+  pg = vm_page_find(pa);
+  assert(pg != NULL);
 
-    if ((prot & VM_PROT_READ) && !(pte & PTE_SW_READ))
-      return EACCES;
+  if (vm_object_is_backing(pg->object))
+    return EFAULT;
```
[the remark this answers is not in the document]
bkjg: I forgot it in `aarch64`. I added it.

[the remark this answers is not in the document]
bkjg: The logic is that if I didn't check if the page belongs to backing object, this function returned this EACCES and in the result SIG SEGV was raised. So to fix it, I had to add checking if owner of this page is backing object and if so then we have to call vm_page_fault.

include/sys/vm_object.h
bkjg: Oh right. My bad. I fixed it.

sys/aarch64/pmap.c
bkjg: I moved `vm_page_find` before access checking because in copy-on-write was the case where we checked access and raised segfault after that. However because the page we checked access for, belonged to the backing object, we shouldn't return `EACCES` but` EFAULT` to make sure `vm_page_fault` will be executed. That's the reason for change.

pj1031999: But here you don't check if page belongs to backing object.

bkjg: I forgot to but I fixed it and now I check it.

sys/mips/pmap.c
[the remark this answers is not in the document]
j-piecuch: I think we need to return `EFAULT` only if the thread attempted to write to a page in a backing object. We could merge that logic into the access check:
```c
if ((prot & VM_PROT_WRITE) && !(pte & PTE_SW_WRITE))
  return (vm_object_is_backing(pg->object) ? EFAULT : EACCES);
```

sys/kern/vm_map.c
bkjg: Okay, I added checking if segment this object originally belonged to, has `VM_PROT_READ` set and if so, then I set all pages as readonly.

cahirwpz:

franciscozdo:
Copy-on-write will be handled in another way.

cahirwpz:
Hence we're going to handle CoW in a different way I'm closing this PR.

PR #833, opened 2020-11-16
TTY: Introduce concepts of controlling terminal and foreground process group

j-piecuch:
- Add concepts of controlling tty and foreground pgrp
- Implement ioctls for getting and setting foreground pgrp
- Implement background wait in `tty_{read,write}()`
- Add `SIGTTIN` and `SIGTTOU` signals

### DIFF
--- a/include/sys/proc.h
+++ b/include/sys/proc.h
@@ -15,6 +15,7 @@ typedef struct cred cred_t;
 typedef struct pgrp pgrp_t;
 typedef struct fdtab fdtab_t;
 typedef struct vnode vnode_t;
+typedef struct tty tty_t;
 typedef TAILQ_HEAD(, proc) proc_list_t;
 typedef TAILQ_HEAD(, pgrp) pgrp_list_t;
 typedef TAILQ_HEAD(, session) session_list_t;
@@ -40,6 +41,7 @@ typedef struct session {
   proc_t *s_leader;            /* (a) Session leader */
   int s_count;                 /* (a) Count of pgrps in session */
   sid_t s_sid;                 /* (!) PID of session leader */
+  tty_t *s_tty;                /* (a) Controlling terminal (if any) */
 } session_t;
 
 /*! \brief Structure allocated per process group.
@@ -101,6 +103,7 @@ struct proc {
   vm_map_t *p_uspace;             /* ($) process' user space map */
   fdtab_t *p_fdtable;             /* ($) file descriptors table */
   sigaction_t p_sigactions[NSIG]; /* (@) description of signal actions */
+  signo_t p_stopsig;              /* (@) signal that stopped the process */
   condvar_t p_waitcv;             /* (a) processes waiting for this one */
   int p_exitstatus;               /* (@) exit code to be returned to parent */
   volatile proc_flags_t p_flags;  /* (@) PF_* flags */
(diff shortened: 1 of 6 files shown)


sys/kern/signal.c
```diff
@@ -342,15 +347,27 @@ void sig_kill(proc_t *p, ksiginfo_t *ksi) {
     return;
   }
 
+  sigprop_t prop = sig_properties[sig];
+
+  /* Don't send TTY stop signals to members of an orphaned process group
+   * with a default handler.
+   * Theoretically, we should hold all_proc_mtx for reading pg_jobc,
+   * but the race here is harmless. */
+  if ((prop & SA_TTYSTOP) && (p->p_pgrp->pg_jobc == 0) && (handler == SIG_DFL))
```
franciscozdo, line 356: Maybe we can wrap this check and comment into simple function, because it repeats also in line 458.

franciscozdo, line 356: I don't understand why we can do this without holding `all_proc_mtx`. Could you explain more?

franciscozdo (changes requested):

include/sys/proc.h
```diff
@@ -101,6 +103,7 @@ struct proc {
   vm_map_t *p_uspace;             /* ($) process' user space map */
   fdtab_t *p_fdtable;             /* ($) file descriptors table */
   sigaction_t p_sigactions[NSIG]; /* (@) description of signal actions */
+  signo_t p_stopsig;              /* (@) signal that stopped the process */
```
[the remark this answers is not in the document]
j-piecuch: Right, I forgot to use it in `do_waitpid()`. It's fixed now.

sys/kern/signal.c
j-piecuch: Actually, when writing my response I realized that this race can result in unwanted behavior: a process in an orphaned process group stopping on a terminal stop signal. However, the second race (the one in `sig_post()`) is actually harmless, here's why:

In order for a process in an orphaned pgrp to stop on a terminal stop signal, it must observe in `sig_post()` that its `jobc` is greater than 0. Suppose that immediately after reading `jobc`, the process group becomes orphaned. In `pgrp_maybe_orphan()`, we iterate over all processes in the orphaned group and send `SIGHUP` and `SIGCONT` to all stopped processes. The process state check is done while holding the process's `p_lock`, which `sig_post()` also holds. So, `pgrp_maybe_orphan()` won't examine the state of the process until `sig_post()` releases the lock, which it does __after__ stopping the process. So, when `pgrp_maybe_orphan()` acquires the lock, it will see that the process is stopped, and it will send `SIGCONT` to it, which is what we want to happen.

include/sys/tty.h
```diff
@@ -122,4 +128,12 @@ void tty_input(tty_t *tty, uint8_t c);
  */
 void tty_getc_done(tty_t *tty);
 
+/*
+ * Returns whether `tty` is the controlling terminal of process `p`.
+ * Must be called with `tty->t_lock` and `p->p_lock` held.
+ */
+static inline bool tty_is_ctty(tty_t *tty, proc_t *p) {
+  return (tty->t_session == p->p_pgrp->pg_session);
```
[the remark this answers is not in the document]
j-piecuch: I can, but usually these small inline functions like `proc_is_alive()` don't have locking assertions... meh, I guess it won't hurt.

sys/kern/signal.c
j-piecuch: Done.

franciscozdo (approved):

PR #834, opened 2020-11-18
Fix assert in pmap_enter.

pj1031999:
Use virtual address instead of physical address.


cahirwpz:
Could you fix `pmap_enter` in _MIPS_ as well ?

pj1031999:
> Could you fix `pmap_enter` in _MIPS_ as well ?

Done.

PR #835, opened 2020-11-18
Add support for ELF on AArch64.

pj1031999:
Add support for ELF on AArch64.

### DIFF
--- a/sys/kern/exec_elf.c
+++ b/sys/kern/exec_elf.c
@@ -9,6 +9,7 @@
 #include <sys/errno.h>
 #include <sys/vnode.h>
 #include <sys/proc.h>
+#include <machine/elf_arch.h>
 
 int exec_elf_inspect(vnode_t *vn, Elf_Ehdr *eh) {
   int error;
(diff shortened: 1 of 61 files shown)


sys/kern/exec_elf.c
```diff
@@ -132,7 +132,6 @@ static int load_elf_segment(proc_t *p, vnode_t *vn, Elf_Phdr *ph) {
 int exec_elf_load(proc_t *p, vnode_t *vn, Elf_Ehdr *eh) {
   int error;
 
-  assert(eh->e_phoff < 64);
```
cahirwpz, line 135: Maybe `sizeof(Elf_Phdr)`?

pj1031999: I don't think so.
In our case for `ksh` – `e_phoff` is `0x40` and `sizeof(Elf_Phdr)` is `0x38`.

cahirwpz: Anyway I'd expect this assert to come back with right value for both MIPS and AArch64.

pj1031999: I still can't find any reason for that assert.
First of all – userland should not have a chance for make kernel panic.
I have found that this assert was introduced by @rafalcieslak in #259 and if I understand correctly main reason for that assert is stack limit.

cahirwpz: Maybe we should pick arbitrary limits for these values and fail with `ENOMEM` if they're exceeded ? Could you quickly check how BSDs behave wrt. program headers size ?

pj1031999: Maybe we should remove all asserts and add `M_NOWAIT` for kmalloc here and then return `ENOMEM` on case of failure?
It looks like BSDs don't care about that. They just try to allocate memory and then try to read data from `e_phoff` offset and if case of failure propagate error to caller.

PR #837, opened 2020-11-21
Restore link register in copy.S.

pj1031999:
Restore link register in `copyin`, `copyout`, `copyinstr`, `copyerr`.


cahirwpz (approved):
LGTM

PR #840, opened 2020-11-21
Starting procedure for user ELF images.

pj1031999:
Starting procedure for user ELF images.

### DIFF
--- a/lib/csu/aarch64/crt0.S
+++ b/lib/csu/aarch64/crt0.S
@@ -8,6 +8,17 @@
 # to prepare_program_stack in exec.c.
 
 ENTRY(_start)
+        # Grab argc from stack.
+        ldr     w0, [sp, #0]
+
+        # Prepare argv.
+        add     x1, sp, #8
+
+        # Prepare envp.
+        add     x2, sp, #16
+
+        # Jump to start in crt0-common.
+        # void ___start(int argc, char **argv, char **envp)
         b       ___start
         nop
 END(_start)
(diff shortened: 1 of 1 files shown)


lib/csu/aarch64/crt0.S
```diff
@@ -8,8 +8,18 @@
...
 
 ENTRY(_start)
+        # Grab argc from stack.
+        ldr     w0, [sp, #0]
+
+        # Prepare argv.
+        add     x1, sp, #8
+
+        # Prepare envp.
+        add     x2, sp, #16
```
cahirwpz, line 18: `x2` points to `argv[1]`

cahirwpz (changes requested):
`envp` is not calculated correctly.

lib/csu/aarch64/crt0.S
pj1031999: I'm confused. In `exec_args_copyout` it is said that layout is sp, argc, argv, envp, argv[0..n], envp[0..m] so first value above argv – which is x2 – should be envp. What did I miss?

cahirwpz: Hmm... could you check if _MIPS_ implementation is correct then?

pj1031999: It looks like _MIPS_ implementation works. It makes sense for me because we still calculate correct address according to `exec_args_copyout` comments but in different way.

cahirwpz (approved):
LGTM

PR #844, opened 2020-11-23
Interrupt threads implementation

ThomasStachovsky:
Interrupt threads implementation.

### DIFF
--- a/include/sys/interrupt.h
+++ b/include/sys/interrupt.h
@@ -96,7 +96,7 @@ typedef struct intr_event {
   const char *ie_name;     /* individual event name */
   unsigned ie_irq;         /* physical interrupt request line number */
   unsigned ie_count;       /* number of handlers attached */
-  intr_thread_t ie_ithread; /* Associated interrupt thread */
+  intr_thread_t *ie_ithread; /* Associated interrupt thread */
 } intr_event_t;
 
 typedef TAILQ_HEAD(, intr_event) ie_list_t;
--- a/include/sys/interrupt.h
+++ b/include/sys/interrupt.h
@@ -67,7 +67,6 @@ typedef void ie_action_t(intr_event_t *);
 
 struct intr_handler {
   TAILQ_ENTRY(intr_handler) ih_link;
-  TAILQ_ENTRY(intr_handler) ih_ithread_link;
   ih_filter_t *ih_filter;   /* interrupt filter routine (run in irq ctx) */
   ih_service_t *ih_service; /* interrupt service routine (run in thread ctx) */
   intr_event_t *ih_event;   /* event we are connected to */
--- a/include/sys/interrupt.h
+++ b/include/sys/interrupt.h
@@ -75,8 +75,7 @@ typedef struct intr_event {
   void *ie_source;           /* additional argument for actions */
   const char *ie_name;       /* individual event name */
   unsigned ie_irq;           /* physical interrupt request line number */
-  unsigned ie_count;         /* number of handlers attached */
-  intr_thread_t *ie_ithread; /* Associated interrupt thread */
+  intr_thread_t *ie_ithread; /* associated interrupt thread */
 } intr_event_t;
 
 intr_event_t *intr_event_create(void *source, int irq, ie_action_t *disable,
--- a/include/sys/interrupt.h
+++ b/include/sys/interrupt.h
@@ -47,7 +47,6 @@ static inline void __intr_enable(void *data) {
 
 typedef struct intr_event intr_event_t;
 typedef struct intr_handler intr_handler_t;
-typedef struct intr_thread intr_thread_t;
 
 typedef enum {
   IF_STRAY = 0,    /* this device did not trigger the interrupt */
@@ -70,12 +69,12 @@ typedef struct intr_event {
   spin_t ie_lock;
   TAILQ_ENTRY(intr_event) ie_link; /* link on list of all interrupt events */
   TAILQ_HEAD(, intr_handler) ie_handlers; /* sorted by descending ih_prio */
-  ie_action_t *ie_disable;   /* called before ithread delegation (mask irq) */
-  ie_action_t *ie_enable;    /* called after ithread delagation (unmask irq) */
-  void *ie_source;           /* additional argument for actions */
-  const char *ie_name;       /* individual event name */
-  unsigned ie_irq;           /* physical interrupt request line number */
-  intr_thread_t *ie_ithread; /* associated interrupt thread */
+  ie_action_t *ie_disable; /* called before ithread delegation (mask irq) */
+  ie_action_t *ie_enable;  /* called after ithread delagation (unmask irq) */
+  void *ie_source;         /* additional argument for actions */
+  const char *ie_name;     /* individual event name */
+  unsigned ie_irq;         /* physical interrupt request line number */
+  thread_t *ie_ithread;    /* associated interrupt thread */
 } intr_event_t;
 
 intr_event_t *intr_event_create(void *source, int irq, ie_action_t *disable,
--- a/include/sys/interrupt.h
+++ b/include/sys/interrupt.h
@@ -9,9 +9,6 @@
 typedef struct ctx ctx_t;
 typedef struct device device_t;
 
-/*! \brief Called during kernel initialization. */
-void init_ithreads(void);
-
 /*! \brief Disables hardware interrupts.
  *
  * Calls to \fn intr_disable can nest, you must use the same number of calls to
@@ -50,6 +47,7 @@ static inline void __intr_enable(void *data) {
 
 typedef struct intr_event intr_event_t;
 typedef struct intr_handler intr_handler_t;
+typedef struct intr_thread intr_thread_t;
 
 typedef enum {
   IF_STRAY = 0,    /* this device did not trigger the interrupt */
@@ -69,6 +67,7 @@ typedef void ie_action_t(intr_event_t *);
 
 struct intr_handler {
   TAILQ_ENTRY(intr_handler) ih_link;
+  TAILQ_ENTRY(intr_handler) ih_ithread_link;
   ih_filter_t *ih_filter;   /* interrupt filter routine (run in irq ctx) */
   ih_service_t *ih_service; /* interrupt service routine (run in thread ctx) */
   intr_event_t *ih_event;   /* event we are connected to */
@@ -79,6 +78,12 @@ struct intr_handler {
 };
 
 typedef TAILQ_HEAD(, intr_handler) ih_list_t;
+
+typedef struct intr_thread {
+  intr_event_t *it_event; /* Associated event */
+  thread_t *it_thread;    /* Kernel thread. */
+  ih_list_t it_delegated;
+} intr_thread_t;
 
 /* Software representation of interrupt line. */
 typedef struct intr_event {
@@ -91,6 +96,7 @@ typedef struct intr_event {
   const char *ie_name;     /* individual event name */
   unsigned ie_irq;         /* physical interrupt request line number */
   unsigned ie_count;       /* number of handlers attached */
+  intr_thread_t ie_ithread; /* Associated interrupt thread */
 } intr_event_t;
 
 typedef TAILQ_HEAD(, intr_event) ie_list_t;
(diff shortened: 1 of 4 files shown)


pj1031999:
@ThomasStachovsky is it a final version of PR? You requested review but we still have a WIP label here.

sys/kern/interrupt.c
```diff
@@ -127,8 +134,21 @@ void intr_root_handler(ctx_t *ctx) {
     on_user_exc_leave();
 }
 
-/* interrupt handlers delegated to be called in the interrupt thread */
-static ih_list_t delegated = TAILQ_HEAD_INITIALIZER(delegated);
+void intr_thread_create(intr_event_t *ie) {
+  intr_thread_t *it = kmalloc(M_INTR, sizeof(intr_thread_t), M_WAITOK | M_ZERO);
+  ie->ie_ithread = it;
+
+  assert(ie->ie_ithread != NULL);
```
pj1031999, line 141: Maybe calling `kmalloc` without `M_WAITOK` will be better here?

sys/kern/interrupt.c
```diff
@@ -157,30 +180,44 @@ void intr_event_run_handlers(intr_event_t *ie) {
 }
 
 static void intr_thread(void *arg) {
+  intr_thread_t *it = (intr_thread_t *)arg;
+
   while (true) {
-    intr_handler_t *ih;
+    intr_handler_t *ih = NULL, *next, *curr;
```
pj1031999, line 186: I think that mixing declarations with assignment and without assignment is not a good idea.

sys/kern/interrupt.c
```diff
@@ -157,30 +180,44 @@ void intr_event_run_handlers(intr_event_t *ie) {
...
+        ie->ie_count--;
+      } else {
+        ih->ih_flags &= ~IH_DELEGATE;
+        ie_insert_handler(ie, ih);
+        if (ie->ie_enable)
+          ie->ie_enable(ie);
+      }
+    }
+    if (ih->ih_flags & IH_REMOVE) {
+      kfree(M_INTR, ih);
```
pj1031999, line 220: Comments would be useful.

mohrcore (approved):
LGTM

pj1031999 (approved):
LGTM

sys/kern/interrupt.c
```diff
@@ -157,30 +183,61 @@ void intr_event_run_handlers(intr_event_t *ie) {
...
+        if (curr->ih_flags & IH_DELEGATE) {
+          ih = curr;
+          break;
+        }
+      }
+      if (ih == NULL)
+        sleepq_wait(&it->it_wchan, NULL);
     }
 
     ih->ih_service(ih->ih_argument);
```
MichalBlk, line 208: Wait a minute. Do I understand it correctly? I think that's what you do:
1. You look for a handler within an interrupt event which needs to be serviced.
2. Now, let's say you won't find any (i.e. `ih == NULL`).
3. You go to sleep on the wait channel.
4. You've reached line 208, and... `ih` is stlii NULL.

If I am right, what you need to do is to put in to a `do {} while (it == NULL);` loop.

sys/kern/interrupt.c
```diff
@@ -157,30 +183,61 @@ void intr_event_run_handlers(intr_event_t *ie) {
...
     WITH_SPIN_LOCK (&ie->ie_lock) {
-      ie_insert_handler(ie, ih);
-      if (ie->ie_enable)
-        ie->ie_enable(ie);
+
+      /* If the handler is flagged as IH_REMOVE, we remove it similarly to how
+       * we do it in intr_event_remove_handler. We are under a spin lock so we
+       * don't call kfree just yet.
+       */
+      if (ih->ih_flags & IH_REMOVE) {
```
MichalBlk, line 218: This part is common with `intr_event_remove_handler`. Maybe some succinct static function?

sys/kern/interrupt.c
```diff
@@ -157,30 +183,61 @@ void intr_event_run_handlers(intr_event_t *ie) {
...
+
+        if (ie->ie_count == 1 && ie->ie_disable)
+          ie->ie_disable(ie);
+
+        TAILQ_REMOVE(&ie->ie_handlers, ih, ih_link);
+        ih->ih_event = NULL;
+        ie->ie_count--;
+      }
+      /* If the handler is not flagged as IH_REMOVE, we
```
MichalBlk, line 227: It would be more convenient to read if the comment were inside the else clause.

include/sys/interrupt.h
```diff
@@ -75,11 +86,19 @@ struct intr_handler {
...
+  uint8_t ih_flags; /* flags (currently if to be removed or delegated to a
+                       private thread) */
 };
 
 typedef TAILQ_HEAD(, intr_handler) ih_list_t;
 
+typedef struct intr_thread {
+  intr_event_t *it_event; /* Associated event */
+  thread_t *it_thread;    /* Kernel thread. */
+  ih_list_t it_wchan;
```
MichalBlk, line 99: Why do we need this field? This is used only as a sleep channel. We can safely use the `intr_thread` structure itself as a sleep channel for an interrupt thread. This is very confusing.

sys/kern/interrupt.c
```diff
@@ -146,9 +169,12 @@ void intr_event_run_handlers(intr_event_t *ie) {
...
 
-      TAILQ_REMOVE(&ie->ie_handlers, ih, ih_link);
-      TAILQ_INSERT_TAIL(&delegated, ih, ih_link);
-      sleepq_signal(&delegated);
+      if (ie->ie_ithread == NULL)
+        intr_thread_create(ie);
+
+      ih->ih_flags = IH_DELEGATE;
+      sleepq_signal(&ie->ie_ithread->it_wchan);
+
```
MichalBlk, line 177: Note that we can have more handlers within an interrupt event which require attention when `intr_event_run_handlers` is called. The problem is that this procedure isn't adapted to such a case. Please see #794 .

MichalBlk (changes requested):

sys/kern/interrupt.c
```diff
@@ -127,8 +134,24 @@ void intr_root_handler(ctx_t *ctx) {
     on_user_exc_leave();
 }
 
-/* interrupt handlers delegated to be called in the interrupt thread */
-static ih_list_t delegated = TAILQ_HEAD_INITIALIZER(delegated);
+void intr_thread_create(intr_event_t *ie) {
+  intr_thread_t *it = kmalloc(M_INTR, sizeof(intr_thread_t), M_ZERO);
+
+  assert(it && "We couldn't allocate memory for an ithread");
```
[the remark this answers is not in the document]
cahirwpz: Not quite, in fact we prefer bare `assert(it)`. You can leave a comment here what the codes assumes.

sys/kern/interrupt.c
```diff
@@ -127,60 +141,103 @@ void intr_root_handler(ctx_t *ctx) {
...
 }
 
-/* interrupt handlers delegated to be called in the interrupt thread */
-static ih_list_t delegated = TAILQ_HEAD_INITIALIZER(delegated);
+static void intr_thread_create(intr_event_t *ie) {
+  intr_thread_t *it = kmalloc(M_INTR, sizeof(intr_thread_t), M_ZERO);
+
+  assert(it);
+
+  ie->ie_ithread = it;
```
MichalBlk, line 149: You're missing `it->it_event = ie`. For now every interrupt thread has associated interrupt event set to NULL.

sys/kern/interrupt.c
```diff
@@ -127,60 +141,103 @@ void intr_root_handler(ctx_t *ctx) {
...
   }
 
+  if (it_wakeup) {
+    if (ie->ie_ithread == NULL)
+      intr_thread_create(ie);
+
+    sleepq_signal(ie->ie_ithread);
+  }
+
   klog("Spurious %s interrupt!", ie->ie_name);
```
MichalBlk, line 190: You will always reach this line, hence every correct interrupt request will yield the "Spurious interrupt!" message. To prevent this, you can add a `bool stray = true;` value and clear it if a handler is filtered or delegated. Then at this point do `if (stray) klog(...)`.

sys/kern/interrupt.c
```diff
@@ -127,60 +141,103 @@ void intr_root_handler(ctx_t *ctx) {
...
+     * go to sleep.
+     */
+    do {
+      WITH_INTR_DISABLED {
+        TAILQ_FOREACH (ih, &ie->ie_handlers, ih_link) {
+          if (ih->ih_flags & IH_DELEGATE)
+            break;
+        }
+        if (ih == NULL)
+          sleepq_wait(it, NULL);
```
MichalBlk, line 212: Are you going to sleep with disabled interrupts?

MichalBlk (changes requested):
Please test the code.

MichalBlk (approved):

MichalBlk:
LGTM

sys/kern/interrupt.c
```diff
@@ -127,60 +141,80 @@ void intr_root_handler(ctx_t *ctx) {
...
+    if (ie->ie_disable)
+      ie->ie_disable(ie);
 
-      TAILQ_REMOVE(&ie->ie_handlers, ih, ih_link);
-      TAILQ_INSERT_TAIL(&delegated, ih, ih_link);
-      sleepq_signal(&delegated);
-      return;
-    }
+    if (ie->ie_ithread == NULL)
+      intr_thread_create(ie);
```
jpszczolowski, line 179: `intr_thread_create` calls `kmalloc`, which we **must not** do in the bottom-half. I'd suggest calling `kmalloc` with `M_NOWAIT` flag, but it looks like it'd try to acquire a mutex anyway, so it isn't allowed either.

What about:
* creating the ithread (or a pool of them) somewhere else, earlier, in the top-half?
* or if you want to create the ithread here, consider using some pre-allocated static memory (but please note that `thread_create` also internally uses mutexes)

sys/kern/interrupt.c
```diff
@@ -127,60 +141,80 @@ void intr_root_handler(ctx_t *ctx) {
...
 
-    ih->ih_service(ih->ih_argument);
-
-    intr_event_t *ie = ih->ih_event;
-
-    WITH_SPIN_LOCK (&ie->ie_lock) {
-      ie_insert_handler(ie, ih);
-      if (ie->ie_enable)
+    /* If there still handlers assign to the interrupt event, enable interrupts
+     * and wait for wakeup. We do it with interrupts disable to prevent a wakeup
```
jpszczolowski, line 211: Probably:
* `If there still handlers assign` -> `If there are still handlers assigned`
* `with interrupts disable` -> `with interrupts disabled`

jpszczolowski (changes requested):

sys/kern/interrupt.c
cahirwpz: Done.

sys/kern/interrupt.c
cahirwpz: Oops... thank you for spotting that. Indeed this is wrong place to create a thread. I'll move it to `ie_add_handler` most likely.

sys/kern/interrupt.c
```diff
@@ -127,60 +141,80 @@ void intr_root_handler(ctx_t *ctx) {
...
 
-      TAILQ_REMOVE(&ie->ie_handlers, ih, ih_link);
-      TAILQ_INSERT_TAIL(&delegated, ih, ih_link);
-      sleepq_signal(&delegated);
-      return;
-    }
+    if (ie->ie_ithread == NULL)
+      intr_thread_create(ie);
+
+    sleepq_signal(ie->ie_ithread);
```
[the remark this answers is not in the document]
cahirwpz: Not needed anymore. Interrupt thread is create elsewhere.

sys/kern/interrupt.c
```diff
@@ -82,21 +85,32 @@ intr_handler_t *intr_event_add_handler(intr_event_t *ie, ih_filter_t *filter,
   ih->ih_argument = arg;
   ih->ih_name = name;
   ih->ih_prio = 0;
+  ih->ih_flags = 0;
   ie_add_handler(ie, ih);
   return ih;
 }
 
+/* XXX: This should always be called in a spin lock */
+static void intr_event_delete_handler(intr_handler_t *ih) {
```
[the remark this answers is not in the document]
cahirwpz: Done.

PR #845, opened 2020-11-23
Improve qemu-mimiker cleaning

komik0:
To avoid make error caused by quilt - in case when we want to import
new patches, we need to clean also _.pc_

We can be more gentle and use `quilt` here (to clean his files) but this approach is also correct
and covers our needs

### DIFF
--- a/toolchain/qemu-mimiker/Makefile
+++ b/toolchain/qemu-mimiker/Makefile
@@ -69,7 +69,7 @@ binary-stamp: install-stamp
 
 clean: unpatch
 	dh_clean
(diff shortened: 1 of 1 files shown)


komik0:
my mistake

komik0:
because it was my mistake. we can consider it as a small improvement.

cahirwpz:
If we can use `quilt` for that instead of forcefully remove its directory, then please do so.

komik0:
not needed

cahirwpz:
If the PR is not needed as you stated in your previous comment then I'll close it.

PR #847, opened 2020-11-25
Chown

franciscozdo:
I have implemented chown syscall and added privilege checks in chmod. 

Process can chmod if:
* is super-user
* is owner of file

_I didn't involved checks while setting S_ISVTX bit on files, because we have no interpretation of sticky-bit on files_

Process can chown when:
* is super-user
* is owner and NOT setting new owner and new group is in his supplementary groups

When changing owner or group of file bits S_ISUID or S_ISGID are cleared.

I set sticky-bit on tmpfs dir (as it is in all systems) for future use.


sys/kern/tmpfs.c
```diff
@@ -284,16 +285,44 @@ static int tmpfs_vop_getattr(vnode_t *v, vattr_t *va) {
...
   tmpfs_mount_t *tfm = TMPFS_ROOT_OF(v->v_mount);
   tmpfs_node_t *node = TMPFS_NODE_OF(v);
+  int error = 0;
 
   if (va->va_size != (size_t)VNOVAL)
     tmpfs_reg_resize(tfm, node, va->va_size);
   if (va->va_mode != (mode_t)VNOVAL)
-    node->tfn_mode = (node->tfn_mode & ~ALLPERMS) | (va->va_mode & ALLPERMS);
+    error = tmpfs_chmod(node, va->va_mode, cred);
+  if (error == 0 &&
```
cahirwpz, line 321: Lines 321-325 are rather complex. Please simplify this piece of code.

sys/kern/tmpfs.c
```diff
@@ -743,7 +772,7 @@ static int tmpfs_mount(mount_t *mp) {
   /* Allocate the root node. */
   vattr_t va;
   vattr_null(&va);
-  va.va_mode = S_IFDIR | (ACCESSPERMS & ~CMASK);
+  va.va_mode = S_IFDIR | (ACCESSPERMS & ~CMASK) | S_ISVTX;
```
cahirwpz, line 775: Why sticky bit here?

include/sys/cred.h
```diff
@@ -53,6 +54,13 @@ int cred_cansignal(proc_t *p, cred_t *cred);
  * with lock held. */
 int proc_cansignal(proc_t *target, signo_t sig);
 
+/* Checks if given gid is in supplementary groups of given credentials set */
+bool cred_groupmember(gid_t gid, cred_t *cred);
+
+/* VFS checks */
+bool cred_can_chmod(uid_t f_owner, gid_t f_group, cred_t *cred, mode_t mode);
```
cahirwpz, line 61: Maybe it'll save you some code if you returned `errno` (like `EPERM`) instead of boolean? Please reevaluate if using standard unix error reporting trick could be better in this case.

cahirwpz (approved):
LGTM but please address my comments before merging.

include/sys/cred.h
franciscozdo: I don't think so. I use it only in `tmpfs.c:297` in if condition. There is no need to return errorcode because it is only error code possible here.

sys/kern/tmpfs.c
franciscozdo: Removed. I added it because in every UNIX like system tmp dir is marked with sticky bit. However, there is no meaning for sticky bit at all in mimiker, so I will add it in future changes.

PR #848, opened 2020-11-27
RTL8139: Alloc registers resource and perform software reset

komik0:
With this commit the following changes are introduced:

- resource for rtl8139 regs is added
- software reset is performed


sys/drv/rtl8139.c
```diff
@@ -21,7 +22,31 @@ static int rtl8139_probe(device_t *dev) {
...
 
+static void rtk_reset(rtl8139_state_t *state) {
+  int i;
+
+  bus_write_1(state->regs, RL_8139_CFG1, RL_CFG1_PWRON);
+  bus_write_1(state->regs, RL_COMMAND, RL_CMD_RESET);
+  for (i = 0; i < RL_TIMEOUT; i++) {
+    if (!(bus_read_1(state->regs, RL_COMMAND) & RL_CMD_RESET))
+      break;
+  }
```
pj1031999, line 33: You can write
`for (int i = 0; i < RL_TIMEOUT; i++)`

sys/drv/rtl8139.c
```diff
@@ -21,7 +22,31 @@ static int rtl8139_probe(device_t *dev) {
...
 static int rtl8139_attach(device_t *dev) {
+  assert(dev->parent->bus == DEV_BUS_PCI);
+  rtl8139_state_t *state = dev->state;
+
+  state->regs = bus_alloc_resource_any(dev, RT_MEMORY, 1, RF_ACTIVE);
+
+  if (state->regs == NULL)
+    return -1;
+
+  rtk_reset(state);
```
pj1031999, line 48: I don't know how that device works but maybe we should return some error here in case of timeout?

sys/mips/gt64120.c
```diff
@@ -346,6 +346,8 @@ static resource_t *gt_pci_alloc_resource(device_t *dev, res_type_t type,
     /* The size of allocated resource must not be larger than what was reported
      * by corresponding BAR. */
     size = min(size, bar->size);
+    /* XXX: Perhaps, the rman_alloc_resource should take this into account */
+    size = roundup(size, PAGESIZE);
```
pj1031999, line 350: Please consult that with @mohrcore, @ThomasStachovsky, @MichalBlk or @cahirwpz.

MichalBlk: This PR is blocked until merging #826 .

MichalBlk:
Please adjust to #826

sys/kern/kmem.c
```diff
@@ -107,7 +107,7 @@ void kmem_free(void *ptr, size_t size) {
 }
 
 vaddr_t kmem_map(paddr_t pa, size_t size, unsigned flags) {
-  assert(page_aligned_p(pa) && page_aligned_p(size));
+  assert(page_aligned_p(pa));
```
pj1031999, line 110: Why you removed `page_aligned_p` from `assert`?

sys/drv/rtl8139.c
```diff
@@ -7,6 +7,7 @@
 #include <sys/libkern.h>
 #include <sys/devfs.h>
 #include <sys/devclass.h>
+#include <dev/rtl8139_reg.h>
```
MichalBlk, line 10: The `rtl8139_state` structure absolutely can't contain an `intr_handler_t` structure. Please note that interrupt handlers are only created inside the interrupt module and are allocated form the `M_INTR` kmalloc pool. Please remove this field. If you ever need an irq, you should use the `resource_t` structure.

MichalBlk (changes requested):

sys/drv/rtl8139.c
```diff
@@ -21,7 +22,29 @@ static int rtl8139_probe(device_t *dev) {
   return pci_device_match(pcid, RTL8139_VENDOR_ID, RTL8139_DEVICE_ID);
 }
 
+static int rtl_reset(rtl8139_state_t *state) {
+  bus_write_1(state->regs, RL_8139_CFG1, RL_CFG1_PWRON);
+  bus_write_1(state->regs, RL_COMMAND, RL_CMD_RESET);
+  for (int i = 0; i < RL_TIMEOUT; i++) {
```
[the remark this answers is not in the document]
MichalBlk: Ok, for now there is no good way to wait without doing busy waiting.

sys/drv/rtl8139.c
```diff
@@ -21,7 +22,29 @@ static int rtl8139_probe(device_t *dev) {
...
+  }
+
+  return 1;
+}
+
 static int rtl8139_attach(device_t *dev) {
+  assert(dev->parent->bus == DEV_BUS_PCI);
+  rtl8139_state_t *state = dev->state;
+
+  state->regs = bus_alloc_resource_any(dev, RT_MEMORY, 1, RF_ACTIVE);
```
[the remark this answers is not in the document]
komik0: ok, it is just outdated

komik0 (commented):
cs is ready for review

sys/drv/rtl8139.c
```diff
@@ -21,7 +22,29 @@ static int rtl8139_probe(device_t *dev) {
...
+
+  if (state->regs == NULL)
+    return -1;
+
+  if (rtl_reset(state))
```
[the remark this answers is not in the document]
MichalBlk: It's interesting. I presume that reset initializes some internal state (like counters and registers). If it fails then what guarantee do you have about consistency of the state?
It would be nice to add a comment which explains that this kind of failure isn't fatal.

include/dev/rtl8139_reg.h
```diff
@@ -0,0 +1,22 @@
...
+
+/*
+ * Command register.
+ */
+#define RL_CMD_RESET 0x0010
+
+/*
+ * Config 1 register
+ */
+#define RL_CFG1_PWRON 0x00
```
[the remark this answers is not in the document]
MichalBlk: Why not just `RL_CFG1` and `RL_CFG1_...`? Do you really want to literally copy all definitions from their header file? Maybe they have other devices starting with prefix "RL" and hence they use `RF_8139_CFG1` instead of `RF_CFG1`.

include/dev/rtl8139_reg.h
```diff
@@ -0,0 +1,22 @@
+#ifndef _RTL8139_REG_H_
+#define _RTL8139_REG_H_
+/*
+ * Copied from FreeBSD: sys/dev/rl/if_rlreg.h
+ */
+
+#define RL_COMMAND 0x0037 /* command register */
```
MichalBlk, line 7: The layout of this header file is strange. You labeled the definition of `RL_CMD_RESET` with notable comment
```C
/*
 * Command register.
 */
```
but totally separated it from the definition of `RL_COMMAND`. Why not
```C
/*
 * Command register.
 */
#define RL_COMMAND 0x0037
#define RL_CMD_RESET 0x0010
```

include/dev/rtl8139_reg.h
```diff
@@ -0,0 +1,22 @@
...
+
+#define RL_8139_CFG1 0x0052 /* config register #1 */
```
MichalBlk, line 9: The same thing, why not
```C
/*
 * Config register #1
 */
#define RL_CFG1 0x0052
#define RL_CFG1_PWRON 0x00
```

include/dev/rtl8139_reg.h
```diff
@@ -0,0 +1,22 @@
...
+ * Command register.
+ */
+#define RL_CMD_RESET 0x0010
+
+/*
+ * Config 1 register
+ */
+#define RL_CFG1_PWRON 0x00
+
+#define RL_TIMEOUT (10 * 1000)
```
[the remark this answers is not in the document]
MichalBlk: I agree. Some bits require a few writes to actually obtain some effect. This driver isn't the only one which uses this approach and the value (i.e. `TIMEOUT`) is most probably magic.

include/dev/rtl8139_reg.h
MichalBlk, line 19: I guess that this define describes some particular setting of the first configuration register (probably "power on"?).
It would be nice to add a brief comment beside the definition, like this
```C
#define  RL_CFG1_PWRON 0x00 /* power on */
```

sys/kern/kmem.c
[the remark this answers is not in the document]
MichalBlk: It absolutely doesn't and you can't remove this assert. I think that this assert must be here and is essential form the perspective of the `kmem_map` function. We need to change logic in bus driver (probably in activate function).

include/dev/rtl8139_reg.h
komik0: as stated in comment - it is copied from freebsd - I only removed not needed constants.
I think in such cases it is expected to do not reorder registers - that approach is useful if we want to use difftool in future.

include/dev/rtl8139_reg.h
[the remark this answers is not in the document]
komik0: It was idea to copy whole file and do not introduce changes wherever possible

komik0: done

include/dev/rtl8139_reg.h
komik0: as above

MichalBlk (approved):

cahirwpz (approved):
LGTM after some fixes.

PR #849, opened 2020-11-29
Fix sequential PCI bar RIDs

mohrcore:
First of all, I'm not very familiar with the concept of PCI bars yet, so it's quite possible there's a better solution.

#817 was meant to free PCI device drivers from being responsible for choosing the right PCI bar RIDs by letting them just pick _n_ natural numbers beginning from 0.

**I am surprised it actually ended up getting merged to master branch.** There's just one driver which benefits from this change and it is _stdvga_ on _MIPS_ board. Yet, after this commit, anyone who had run mimiker with `-g` option would notice that the VGA is broken.

Currently, my solution makes the drivers access PCI bar resources using RIDs as if they were sequential while retaining the underlying physical order. Which bar is valid for use depends on its config which is read in `pci_enumerate` procedure and it looks like reordering them (if possible) would require changing their configurations, which I _believe_ are written by the device connected to the PCI bus.

### DIFF
--- a/sys/kern/pci.c
+++ b/sys/kern/pci.c
@@ -108,9 +108,12 @@ void pci_bus_enumerate(device_t *pcib) {
 
         size = -size;
         uint8_t rid = pcid->nbars++;
-        pcid->bar[rid] = (pci_bar_t){
-          .owner = dev, .type = type, .flags = flags, .size = size, .rid = rid,
-          .id = i };
+        pcid->bar[rid] = (pci_bar_t){.owner = dev,
+                                     .type = type,
+                                     .flags = flags,
+                                     .size = size,
+                                     .rid = rid,
+                                     .id = i};
       }
     }
   }
--- a/sys/kern/pci.c
+++ b/sys/kern/pci.c
@@ -50,6 +50,8 @@ static uint32_t pci_bar_size(device_t *pcid, int bar, uint32_t addr) {
 
 DEVCLASS_CREATE(pci);
 
+#define INVALID_BAR 0xff
+
 void pci_bus_enumerate(device_t *pcib) {
   pci_addr_t pcia = {.bus = 0, .device = 0};
   device_t pcid = {.parent = pcib, .bus = DEV_BUS_PCI, .instance = &pcia};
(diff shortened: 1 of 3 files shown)


sys/kern/pci.c
```diff
@@ -50,6 +50,8 @@ static uint32_t pci_bar_size(device_t *pcid, int bar, uint32_t addr) {
 
 DEVCLASS_CREATE(pci);
 
+#define INVALID_BAR 0xff
```
ThomasStachovsky, line 53: I know PCI_BAR_MAX is equal to 6, but maybe it would be better to define invalid bar to be (-1)? 0xff seems a bit random, since rid is an int.

ThomasStachovsky (approved):
Looks good to me

MichalBlk:
We can make distinction between real bar id and sequential bar id in the `pci_bar` structure (e.g. `rrid` and `sriq`).
Then in `pci_bus_enumerate` do `pcid->bar[id].rriq = i` and use it in `gt_pci_activate_resource`. It would only require minor changes.

sys/kern/pci.c
mohrcore: I chose `0xff` because `nbars` field of `pci_device_t` is an `uint8_t`.

Speaking of which, I'm unsure whether I should keep it. Technically it isn't really used by any of existing drivers, but it might be considered a useful information.

mohrcore:
> We can make distinction between real bar id and sequential bar id in the `pci_bar` structure (e.g. `rrid` and `srid`).
> Then in `pci_bus_enumerate` do `pcid->bar[id].rrid = i` and use it in `gt_pci_activate_resource`. It would only require minor changes.

Ok, but why? Is there any use for keeping real bar id there?

MichalBlk:
It eliminates the need to create `gt_pci_bar_of_rid` and additional logic in `gt_pci_alloc_resource`.

mohrcore:
I wrote a different solution which, much like what @MichalBlk has suggested. I'm not sure wheter I like it more. It feels more intuitive to store `pci_bar_t` in order of those bars instead of ordering them by their RIDs. I feel like we are mixing information that's related to drivers (RIDs) with information that describes connected hardware a bit too much here. Ideally, I think we should neither store RID in `pci_bar_t` or put these `pci_bar_t`s in order that's unrelated to hardware. Resource lists should take care of that when they are implemented.

ThomasStachovsky (approved):
lgtm

MichalBlk:
I'm sorry. I didn't think it through. We only need one identifier in the `bar` structure (let's call it `rid`). It will remember the physical position of the bar. Then we do `pcid->bar[pcid->nbars++] = (pci_bar_t){..., .rid = i, ...}`.

cahirwpz:
We're back to original solution. We don't want to renumber BARs.

PR #854, opened 2020-12-01
Implement /dev/tty pseudo-device.

j-piecuch:
This device always refers to the opening process's controlling terminal.


sys/kern/tty.c
```diff
@@ -910,3 +911,31 @@ vnodeops_t tty_vnodeops = {
   .v_close = tty_vn_close,
   .v_getattr = tty_vn_getattr,
 };
+
+/* Controlling terminal pseudo-device (/dev/tty) */
+
+static int dev_tty_open(vnode_t *v, int mode, file_t *fp) {
```
franciscozdo, line 917: Why we don't use first argument in this function?

sys/kern/tty.c
```diff
@@ -910,3 +911,31 @@ vnodeops_t tty_vnodeops = {
...
+  proc_t *p = proc_self();
+  int error;
+
+  SCOPED_MTX_LOCK(all_proc_mtx);
+  tty_t *tty = p->p_pgrp->pg_session->s_tty;
+  if (!tty)
```
franciscozdo, line 923: Could you shortly explain this check?

franciscozdo (commented):
LGTM. I would be grateful if you explain purpose of `dev_tty_open()`. (see my comments)

sys/kern/tty.c
j-piecuch: `/dev/tty` refers to the controlling terminal of the current process, if it has one. See https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap10.html. I return `ENXIO` here because that's what NetBSD does: http://bxr.su/NetBSD/sys/kern/tty_tty.c#62

sys/kern/tty.c
j-piecuch: Because we don't need it. What we need is the vnode of the controlling terminal.

sys/kern/tty.c
franciscozdo: I got it. Thanks

franciscozdo (approved):

cahirwpz (approved):
LGTM For now I cannot see how to achieve the same but in a cleaner way.

PR #856, opened 2020-12-03
tty: don't report errors on partial reads/writes.

j-piecuch:
If we report an error, the number of bytes read/written is not returned to userspace.


franciscozdo (approved):
LGTM

cahirwpz (approved):
LGTM

PR #858, opened 2020-12-04
Only use physical bar numbers.

MichalBlk:
Abandon the current approach which (incorrectly) tries to provide sequential bar numbers to a device driver. We should only rely on physical bar identifiers, which drivers must be aware of.


mohrcore (approved):
LGTM, please provide a brief description for this PR tho.

ThomasStachovsky (approved):

PR #859, opened 2020-12-05
e.MMC/SD(SC/HC) drivers for RPi 3

mohrcore:
~Working on _EMMC_ driver for RPi3.~

~Currently porting some stolen code and trying to get it to work.~

This PR compiles all the work related to making drivers based on e.MMC.

A driver for e.MMC controller on BCM2835

- [x] e.MMC interface (completed in #1057)
- [x] Driver for e.MMC controller on BCM2835 (#1101 in progress: review)
- [x] Machine-independent driver for SD(SC/HC) bus (#1135 in progress: review)


mohrcore:
_**f u n  f a c t**_: Apparently there are **two** _eMMC_ controllers onboard and the one that's documented by Broadcom doesn't even implement full specs. The other one seems to be undocumented but _Linux_ and _FreeBSD_'s drivers use this one instead. However it seems to not be enabled (at least for the SD card reader) by default. Also it looks like there's a [confusion regarding the matter of which one should be called "Arasan controller"](https://www.raspberrypi.org/forums/viewtopic.php?t=240243&start=62), so I'm just going to call the one at `MMIO_BASE + 0x00300000` a _****_ controller, because supposedly it doesn't comply with the standard, and the other one (`MMIO_BASE +  0x202000`) is going to be called _The Snobby Controller_, because it seems that only a select elite is allowed to read its docs.

So right now, this is a WIP driver for the **** controller, but hopefully it won't be a big problem to move to the Snobby one after I get this one working.

mohrcore:
So I've decided to merge #1057 into this branch already. If any adjustments will be done to the interface, I'll just merge them here later.

mohrcore:
Something really weird has happened and the driver no longer recognizes the card as SDHC like it used to, but as SDSC. I'm pretty sure the code responsible for doing that in `sd.c` reamained unchanged. I'll have to look into that.

mohrcore:
I wanted to make two PRs that would feature e.MMC controller  and SD drivers separately, but due to the fact that separating them leaves `emmc` devclass empty which results in linker failing I'm not exactly sure what to do with this.

cahirwpz:
All parts that were constituents of this PR have been already merged into `master`. Hence it's safe to close this PR.

PR #861, opened 2020-12-06
Implement syscall restarting

j-piecuch:
- Move stop & kill signal handling to `sig_check()`
- Add `ERESTARTSYS` and `ERESTARTNOHAND` error codes, remove `ERESTART`
- Move setting syscall return value to `on_user_exc_leave()`


include/sys/errno.h
```diff
@@ -176,8 +176,13 @@
 /* pseudo-errors returned inside kernel to modify return to process */
 #define EJUSTRETURN 256  /* don't modify regs, just return to userspace with */
                          /* current uctx (used by sigreturn and execve) */
-#define ERESTART 257     /* restart syscall */
 #define EPASSTHROUGH 258 /* ioctl not handled by this layer */
+/* These errors indicate when the current syscall should be restarted after
+ * being interrupted by a signal. Their semantics are the same as in Linux. */
+#define ERESTARTSYS                                                            \
+  259                      /* Restart iff no signal was caught                 \
```
pj1031999, line 183: Could you do something with formatting here?
`#define ERESTARTSYS 259` is much more readable for me.

include/sys/errno.h
```diff
@@ -176,8 +176,13 @@
...
+                            * or caught signal has SA_RESTART set */
+#define ERESTARTNOHAND 260 /* Restart iff no signal was caught */
```
pj1031999, line 185: Please correct me if I am wrong but on _BSDs_ and on _Linux_ `ERESTART` exists end it is exposed to user. On the other hand `ERESTARTSYS` and `ERESTARTNOHAND` are not visible for user.
In `errno.h` we just include `sys/errno.h` so that errors are visible for user.
Could you comment that?

sys/kern/exception.c
```diff
@@ -17,18 +21,50 @@ void on_exc_leave(void) {
...
     WITH_PROC_LOCK(p) {
       /* Calling sig_post() multiple times before returning to userspace
-       * will not make us lose signals, see comment on sig_post() in signal.h */
-      ksiginfo_t ksi;
-      while (sig_check(td, &ksi))
-        sig_post(&ksi);
+       * will not make us lose signals, see comment on sig_post() in signal.h
+       */
+      sig_post(&ksi);
+      sig = sig_check(td, &ksi);
```
pj1031999, line 67: I don't understand. What is the difference?

sys/kern/signal.c
```diff
@@ -445,27 +445,47 @@ void sig_onexec(proc_t *p) {
   }
 }
 
+bool sig_should_stop(sigaction_t *sigactions, signo_t sig) {
+  return (sigactions[sig].sa_handler == SIG_DFL && defact(sig) == SA_STOP);
+}
```
pj1031999, line 450: Is it used outside of `signal.c`?
If not maybe changing it into static function will be a good idea?

include/sys/errno.h
j-piecuch: Oh, I didn't notice it was formatted like this. `clang-format` is weird sometimes.

include/sys/errno.h
j-piecuch: In FreeBSD, `ERESTART` is internal to the kernel, as can be seen here: http://bxr.su/FreeBSD/sys/sys/errno.h#192. On Linux, `ERESTART` does seem to be exposed to userspace, but it is used very rarely.

In our case, the `ERESTART*` codes are not exposed to userspace, since they are guarded by `#ifdef _KERNEL`.

sys/kern/exception.c
j-piecuch: This is a bit ugly. We need to know whether a signal will be caught in order to set the syscall return value (and maybe restart the syscall), so we must call `sig_check()` before `set_syscall_retval()`.  But we also must call `set_syscall_retval()` before `sig_post()`, as `sig_post()` modifies the user context.

sys/kern/signal.c
j-piecuch: Done.

include/sys/errno.h
pj1031999: Ok. Now I see.

sys/kern/exception.c
pj1031999: Thanks for your explanation.

pj1031999 (approved):
LGTM

sys/kern/exception.c
```diff
@@ -17,18 +21,50 @@ void on_exc_leave(void) {
...
+    return;
 
+  if (error == ERESTARTSYS || error == ERESTARTNOHAND) {
+    if (!sig || (error == ERESTARTSYS &&
+                 (p->p_sigactions[sig].sa_flags & SA_RESTART))) {
+      mcontext_restart_syscall(ctx);
+      return;
+    }
+    error = EINTR;
+  }
```
pj1031999, line 39: It would be nice if it could be simplified. Otherwise, it's worth adding a comment here.

sys/aarch64/trap.c
```diff
@@ -123,7 +125,7 @@ void user_trap_handler(mcontext_t *uctx) {
       break;
 
     case EXCP_SVC64:
-      syscall_handler(ESR_ELx_SYSCALL(esr), ctx);
+      syscall_handler(ESR_ELx_SYSCALL(esr), ctx, &result);
```
wiklam, line 128: Maybe it's a good idea instead using `ESR_ELx_EXCEPTION(esr)`, use `exc_code` like you have done in switch above.

sys/kern/exception.c
j-piecuch: I think I've made it more readable now. Do you agree?

sys/aarch64/trap.c
j-piecuch: This is line uses `ESR_ELx_SYSCALL`, not `ESR_ELx_EXCEPTION`.

wiklam: My bad.

sys/kern/exception.c
pj1031999: Yes, thanks.
You may consider adding `{`, `}` for maximum readability.

j-piecuch: Done.

cahirwpz (approved):
Though I find the changes somewhat convoluted I have not found a way to simplify them significantly. Since the review has lasted for too long I feel compelled to merge the PR as is.

PR #863, opened 2020-12-07
Kernel profiling

wiklam:
Kernel profiling enable to create a graph of the called functions.
- [x] Profiling data initialization.

- [x] Function which will collect information at the beginning of each function call.

- [x] Compiler instrumentalization
- [x] Save the profiled data

- [x] Profiled data parser


wiklam:
Introduced in #979  and #1062

PR #866, opened 2020-12-07
Access checks during open

franciscozdo:
I implemented checks performed while opening existing file. For now we don't check if we have permissions to perform operations on directories on path to given file.

It required some fixes in tests. Some files should be created with RW permissions


sys/kern/cred_checks.c
```diff
@@ -74,3 +75,47 @@ bool cred_can_setlogin(cred_t *cred) {
...
+    granted |= VWRITE;
+  if (va->va_mode & S_IXOTH)
+    granted |= VEXEC;
+
+check:
+  if (cred->cr_euid == 0) {
+    granted |= VREAD | VWRITE;
+    /* root has exec permission when:
+     *   - file is directory
+     *   - at lest one exec bit is se
```
xThaid, line 114: typo

xThaid (commented):
Looks goods. There's a small typo, but it is not that important.

sys/kern/cred_checks.c
```diff
@@ -74,3 +75,47 @@ bool cred_can_setlogin(cred_t *cred) {
...
+     */
+    if (S_ISDIR(va->va_mode) || va->va_mode & (S_IXUSR | S_IXGRP | S_IXOTH))
+      granted |= VEXEC;
+  }
```
j-piecuch, line 118: Minor thing: I would move this root check to the very beginning of the function, so that the code at the `check` label does as little as possible. I think this would improve readability.

include/sys/vnode.h
```diff
@@ -24,7 +24,7 @@ typedef struct componentname componentname_t;
 #define VNOVAL (-1)
 
 /* vnode access modes */
-typedef enum { VEXEC = 1, VWRITE = 2, VREAD = 4 } accmode_t;
+typedef enum { VEXEC = 1, VWRITE = 2, VREAD = 4, VADMIN = 8 } accmode_t;
```
j-piecuch, line 27: What does `VADMIN` mean? Also, why isn't it granted to root in `cred_can_access()`?

sys/kern/tty.c
```diff
@@ -862,7 +862,11 @@ static int tty_ioctl(file_t *f, u_long cmd, void *data) {
 
 static int tty_vn_getattr(vnode_t *v, vattr_t *va) {
   memset(va, 0, sizeof(vattr_t));
-  va->va_mode = S_IFCHR;
+  /* XXX assume root owns tty and only root can read and write to it */
+  va->va_mode =
+    S_IFCHR | S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
```
j-piecuch, line 867: Doesn't `S_IROTH | S_IWOTH` mean that anyone can read and write to it?

sys/kern/vfs_syscalls.c
```diff
@@ -79,6 +80,7 @@ static int vfs_create(proc_t *p, int fdat, char *pathname, int flags, int mode,
       vnode_drop(vs.vs_vp);
       error = EEXIST;
     }
+    created = 0;
```
j-piecuch, line 83: `*created`

sys/kern/vfs_syscalls.c
```diff
@@ -87,19 +89,42 @@ static int vfs_create(proc_t *p, int fdat, char *pathname, int flags, int mode,
...
       return error;
   } else {
     if ((error = vfs_namelookupat(p, fdat, VNR_FOLLOW, pathname, &v)))
       return error;
   }
 
+  if (!created)
+    if ((error = vfs_check_open(v, flags, &p->p_cred)))
+      return error;
+
```
j-piecuch, line 127: Just making sure I understand it:
This is an optimization, right? We could always do `vfs_check_open()`, but it will always succeed if we created the file. So we don't _need_ the `created` parameter in `vfs_create()`, correct?

j-piecuch (changes requested):

include/sys/vnode.h
franciscozdo: `VADMIN` means "owner of the file". I am not sure if root shoud get VADMIN access. I will check that. Also we don't use it now so I will introduce it later.

sys/kern/tty.c
franciscozdo: Yes. Should I change mode of tty?

sys/kern/vfs_syscalls.c
franciscozdo: No. We don't call `vfs_check_open()` when we created a file. That means if you created a file and you won't have access to it you can still write or read it unless you close it in your program. It is done like that in BSD and in Linux.

sys/kern/cred_checks.c
franciscozdo: Thanks!

sys/kern/vfs_syscalls.c
franciscozdo: Thanks!

sys/kern/tty.c
j-piecuch: The comment above says that only root can read and write to the tty, so I suggest changing the comment.

sys/kern/vfs_syscalls.c
j-piecuch: I see. Thanks!

sys/kern/tty.c
franciscozdo: Yes. I forgot to change comment. Thanks!

j-piecuch (approved):

cahirwpz (approved):
LGTM

PR #867, opened 2020-12-07
KCSAN

xThaid:


xThaid:
OK, so in theory this is a working version of KCSAN, which could be reviewed. It is only working in theory, because it requires higher version of the tool chain and as long as Mimiker doesn't support many threads running at the same time - this tool will be useless.

cahirwpz:
Please move changes unrelated to KCSAN to another PR.

sys/kern/main.c
```diff
@@ -110,6 +114,10 @@ __noreturn void kernel_init(void) {
 
   klog("Kernel initialized!");
 
+#if KCSAN
+  kcsan_ready = 1;
+#endif
```
j-piecuch, line 119: Perhaps having a `kcsan_enable()` function would look cleaner? The function would be defined as `__nothing` if `!KCSAN`.

jpszczolowski: Or `init_kcsan()` to stay consistent with KASAN

sys/kern/kcsan.c
```diff
@@ -0,0 +1,262 @@
...
+      break;
+    case 8:
+      new_val.b8 = *((uint64_t *)addr);
+      break;
+    default:
+      break;
+  }
+
+  if (prev_val.b8 != new_val.b8) {
+    panic("===========KernelConcurrencySanitizer===========\n"
```
jpszczolowski, line 160: I'm not sure how KCSAN works but I guess this code will be called very often. If that's the case, you can consider adding `__predict_true()` / `__predict_false()` to branches that will be almost always/never taken, like in KASAN. E.g. I'd change this to `if (__predict_false(prev_val.b8 != new_val.b8))`.

sys/kern/kcsan.c
```diff
@@ -0,0 +1,262 @@
...
+#include <sys/vm.h>
+#include <sys/thread.h>
+#include <machine/interrupt.h>
+#include <sys/klog.h>
+
+#define WATCHPOINT_READ_BIT (1 << 31)
+#define WATCHPOINT_SIZE_SHIFT 28
+#define WATCHPOINT_SIZE_MASK (~(WATCHPOINT_ADDR_MASK | WATCHPOINT_READ_BIT))
+#define WATCHPOINT_ADDR_MASK ((1 << WATCHPOINT_SIZE_SHIFT) - 1)
+
```
jpszczolowski, line 11: Could you please add a little bit more comments to non-obvious parts of this file? And if you know a good explanation of the algorithm online (LWN?), a link inside a comment could be helpful for future maintainers of the code.

sys/kern/sched.c
```diff
@@ -218,7 +218,7 @@ void sched_maybe_preempt(void) {
   }
 }
 
-bool preempt_disabled(void) {
+__no_sanitize bool preempt_disabled(void) {
   thread_t *td = thread_self();
```
jpszczolowski, line 222: Please consider adding a short comment explaining why `__no_sanitize` is required here.

sys/kern/thread.c
```diff
@@ -131,7 +131,7 @@ void thread_delete(thread_t *td) {
   pool_free(P_THREAD, td);
 }
 
-thread_t *thread_self(void) {
+__no_sanitize thread_t *thread_self(void) {
   return PCPU_GET(curthread);
```
jpszczolowski, line 135: Please consider adding a short comment explaining why `__no_sanitize` is required here.

sys/kern/kcsan.c
```diff
@@ -0,0 +1,262 @@
...
+void __tsan_write_range(void *ptr, size_t size) {
+  kcsan_check((uintptr_t)ptr, size, false);
+}
+
+#define DEFINE_KCSAN_VOLATILE_READ_WRITE(size)                                 \
+  void __tsan_volatile_read##size(void *ptr) {                                 \
+  }                                                                            \
+  void __tsan_volatile_write##size(void *ptr) {                                \
+  }
```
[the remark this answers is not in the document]
jpszczolowski: (No matter what the answer is, would be good to explain that in a short comment in the source code)

sys/kern/kcsan.c
```diff
@@ -0,0 +1,262 @@
...
+    case 2:
+      new_val.b2 = *((uint16_t *)addr);
+      break;
+    case 4:
+      new_val.b4 = *((uint32_t *)addr);
+      break;
+    case 8:
+      new_val.b8 = *((uint64_t *)addr);
+      break;
+    default:
```
jpszczolowski, line 155: You can consider moving the switch to a separate function so that it's not duplicated here. Not sure if worth the effort, though.

sys/kern/kcsan.c
```diff
@@ -0,0 +1,262 @@
...
+#define WATCHPOINT_READ_BIT (1 << 31)
+#define WATCHPOINT_SIZE_SHIFT 28
+#define WATCHPOINT_SIZE_MASK (~(WATCHPOINT_ADDR_MASK | WATCHPOINT_READ_BIT))
+#define WATCHPOINT_ADDR_MASK ((1 << WATCHPOINT_SIZE_SHIFT) - 1)
+
+#define WATCHPOINT_INVALID 0
+
+#define WATCHPOINT_NUM 32
+#define SKIP_COUNT 500
+#define WATCHPOINT_DELAY 10000
```
jpszczolowski, line 16: How was this delay chosen?

jpszczolowski (changes requested):
Nice :+1: but where are the comments? :joy:

sys/kern/kcsan.c
```diff
@@ -0,0 +1,262 @@
+#include <sys/mimiker.h>
+#include <sys/vm.h>
+#include <sys/thread.h>
+#include <machine/interrupt.h>
+#include <sys/klog.h>
+
```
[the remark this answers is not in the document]
xThaid: Yep, added.

sys/kern/kcsan.c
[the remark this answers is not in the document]
xThaid: I've added a description of the algorithm with the motivation why volatile accesses are discarded.

sys/kern/main.c
xThaid: I like the idea. Thanks

sys/kern/kcsan.c
xThaid: It won't be called that often (approximately every 500th memory access or something), but it doesn't cost to add that predictions. I've added it to other places as well.

sys/kern/kcsan.c
xThaid: I've added a short description of the algorithm that should clarify some ideas. Please let me know if you find it not explanatory enough.

sys/kern/kcsan.c
xThaid: It turned out the effort wasn't too big, so I've changed it.

sys/kern/kcsan.c
```diff
@@ -0,0 +1,262 @@
...
+  return atomic_fetch_sub(&skip_counter, 1) == 1;
+}
+
+static void delay(int count) {
+  for (int i = 0; i < count; i++) {
+    __asm__ volatile("nop");
+  }
+}
+
+static inline atomic_int *find_watchpoint(uintptr_t addr, size_t size,
```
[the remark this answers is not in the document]
xThaid: The function isn't that complex, so I didn't split it. I've only changed its name and added a short comment.

sys/kern/kcsan.c
xThaid: These constants are chosen rather arbitrarily. That delay makes tests execute within a reasonable time.

sys/kern/thread.c
```diff
@@ -131,6 +131,11 @@ void thread_delete(thread_t *td) {
   pool_free(P_THREAD, td);
 }
 
+/*
+ * Instrumentation in this function makes KCSAN crash. I don't know the cause,
+ * but the function is so simple that there is no need to instrumentate it
+ * anyway.
+ */
```
j-piecuch, line 138: Maybe it crashes for the same reason as `preemt_disabled()` - infinite recursion. Notice that `preempt_disabled()` calls `thread_self()`.

include/sys/kcsan.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * When KCSAN is enabled, the code is instrumented to allow the monitoring of
+ * memory accesses. Specifically, each memory access will be augmented by a
+ * function call. The compiler distinguish different memory accesses depending
+ * of the type of the variable: plain, volatile or atomic. In our implementation
+ * last two are discarded, because they don't meet the conditions of the
+ * previously defined data race.
+ *
+ * These function calls are ignored (SKIP_COUNT - 1) out of SKIP_COUNT times; to
+ * do otherwise would slow the kernel to a point of complete unusability. On the
```
j-piecuch, line 23: This isn't quite true: `kcsan_check()` always checks for a race, but only sets up a watchpoint once every `SKIP_COUNT` calls. I see that you clarify this in the next paragraph, but I would like the reader to know it after reading only this paragraph.

sys/kern/kcsan.c
```diff
@@ -0,0 +1,252 @@
...
+
+  *addr = watchpoint & WATCHPOINT_ADDR_MASK;
+  *size = (watchpoint & WATCHPOINT_SIZE_MASK) >> WATCHPOINT_SIZE_SHIFT;
+  *is_read = (watchpoint & WATCHPOINT_READ_BIT) == WATCHPOINT_READ_BIT;
+  return true;
+}
+
+static inline bool address_match(uintptr_t addr1, size_t size1, uintptr_t addr2,
+                                 size_t size2) {
+  return (addr1 <= addr2 + size2 - 1) && (addr2 <= addr1 + size1 - 1);
```
j-piecuch, line 54:
```suggestion
  return (addr1 < addr2 + size2) && (addr2 < addr1 + size1);
```

sys/kern/thread.c
xThaid: Oh right, I missed that.

include/sys/kcsan.h
xThaid: Ok, I've added a small clarification at the beginning of that paragraph.

sys/kern/kcsan.c
```diff
@@ -0,0 +1,262 @@
...
+static atomic_int kcsan_slot_taken_count;
+static atomic_int kcsan_set_watchpoint_count;
+
+/* How many accesses should be skipped before we setup a watchpoint. */
+static atomic_int skip_counter = SKIP_COUNT;
+
+int kcsan_ready;
+
+static inline int encode_watchpoint(uintptr_t addr, size_t size, bool is_read) {
+  return (is_read ? WATCHPOINT_READ_BIT : 0) | (size << WATCHPOINT_SIZE_SHIFT) |
```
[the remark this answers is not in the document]
xThaid: Ok, we've gained 1 extra bit for an address by storing the logarithm of the size.

sys/kern/kcsan.c
```diff
@@ -0,0 +1,252 @@
...
+  if (!kcsan_ready || cpu_intr_disabled() || preempt_disabled() ||
+      size > MAX_ENCODABLE_SIZE)
+    return;
+
+  if (addr < KERNEL_SPACE_BEGIN)
+    return;
+
+  atomic_int *watchpoint = search_for_race(addr, size, is_read);
+  if (__predict_true(watchpoint == NULL)) {
+    if (should_watch())
```
j-piecuch, line 169: `size` isn't guaranteed to be a power of 2 due to `__tsan_{read,write}_range()`. We could either check that it is before setting up a watchpoint, or round it down to the nearest power of 2 (which is what is done currently, since `log2()` rounds down). If we're going to stick with rounding down, there should probably be a comment about it somewhere, maybe in `setup_watchpoint()`.

sys/kern/kcsan.c
jpszczolowski: :+1:

xThaid:
> Could you please also mention `KCSAN=1` flag in https://github.com/cahirwpz/mimiker/blob/master/README.md?

Done

sys/kern/sched.c
```diff
@@ -218,7 +218,11 @@ void sched_maybe_preempt(void) {
   }
 }
 
-bool preempt_disabled(void) {
+/*
+ * Instrumentation in this function would cause KCSAN to fall into an infinite
+ * recursion.
+ */
+__no_sanitize bool preempt_disabled(void) {
```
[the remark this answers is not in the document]
xThaid: +1

sys/kern/thread.c
```diff
@@ -131,7 +131,11 @@ void thread_delete(thread_t *td) {
   pool_free(P_THREAD, td);
 }
 
-thread_t *thread_self(void) {
+/*
+ * Instrumentation in this function would cause KCSAN to fall into an infinite
+ * recursion.
+ */
+__no_sanitize thread_t *thread_self(void) {
```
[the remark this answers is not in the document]
xThaid: +1

sys/kern/kcsan.c
```diff
@@ -0,0 +1,256 @@
+#include <sys/mimiker.h>
+#include <sys/vm.h>
+#include <sys/thread.h>
+#include <machine/interrupt.h>
+#include <sys/klog.h>
+#include <sys/kcsan.h>
+
+#define WATCHPOINT_READ_BIT (1 << 31)
```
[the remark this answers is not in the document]
xThaid: We always want to use the leftmost bit as a read bit, so I don't think it would be useful to make it dependent on another value.

sys/kern/kcsan.c
```diff
@@ -0,0 +1,256 @@
...
+}
+
+static inline bool decode_watchpoint(int watchpoint, uintptr_t *addr,
+                                     size_t *size, bool *is_read) {
+  if (watchpoint == WATCHPOINT_INVALID)
+    return false;
+
+  *addr = watchpoint & WATCHPOINT_ADDR_MASK;
+  *size = 1 << ((watchpoint & WATCHPOINT_SIZE_MASK) >> WATCHPOINT_SIZE_SHIFT);
+  *is_read = (watchpoint & WATCHPOINT_READ_BIT) == WATCHPOINT_READ_BIT;
```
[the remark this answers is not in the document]
xThaid: This way we make sure it is properly casted to boolean and it doesn't affect the readability that much.

cahirwpz (approved):

PR #869, opened 2020-12-08
Add launcher option to select UARTs that should be run in raw mode.

j-piecuch:
Add support for selecting raw mode UARTs from the command line.
We should also consider running `/dev/tty1` in raw mode by default, since we now have a working TTY subsystem.


j-piecuch:
Superseded by #870

PR #870, opened 2020-12-08
launcher: run socat on /dev/tty1 in raw mode by default.

j-piecuch:
At this point our TTY subsystem has enough features to handle raw input.


cahirwpz (approved):
LGTM

PR #871, opened 2020-12-11
vm_object_t: add merging shadow objects

bkjg:
This PR depends on #832.


cahirwpz:

franciscozdo:
This PR will not be used, because we will redefine meaning of objects in VM system.

cahirwpz:
Since we're going to use `amap`-s instead of shadow objects I'm closing this PR.

PR #872, opened 2020-12-11
Further unification of context handling

cahirwpz:
Let's simplify and unify all context handling in the kernel to make it as portable as possible.


sys/aarch64/rootdev.c
```diff
@@ -270,19 +271,22 @@ static resource_t *rootdev_alloc_resource(device_t *dev, res_type_t type,
...
+                                     resource_t *r) {
   if (type == RT_MEMORY)
     return bus_space_map(r->r_bus_tag, r->r_bus_handle, resource_size(r),
                          &r->r_bus_handle);
   return 0;
 }
 
-static void rootdev_deactivate_resource(device_t *dev, res_type_t type, resource_t *r) {
+static void rootdev_deactivate_resource(device_t *dev, res_type_t type,
+                                        resource_t *r) {
```
pj1031999, line 289: We need to add `verify_formatting` for _AArch64_.

pj1031999 (approved):
LGTM

PR #836, opened 2020-11-19
ISA bus driver

mohrcore:
WIP ISA bus driver.

Should solve #380 in future.

PR #873, opened 2020-12-11
ISA bus driver for Malta

mohrcore:
Working on ISA bus driver.

This is a an update and a continuation of #836, I had to switch to another branch, due to rebasing problems.


mohrcore:
Blocked by #888

ThomasStachovsky (changes requested):
Looks okay to me. You just have to make a few changes imposed by recent changes to resources and rmans.
Edit: Didn't notice the review label has been removed

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  resource_t *r =
+    rman_reserve_resource(rman, type, rid, start, end, size, 0, flags);
+
+  if (r == NULL)
+    return NULL;
+  r->r_rid = rid;
+
+  if (type != RT_IRQ) {
```
MichalBlk, line 63: Please remove the whole if statement an place the bare body of the statement (at this point `type` is always equal to `RT_IOPORTS`).

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+    r->r_bus_tag = generic_bus_space;
+    r->r_bus_handle = bh + resource_start(r);
+  }
+
+  if (flags & RF_ACTIVE) {
```
MichalBlk, line 68: Please remove the if condition and place the bare body of the statement. Note that in attach, you've allocated and activated the IO ports hence any IO ports you allocate must be marked as active (please check out #898 for latest changes regarding the `RF_ACTIVE` flag).

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+static int intel_isa_activate_resource(device_t *dev, resource_t *r) {
+  return 0;
+}
+
+static void intel_isa_deactivate_resource(device_t *dev, resource_t *r) {
+  return;
+}
+
+static void intel_isa_release_resource(device_t *dev, resource_t *r) {
+  /* No need to deactivate resource, because deactivation is a NOP anyway */
```
MichalBlk, line 87: That's not true. Please note that the function you've presented is wrapped by the `bus_deactivate_resource`.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+    if (bus_activate_resource(dev, r)) {
+      resource_release(r);
+      return NULL;
+    }
+  }
+
+  return r;
+}
+
+static int intel_isa_activate_resource(device_t *dev, resource_t *r) {
```
MichalBlk, line 78: What if a PCI device allocates a memory resource without the `RV_ACTIVE` flag and later calls `bus_activate_resource`?
I think you should delegate this methods to the parent (I am aware of the bar problems).

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+#include <sys/pci.h>
+#include <sys/interrupt.h>
+#include <dev/piixreg.h>
+#include <dev/isareg.h>
+#include <sys/libkern.h>
+#include <sys/devclass.h>
+
+#define IO_ISASIZE 1024
+
+#define ICU_ADDR(x) ((x) + 0)
```
MichalBlk, line 17: Why do we need all the defines?

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+
+static int intel_isa_activate_resource(device_t *dev, resource_t *r) {
+  return 0;
+}
+
+static void intel_isa_deactivate_resource(device_t *dev, resource_t *r) {
+  return;
+}
+
+static void intel_isa_release_resource(device_t *dev, resource_t *r) {
```
MichalBlk, line 86: If `r->r_type != RT_IOPORTS` then you should delegate this request to the parent.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+  /* No need to deactivate resource, because deactivation is a NOP anyway */
+  resource_release(r);
+}
+
+static int intel_isa_probe(device_t *d) {
+  return d->unit == 0;
+}
+
+static int intel_isa_attach(device_t *isab) {
```
MichalBlk, line 95: Shouldn't this function perform some action on the 8259A interrupt controller?

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+
+bus_methods_t intel_isa_bus_bus_if = {
+  .intr_setup = NULL,    /* Dispatched */
+  .intr_teardown = NULL, /* Dispatched */
+  .alloc_resource = intel_isa_alloc_resource,
+  .activate_resource = intel_isa_activate_resource,
+  .deactivate_resource = intel_isa_deactivate_resource,
+  .release_resource = intel_isa_release_resource};
+
+/* clang-format off */
```
MichalBlk, line 140: Please remove it (it's not needed and there is no matching `/* clang-format on */).

MichalBlk (changes requested):
Some major changes are needed.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+  resource_release(r);
+}
+
+static int intel_isa_probe(device_t *d) {
+  return d->unit == 0;
+}
+
+static int intel_isa_attach(device_t *isab) {
+  intel_isa_state_t *isa = isab->state;
+  isa->io = device_take_ioports(isab, IO_ICUSIZE, RF_ACTIVE);
```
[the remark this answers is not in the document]
mohrcore: No. This is ISA's IO ports RID, not size.
There was a point at which I considered allocating an interrupt resource for each ISA interrupt and give each interrupt its number as its RID, so everything else had its RIDs >= IO_ICUSIZE. This doesn't bear any meaning anymore and I believe resources are now identified uniquely not only by their RID, but by a pair of their type an RID, so I think I'll just change to 0.

sys/mips/82371AB.c
mohrcore: No. Unless this is a philosophical question. In that case - yes, probably.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+#define ICU1_DATA ICU_DATA(IO_ICU1)
+#define ICU2_ADDR ICU_ADDR(IO_ICU2)
+#define ICU2_DATA ICU_DATA(IO_ICU2)
+
+#define ISA_IRQ_BASE (0)
+
+typedef struct intel_isa_state {
+  resource_t *io;
+  rman_t io_rman;
+} intel_isa_state_t;
```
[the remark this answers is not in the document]
mohrcore: Idk. Device names in drivers seem to be quite specific. I know the "_intel_" prefix isn't the most specific, but I can't have the name start with "_82371ab_". Starting a name with just _"isa_" would look like this is a part of a generalized ISA interface, just like things that start with "_pci_" are related to the generalized PCI interface, hence the names of internal procedures and structures of _gt64120_ driver begin with "_gt_pci"_ instead of just "_pci_". I think that giving it "_intel_" as a prefix is a nice compromise between having no distinguishable prefix at all and having an awfully specific one that would result in horrible function names like like `intel82371ab_intr_teardown`. That's at least the kind of compromise _gt64120_ driver goes for.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+    else
+      panic("Can't allocate resource, no dispatch available!");
+  }
+
+  resource_t *r =
+    rman_reserve_resource(rman, type, rid, start, end, size, 0, flags);
+
+  if (r == NULL)
+    return NULL;
+  r->r_rid = rid;
```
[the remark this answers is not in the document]
mohrcore: Am I missing something? `rman_reserve_resource` seems to return `NULL` if it there is an overlap, so this check makes sense to be here.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,153 @@
...
+#include <sys/bus.h>
+#include <mips/malta.h>
+#include <sys/pci.h>
+#include <sys/interrupt.h>
+#include <dev/piixreg.h>
+#include <dev/isareg.h>
+#include <sys/libkern.h>
+#include <sys/devclass.h>
+
+#define IO_ISASIZE 1024
```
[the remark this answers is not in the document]
mohrcore: We don't. It's gone now.

sys/mips/82371AB.c
mohrcore: Done.

sys/mips/82371AB.c
mohrcore: Fixed.

sys/mips/82371AB.c
mohrcore: Again, we don't. No more pointless defines.

sys/mips/82371AB.c
mohrcore: Yes. This is fixed now.

sys/mips/82371AB.c
mohrcore: Done.

sys/mips/82371AB.c
mohrcore: idk wym.
If this is about deactivating interrupts which can and should be deactivated through gt64120 driver, then it's fixed now.

sys/mips/82371AB.c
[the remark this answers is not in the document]
mohrcore: Ooooh, you were talking about the RID setting! It's removed now.

sys/kern/isa.c
```diff
@@ -0,0 +1,3 @@
+#include <sys/devclass.h>
+
+DEVCLASS_CREATE(isa);
```
MichalBlk, line 3: I think you could just move it to the end of `82371AB.c` (and remove this file).

sys/mips/82371AB.c
MichalBlk: You really should remove the condition and leave just
```C
   if (bus_activate_resource(dev, r)) {
      resource_release(r);
      return NULL;
    }
```
(See `gt_pci_alloc_resource`).

sys/mips/82371AB.c
```diff
@@ -0,0 +1,148 @@
...
+      return NULL;
+    }
+  }
+
+  return r;
+}
+
+static int intel_isa_activate_resource(device_t *dev, resource_t *r) {
+  if (r->r_type != RT_IOPORTS)
+    return bus_activate_resource(dev->parent, r);
```
MichalBlk, line 65: Please don't do that. The problem is that `bus_activate_resource` is a non trivial wrapper for a real activation function. Because of this delegation the wrapper will be called twice and thereby the `resource_activate` function will be called twice which bumps up an internal `range` activation counter twice. I know that you do the same for the deactivation function and hence it should work properly, but it's a dirty and confusing solution. I'm afraid you have to do the following
```C
  device_t *isab = dev->parent;
  return BUS_METHODS(isab->parent).activate_resource(isab, r);
```

sys/mips/82371AB.c
```diff
@@ -0,0 +1,148 @@
...
+  /* There are no resource activation procedures for ISA ioports  */
+  return 0;
+}
+
+static void intel_isa_deactivate_resource(device_t *dev, resource_t *r) {
+  if (r->r_type != RT_IOPORTS)
+    return bus_deactivate_resource(dev->parent, r);
```
MichalBlk, line 72: Same problem as with the activation function.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,148 @@
...
+  return;
+}
+
+static void intel_isa_release_resource(device_t *dev, resource_t *r) {
+  if (r->r_type != RT_IOPORTS) {
+    /* Dispatch release for inderectly allocated resources */
+    bus_deactivate_resource(dev->parent, r);
```
MichalBlk, line 79: There should be `bus_release_resource`.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,148 @@
...
+  } else {
+    /* No need to deactivate resource, because deactivation is a NOP anyway */
```
MichalBlk, line 81: You must call the deactivation function because it degrades the `range` activation counter. You're right that ` intel_isa_deactivate_resource` does not do anything, but `bus_deactivate_resource` does and it should be called.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,148 @@
...
+                                            int rid, rman_addr_t start,
+                                            rman_addr_t end, size_t size,
+                                            rman_flags_t flags) {
+  assert(dev->bus == DEV_BUS_ISA);
+
+  rman_t *rman = NULL;
+  intel_isa_state_t *isa = dev->parent->state;
+  bus_space_handle_t bh = 0;
+
+  if (type == RT_IOPORTS) {
```
MichalBlk, line 31: Maybe
```C
/* The long comment below. */
if (type != RT_IOPORTS)
  return us_alloc_resource(dev->parent, type, rid, start, end, size, flags);

intel_isa_state_t *isa = dev->parent->state;
man_t *rman = &isa->io_rman;
bus_space_handle_t bh = isa->io->r_bus_handle;

...
```

sys/mips/82371AB.c
```diff
@@ -0,0 +1,148 @@
...
+  if (r->r_type != RT_IOPORTS)
+    return bus_activate_resource(dev->parent, r);
+  /* There are no resource activation procedures for ISA ioports  */
+  return 0;
+}
+
+static void intel_isa_deactivate_resource(device_t *dev, resource_t *r) {
+  if (r->r_type != RT_IOPORTS)
+    return bus_deactivate_resource(dev->parent, r);
+  return;
```
MichalBlk, line 73: I would remove the return instructions (especially the one with an operand on the right) and leave
```C
if (r->r_type != RT_IOPORTS) {
  device_t *isab = dev->parent;
  return BUS_METHODS(isab->parent).deactivate_resource(isab, r);
}
```

MichalBlk (changes requested):

sys/mips/82371AB.c
```diff
@@ -0,0 +1,148 @@
...
+  /* i8254 timer device */
+  dev = device_add_child(isab, 3);
+  dev->bus = DEV_BUS_ISA;
+  device_add_ioports(dev, 0, IO_TIMER1, IO_TMRSIZE);
+  device_add_irq(dev, 0, 0);
+
+  return bus_generic_probe(isab);
+}
+
+bus_methods_t intel_isa_bus_bus_if = {
```
MichalBlk, line 127: Please add `static`.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,148 @@
...
+  .intr_setup = NULL,    /* Dispatched */
+  .intr_teardown = NULL, /* Dispatched */
+  .alloc_resource = intel_isa_alloc_resource,
+  .activate_resource = intel_isa_activate_resource,
+  .deactivate_resource = intel_isa_deactivate_resource,
+  .release_resource = intel_isa_release_resource,
+};
+
+driver_t intel_isa_bus = {
```
MichalBlk, line 136: Please add `static`.

sys/mips/82371AB.c
MichalBlk: Of course you can just remove (thereby don't implement) the function (the same for deactivation).

MichalBlk: Of course you can just remove (thereby don't implement) the function (the same for deactivation) and handle the "ISA" case in `gt_pci_activate_resource` (`gt_pci_deactivate_resource`).

sys/kern/isa.c
mohrcore: I could, but I'm not fully convinced that a generalized `isa` devclass should be created insde a specific driver.

MichalBlk: Okay, I understand. Let it be that way.

sys/mips/82371AB.c
mohrcore: Done.

sys/mips/82371AB.c
mohrcore: Done.

sys/mips/82371AB.c
mohrcore: Done.

sys/mips/82371AB.c
mohrcore: Yeah, this is way better.

sys/mips/82371AB.c
mohrcore: Done. Now it's always being called.

sys/mips/82371AB.c
mohrcore: This is now fixed addresed with a more robust approach.

sys/mips/82371AB.c
mohrcore: Fixed.

sys/mips/82371AB.c
mohrcore: I decided to not remove these functions as I don't want to rely on behaviour of other drivers too much, but I left them empty which should cover all the scenarios correctly.

sys/mips/82371AB.c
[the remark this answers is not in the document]
komik0: why not just ` piix4_isa_state_t` - it is more specific?

sys/mips/82371AB.c
```diff
@@ -0,0 +1,141 @@
...
+
+  return bus_generic_probe(isab);
+}
+
+static bus_methods_t intel_isa_bus_bus_if = {
+  .intr_setup = NULL,    /* Dispatched */
+  .intr_teardown = NULL, /* Dispatched */
+  .alloc_resource = intel_isa_alloc_resource,
+  .activate_resource = intel_isa_activate_resource,
+  .deactivate_resource = intel_isa_deactivate_resource,
```
komik0, line 125: can it be null? with above (line 65) comment

mohrcore: Could you elaborate? Line 65 is currently
```C
static void intel_isa_deactivate_resource(device_t *dev, resource_t *r) {
```
It's `NULL` because these procedure are dispatched to the gt driver. If your concern is that the information about the original caller might get lost due to insufficient method semantics then don't worry - it's actually still there, just hidden within auxiliary `arg` argument.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,141 @@
...
+} intel_isa_state_t;
+
+static resource_t *intel_isa_alloc_resource(device_t *dev, res_type_t type,
+                                            int rid, rman_addr_t start,
+                                            rman_addr_t end, size_t size,
+                                            rman_flags_t flags) {
+  assert(dev->bus == DEV_BUS_ISA);
+
+  if (type != RT_IOPORTS) {
+    device_t *idev = BUS_METHOD_PROVIDER(dev->parent, alloc_resource);
```
MichalBlk, line 28: Why not just `bus_alloc_resource`? You have literally inlined the `bus_alloc_resource` procedure.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,141 @@
...
+    return BUS_METHODS(idev->parent)
+      .alloc_resource(idev, type, rid, start, end, size, flags);
+  }
+
+  assert(start >= IO_ICUSIZE);
+
+  rman_t *rman = NULL;
+  intel_isa_state_t *isa = dev->parent->state;
+  rman = &isa->io_rman;
```
MichalBlk, line 37: Why not just `rman_t *rman = &isa->io_rman`? This separation of definition and (only) assignment looks wird.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,141 @@
...
+  if (bus_activate_resource(dev, r)) {
+    resource_release(r);
+    return NULL;
+  }
+
+  return r;
+}
+
+static int intel_isa_activate_resource(device_t *dev, resource_t *r) {
+  /* Neither IRQs in general or IOports on ISA require activation */
```
MichalBlk, line 61: Unfortunately, it won't handle all the scenarios correctly. Again, what if an ISA device allocates an inactive memory and wants to activate it later? Your solution is simply inconsistent. If the device allocates memory with the `RF_ACTIVE` flag, then you delegate the call to the parent which calls **its** activation function (which maps the resource). But if the device makes the same call without the mentioned flag and activates it later, then you call your **empty** activation function. As I said, I see two options:
1. If you don't want to remove the function, the you have to to the following
```C
{
  if (dev->r_type != RT_IOPORTS) {
    device_t *isab = dev->parent;
    return BUS_METHODS(isab->parent).activate_resource(isab, r);
  }
}
```
(As I mentioned in some previous comment, here we must do the explicit unwinding (not `bus_activate_resource`))
2. Just delegate the call to the parent.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,141 @@
...
+  return 0;
+}
+
+static void intel_isa_deactivate_resource(device_t *dev, resource_t *r) {
+  /* Neither IRQs in general or IOports on ISA require deactivation */
```
MichalBlk, line 66: The same problem as with the activation.

sys/mips/82371AB.c
```diff
@@ -0,0 +1,141 @@
...
+}
+
+static void intel_isa_release_resource(device_t *dev, resource_t *r) {
+  if (r->r_type != RT_IOPORTS) {
+    device_t *idev = BUS_METHOD_PROVIDER(dev->parent, release_resource);
```
MichalBlk, line 71: Why not just `bus_release_resource`?

MichalBlk (changes requested):
We're almost there.

sys/mips/82371AB.c
mohrcore: Given the past issues with `bus_activate_resource` and `bus_deactivate_resource` I consider these wrappers to be untrustworthy when used in dispatch scenarios inside drivers.

sys/mips/82371AB.c
mohrcore: ISA devices are not supposed to allocate memory under current circumstances and allowing them to do so was not a scope of this driver, so I didn't bother with writing what would be a dead code.

In fact, what you propose is not just unnecessary - it's an inherently flawed solution. Even if we were to introduce memory allocations for ISA devices, no dispatching should be done. All procedures for managing memory resources should be handled locally within ISA driver, just like IO ports are. Dispatching activation procedures to another bus risks messing up a configuration for a device that's directly underneath it. To be precise - we don't want gt64120 driver to mess with any BARs when activating memory for ISA devices (Actually I'm pretty sure it would just panic on one of the asserts inside pci driver, because ISA bus is not a PCI device and doesn't have any BARs).

sys/mips/82371AB.c
mohrcore: Yes, and my answer is the same too.

sys/mips/82371AB.c
mohrcore: Same reaso as with alloc procedure.

sys/mips/82371AB.c
mohrcore: Fixed.

sys/mips/82371AB.c
MichalBlk: If we don't allow ISA devices to allocate memory resources, then please add `assert(r->r_type != RT_MEMORY)` in activation/deactivation/release and `assert(type != RT_MEMORY)` in allocation.

mohrcore: Here you go. I added some asserts. I do believe tho that adding them to activation/deactivation procedures was an overkill as allocating memory from ISA bus is impossible anyway.

MichalBlk (approved):
LGTM

ThomasStachovsky (approved):
It looks okay. One thing that will probably need to be changed later is that the driver takes the whole interval of addresses from pci when in reality a few of them will be used by the ide driver. But that's a task for another pull request.

cahirwpz (approved):
LGTM I'm happy to see this change in `master` finally...

PR #875, opened 2020-12-12
Multiple independent driver interfaces

mohrcore:
This small PR allows drivers to implement any combination of interfaces.

Currently we have only two interfaces (_bus_ and _PCI_), however we will get more of them, eg. _block device_ and perhaps _ICU_.

The reason for which I'm making this PR now is because I'll be needing it soon.


include/sys/bus.h
```diff
@@ -151,16 +151,18 @@ struct bus_driver {
   bus_methods_t bus;
 };
 
-#define BUS_METHODS(dev) ((bus_driver_t *)((dev)->driver))->bus
+#define bus_get_interface(dev)                                                 \
```
komik0, line 154: I am not sure what's correct but it seems that in general we use capital letters to name __#define__ functions
```
~/mimiker$ sum=0; for c in {a..z}; do y=`grep -Ir "#define $c" --exclude-dir=toolchain | wc -l`; sum=$((sum+y)); done; echo $sum
3202
```
```
~/mimiker$ sum=0; for c in {A..Z}; do y=`grep -Ir "#define $c" --exclude-dir=toolchain | wc -l`; sum=$((sum+y)); done; echo $sum
18058
```

It is just a remark

sys/mips/gt64120.c
```diff
@@ -466,6 +466,10 @@ pci_bus_driver_t gt_pci_bus = {
     .size = sizeof(gt_pci_state_t),
     .attach = gt_pci_attach,
     .probe = gt_pci_probe,
+    .interfaces = {
+      INTERFACE(struct pci_bus_driver, DEV_INTERFACE_BUS, bus),
```
komik0, line 470: INTERFACE(struct bus_driver, DEV_INTERFACE_BUS, bus),

include/sys/bus.h
mohrcore: There are plenty of non-capital defines as well usually used for macros that behave like functions but are not worth even writing an inline static function. Eg. `bus_write_1`.

sys/mips/gt64120.c
mohrcore: I take it you suggest to change `struct pci_bus_driver` to `struct bus_driver`, right?
Why? With this new interface it's purely incidental that the offset of the `bus` field of `bus_driver` matches with with the one in `pci_bus_driver_t`.

mohrcore:
I made some changes (aba30d3), probably while @komik0 was reviewing my PR.

mohrcore:
I'm thinking about completely dropping driver sub-classes like `bus_driver_t` and `pci_bus_driver_t` an instead of that just go with per-driver type definitions. The only reason for exposing them was to make use of simple inheritance pattern which, as for this PR, is no longer the way driver interfaces work.

sys/mips/gt64120.c
komik0: In my opinion it much better to change INTERFACE - to make it simpler and easier to read.
- the first what has been confusing (for me) is the argument order (I think the field like pci_bus, bus should be at the beginning) because other arguments are almost meaningless for reader.

from my point of view it would be much better when the driver interfaces are moved outside and sub-structs are just registered in ```.interface```
I mean something like
```
 bus_interface... bus_if = {
    .intr_setup = gt_pci_intr_setup,
    .intr_teardown = gt_pci_intr_teardown,
    .alloc_resource = gt_pci_alloc_resource,
    .release_resource = gt_pci_release_resource,
    .activate_resource = gt_pci_activate_resource,
    .deactivate_resource = gt_pci_deactivate_resource,
}

pci_interface... pci_bus_if = {
    .read_config = gt_pci_read_config,
    .write_config = gt_pci_write_config,
    .enable_busmaster = gt_pci_enable_busmaster,
}

pci_bus_driver_t gt_pci_bus = {
  .driver = {
    .desc = "GT-64120 PCI bus driver",
    .size = sizeof(gt_pci_state_t),
    .attach = gt_pci_attach,
    .probe = gt_pci_probe,
    .interfaces = {
      INTERFACE(DEV_INTERFACE_BUS, &bus_if),
      INTERFACE(DEV_INTERFACE_PCI, &pci_bus_if)
    }
}
```

imo it is better - because in this approach you drop all hard-to-read logic like ```offsetof```
because it is only philosophicalI - I don't insist. If you want to proceed with this approach I will give you aproval.
could you please let me know when this cs is ready for review. (or to label this PR as WIP)

komik0: Maybe I don't see something obvious why it cannot be implemented in this way - please correct me if I am wrong

include/sys/bus.h
komik0: I agree - it is just a remark

sys/aarch64/rootdev.c
```diff
@@ -290,15 +290,17 @@ static void rootdev_deactivate_resource(device_t *dev, res_type_t type,
   /* TODO: unmap mapped resources. */
 }
 
-/* clang-format off */
-static bus_driver_t rootdev_driver = {
+static struct rootdev_driver {
+  driver_t driver;
+  bus_methods_t bus;
+} rootdev_driver = {
```
komik0, line 296: NIT:
imo, it is easier to read when definition and declarations are separated - I mean:

```
typedef struct  {
  driver_t driver;
  bus_methods_t bus;
} rootdev_driver_t;

static rootdev_driver_t rootdev_driver = {
```
or something

sys/mips/gt64120.c
mohrcore: The reason why I did it they way I did was because it's a slightly more general solution that works also for dynamically-allocated structures. I also liked having those interfaces in the same block and because at the time it was compatible with the inheritance-based solution (technically it still is in terms of structure layout but I removed sub-type definitions).

It's also _possibly_ slightly faster than using references, as all the data will most likely be in one block, however I doubt the compiler would put these structs far apart of each other and even if it did it would probably bear no real significance.

cahirwpz (commented):
I'm afraid that you introduced extra complexity without removing existing one. This is an unwanted move, but hopefully we will resolve that during Monday's meeting.

sys/aarch64/rootdev.c
```diff
@@ -290,22 +290,22 @@ static void rootdev_deactivate_resource(device_t *dev, res_type_t type,
...
+  .activate_resource = rootdev_activate_resource,
+  .deactivate_resource = rootdev_deactivate_resource,
+};
+
+static driver_t rootdev_driver = {
+  .size = sizeof(rootdev_t),
+  .desc = "RPI3 platform root bus driver",
+  .attach = rootdev_attach,
+  .interfaces = {
+    [DIF_BUS] = &rootdev_bus_if,
```
komik0, line 307: NIT: I am not sure if it is better to add functions like:
```
____ bus_register_interface(driver_t drv)
____ pci_bus_register_interface(driver_t drv)
```
and to register that interfaces in driver's attach function - I think it is more linux-style

komik0 (approved):
LGTM

sys/aarch64/rootdev.c
cahirwpz: FTTB I don't think I can come up with use case for such interface.

include/sys/device.h
```diff
@@ -22,12 +22,20 @@ typedef int (*d_probe_t)(device_t *dev);
...
+  DIF_COUNT /* this must be the last item */
+} drv_if_t;
+
 struct driver {
   const char *desc;  /* short driver description */
   size_t size;       /* device->state object size */
   d_probe_t probe;   /* probe for specific device(s) */
   d_attach_t attach; /* attach device to system */
   d_detach_t detach; /* detach device from system */
+  void *interfaces[DIF_COUNT];
```
MichalBlk, line 38: Maybe add a brief comment.

MichalBlk (approved):
LGTM

include/sys/device.h
cahirwpz: Done.

PR #876, opened 2020-12-13
Delete red-black trees

bkjg:
Red-black trees inside vm_object didn't work properly so the time has come to remove them.


sys/kern/vm_object.c
```diff
@@ -14,38 +14,28 @@ static inline int vm_page_cmp(vm_page_t *a, vm_page_t *b) {
...
   vm_object_t *obj = pool_alloc(P_VMOBJ, M_ZERO);
   TAILQ_INIT(&obj->list);
-  RB_INIT(&obj->tree);
   mtx_init(&obj->mtx, 0);
   obj->pager = &pagers[type];
   obj->ref_counter = 1;
   return obj;
 }
 
-void vm_object_free(vm_object_t *obj) {
```
cahirwpz, line 30: Please do not move the code around if it isn't strictly necessary. That artificially increases amount of changes and confuses a reader! I initially though that you remove `vm_object_free`. Please **always** minimize the changes.

sys/kern/vm_object.c
```diff
@@ -57,34 +47,39 @@ bool vm_object_add_page(vm_object_t *obj, off_t offset, vm_page_t *page) {
...
 
-  return false;
+  /* offset of page is greater than the offset of any other page */
+  TAILQ_INSERT_TAIL(&obj->list, page, obj.list);
+  obj->npages++;
+
+  return true;
 }
 
-static void vm_object_remove_page_nolock(vm_object_t *obj, vm_page_t *page) {
```
cahirwpz, line 74: I think in other places we do use `_nolock` convention. Please revert that change.

cahirwpz (changes requested):
As mentioned earlier I expect students to introduce changes that are strictly necessary to complete a task.

While removing the use of rb-trees from `vm_object` was a good move, I don't think we need to remove `tree.h` now – that would be premature. There's slight chance that some tools in user-space will need that header file as well.

sys/kern/vm_object.c
bkjg: In this place moving the code was strictly necessary - to not copy the same code, now in vm_object_free we are calling the vm_object_remove_page_nolock function. Previously, vm_object_free was above this additional function , so the code wouldn't compile, so I had to move the code to some lower position to be below it.

sys/kern/vm_object.c
bkjg: done

PR #877, opened 2020-12-13
Implement locking mechanism for pv_list in vm_page_t structure

bkjg:
Now pv_list in vm_page_t structure isn't protected by any mechanism. That can lead to many races when we will try to share our resources. This PR will introduce lock for all pv_list to protect our data and in the same time - to save some memory by using only one lock for all pv_list data structures.


cahirwpz (approved):
EDIT: removed duplicate review.

cahirwpz (approved):
LGTM Although a comment would be welcome here. There's no mention that the lock actually protects `vm_page::pv_list` which may elude some readers.

PR #878, opened 2020-12-13
Enable verify_formatting for AArch64

pj1031999:
Enable _verify_formatting_ for _AArch64_.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,9 +8,14 @@ defaults: &defaults
 jobs:
   verify_formatting:
     <<: *defaults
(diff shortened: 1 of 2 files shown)


cahirwpz (changes requested):
While the change does solve the problem in CI, it doesn't help a developer. I expect that running `make format` locally would solve the problem for both _AArch64_ and _MIPS_. With that done you wouldn't have to modify CI scripts at all.

cahirwpz:
Since we switched to _GH Actions_ this PR is no longer valid.

PR #879, opened 2020-12-14
Implement do_setcontext for AArch64.

pj1031999:
Implement `do_setcontext` for _AArch64_.


cahirwpz (approved):
LGTM But please introduce similar changes for MIPS counterpart.

PR #882, opened 2020-12-15
tty: notify driver when there's space in the input queue.

j-piecuch:
This PR kicks off a series of (hopefully small) PRs with an end goal of implementing pseudoterminals.
Why we need this PR:
`write()` calls on a PTY master device can block due to the slave device's input queue being full. We need some means of notifying the master that space has become available.

### DIFF
--- a/sys/kern/tty.c
+++ b/sys/kern/tty.c
@@ -169,7 +169,7 @@ tty_t *tty_alloc(void) {
   tty->t_line.ln_size = LINEBUF_SIZE;
   tty->t_line.ln_count = 0;
   tty_init_termios(&tty->t_termios);
-  tty->t_ops.t_notify_out = NULL;
+  tty->t_ops = (ttyops_t){0};
   tty->t_data = NULL;
   tty->t_column = 0;
   tty->t_rocol = tty->t_rocount = 0;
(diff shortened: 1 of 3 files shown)


wiklam (approved):
LGTM

sys/kern/tty.c
```diff
@@ -493,17 +495,28 @@ void tty_input(tty_t *tty, uint8_t c) {
...
+      return false;
     }
 
     ringbuf_putb(&tty->t_inq, c);
     tty_wakeup(tty);
+    return true;
+  }
+}
+
+static void tty_check_in_lowat(tty_t *tty) {
```
cahirwpz, line 512: Could you adda comment that `&tty->t_lock` should be held ?

sys/kern/tty.c
```diff
@@ -493,17 +495,28 @@ void tty_input(tty_t *tty, uint8_t c) {
...
+  if (!(tty->t_flags & TF_IN_HIWAT))
     return;
+
+  if (tty->t_inq.count < TTY_IN_LOW_WATER) {
+    tty->t_flags &= ~TF_IN_HIWAT;
+    if (tty->t_ops.t_notify_in)
```
cahirwpz, line 518: Consider introducing `tty` method wrapper as mentioned in the other review.

cahirwpz (approved):
LGTM though two small improvements could be added.

sys/kern/tty.c
j-piecuch: Done

sys/kern/tty.c
j-piecuch: I have added an assertion instead.

PR #883, opened 2020-12-15
Implement removing devfs nodes.

j-piecuch:
PTY slave devices must be removed when the corresponding master device is closed. For this, it's necessary to be able to remove devfs nodes.

### DIFF
--- a/sys/kern/devfs.c
+++ b/sys/kern/devfs.c
@@ -81,20 +81,24 @@ static int devfs_add_entry(devfs_node_t *parent, const char *name,
   return 0;
 }
 
-int devfs_remove(devfs_node_t *dn) {
+int devfs_unlink(devfs_node_t *dn) {
   SCOPED_MTX_LOCK(&devfs.lock);
 
   assert(dn->dn_parent != NULL);
 
   /* Only allow removal of empty directories. */
   if (dn->dn_vnode->v_type == V_DIR && !TAILQ_EMPTY(&dn->dn_children))
-    return EINVAL;
+    return ENOTEMPTY;
 
   TAILQ_REMOVE(&dn->dn_parent->dn_children, dn, dn_link);
   vnode_drop(dn->dn_vnode);
+  return 0;
+}
+
+void devfs_free(devfs_node_t *dn) {
+  assert(dn->dn_vnode->v_usecnt == 0);
   kfree(M_STR, dn->dn_name);
   kfree(M_DEVFS, dn);
-  return 0;
 }
 
 int devfs_makedev(devfs_node_t *parent, const char *name, vnodeops_t *vops,
--- a/sys/kern/devfs.c
+++ b/sys/kern/devfs.c
@@ -11,6 +11,7 @@
 #include <sys/dirent.h>
 #include <sys/vfs.h>
 #include <sys/queue.h>
+#include <sys/stat.h>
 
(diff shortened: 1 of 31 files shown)


sys/kern/devfs.c
```diff
@@ -80,8 +81,24 @@ static int devfs_add_entry(devfs_node_t *parent, const char *name,
...
+  assert(dn->dn_parent != NULL);
+
+  /* Only allow removal of empty directories. */
+  if (dn->dn_vnode->v_type == V_DIR && !TAILQ_EMPTY(&dn->dn_children))
+    return EINVAL;
+
+  TAILQ_REMOVE(&dn->dn_parent->dn_children, dn, dn_link);
+  vnode_drop(dn->dn_vnode);
+  kfree(M_STR, dn->dn_name);
+  kfree(M_DEVFS, dn);
```
xThaid, line 96: Shouldn't we wait with freeing memory until the reference counter of `vnode` hits 0 (or equally, until `VOP_RECLAIM` is called)? That dev file can possibly be opened in some processes.

j-piecuch: Yes, the file could still be opened, but the devfs node is not accessible from the vnode (at least not directly), so we're good as long as the device driver doesn't access the devfs node, which it shouldn't, since it presumably called `devfs_remove()` in the first place.

j-piecuch: The goal of `devfs_remove` is to remove the device from the filesystem, so that it's no longer visible. The device driver implements `VOP_RECLAIM()` and frees its data structures there.

xThaid: I'm still concerned about that situation:
1. a userspace program opens a device
2. then the device is removed by the kernel
3. user does `fstat(fd, ..)` on that opened device
4. fstat accesses the fields in devfs node, but it is already freed

j-piecuch: I think I see what you mean: the devfs node has some attributes (e.g. owner's uid and gid) that need to be accessible as long as the file descriptor is open. You're right: we can't free the devfs node until the vnode's refcount drops to 0 - we need to do it in `VOP_RECLAIM()`.

j-piecuch: ...which poses some problems, as device drivers overwrite the vnode's vnodeops with their own implementation... so every device driver that supports device removal must implement `VOP_RECLAIM()` and free the devfs node there.

j-piecuch: Ok, I have now separated unlinking the devfs node from the tree and freeing it.

xThaid (approved):
Looks good.

franciscozdo (approved):
LGTM

sys/kern/tty.c
```diff
@@ -944,9 +929,7 @@ static int dev_tty_open(vnode_t *v, int mode, file_t *fp) {
   return error;
 }
 
-vnodeops_t dev_tty_vnodeops = {.v_open = dev_tty_open,
-                               .v_getattr = tty_vn_getattr,
-                               .v_access = vnode_access_generic};
+vnodeops_t dev_tty_vnodeops = {.v_open = dev_tty_open};
```
cahirwpz, line 932: Should that not be `static`?

j-piecuch: Sure, why not. It's not exposed in the header anyway

cahirwpz (approved):

PR #884, opened 2020-12-16
Track TTY open() and close() calls.

j-piecuch:
- Add an open count to the `tty_t` structure
- Notify the driver when the open count goes from 0 to 1 and from 1 to 0
- Unify common parts of `tty_vn_open()` and `dev_tty_open()`
- Add helper function for making TTY devfs entries

### DIFF
--- a/sys/kern/tty.c
+++ b/sys/kern/tty.c
@@ -909,7 +909,7 @@ static void maybe_assoc_ctty(proc_t *p, tty_t *tty) {
   tty->t_pgrp = p->p_pgrp;
 }
 
-static int __tty_vn_open(vnode_t *v, int mode, file_t *fp) {
+static int _tty_vn_open(vnode_t *v, int mode, file_t *fp) {
   tty_t *tty = devfs_node_data(v);
   proc_t *p = proc_self();
   int error;
@@ -933,7 +933,7 @@ static int __tty_vn_open(vnode_t *v, int mode, file_t *fp) {
 
 static int tty_vn_open(vnode_t *v, int mode, file_t *fp) {
   SCOPED_MTX_LOCK(all_proc_mtx);
-  return __tty_vn_open(v, mode, fp);
+  return _tty_vn_open(v, mode, fp);
 }
 
 static vnodeops_t tty_vnodeops = {.v_open = tty_vn_open,
@@ -953,7 +953,7 @@ static int dev_tty_open(vnode_t *v, int mode, file_t *fp) {
   if (!tty)
     return ENXIO;
 
-  return __tty_vn_open(tty->t_vnode, mode, fp);
+  return _tty_vn_open(tty->t_vnode, mode, fp);
 }
 
 static vnodeops_t dev_tty_vnodeops = {.v_open = dev_tty_open};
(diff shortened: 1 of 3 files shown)


franciscozdo (approved):
LGTM

wiklam (approved):
LGTM

sys/kern/tty.c
```diff
@@ -737,7 +738,12 @@ static void tty_drain_out(tty_t *tty) {
 }
 
 static int tty_vn_close(vnode_t *v, file_t *fp) {
-  /* TODO implement. */
+  tty_t *tty = v->v_data;
+  SCOPED_MTX_LOCK(&tty->t_lock);
+  assert(tty->t_opencount > 0);
+  tty->t_opencount--;
+  if (tty->t_opencount == 0 && tty->t_ops.t_notify_inactive)
```
cahirwpz, line 745: Please wrap calls to methods in following manner:

```c
static inline void tty_notify_inactive(tty_t *tty) {
  if (tty->t_ops.t_notify_inactive)
    tty->t_ops.t_notify_inactive(tty);
}
```

sys/kern/tty.c
```diff
@@ -876,57 +882,67 @@ static fileops_t tty_fileops = {
...
+  session_t *s = p->p_pgrp->pg_session;
+  if (s->s_tty != NULL)
+    return;
+  tty->t_session = s;
+  s->s_tty = tty;
+  tty->t_pgrp = p->p_pgrp;
 }
 
-static int tty_vn_open(vnode_t *v, int mode, file_t *fp) {
+static int __tty_vn_open(vnode_t *v, int mode, file_t *fp, bool assoc) {
```
cahirwpz, line 900: Double underscore is usually used for functions and symbols provided by compiler and linker. Single underscore is ok though.

cahirwpz (changes requested):
As you already know I'm always hesitant about using `bool` as an argument to a procedure. To me it  indicates that control flow could possibly be decoupled for two similar but different cases. Could you use other information than `assoc` argument to perform the same task? Perhaps `maybe_assoc_ctty` could detect the case for its own.

Observation: you make a decision somewhere down the call chain and then you apply it in 
`__tty_vn_open`. What prevents you from doing both in the same place?

To summarize: I don't understand why there's a subtle difference between `dev_tty_open` and `tty_vn_open` and I guess it will confuse other readers too.

sys/kern/tty.c
j-piecuch: I don't know, functions prefixed with double underscores are very common in Linux

sys/kern/tty.c
j-piecuch: Done

sys/kern/tty.c
j-piecuch: I changed it to a single underscore.

j-piecuch:
@cahirwpz The `assoc` parameter in `__tty_vn_open()` was there to avoid the redundant call to `maybe_assoc_ctty()`, since when opening `/dev/tty` we know that the terminal is our controlling terminal. It was a questionable optimization, since calling `maybe_assoc_ctty()` is not expensive at all. I have removed the `assoc` parameter.

cahirwpz (approved):
LGTM

cahirwpz:
Ops... please resolve conflicts.

PR #885, opened 2020-12-16
Check permissions during lookup

franciscozdo:


include/sys/vfs.h
```diff
@@ -100,21 +102,22 @@ int do_fchown(proc_t *p, int fd, uid_t uid, gid_t gid);
 int do_fchownat(proc_t *p, int fd, char *path, uid_t uid, gid_t gid, int flag);
 
 /* Mount a new instance of the filesystem named fs at the requested path. */
-int do_mount(const char *fs, const char *path);
```
j-piecuch, line 103: Maybe change the signature to `do_mount(proc_t *p, const char *fs, const char *path)` for the sake of consistency with the functions below?

j-piecuch (approved):
One nitpick, otherwise LGTM

xThaid (approved):

include/sys/vfs.h
franciscozdo: Okay. I was considering this and if you think i would be better it is probably the right way.

PR #886, opened 2020-12-17
Reclaim devfs directory nodes.

j-piecuch:
There was an oversight in #883: when a devfs directory node is unlinked, the devfs layer is responsible for freeing the directory node after all references to the vnode are gone (i.e. in `VOP_RECLAIM()`), but we didn't provide an implementation of `VOP_RECLAIM()`.
I have also thrown in a devfs directory removal test.


xThaid (approved):

franciscozdo (approved):
LGTM

cahirwpz (approved):
LGTM

PR #888, opened 2020-12-17
ICU interface and searching

mohrcore:
Made a separate interface for ICUs. Also ICU methods now search the tree upwards until they find the first device with a driver that implements ICU interface.


include/sys/icu.h
```diff
@@ -0,0 +1,36 @@
...
+} icu_methods_t;
+
+#define ICU_METHODS(dev) (*(icu_methods_t *)(dev)->driver->interfaces[DIF_ICU])
+
+static inline void icu_intr_setup(device_t *dev, resource_t *irq,
+                                  ih_filter_t *filter, ih_service_t *service,
+                                  void *arg, const char *name) {
+  assert(dev->parent);
+  if (!dev->parent->driver->interfaces[DIF_ICU]) {
+    icu_intr_setup(dev, irq, filter, service, arg, name);
```
MichalBlk, line 21: It's broken and leads to an infinite loop. You probably want to call `icu_intr_setup(dev->parent, irq, filter, service, arg, name)`, but it's not that simple. For now, rootdev and GT64120 assume that a device which requests an interrupt setup is a direct child of the bus, eg:
```C
gt_pci_intr_setup(...) {
  assert(dev->parent->driver == &gt_pci_bus);
  ...
}
rootdev_intr_setup(...) {
  rootdev_t *rd = dev->parent->state;
  ...
}
```

include/sys/icu.h
```diff
@@ -0,0 +1,36 @@
...
+    return;
+  }
+  ICU_METHODS(dev->parent).intr_setup(dev, irq, filter, service, arg, name);
+}
+
+static inline void icu_intr_teardown(device_t *dev, resource_t *irq) {
+  assert(dev->parent);
+  if (!dev->parent->driver->interfaces[DIF_ICU]) {
+    icu_intr_teardown(dev, irq);
```
MichalBlk, line 30: Same as above. Infinite loop.

include/sys/icu.h
```diff
@@ -0,0 +1,36 @@
...
+  void (*intr_teardown)(device_t *dev, resource_t *irq);
+} icu_methods_t;
+
+#define ICU_METHODS(dev) (*(icu_methods_t *)(dev)->driver->interfaces[DIF_ICU])
+
+static inline void icu_intr_setup(device_t *dev, resource_t *irq,
+                                  ih_filter_t *filter, ih_service_t *service,
+                                  void *arg, const char *name) {
+  assert(dev->parent);
+  if (!dev->parent->driver->interfaces[DIF_ICU]) {
```
MichalBlk, line 20: Maybe add something like 
```C
#define DIF_IMPLEMENST(dev, interface) ((dev)->interfaces[(interface)] != NULL)
```

MichalBlk (changes requested):

include/sys/icu.h
mohrcore: You are right! I must've accidentally change it from `dev->parent` to `dev` when I was willing to change something else. I'll correct it soon.

include/sys/icu.h
mohrcore: It's not a secret interfaces are stored simply as fields of an array, as simple array initialization is the way they are being assigned to drivers. I think that such macro would only obfuscate the code. Checking for null pointer is a very common pattern in C.

I believe the most elegant solution would be to change `..._METHODS` macros so they wouldn't dereference structs. Then, this check could be converted to something like
```
if(!ICU_METHODS(dev->parent))
```

Tbh I never liked that dereferencing inside those macros. It just seemed weird as most of the dereferencing in the code base seems to be done explicitly.

MichalBlk: I agree. We should change `..._METHODS` definitions. I also think that the following control flow is more coherent
```C
icu_intr_...(dev, ...) {
  device_t *parent = dev->parent;
  assert(parent);
  if (!ICU_METHODS(parent))
    icu_intr_...(parent, ...);
  else
    ICU_METHODS(parent)->intr_...(dev, ...);
}
```
This removes the inner `return` instruction.

include/sys/device.h
```diff
@@ -26,6 +26,7 @@ typedef int (*d_detach_t)(device_t *dev);
 typedef enum {
   DIF_BUS,
   DIF_PCI_BUS,
+  DIF_ICU,
```
komik0, line 29: could you please expand ICU acronym in comment?

sys/mips/rootdev.c
```diff
@@ -173,9 +174,12 @@ static int rootdev_attach(device_t *bus) {
   return bus_generic_probe(bus);
 }
 
-static bus_methods_t rootdev_bus_if = {
+static icu_methods_t rootdev_icu_if = {
```
komik0, line 177: could you please help me to understand why do we want to have intr methods in separate struct?

mohrcore: Because we are going to have bus drivers which do not manage any ICUs. ISA is going to have a driver separate from the PCI one, yet they share the same interrupt controller unit. Either GT PCI driver will reamain responsible for managing ICU state and any interrupt-related routines from ISA devices will be propagated up to it, or the whole logic will be moved into ISA driver, which would leave PCI without the ability to manage interrupts.

TLDR; there's no one-to-one relation between Interrupt controllers and buses.

include/sys/device.h
mohrcore: ok.

sys/mips/rootdev.c
komik0: ok, thank you for the explanation

include/sys/device.h
mohrcore: Done.

include/sys/icu.h
mohrcore: Fixed.

include/sys/icu.h
mohrcore: Fixed.

include/sys/icu.h
mohrcore: I changed `..._METHODS` macros which made the check cleaner.

include/sys/device.h
```diff
@@ -26,7 +26,7 @@ typedef int (*d_detach_t)(device_t *dev);
 typedef enum {
   DIF_BUS,
   DIF_PCI_BUS,
-  DIF_ICU,
+  DIF_ICU,  /* Interrupt Controller Unit *
```
komik0, line 29: ```*/```

mohrcore:
I rewrote `icu_intr_setup` and `icu_intr_teardown` procedures so now they are iterative instead of being recursive. Disassembling the binaries prior to that change revealed that there seems to be no tail call optimisation applied. This allowed for a potential stack overflow in kernel mode, which is obviously unacceptable. Also it nullified inlining.

include/sys/device.h
komik0: thanks

include/sys/device.h
mohrcore: Fixed.

ThomasStachovsky (approved):
It looks okay.

include/sys/device.h
```diff
@@ -26,6 +26,7 @@ typedef int (*d_detach_t)(device_t *dev);
 typedef enum {
   DIF_BUS,
   DIF_PCI_BUS,
+  DIF_ICU,  /* Interrupt Controller Unit */
```
cahirwpz, line 29: More commonly used name for that is _Programmable Interrupt Controller_ (PIC for short). 
https://en.wikipedia.org/wiki/Programmable_interrupt_controller

include/sys/icu.h
```diff
@@ -0,0 +1,34 @@
...
+
+typedef struct icu_methods {
+  void (*intr_setup)(device_t *dev, resource_t *irq, ih_filter_t *filter,
+                     ih_service_t *service, void *arg, const char *name);
+  void (*intr_teardown)(device_t *dev, resource_t *irq);
+} icu_methods_t;
+
+#define ICU_METHODS(dev) ((icu_methods_t *)(dev)->driver->interfaces[DIF_ICU])
+
+static inline void icu_intr_setup(device_t *dev, resource_t *irq,
```
cahirwpz, line 16: I don't think this is a good candidate for `static inline` function. If it has more complex control flow than just single `if` statement it should be moved to `.c` file.

include/sys/bus.h
```diff
@@ -145,17 +142,7 @@ struct bus_methods {
...
-static inline void bus_intr_setup(device_t *dev, resource_t *irq,
-                                  ih_filter_t *filter, ih_service_t *service,
-                                  void *arg, const char *name) {
-  BUS_METHODS(dev->parent).intr_setup(dev, irq, filter, service, arg, name);
-}
-
-static inline void bus_intr_teardown(device_t *dev, resource_t *irq) {
-  BUS_METHODS(dev->parent).intr_teardown(dev, irq);
-}
+#define BUS_METHODS(dev) ((bus_methods_t *)(dev)->driver->interfaces[DIF_BUS])
```
cahirwpz, line 145: I agree this is a better approach, but still you should not tackle two issues in the same PR. Please split such changes in the future.

cahirwpz (changes requested):
There's a serious architectural problem with the changes you propose. Since we want to decouple `pic` from `bus` (and I'm in favour of doing that) I'd expect interrupt resource allocation and so on should be handled by `pic` code and not `bus`. Obviously `gtpci` driver need to delegate these tasks to `pic` controller, but PCI interrupt routing is somewhat specific to `gtpci` in this case (ask @komik0 about it). We should talk it through during next meeting.

BTW Have you had a look at https://grok.dragonflybsd.org/xref/freebsd/sys/kern/pic_if.m ?

mohrcore:
Uh, so the solution we came up with was to make dynamic dispatch more granular, ie. make ALL methods dispatchable instead of dispatching interfaces in some method wrappers. However there's another problem with this solution. There's a tremendous amount of code that relies on an assumption that the `dev` parameter of driver methods is a direct child of the device that uses the driver. Unfortunately, allowing every method to be dispatched dynamically means that a lot of code that should be rewritten, even tho it's gonna do exactly the same as before and the whole dispatch part of it will be a sort of dead code.

However, I'm in favor of such solution anyway as passing a child as a `dev` parameter feels extremely awkward, especially now that we are going to have a dispatching system.

I think the signatures of some methods could be changed to make `dev` the device that actually uses the driver and some methods should have additional `target` parameter if they (may) need to know what device has a driver that made the call (eg. determine PCI bar for some PCI bus methods or or interrupt handlers that do some stuff with the device). The wrappers (such as `bus_intr_setup`) would still be used in the same way, as they would perform dynamic dispatch starting from `dev`'s parent.

That's at least the solution I'm working on right now. Since there are a lot of changes involved here, I'd appreciate some opinions on it.

mohrcore:
I'm closing this PR as the problem it addressed has been solved in #906. We chose a different approach and instead of dispatching interfaces we dispatch methods.
###### _(which is an approach I don't like, but whatever, it's good enough and allows PIC methods to be a part of bus interface, just like in FreeBSD)_

PR #890, opened 2020-12-19
RTL8139: first version of rx-interrupt handler

komik0:
This cs alloc irq resource and outlines rx-interrupt handler


sys/drv/rtl8139.c
```diff
@@ -32,18 +37,72 @@ static int rtl_reset(rtl8139_state_t *state) {
...
     return ENXIO;
   }
 
+  /* set-up address for dma */
+  bus_write_4(state->regs, RL_RXADDR, state->paddr);
+  /* we want to receive all packets - promiscious mode */
+  /* 0xf - accept broadcast, multicast, physical match, all other packets
+   * (promisc) */
+  /* 0x80 - wrap bit */
+  bus_write_4(state->regs, RL_RXCFG, 0x8f);
```
pj1031999, line 101: Try to avoid magic constants in code. Please define wrap and promisc flags.

sys/drv/rtl8139.c
```diff
@@ -32,18 +37,72 @@ static int rtl_reset(rtl8139_state_t *state) {
   return 1;
 }
 
-static int rtl8139_attach(device_t *dev) {
+static intr_filter_t rtl8139_intr(void *data) {
+  rtl8139_state_t *state = data;
+  // TODO: add enqueue routines
+  bus_write_2(state->regs, RL_ISR, RL_ISR_RX_OK);
+  return IF_FILTERED;
+}
```
pj1031999, line 45: Do I understand correctly that for each interrupt you **only** write `RL_ISR_RX_OK`?
Shouldn't we be doing something more with this data?

sys/drv/rtl8139.c
```diff
@@ -32,18 +37,72 @@ static int rtl_reset(rtl8139_state_t *state) {
...
+  device_add_irq(dev, 0, 10);
+  state->irq_res = device_take_irq(dev, 0, RF_ACTIVE);
+  if (!state->regs || !state->irq_res) {
+    klog("rtl8139 failed to init resources");
     return ENXIO;
+  }
+  bus_intr_setup(dev, state->irq_res, rtl8139_intr, NULL, state, "RTL8139");
+
+  // TODO: mark this memory as PMAP_NOCACHE
+  state->rx_buf = (vaddr_t)kmem_alloc(2 * PAGESIZE, M_ZERO);
```
pj1031999, line 83: Something like `RX_BUF_SIZE` is more readable than `2 * PAGESIZE` for me.

sys/drv/rtl8139.c
komik0: you are right, all expected routines are implemented in #900 - I wanted to separate this PR into 2 PRs to be more readable (one for setting-up interrupts, 2nd for the implementation of rx-path)

sys/drv/rtl8139.c
```diff
@@ -32,18 +37,72 @@ static int rtl_reset(rtl8139_state_t *state) {
   return 1;
 }
 
-static int rtl8139_attach(device_t *dev) {
+static intr_filter_t rtl8139_intr(void *data) {
+  rtl8139_state_t *state = data;
+  // TODO: add enqueue routines
+  bus_write_2(state->regs, RL_ISR, RL_ISR_RX_OK);
+  return IF_FILTERED;
```
MichalBlk, line 44: It's a bad idea to always return `IF_FILTERED`. An interrupt event may consist of multiple handlers each of which is called during a call to `intr_event_run_handlers`. If the network card isn't the actual source of an interrupt, then you should return `IF_STRAY`.

sys/drv/rtl8139.c
```diff
@@ -32,18 +37,72 @@ static int rtl_reset(rtl8139_state_t *state) {
...
+
+int rtl8139_attach(device_t *dev) {
   rtl8139_state_t *state = dev->state;
 
+  pci_enable_busmaster(dev);
   state->regs = device_take_memory(dev, 1, RF_ACTIVE);
-  if (state->regs == NULL)
+
+  /*
+   * XXX: there is no irq routing in mimiker
```
MichalBlk, line 71: Due to #961 and #895, there is GT PCI interrupt routing, thus you can remove the comment and the call `device_add_irq`.

sys/drv/rtl8139.c
```diff
@@ -32,18 +37,72 @@ static int rtl_reset(rtl8139_state_t *state) {
...
+  // TODO: introduce ring buffer
 
   if (rtl_reset(state)) {
     klog("RTL8139: Failed to reset device!");
     return ENXIO;
   }
 
+  /* set-up address for dma */
+  bus_write_4(state->regs, RL_RXADDR, state->paddr);
+  /* we want to receive all packets - promiscious mode */
```
MichalBlk, line 97: I'd use a larger continuous comment.
```C
/* ...
 * ...
 */
```

sys/drv/rtl8139.c
```diff
@@ -32,18 +37,72 @@ static int rtl_reset(rtl8139_state_t *state) {
...
 
-static int rtl8139_attach(device_t *dev) {
+static intr_filter_t rtl8139_intr(void *data) {
+  rtl8139_state_t *state = data;
+  // TODO: add enqueue routines
+  bus_write_2(state->regs, RL_ISR, RL_ISR_RX_OK);
+  return IF_FILTERED;
+}
+
+static int set_mem_contig_address(rtl8139_state_t *state) {
```
MichalBlk, line 47: We must get rid of it in the future (in some driver infrastructure related PR).

MichalBlk (changes requested):
Few changes are required.

MichalBlk:
Pleas update the description.

sys/drv/rtl8139.c
komik0: Done

sys/drv/rtl8139.c
komik0: Done

sys/drv/rtl8139.c
komik0: in the current PR I will return IF_STRAY - it will be improved in further PR.  - done

sys/drv/rtl8139.c
komik0: Done

sys/drv/rtl8139.c
komik0: Done

MichalBlk (approved):
LGTM

MichalBlk:
It'd nice to update the description of this PR.

sys/drv/rtl8139.c
cahirwpz: This function is so ugly that I insist it should be done sooner than later.

sys/drv/rtl8139.c
```diff
@@ -32,18 +39,63 @@ static int rtl_reset(rtl8139_state_t *state) {
...
+    klog("rtl8139 failed to init resources");
+    return ENXIO;
+  }
+  bus_intr_setup(dev, state->irq_res, rtl8139_intr, NULL, state, "RTL8139");
+
+  // TODO: mark this memory as PMAP_NOCACHE
+  state->rx_buf = (vaddr_t)kmem_alloc(RX_BUF_SIZE, M_ZERO);
+  if (set_mem_contig_address(state)) {
+    klog("failed to alloc contig memory");
     return ENXIO;
```
pj1031999, line 83: Here you always allocate memory with success - it's ok, but in case of non contiguous allocation you get error - it's still ok - but you have memory leak.
Your log message is confusing here - "failed to alloc ... memory" and you return `ENXIO`.
- `ENXIO` - Device not configured
- `ENOMEM` - Cannot allocate memory

I would be grateful if you could make that part of your code more obvious for other people.

sys/drv/rtl8139.c
```diff
@@ -32,18 +39,63 @@ static int rtl_reset(rtl8139_state_t *state) {
...
+  }
+  // TODO: introduce ring buffer
 
   if (rtl_reset(state)) {
     klog("RTL8139: Failed to reset device!");
     return ENXIO;
   }
```
pj1031999, line 90: Memory leak?

komik0: thanks, done

sys/drv/rtl8139.c
komik0: @cahirwpz has improved this part of code with kmem_alloc_contig - thank you
@pj1031999 done - thanks

sys/drv/rtl8139.c
```diff
@@ -32,18 +39,63 @@ static int rtl_reset(rtl8139_state_t *state) {
...
+  // TODO: mark this memory as PMAP_NOCACHE
+  state->rx_buf = (vaddr_t)kmem_alloc(RX_BUF_SIZE, M_ZERO);
+  if (set_mem_contig_address(state)) {
+    klog("failed to alloc contig memory");
     return ENXIO;
+  }
+  // TODO: introduce ring buffer
 
   if (rtl_reset(state)) {
     klog("RTL8139: Failed to reset device!");
```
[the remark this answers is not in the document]
komik0: done

sys/drv/rtl8139.c
komik0: @cahirwpz thank you for `kmem_alloc_contig`

sys/drv/rtl8139.c
```diff
@@ -32,19 +39,55 @@ static int rtl_reset(rtl8139_state_t *state) {
...
 
+  /* set-up address for DMA */
+  bus_write_4(state->regs, RL_RXADDR, state->rx_buf_physaddr);
+  bus_write_4(state->regs, RL_RXCFG, RL_RXCFG_CONFIG);
+  bus_write_1(state->regs, RL_COMMAND, RL_CMD_RX_ENB | RL_CMD_TX_ENB);
+  /* IMR - mask rx enable */
+  bus_write_2(state->regs, RL_IMR, RL_ISR_RX_OK);
+
   return 0;
+error:
```
komik0, line 86: I placed it because of coming PR

sys/drv/rtl8139.c
```diff
@@ -32,19 +39,55 @@ static int rtl_reset(rtl8139_state_t *state) {
...
+  pci_enable_busmaster(dev);
 
   state->regs = device_take_memory(dev, 1, RF_ACTIVE);
-  if (state->regs == NULL)
+  state->irq_res = device_take_irq(dev, 0, RF_ACTIVE);
+  if (!state->regs || !state->irq_res) {
+    klog("Failed to init resources!");
     return ENXIO;
+  }
+  bus_intr_setup(dev, state->irq_res, rtl8139_intr, NULL, state, "RTL8139");
```
cahirwpz, line 61: You should call `bus_intr_teardown` if you return from `attach` with value `!= 0`.

komik0: is current solution acceptable?

cahirwpz (changes requested):
Please fix `rtl8139_attach`. You don't release allocated resources (memory, interrupt, etc.) in case return value is not 0.

sys/drv/rtl8139.c
cahirwpz: No.

cahirwpz (approved):
LGTM

PR #893, opened 2020-12-21
Allow for detaching serial device driver from TTY.

j-piecuch:
When the PTY master device file is closed, the corresponding slave device should disappear.
This PR implements the disappearing part.

### DIFF
--- a/sys/kern/tty.c
+++ b/sys/kern/tty.c
@@ -762,7 +762,8 @@ static int tty_do_write(tty_t *tty, uio_t *uio) {
   while (uio->uio_resid > 0) {
     if ((error = uiomove(&c, 1, uio)))
       break;
-    tty_output_sleep(tty, c);
+    if ((error = tty_output_sleep(tty, c)))
+      break;
     tty->t_rocount = 0;
   }
   tty_notify_out(tty);
--- a/sys/kern/tty.c
+++ b/sys/kern/tty.c
@@ -182,6 +182,11 @@ tty_t *tty_alloc(void) {
 void tty_free(tty_t *tty) {
   assert(!tty_opened(tty));
   assert(tty_detached(tty));
+  assert(!mtx_owned(&tty->t_lock));
+  mtx_destroy(&tty->t_lock);
+  cv_destroy(&tty->t_incv);
+  cv_destroy(&tty->t_outcv);
+  cv_destroy(&tty->t_serialize_cv);
   kfree(M_DEV, tty->t_line.ln_buf);
   kfree(M_DEV, tty->t_inq.data);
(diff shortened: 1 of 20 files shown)


sys/kern/tty.c
```diff
@@ -902,6 +962,54 @@ static int tty_ioctl(file_t *f, u_long cmd, void *data) {
   }
 }
 
+static void tty_hangup(tty_t *tty) {
+  assert(mtx_owned(&tty->t_lock));
+
+  if (!tty_opened(tty) || (tty->t_lflag & CLOCAL))
+    return;
```
wiklam, line 969: Could you provide a description why we check CLOCAL flag?

sys/kern/tty.c
```diff
@@ -179,6 +179,29 @@ tty_t *tty_alloc(void) {
   return tty;
 }
 
+void tty_free(tty_t *tty) {
+  assert(!tty_opened(tty));
+  assert(tty_detached(tty));
+  kfree(M_DEV, tty->t_line.ln_buf);
```
wiklam, line 185: It is not implemented yet, but `mtx_destroy` is a good idea.

sys/kern/tty.c
```diff
@@ -902,6 +962,54 @@ static int tty_ioctl(file_t *f, u_long cmd, void *data) {
...
+  if (!tty_opened(tty) || (tty->t_lflag & CLOCAL))
+    return;
+
+  session_t *s = tty->t_session;
+
+  /* Send SIGHUP to session leader.
+   * NOTE: access to s_leader is safe here,
+   * even though we're not holding all_proc_mtx. See session_leave(). */
+  if (s && s->s_leader) {
+    proc_t *leader = s->s_leader;
```
franciscozdo, line 977: I don't get it. Why `s->s_leader` cannot became `NULL` between this lines? Does only session leader can call this function? If not i think it is possible that this function is triggered by some process and `session_leave()` is called in session leader in the same time.

franciscozdo (approved):
Looks good, but I have one question about locking.

sys/kern/tty.c
franciscozdo: Okay, i missed one change. We are holding `tty->t_lock` here and in `session_leave()`. Now I see it.

sys/kern/tty.c
j-piecuch: Done.

sys/kern/tty.c
j-piecuch: Done.

wiklam (approved):
LGTM

cahirwpz (approved):
LGTM

PR #894, opened 2020-12-21
Properly handle stop signals in sigsuspend()

j-piecuch:
Stopping a process that's sleeping in `sigsuspend()` with a signal should not return `EINTR` to the user, but we currently do that.
In this PR we leverage `ERESTARTNOHAND` to restart the syscall if no signal with an installed handler is caught.


cahirwpz (approved):
LGTM

PR #898, opened 2020-12-22
Interrupt sharing.

MichalBlk:
Implement interrupt resource sharing.


sys/kern/rman.c
```diff
@@ -142,11 +150,33 @@ resource_t *rman_reserve_resource(rman_t *rm, rman_addr_t start,
...
+        r->r_sharelist = kmalloc(M_RES, sizeof(share_list_t), M_WAITOK);
+        TAILQ_INIT(r->r_sharelist);
+        TAILQ_INSERT_HEAD(r->r_sharelist, r, r_sharelink);
+      }
+
+      new->r_sharelist = r->r_sharelist;
+      TAILQ_INSERT_TAIL(r->r_sharelist, new, r_sharelink);
+
+      return new;
+    }
```
mohrcore, line 177: Can't it all be done in the previous loop?

mohrcore (changes requested):
I don't really understand what's the point of _share lists_ here.

You just allowed drivers to claim the same resource (I'm refering to the physical memory address range here, not its representation that is `resource_t`) multiple times if it's marked as `RF_SHAREABLE`. I don't see any actual use for those _share lists_ here. I guess that would be useful when checking for example wether the resource is active on any of sharer's `resource_t`s, but I don't see any of that here, all checks are local ie. they only check flags for single given `resource_t`.

Am I missing something here? If I'm not then I think those _share lists_ are completely unnecessary.

MichalBlk:
I made some major changes to the code regarding moving `auxiliary data` and `rid` to the `resource_list_entry` structure.

sys/kern/rman.c
```diff
@@ -97,6 +101,13 @@ void rman_fini(rman_t *rm) {
   /* TODO: destroy the `rm_lock` after implementing `mtx_destroy`. */
 }
 
+static int r_share_hit(resource_t *r, rman_addr_t start, size_t count,
+                       res_flags_t flags) {
+  size_t rsize = r->r_end - r->r_start + 1;
+  return r_shareable(r) && (flags & RF_SHAREABLE) && rsize == count &&
+         r->r_start <= start;
+}
```
ThomasStachovsky, line 109: What does 'count' mean? Note that if r_start is strictly smaller than start, than  start+count > r_end, because count == rsize.

sys/kern/rman.c
```diff
@@ -166,6 +182,9 @@ void rman_release_resource(resource_t *r) {
   assert(!r_active(r));
   assert(r_reserved(r));
 
+  if (r_shareable(r) && --r->r_refcnt)
+    return;
```
mohrcore, line 186: I think you might be abusing `rm_lock` to protect the reference counter. If it was only the counter that was shared between resource sharers, each sharer could hold its own `resource_t`, fields such as `rid` or `handler` wouldn't have to be pushed out of `resource_t` to `resource_list_entry_t` and decrementing `r_refcnt` wouldn't block the entire rman.

To be clear - I'm not 100% sure that this can be easily solved by adding a separate lock on `r_refcnt`, but it it can be solved that way, I'd be much more in favor of such solution.

mohrcore (changes requested):

sys/kern/rman.c
mohrcore: Ok, I just realized that in the previous solution share lists were present for this very reason to allow devices to hold their own `resource_t`s. They ensured that after releasing a resource rman would still hold some `resource_t` to mark the physical resource as claimed. Going back to having multiple `resource_t`s for one shared resource would be in fact going back to the previous solution.

After thinking about it for a while, I decided that this new approach is indeed much better after all! I just reimagined `resource_list_entry_t` to be named as something like `resource_handle_t` and it all made sense! There's a clear dichotomy ie. `resource_t` holds the data that defines the resource along with user count, and `resource_list_entry_t` hold any auxiliary data that a driver ties to the resource. I wonder if we could use `resource_list_entry_t`s as handles to resources iinstead of `resource_t *`s in drivers and just get rid of `resource_rle`. This feels like a natural step after moving all that auxiliary data to `resource_list_entry_t`.

However, all this weghing doesn't really affect my point about abusing `rm_lock`. Correct me, if I'm wrong, but I'm pretty sure that it would be a good idead to introduce a per-resource lock that would protect the reference counter. without locking the entire rman if not necessary.

sys/kern/rman.c
MichalBlk: It's broken. There should be `r->r_start >= start`.  'count' expresses the amount of memory we'd like to allocate.

sys/kern/rman.c
```diff
@@ -97,6 +101,13 @@ void rman_fini(rman_t *rm) {
   /* TODO: destroy the `rm_lock` after implementing `mtx_destroy`. */
 }
 
+static int r_share_hit(resource_t *r, rman_addr_t start, size_t count,
+                       res_flags_t flags) {
+  size_t rsize = r->r_end - r->r_start + 1;
+  return r_shareable(r) && (flags & RF_SHAREABLE) && rsize == count &&
+         r->r_start <= start;
```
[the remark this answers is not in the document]
MichalBlk: I was wrong. There should be  `r->r_start >= start`.

mohrcore (approved):
All my major issues with this code have been resolved. As for more granular locks, we agreed to let go of of this idea and maybe introduce it later if necessary.

The matter of making `resource_t` an internal structure of _rman_ and using `resource_list_entry_t` instead to reference resources in drivers can be addressed in a separate PR (maybe I'll do it).

So, keeping all this in mind, this is a LGTM from me.

MichalBlk:
The solution has been updated.

sys/kern/rman.c
```diff
@@ -53,10 +55,21 @@ static int r_reserved(range_t *r) {
...
+  return r->flags & RF_SHAREABLE;
+}
+
 static int r_canmerge(range_t *r, range_t *next) {
   return (r->end + 1 == next->start) && !r_reserved(next);
 }
 
+static int r_canshare(range_t *r, rman_addr_t start, size_t count,
+                      rman_flags_t flags) {
+  size_t rsize = r->end - r->start + 1;
```
cahirwpz, line 68: Maybe inline function `r_size` ?

cahirwpz (commented):
What about `resource_activate` & `resource_deactivate` - should we take `refcnt` for them into account as well ?

sys/kern/rman.c
MichalBlk: Done.

MichalBlk:
I don't think that `resource_activate` and `resource_deactivate` require some additional logic regarding `refcnt`.

MichalBlk:
Ok, I've changed my mind. We do need to care about `refcnt` during activation and deactivation.

MichalBlk:
Ok, it's ready.

cahirwpz (approved):
LGTM

PR #899, opened 2020-12-29
Add atto and incurses to /contrib.

j-piecuch:
`atto` is a minimal emacs-like text editor.
`incurses` is a minimal implementation of an `ncurses` library, just enough to make `atto` work.


cahirwpz:
IIUC `incurses` is supposed to handle only this subset of `ncurses` API which is used solely by `atto`. Hence I think we can safely make `incurses` compilation a part of `atto` build process. There's no need to build these separately.

j-piecuch:
I thought you wanted them built separately, but ok, I have integrated them.

PR #902, opened 2021-01-02
Handle signals in the TTY layer using ERESTARTSYS

j-piecuch:
When a signal is waiting to be delivered, the TTY layer immediately returns `ERESTARTSYS` to handle the signal. If it was a stop signal and no handler was run, then the system call will be restarted, and if a handler was run, the system call returns `EINTR`, which is exactly what we want. Now we don't need `t_background_cv`.

### DIFF
--- a/bin/Makefile
+++ b/bin/Makefile
@@ -2,7 +2,7 @@
 
 TOPDIR = $(realpath ..)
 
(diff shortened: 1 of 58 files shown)


cahirwpz:
Please resolve conflicts.

cahirwpz (approved):
LGTM

PR #903, opened 2021-01-03
Checking permission to perform write operation on directory

franciscozdo:
We should check permission to write to directory when we create or remove files and directories.

Now we also should change permissions of `/tmp` dir because everyone should have permission to create files in it. (In the future this directory should have also sticky bit set so users can't delete others' files).


wiklam (approved):
LGTM, but the PR does more then it is provided by its title.

franciscozdo:
Thanks! I provided description so it will be clear now.

cahirwpz (approved):
LGTM

PR #904, opened 2021-01-03
Port chown program

franciscozdo:
I added functions:
* `uid_from_user()` 
* `gid_from_group()`

And added implementation for:
* `user_from_uid()`
* `group_from_gid()`

### DIFF
--- a/bin/chown/chown.c
+++ b/bin/chown/chown.c
@@ -1,0 +1,289 @@
+/*	$NetBSD: chown.c,v 1.8 2012/10/24 01:12:51 enami Exp $	*/
+
+/*
+ * Copyright (c) 1988, 1993, 1994, 2003
+ *	The Regents of the University of California.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ * 3. Neither the name of the University nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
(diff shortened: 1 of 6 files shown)


lib/libc/gen/pwcache.c
```diff
@@ -2,9 +2,39 @@
 #include <grp.h>
 
 const char *user_from_uid(uid_t uid, int noname) {
```
cahirwpz, line 4: Since implementation of these is not based on `lib/libc/gen/pwcache.c` from _NetBSD_ it's worth to mention it here.

OTOH I guess it'd be possible to take [pwcache.c](http://bxr.su/NetBSD/lib/libc/gen/pwcache.c#260) as is and add required functions from [getpwent.c](http://bxr.su/NetBSD/distrib/utils/libhack/getpwent.c). Could you verify that?

bin/chown/chown.c
```diff
@@ -0,0 +1,289 @@
+/*	$NetBSD: chown.c,v 1.8 2012/10/24 01:12:51 enami Exp $	*/
```
cahirwpz, line 1: Have you changed anything in this file or is it taken verbatim from _NetBSD_?

franciscozdo: Almost nothing. This is whole diff:
```
33,45d32
< #ifndef lint
< __COPYRIGHT("@(#) Copyright (c) 1988, 1993, 1994, 2003\
<  The Regents of the University of California.  All rights reserved.");
< #endif /* not lint */
<
< #ifndef lint
< #if 0
< static char sccsid[] = "@(#)chown.c   8.8 (Berkeley) 4/4/94";
< #else
< __RCSID("$NetBSD: chown.c,v 1.8 2012/10/24 01:12:51 enami Exp $");
< #endif
< #endif /* not lint */
<
66c53
< __dead static void    usage(void);
---
> static void   usage(void);
```

lib/libc/gen/pwcache.c
franciscozdo: I will check that and push it in another PR.

cahirwpz (approved):
LGTM

PR #906, opened 2021-01-03
Dynamic method dispatch for device drivers

mohrcore:
See #888 for some background.

This PR aims to add a dynamic dispatch for device driver methods implemented as interfaces. If a method is not implemented by device's parent, the dispatch mechanism tries to find a suitable device higher in device tree that implements such method.

This also means that some method signatures have to change as the assumption that a direct child of a bus calls the method does not hold true anymore (unless it has to, then an assert is being used). The `dev` parameter is now the device that is the method actually operates on instead of being it's direct child (which is something I considered to be very confusing in previous design). The new `target` parameter is a device whose driver called the method.


cahirwpz:
This PR attempts to introduce two major changes at once. It's difficult to separate them with my eyes. Hence I'm reluctant to review the code until two functionalities are split into separate PRs.

mohrcore:
oh no

mohrcore:
I've simplified this PR to exclude the part with changing method signatures. It will be necessary to provide a proper dispatch mechanism that will work in _all_ scenarios, however this solution will work in _some_ scenarios (ie. the bus device does about care wether the `dev` device is the caller).

sys/kern/device.c
```diff
@@ -91,3 +91,14 @@ resource_t *device_take_resource(device_t *dev, res_type_t type, int rid,
 
   return rle->res;
 }
+
+device_t *device_if_find_impl(device_t *dev, size_t iface,
```
MichalBlk, line 95: The `iface` argument should be of the `drv_if_t` type.

sys/kern/device.c
```diff
@@ -91,3 +91,14 @@ resource_t *device_take_resource(device_t *dev, res_type_t type, int rid,
...
+                              size_t method_offset) {
+  while ((!dev->parent->driver->interfaces[iface]) ||
```
MichalBlk, line 97: Maybe the following would be more readable
```C
static bool device_if_implemented(void *interface, size_t offset) {
  return *(void **)(interface + offset) != NULL;
}

...
for (;;) {
  device_t *parent = dev->parent;
  assert(parent);
  void *interface = parent->interfaces[iface];
  if (interface && device_if_implemented(interface, method_offset))
    return dev;
  dev = parent;
}
return NULL;
```

sys/kern/device.c
```diff
@@ -91,3 +91,14 @@ resource_t *device_take_resource(device_t *dev, res_type_t type, int rid,
...
+         (!*(vaddr_t **)(dev->parent->driver->interfaces[iface]
```
MichalBlk, line 98: I think that `void **` is more pertinent.

MichalBlk (approved):
I've suggested some stylistic changes, otherwise LGTM.

sys/kern/device.c
mohrcore: Wait, we have that??? Ok then, I'll fix it soon!

sys/kern/device.c
mohrcore: Idk. I'll think about it. It looks like a sligh overkill to make a separate function for that, but maaybe that makes it more readable. The name is a bit misleading tho. Also, this should be inline. How about a comment explaining the check?

sys/kern/device.c
mohrcore: Yes, probably. I'll change it to `void **`.

cahirwpz (approved):
Looks ok. I decided to take over this PR. I'll finish it.

sys/kern/device.c
cahirwpz: Did that.

sys/kern/device.c
cahirwpz: Done!

sys/kern/device.c
cahirwpz: Done!

mohrcore:
Looks like @cahirwpz outran me with those changes.
Anyway, I'm glad this was merged quickly.

This should unblock #873.

PR #907, opened 2021-01-04
mmap: allow replacing mappings when MAP_FIXED flag is specified

bkjg:
This PR introduces changes which will make mmap posix friendly namely allowing replacing existing mappings by the new ones when MAP_FIXED flag is specified.


sys/tests/vm_map.c
```diff
@@ -153,5 +153,58 @@ static int findspace_demo(void) {
   return KTEST_SUCCESS;
 }
 
+static int vm_map_insert_fixed(void) {
```
wiklam, line 156: What about any cases with error occurrences?

wiklam (changes requested):

sys/kern/vm_map.c
```diff
@@ -230,18 +230,40 @@ int vm_map_findspace(vm_map_t *map, vaddr_t *start_p, size_t length) {
...
+
 int vm_map_insert(vm_map_t *map, vm_segment_t *seg, vm_flags_t flags) {
   SCOPED_MTX_LOCK(&map->mtx);
   vm_segment_t *after;
   vaddr_t start = seg->start;
   size_t length = seg->end - seg->start;
   vm_seg_flags_t seg_flags = 0;
 
+  if ((flags & VM_FIXED)) {
+    vm_map_preparespace_nolock(map, start, length);
```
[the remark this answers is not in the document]
bkjg: I return an error only when vm_map doesn't contain this range. And the same error can be found by vm_map_findspace_nolock, which is called next, so to keep the code simpler, I resigned from handling the value returned from this function.

sys/tests/vm_map.c
bkjg: okay, I added one simple test case with an error

sys/kern/vm_map.c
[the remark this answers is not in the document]
bkjg: I changed the signature to `void` and now if it is possible, then it prepares map for inserting new segment, and if not, then just return

sys/kern/vm_map.c
```diff
@@ -230,18 +230,39 @@ int vm_map_findspace(vm_map_t *map, vaddr_t *start_p, size_t length) {
...
+  }
+
   int error = vm_map_findspace_nolock(map, &start, length, &after);
+
   if (error)
     return error;
-  if ((flags & VM_FIXED) && (start != seg->start))
-    return ENOMEM;
+
+  assert(((flags & VM_FIXED) && (start == seg->start)) || !(flags & VM_FIXED));
```
wiklam, line 265: As far as I understand it, the first part of assert will fail when the start argument is not in our map (according to `vm_map_findspace_nolock` implementation cause of `start = max(start, vm_map_start(map));`).
Are you sure about this [comment](https://github.com/cahirwpz/mimiker/pull/907#discussion_r554988238)?

sys/tests/vm_map.c
```diff
@@ -153,5 +153,107 @@ static int findspace_demo(void) {
...
+  assert(n == 0 && t == addr1 + 5 * PAGESIZE);
+
+  vm_map_delete(umap);
+
+  /* Restore original vm_map */
+  vm_map_activate(orig);
+  return KTEST_SUCCESS;
+}
+
+static int vm_map_fixed_bad(void) {
```
wiklam, line 208: I am sorry if I mislead you, but I was thinking about the situation when  `vm_map_preparespace_nolock` returns an error (I am not sure if we need this test), but the comment above describes the situation, about which I have been thinking before.

wiklam (changes requested):
Please provide a description for this PR.

sys/kern/vm_map.c
```diff
@@ -230,18 +230,39 @@ int vm_map_findspace(vm_map_t *map, vaddr_t *start_p, size_t length) {
...
+}
+
 int vm_map_insert(vm_map_t *map, vm_segment_t *seg, vm_flags_t flags) {
   SCOPED_MTX_LOCK(&map->mtx);
   vm_segment_t *after;
   vaddr_t start = seg->start;
   size_t length = seg->end - seg->start;
   vm_seg_flags_t seg_flags = 0;
 
+  if ((flags & VM_FIXED)) {
```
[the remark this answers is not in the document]
bkjg: fixed.

sys/kern/vm_map.c
bkjg: Okay, but above the assert is if statement if error is not zero. And if is, then return with it. So, to be honest, I don't know what do you mean exactly.

sys/tests/vm_map.c
bkjg: Hmm I changed the implementation and now vm_map_preparespace_nolock won't return an error, but vm_map_findspace_nolock will. And this test tests it.

sys/kern/vm_map.c
wiklam: I meant situatuion when `vm_map_start(map) > seg->start` and `VM_FIXED` is set in `vm_map_insert`, does `vm_map_findspace_nolock` always return error in such case?

bkjg: Okay, I see. In that case `vm_map_findspace_nolock` will return some another start address that will be in our `vm_map`. Okay, I agree. It can lead to some problems. I will fix it

sys/kern/vm_map.c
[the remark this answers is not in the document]
bkjg: Okay, now I changed it back to int and now I'm handling the error by returning it.

sys/kern/vm_map.c
```diff
@@ -259,18 +259,45 @@ int vm_map_findspace(vm_map_t *map, vaddr_t *start_p, size_t length) {
   return vm_map_findspace_nolock(map, start_p, length, NULL);
 }
 
+static int vm_map_preparespace_nolock(vm_map_t *map, vaddr_t start,
+                                      size_t length) {
+  assert(page_aligned_p(start) && page_aligned_p(length));
+
+  if (!vm_map_contains_p(map, start, start + length))
+    return ENOMEM;
```
wiklam, line 267: I would add a test for this error if it is possible.

wiklam (approved):
LGTM

pj1031999 (changes requested):
Your tests are not working on AArch64.
`vm_map_fixed_bad` - `assert_fail` for `vm_map_fixed_bad () at /sys/tests/vm_map.c:255`
Please fix that.

bkjg:
> Your tests are not working on AArch64.
> `vm_map_fixed_bad` - `assert_fail` for `vm_map_fixed_bad () at /sys/tests/vm_map.c:255`
> Please fix that.

@pj1031999 Should be fixed.

sys/tests/vm_map.c
```diff
@@ -160,5 +161,107 @@ static int findspace_demo(void) {
...
+  n = vm_map_insert(umap, seg, VM_FIXED);
+  assert(n == 0);
+
+  seg = vm_segment_alloc(NULL, addr2, addr3, VM_PROT_NONE, VM_SEG_PRIVATE);
+  n = vm_map_insert(umap, seg, VM_FIXED);
+  assert(n == 0);
+
+  t = addr1;
+  n = vm_map_findspace(umap, &t, 4 * PAGESIZE);
+  assert(n == 0 && t == addr1);
```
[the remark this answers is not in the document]
bkjg: Done.

cahirwpz:

franciscozdo:
I think we will do it in different way because we won't have segments.

cahirwpz:
Superseded by #1357

PR #908, opened 2021-01-05
munmap: allow munmapping pieces of segments

bkjg:
This PR introduces changes which will make munmap posix friendly namely allowing munmapping pieces of segments.

### DIFF
--- a/sys/kern/mmap.c
+++ b/sys/kern/mmap.c
@@ -63,10 +63,10 @@ int do_munmap(vaddr_t addr, size_t length) {
   if (length == 0)
     return EINVAL;
 
-  /* Round boundaries to contain entire pages from range [addr, addr + length)
-   */
-  vaddr_t right_boundary = roundup(addr + length, PAGESIZE);
-  addr = rounddown(addr, PAGESIZE);
+  if (!page_aligned_p(addr) || !page_aligned_p(length))
+    return EINVAL;
+
+  vaddr_t right_boundary = addr + length;
 
   WITH_VM_MAP_LOCK (uspace) {
     while (addr < right_boundary) {
--- a/sys/kern/mmap.c
+++ b/sys/kern/mmap.c
@@ -77,19 +77,11 @@ int do_munmap(vaddr_t addr, size_t length) {
       /* TODO We support unmaping entire segments only! */
       vaddr_t start = vm_segment_start(seg);
       vaddr_t end = vm_segment_end(seg);
-      if ((addr != start) || (addr + length != end))
-        return ENOTSUP;
 
-      /* there are few cases:
-       * - we want to free the whole segment
-       * - we want to free the beginning of the segment and leave the end
-       * without changes
-       * - we want to free the end of the segment and leave the beginning
-       * without changes
-       * - we want to free the middle of the segment, so we want to split it to
-       * the new two segments with the hole between them */
       vm_segment_destroy_range(uspace, seg, max(addr, start),
                                min(right_boundary, end));
+
+      addr = end;
     }
   }
   return 0;
--- a/sys/kern/mmap.c
+++ b/sys/kern/mmap.c
@@ -60,18 +60,37 @@ int do_munmap(vaddr_t addr, size_t length) {
 
   vm_map_t *uspace = proc_self()->p_uspace;
 
+  if (length == 0)
+    return EINVAL;
+
+  /* Round boundaries to contain entire pages from range [addr, addr + length)
+   */
+  vaddr_t right_boundary = roundup(addr + length, PAGESIZE);
+  addr = rounddown(addr, PAGESIZE);
+
   WITH_VM_MAP_LOCK (uspace) {
-    vm_segment_t *seg = vm_map_find_segment(uspace, addr);
-    if (!seg)
-      return EINVAL;
+    while (addr < right_boundary) {
+      vm_segment_t *seg = vm_map_find_segment(uspace, addr);
+      if (!seg)
+        return EINVAL;
 
-    /* TODO We support unmaping entire segments only! */
-    vaddr_t start = vm_segment_start(seg);
-    vaddr_t end = vm_segment_end(seg);
-    if ((addr != start) || (addr + length != end))
-      return ENOTSUP;
+      /* TODO We support unmaping entire segments only! */
+      vaddr_t start = vm_segment_start(seg);
+      vaddr_t end = vm_segment_end(seg);
+      if ((addr != start) || (addr + length != end))
+        return ENOTSUP;
 
-    vm_segment_destroy(uspace, seg);
+      /* there are few cases:
+       * - we want to free the whole segment
+       * - we want to free the beginning of the segment and leave the end
+       * without changes
+       * - we want to free the end of the segment and leave the beginning
+       * without changes
+       * - we want to free the middle of the segment, so we want to split it to
+       * the new two segments with the hole between them */
+      vm_segment_destroy_range(uspace, seg, max(addr, start),
+                               min(right_boundary, end));
+    }
   }
   return 0;
 }
--- a/sys/kern/mmap.c
+++ b/sys/kern/mmap.c
@@ -74,7 +74,6 @@ int do_munmap(vaddr_t addr, size_t length) {
       if (!seg)
         return EINVAL;
 
-      /* TODO We support unmaping entire segments only! */
       vaddr_t start = vm_segment_start(seg);
       vaddr_t end = vm_segment_end(seg);
 
--- a/sys/kern/vm_map.c
+++ b/sys/kern/vm_map.c
@@ -161,6 +161,46 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
   vm_segment_free(seg);
 }
 
+/* there are few cases:
+ * 1) we want to free the whole segment
+ * 2) we want to free the beginning of the segment and leave the end
+ * without changes
+ * 3) we want to free the end of the segment and leave the beginning
+ * without changes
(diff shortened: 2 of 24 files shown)


sys/kern/mmap.c
```diff
@@ -60,18 +60,29 @@ int do_munmap(vaddr_t addr, size_t length) {
 
   vm_map_t *uspace = proc_self()->p_uspace;
 
+  if (length == 0)
+    return EINVAL;
```
wiklam, line 64: I'm not sure, but according to NetBSD manual it is unnecessary. Please consult it with @cahirwpz.

sys/kern/mmap.c
```diff
@@ -60,18 +60,29 @@ int do_munmap(vaddr_t addr, size_t length) {
...
+
+  /* Round boundaries to contain entire pages from range [addr, addr + length)
+   */
+  vaddr_t right_boundary = roundup(addr + length, PAGESIZE);
+  addr = rounddown(addr, PAGESIZE);
```
wiklam, line 69: The address should be already page aligned, check the manual.

sys/kern/mmap.c
```diff
@@ -60,18 +60,29 @@ int do_munmap(vaddr_t addr, size_t length) {
...
   WITH_VM_MAP_LOCK (uspace) {
-    vm_segment_t *seg = vm_map_find_segment(uspace, addr);
-    if (!seg)
-      return EINVAL;
+    while (addr < right_boundary) {
+      vm_segment_t *seg = vm_map_find_segment(uspace, addr);
+      if (!seg)
+        return EINVAL;
+
+      /* TODO We support unmaping entire segments only! */
```
wiklam, line 77: Is this comment up to date?

sys/kern/vm_map.c
```diff
@@ -161,6 +161,46 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
...
 }
 
+/* there are few cases:
+ * 1) we want to free the whole segment
+ * 2) we want to free the beginning of the segment and leave the end
+ * without changes
+ * 3) we want to free the end of the segment and leave the beginning
+ * without changes
+ * 4) we want to free the middle of the segment, so we want to split it to
+ * the new two segments with the hole between them */
```
wiklam, line 171: Imo this comments should be left by the code which they are describing.

sys/kern/vm_map.c
```diff
@@ -161,6 +161,46 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
...
+  /* case 1) */
+  if (seg->start == start && seg->end == end) {
+    vm_segment_destroy(map, seg);
+  } else if (seg->start == start) { /* case 2) */
+    assert(end < seg->end);
+    size_t length = end - seg->start;
+    vm_object_remove_range(seg->object, 0, length);
+    /* TODO there's no reference to pmap in page, so we have to do it here */
+    pmap_remove(map->pmap, seg->start, end);
+    seg->start = end;
```
wiklam, line 185: It is not clear for me, why we can not do it in the same way as in case 4?

wiklam (changes requested):

sys/kern/vm_map.c
```diff
@@ -161,6 +161,46 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
...
+    vm_object_t *obj = vm_object_clone(seg->object);
+    /* nextly, resize the segment, to contain the beginning of it */
+    vm_segment_resize(map, seg, start);
+    /* now remove proper range from obj */
+    vm_object_remove_range(obj, 0, end - seg->start);
+    /* finally, create a new segment with obj */
+    vm_segment_t *new_seg =
+      vm_segment_alloc(obj, end, seg->end, seg->prot, seg->flags);
+    vm_map_insert_after(map, new_seg, seg);
+  }
```
franciscozdo, line 201: Could all this cases end in just one call to `vn_object_remove_range(obj, offset, length)`  where `offset` and `length` are calculated for all cases?

Could you explain why we resize segment in two cases but not in all (2,3,4) cases? It is a bit confusing to me.

franciscozdo (changes requested):

sys/kern/mmap.c
pj1031999: It looks good to me.
`EINVAL` is valid here for _FreeBSD_  and _Linux_ anyway if length is 0 then `munmap` has nothing to do.

sys/kern/mmap.c
pj1031999: We should be more strict here and return `EINVAL` in case of wrong alignment.

sys/kern/mmap.c
bkjg: According to the POSIX:

```
The munmap() function shall fail if:
[...]
[EINVAL]
The len argument is 0.
```

https://pubs.opengroup.org/onlinepubs/9699919799/functions/munmap.html#

sys/kern/mmap.c
bkjg: Hmm the POSIX says:

```
The munmap() function may fail if:

[EINVAL]
The addr argument is not a multiple of the page size as returned by sysconf().
```

"may", not "have to", so I don't really know, what is the best solution.

sys/kern/mmap.c
bkjg: You're right. It's not up to date now. I will remove it.

sys/kern/mmap.c
pj1031999: So could you give **any** example of **real** program on **real** OS that doesn't fail on that condition?
We could implement POSIX in every detail but I don't see any reason for that. This will be a **dead** feature. We don't implement programs in user space - we only care about kernel so nobody will use that combination of parameters except our own tests. Do we have tests for that?

sys/kern/vm_map.c
bkjg: We cannot do it in the same way as in case 4 because in case we destroy the middle of the segment and finally we split it to the separate two segments. Whereas in case 2 we just destroy the beginning of the segment and change the start value to the new one.

sys/kern/vm_map.c
```diff
@@ -161,6 +161,46 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
...
+    vm_segment_resize(map, seg, start);
+  } else { /* case 4) */
+    assert(start > seg->start && end < seg->end);
+    /* at the beginning: clone the vm_object */
+    vm_object_t *obj = vm_object_clone(seg->object);
+    /* nextly, resize the segment, to contain the beginning of it */
+    vm_segment_resize(map, seg, start);
+    /* now remove proper range from obj */
+    vm_object_remove_range(obj, 0, end - seg->start);
+    /* finally, create a new segment with obj */
```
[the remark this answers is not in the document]
bkjg: Ok, removed.

sys/kern/mmap.c
bkjg: Changed.

sys/kern/vm_map.c
bkjg: Okay, I finally did like you suggested: I did one call to vn_object_remove_range(obj, offset, length). And that changed everything so I think the second question is no longer relevant.

sys/kern/vm_map.c
bkjg: I changed code a little and this comment is no longer relevant, so I deleted it.

sys/kern/vm_map.c
```diff
@@ -161,6 +161,34 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
...
+  vm_object_remove_range(seg->object, start - seg->start, length);
+  pmap_remove(map->pmap, start, end);
+
+  if (seg->start == start) {
+    seg->start = end;
+  } else if (seg->end == end) {
+    seg->end = start;
+  } else { /* a hole inside the segment */
+    vm_object_t *obj = vm_object_clone(seg->object);
+    vm_object_remove_range(obj, 0, end - seg->start);
```
franciscozdo, line 183: I think you should remove range of length `start - seg->start`.

The range between `start` and `end` was removed in line 174.

franciscozdo (changes requested):
Looks good, but you are probably freeing some pages twice.

sys/kern/vm_map.c
bkjg: Oh, yes, sorry, my fault, I fixed it.

franciscozdo (approved):
LGTM

sys/kern/vm_map.c
```diff
@@ -161,6 +161,34 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
...
+  pmap_remove(map->pmap, start, end);
+
+  if (seg->start == start) {
+    seg->start = end;
+  } else if (seg->end == end) {
+    seg->end = start;
+  } else { /* a hole inside the segment */
+    vm_object_t *obj = vm_object_clone(seg->object);
+    vm_object_remove_range(obj, 0, start - seg->start);
+    vm_object_remove_range(seg->object, end - seg->start, seg->end - end);
```
wiklam, line 184: Will we use `vm_segment_destroy_range` outside `do_munmap`? If yes I would add assert to check if `start >= seg->start`, the same for end, to prevent future bugs.

wiklam (approved):
LGTM, but check the comment left bellow

sys/kern/vm_map.c
bkjg: Okay, I added it.

cahirwpz (approved):
LGTM

PR #909, opened 2021-01-07
pmap: fix locks

bkjg:


sys/aarch64/pmap.c
```diff
@@ -414,10 +416,12 @@ void pmap_page_remove(vm_page_t *pg) {
     pmap_t *pmap = pv->pmap;
     vaddr_t va = pv->va;
     TAILQ_REMOVE(&pg->pv_list, pv, page_link);
-    TAILQ_REMOVE(&pmap->pv_list, pv, pmap_link);
-    pte_t *ptep = pmap_lookup_pte(pmap, va);
-    assert(ptep != NULL);
-    pmap_write_pte(pmap, ptep, 0);
+    WITH_MTX_LOCK (&pmap->mtx) {
```
pj1031999, line 419: `pmap_page_remove` is called **only** from `pm_free_from_seg` which already holds global `physmem_lock`.

sys/mips/pmap.c
```diff
@@ -386,8 +385,10 @@ void pmap_page_remove(vm_page_t *pg) {
     pmap_t *pmap = pv->pmap;
     vaddr_t va = pv->va;
     TAILQ_REMOVE(&pg->pv_list, pv, page_link);
-    TAILQ_REMOVE(&pmap->pv_list, pv, pmap_link);
-    pmap_pte_write(pmap, va, empty_pte(pmap), 0);
+    WITH_MTX_LOCK (&pmap->mtx) {
+      TAILQ_REMOVE(&pmap->pv_list, pv, pmap_link);
+      pmap_pte_write(pmap, va, empty_pte(pmap), 0);
+    }
```
pj1031999, line 391: Same as in _AArch64_.

pj1031999 (changes requested):
First of all - without any description that PR is unreadable.

The main part of your PR is replacing `WITH_MTX_LOCK` with `SCOPED_MTX_LOCK`.  `assert` shouldn't modify any data and if you look carefully you have two cases:
1. `page_aligned_p` which doesn't touch any `pmap`.
2. `pmap_contains_p` which calls `pmap_start` and `pmap_end` - they return constant value base on asid but we know that asid is set only in `pmap_setup` and it is not modified.

So I don't see any practical reason for that PR. It will be helpful if you could provide any real scenario where you had problems with that locks.

franciscozdo:
> Have you ever run that part of code?  
>You acquire mutex in 281 and then do the same in 291.  
>Now you have deadlock on every `pmap_kenter` call.

@pj1031999 sorry for deleting this comment. It was by accident.

PR #910, opened 2021-01-07
Implement pseudoterminals

j-piecuch:
- `posix_openpt()` returns a an open file descriptor for the master side of a new pseudoterminal.
- `ptsname()` is used to retrieve the path of the slave device from the file descriptor of the master device.

I will add userspace tests in another PR, so that this one doesn't get too big.


franciscozdo (changes requested):
Overall looks good, but I think you should describe some decisions in code.

sys/kern/pty.c
```diff
@@ -0,0 +1,269 @@
...
+  return 0;
+}
+
+static int pty_ioctl(file_t *f, u_long cmd, void *data) {
+  tty_t *tty = f->f_data;
+
+  switch (cmd) {
+    case TIOCPTSNAME: {
+      char *user_buf = *(char **)data;
+      char tmp_buf[16];
```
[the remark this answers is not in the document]
j-piecuch: Done.

lib/libc/stdlib/pty.c
```diff
@@ -0,0 +1,73 @@
...
+#include <fcntl.h>
+#include <string.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <sys/ttycom.h>
+#include <sys/ioctl.h>
+
+int grantpt(int fildes) {
+  return 0;
+}
```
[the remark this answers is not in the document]
j-piecuch: I have added a comment.

franciscozdo (approved):
LGTM

sys/kern/pty.c
```diff
@@ -0,0 +1,273 @@
...
+    if (atomic_compare_exchange_strong(&pty->pt_number, &(int){-1}, i + 1))
+      /* Got a free PTY. */
+      return pty;
+  }
+  return NULL;
+}
+
+static void pty_free(pty_t *pty) {
+  pty->pt_number = -1;
+}
```
wiklam, line 46: I would add an assert for `pty->pt_number `.

sys/kern/pty.c
```diff
@@ -0,0 +1,273 @@
...
+
+  return error;
+}
+
+static int pty_putc_sleep(tty_t *tty, pty_t *pty, uint8_t c) {
+  while (!tty_input(tty, c)) {
+    if (cv_wait_intr(&pty->pt_outcv, &tty->t_lock))
+      return ERESTARTSYS;
+    if (!tty_opened(tty))
+      return ENOTTY;
```
wiklam, line 84: Could you explain to me why we check it in this order?

sys/kern/pty.c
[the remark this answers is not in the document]
j-piecuch: Done.

j-piecuch: Done.

sys/kern/pty.c
j-piecuch: Actually, we should check this before calling `tty_input()`, and after returning from `cv_wait_intr()`, since the tty's open count can change while we're sleeping. I have fixed it. Thanks!

wiklam (approved):
LGTM

sys/kern/pty.c
j-piecuch: Actually, I was wrong: we don't want to check this before calling `tty_input()` the first time. If it succeeds the first time, the whole function should return 0. However, if `tty_input()` fails and we need to sleep, then we should sleep only if the tty has users, i.e. `tty_opened(tty)`. So the check should be before calling `cv_wait_intr()`: the original order was wrong. I think I have fixed it now. Please resolve this conversation once you have read this.

sys/kern/pty.c
```diff
@@ -0,0 +1,277 @@
...
+  pty_t *pty = tty->t_data;
+
+  mtx_lock(&tty->t_lock);
+  tty_detach_driver(tty); /* Releases t_lock. */
+  pty_free(pty);
+  return 0;
+}
+
+static int pty_seek(file_t *f, off_t offset, int whence, off_t *newoffp) {
+  return 0;
```
cahirwpz, line 135: Should it be `ESPIPE`?

sys/kern/pty.c
```diff
@@ -0,0 +1,277 @@
...
+}
+
+static int pty_stat(file_t *f, stat_t *sb) {
+  memset(sb, 0, sizeof(stat_t));
+  sb->st_mode = S_IFCHR;
```
cahirwpz, line 140: If I perform `stat` on pseudoterminal device it reports:

```
  File: /dev/pts/5
  Size: 0         	Blocks: 0          IO Block: 1024   character special file
Device: 13h/19d	Inode: 8           Links: 1     Device type: 88,5
Access: (0620/crw--w----)  Uid: ( 1001/   cahir)   Gid: (    5/     tty)
Access: 2021-01-28 08:03:36.135880424 +0100
Modify: 2021-01-28 08:03:36.135880424 +0100
Change: 2021-01-27 00:32:48.135880424 +0100
```

Aren't we missing something here?

sys/kern/pty.c
```diff
@@ -0,0 +1,277 @@
...
+  size_t start_resid = uio->uio_resid;
+
+  SCOPED_MTX_LOCK(&tty->t_lock);
+
+  while (uio->uio_resid > 0) {
+    if ((error = uiomove(&c, 1, uio)))
+      break;
+    if ((error = pty_putc_sleep(tty, pty, c))) {
+      /* Undo the last uiomove(). */
+      uio->uio_resid++;
```
cahirwpz, line 110: IIRC `uio_t` fields are modified only in `uiomove` and I would like to keep it this way. Is there a simple way to reorganize this code not to break the premise ?

cahirwpz (commented):
Please verify that `/dev/pts` directory and its contents have right attributes and `stat` returns reasonable values for them.

sys/kern/pty.c
j-piecuch: To do so, we would need something like `pty_can_putc()` that checks if `tty_input()` on the slave device will succeed. The thing is, the outcome of `tty_input(c)` is heavily dependent on the value of `c` and terminal settings: e.g. when the `ISIG` flag is set, `tty_input('^C')` will always succeed, since the character isn't actually stored in the input queue. So IMO it's not possible. I also noticed that simply incrementing `uio_resid` is completely broken, as `uio_offset` and other fields need to be adjusted. I suggest adding a function for rolling back `uiomove()` operations, something like `void uio_rollback(uio_t *uio, size_t nbytes)`

sys/kern/pty.c
j-piecuch: Yes, it should. Fixed.

sys/kern/pty.c
j-piecuch: `/dev/pts/5` is a slave tty, `pty_stat()` implements `FOP_STAT()` for the master side. I agree it's somewhat bare-bones... FreeBSD appears to take most attributes from the slave device ([link](https://grok.dragonflybsd.org/xref/freebsd/sys/kern/tty_pts.c?r=77fb93e1#538)), so I will do that too.

sys/kern/pty.c
cahirwpz: OK. I see... thanks for the explanation.

cahirwpz (approved):
LGTM

PR #913, opened 2021-01-08
IDE ATA Driver (PIO)

ThomasStachovsky:


ThomasStachovsky:
The code of the driver is ready for review. The only thing to be changed are the additions in pci.c, but maybe it would be better to do this in another pull request? 
This code after merging the isa driver PR should not pass tests! To make it pass tests we have to make a small PR for the isa driver. This is because the isa driver takes the whole range of addresses in which there are addresses of ports used by the ide controller.

include/dev/atareg.h
```diff
@@ -0,0 +1,571 @@
+/*	$NetBSD: atareg.h,v 1.45 2020/05/15 21:56:14 jdolecek Exp $	*/
```
komik0, line 1: do we need all below registers?

sys/drv/Makefile
```diff
@@ -8,6 +8,7 @@ SOURCES = \
 	rtc.c \
 	rtl8139.c \
 	pit.c \
-	stdvga.c
+	stdvga.c \
+	ide.c
```
komik0, line 12: nit: alphabetical order

sys/drv/ide.c
```diff
@@ -0,0 +1,326 @@
...
+    ide_wait_busy(ide, channel);
+
+    ide_wait_ready(ide, channel);
+
+    for (int j = 0; j < 256; j++)
+      if (action == WDCC_READ)
+        buffer[j] = ide_read2(ide, channel, ATA_REG_DATA);
+      else
+        ide_write2(ide, channel, ATA_REG_DATA, buffer[j]);
+    buffer += 256;
```
komik0, line 151: imo 256 should be defined as a block size or something

PR #914, opened 2021-01-09
Rtl8139 tx path

komik0:
now we only send dummy ack for every packet.
this is just demo version (for monday's meeting) - to show what's needed to send any packet.


cahirwpz:
Since we're going to drop support for Malta in the near future this driver has little to no value for us. Hence I decided to close the PR.

cahirwpz:
Since we're going to drop support for Malta in the near future this driver has little to no value for us. Hence I decided to close the PR.

PR #916, opened 2021-01-11
tty: add support for ISIG flag (i.e. send signals in response to certain control characters)

j-piecuch:
If the `ISIG` terminal flag is set, certain control characters result in a signal being sent to the foreground process group:
- `VINTR` causes `SIGINT` to be sent,
- `VSUSP` causes `SIGTSTP` to be sent,
- `VQUIT` causes `SIGQUIT` to be sent.


sys/kern/tty.c
```diff
@@ -474,6 +475,31 @@ static void tty_in_hiwat(tty_t *tty) {
...
+    tty_notify_in(tty);
+  }
+}
+
+static void tty_discard_input(tty_t *tty) {
+  assert(mtx_owned(&tty->t_lock));
+  ringbuf_reset(&tty->t_inq);
+  tty->t_line.ln_count = 0;
+  tty_check_in_lowat(tty);
+}
```
wiklam, line 495: Maybe is not a big feature/change, but code transfer in the same PR is not a good idea.

wiklam (approved):
LGTM, despite the small code transfer.

sys/kern/tty.c
j-piecuch: Yeah, I've changed the code to declare the needed functions at the start of the file.

cahirwpz (approved):
LGTM

PR #892, opened 2020-12-21
Cbus Driver

ThomasStachovsky:

PR #917, opened 2021-01-11
Reorganize hardware resources handling.

MichalBlk:
- Extract type dependent fields from the `resource_t` structure to the `resource_list_entry_t` structure.
- Make the `resource_t` structure an rman private structure.
- Present the `resource_list_entry_t` structure as a public handle to hardware resources and rename it to `resource_t`.


sys/mips/cbus.c
```diff
@@ -19,6 +19,5 @@ static bus_space_t cbus_bus_space = {
 /* clang-format off */
 resource_t cbus_uart[1] = {{.r_bus_tag = &cbus_bus_space,
                             .r_bus_handle = 0,
-			    .r_start = 0,
-			    .r_end = 7}};
```
mohrcore, line 23: Ok, I know `cbus` is a one hot mess rn, but I this still feels somehow inappropriate...

That said I don't see any better solution. As long as `cbus` works and isn't a focal point of a PR I'm fine with pretending I've never seen its driver's code.

mohrcore (approved):
LGTM, hooray!

sys/mips/cbus.c
MichalBlk: I am looking forward to the new `cbus` driver (#892). The `rootdev` should assign an appropriate resource to the `cbus` in its attach function. Then this awkward definition will expire (and finally we'll get rid of the terrible `RESOURCE_DECLARE(name)` macro).

PR #919, opened 2021-01-12
TTY & PTY userspace tests

j-piecuch:
This PR adds a basic PTY test, and 3 TTY tests that test canonical mode, echo and signals.

It also fixes a small bug: we didn't echo characters in raw mode even if `ECHO` was set.


franciscozdo (approved):
LGTM

cahirwpz (approved):
LGTM

PR #921, opened 2021-01-12
Progressive device tree scanning.

MichalBlk:
Scan the device tree multiple times. Each scan is identified be a pass number and each device driver is assigned a pass number.
At each scan, drivers can be attached to devices. Driver can be probed and attached only if the current system-wide pass number
matches the driver's pass number. The first pass is devoted to detect basic hardware resources like timers and buses. All remaining devices are processed during the second (FTTB also last) pass.


include/sys/device.h
```diff
@@ -29,13 +29,23 @@ typedef enum {
...
+} pass_num_t;
+
 struct driver {
   const char *desc;            /* short driver description */
   size_t size;                 /* device->state object size */
   d_probe_t probe;             /* probe for specific device(s) */
   d_attach_t attach;           /* attach device to system */
   d_detach_t detach;           /* detach device from system */
   void *interfaces[DIF_COUNT]; /* pointers to device methods structures */
+  pass_num_t pass;             /* device tree pass number */
```
cahirwpz, line 48: Could you move it just after `size` - here and in every definition?

include/sys/bus.h
```diff
@@ -210,6 +210,9 @@ static inline void bus_release_resource(device_t *dev, res_type_t type,
   BUS_METHODS(idev->parent).release_resource(idev, type, r);
 }
 
+/* System-wide current pass number. */
+extern pass_num_t current_pass;
```
cahirwpz, line 214: I think in current use scenario you can get rid of it.

sys/aarch64/rootdev.c
```diff
@@ -302,21 +304,27 @@ static bus_methods_t rootdev_bus_if = {
...
-  .state = &(rootdev_t){},
-  .devclass = &DEVCLASS(root),
-};
-
-DEVCLASS_ENTRY(root, rootdev);
+DEVCLASS_CREATE(root);
 
-void init_devices(void) {
-  device_attach(&rootdev);
+void init_devices(pass_num_t pass) {
```
cahirwpz, line 318: Please note that first pass is performed when `rootdev` is `NULL`. Other passes when `rootdev` is already present.

sys/aarch64/trap.c
```diff
@@ -18,7 +18,8 @@ static __noreturn void kernel_oops(ctx_t *ctx) {
   panic();
 }
 
-static void syscall_handler(register_t code, ctx_t *ctx, syscall_result_t *result) {
+static void syscall_handler(register_t code, ctx_t *ctx,
```
cahirwpz, line 21: Random change?

sys/kern/bus.c
```diff
@@ -117,26 +117,41 @@ void bus_deactivate_resource(device_t *dev, res_type_t type, resource_t *r) {
   rman_deactivate_resource(r);
 }
 
+pass_num_t current_pass;
```
cahirwpz, line 120: Please make it `static`.

sys/kern/main.c
```diff
@@ -99,8 +102,8 @@ __noreturn void kernel_init(void) {
   /* Mount filesystems (including devfs). */
   mount_fs();
 
-  /* First (FTTB also last) stage of device init. */
-  init_devices();
+  /* First stage of device init. */
+  init_devices(FIRST_PASS);
```
cahirwpz, line 106: It'd be great if you described what kind of drivers should be initialized by `FIRST_PASS`. Our goal was to start those that do not require `devfs`, but provide first timer and console output.

cahirwpz (commented):
I like the simplicity, but `FreeBSD` does it in different way. Can you look into their sources and figure out why they need such a complex solution? (they keep `pass` field in `driverlink`)

include/sys/device.h
MichalBlk: Done.

include/sys/bus.h
MichalBlk: Done.

sys/aarch64/trap.c
MichalBlk: This is due to make format for aarch64.

sys/kern/bus.c
MichalBlk: Done.

sys/kern/main.c
MichalBlk: Done.

sys/aarch64/rootdev.c
MichalBlk: Done.

cahirwpz (approved):
LGTM

PR #922, opened 2021-01-13
Add an attribute to disable the compiler instrumentation

xThaid:
Some functions can't be instrumented because of stack issues. The compilers (both GCC and Clang) provide an attribute to specify that a particular set of instrumentations should not be applied to a function, so we won't have to write extra Makefile rules anymore.

### DIFF
--- a/include/sys/cdefs.h
+++ b/include/sys/cdefs.h
@@ -73,6 +73,7 @@
 #define __long_call
 #endif
 #define __transparent_union __attribute__((__transparent_union__))
+#define __no_instrument __attribute__((no_sanitize("address,thread,undefined")))
 
 /* Clang introduces support for the fallthrough attribute in C2x. */
 #ifdef __clang__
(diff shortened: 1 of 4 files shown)


include/sys/cdefs.h
```diff
@@ -73,6 +73,8 @@
 #define __long_call
 #endif
 #define __transparent_union __attribute__((__transparent_union__))
+#define __no_instrument                                                        \
```
cahirwpz, line 76: Maybe `__no_sanitize` instead?

cahirwpz (commented):
What about other functions used from `copy.S` like `bcopy`?

xThaid:
> What about other functions used from `copy.S` like `bcopy`?

`copy.S` uses only `bcopy` and `copystr` and `bcopy` is already implemented in assembly.

pj1031999 (approved):
You may consider adding information about this attribute to prototypes.

cahirwpz (approved):
LGTM

PR #923, opened 2021-01-14
Implement TIOCGWINSZ and TIOCSWINSZ ioctls

j-piecuch:
Implement ioctls for getting and setting the terminal window size.
Also, allocate `tty_t` structures with the `M_ZERO` flag, so that we don't need to explicitly zero-initialize most fields.


franciscozdo (approved):
LGTM

pj1031999 (approved):
LGTM

cahirwpz (approved):
LGTM

PR #924, opened 2021-01-14
AArch64: fix buffer overflow in syscall handler

pj1031999:
Fix buffer overflow in syscall handler.
Number of registers copied from context into args buffer should not be greater than `SYS_MAXSYSARGS`.


franciscozdo (approved):
LGTM

cahirwpz (approved):
LGTM

PR #925, opened 2021-01-15
Allow to set set-user-id bit on userspace programs

franciscozdo:
We can specify if program will have set `set-user-id` bit by setting variable `SETUID` in Makefile for that program.


build/build.prog.mk
```diff
@@ -18,6 +18,13 @@ BINDIR ?= $(shell echo $(DIR) | cut -f 1 -d / | tr . /)
 
 BUILD-FILES += $(PROGRAM).uelf
 INSTALL-FILES += $(SYSROOT)/$(BINDIR)/$(PROGRAM)
+SETUID ?= 0
```
cahirwpz, line 21: BINMODE ?= 0755

franciscozdo: done

cahirwpz (approved):
LGTM

PR #926, opened 2021-01-15
Fix files and directory ownership

franciscozdo:
File should have owner and group of creating process effective uid and gid.
I also added `set-group-id` functionality on directories. Files created under such directories will have group of parent dir. Directories will have group of parent dir and also set `set-group-id`.


xThaid (approved):
lgtm

wiklam (approved):
LGTM

cahirwpz (approved):
LGTM

PR #927, opened 2021-01-15
Simplify PIT timer - remove spinlock

wiklam:
Updating the counter when we request time from the timer increase the quality of measured time.


j-piecuch (approved):
LGTM, although I am a bit concerned about performing expensive multiplication, modulo and division every time `binuptime()` is called.

pj1031999 (approved):
LGTM

cahirwpz (approved):
LGTM

PR #928, opened 2021-01-15
Save FPU context on ctx_switch, restore on user_exc_leave.

cahirwpz:
We don't want to needlessly save & restore FPU context. Moreover we would like to perform both things in a uniform way on both MIPS & AArch64. Let us save (where?) an information that current FPU context is valid and has not been saved. We need to save the context only when kernel thread calls `ctx_switch` and mark that FPU context was saved, so we don't accidentally save it the second time. We need to restore FPU context only when `ctx_switch` was performed.

Let's devise private thread flags aka `td_pflags`:

`TDP_FPCTXSAVED`: the context was saved because `ctx_switch` happened
`TDP_FPCTXUSED`: the context is valid and should be saved & restored on demand


sys/mips/ebase.S
```diff
@@ -345,15 +347,14 @@ SNESTED(user_exc_enter, MCONTEXT_SIZE, ra)
...
-        nop
-
-        SAVE_FPU_CTX()
+        # Invalidate current FPU context.
+        LOAD_PCPU(t0)
+        lw      t0, PCPU_CURTHREAD(t0)
+        lw      t1, TD_PFLAGS(t0)
+        nor     t2, zero, TDP_FPCTXSAVED
+        and     t1, t1, t2
+        sw      t1, TD_PFLAGS(t0)
```
j-piecuch, line 356: Couldn't we clear the `TDP_FPCTXSAVED` flag in `user_exc_leave()`, in the same place where we restore the FPU context? That way, we do this only for threads that have the flag set.

sys/mips/ebase.S
```diff
@@ -396,8 +397,11 @@ user_exc_leave:
         sw      t0, PCPU_SR(s0)
 
         # If user has FPU disabled, don't restore its registers.
```
j-piecuch, line 399: This comment needs updating: now we also skip restoring the FPU registers if we know the current FPU context belongs to the current thread (i.e. there were no context switches).

j-piecuch (changes requested):

include/sys/thread.h
```diff
@@ -69,7 +69,10 @@ typedef enum {
 #define TDF_SLPTIMED 0x00000080 /* sleep with timeout */
 
 typedef enum {
-  TDP_OLDSIGMASK = 0x01 /* Pass td_oldsigmask as return mask to send_sig(). */
+  TDP_OLDSIGMASK = 0x01, /* Pass td_oldsigmask as return mask to send_sig(). */
+  TDP_FPCTXSAVED = 0x02, /* The FPU context was saved by `ctx_switch`. */
+  TDP_FPCTXVALID = 0x04  /* The FPU context is valid and should be saved &
```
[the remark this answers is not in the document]
pj1031999: done

sys/mips/ebase.S
pj1031999: done

sys/mips/ebase.S
pj1031999: done

sys/mips/ebase.S
```diff
@@ -395,11 +388,20 @@ user_exc_leave:
...
+        and     t1, t1, TDP_FPCTXSAVED
         beqz    t1, skip_fpu_restore
         nop
 
+        # Clear TD_FPCTXSAVED flag.
+        lw      t1, TD_PFLAGS(t0)
+        nor     t2, zero, TDP_FPCTXSAVED
+        and     t1, t1, t2
+        sw      t1, TD_PFLAGS(t0)
+
```
j-piecuch, line 404: This could be optimized a little bit:
```
        # If user FPU context is not saved, don't restore its registers.
        ext     t1, t0, SR_CU1_SHIFT, 1
        LOAD_PCPU(t0)
        lw      t0, PCPU_CURTHREAD(t0)
        lw      t1, TD_PFLAGS(t0)
        and     t2, t1, TDP_FPCTXSAVED
        beqz    t2, skip_fpu_restore
        nop


        # Clear TD_FPCTXSAVED flag.
        # We know that t2 == TDP_FPCTXSAVED and that the flag is set,
        # so we can just xor t2 with td_pflags.
        xor     t1, t1, t2
        sw      t1, TD_PFLAGS(t0)
```

j-piecuch (approved):

sys/mips/ebase.S
pj1031999: done - thanks

PR #929, opened 2021-01-15
Add various functionality required to port the `script` program.

j-piecuch:
- Add `bswap{16,32,64}()` (implemented as macros using `__builtin_bswap{16,32,64}()`
- Add `login_tty()` from NetBSD and implement `openpty()` in `libutil`
- Implement `TIOCSCTTY` `ioctl()`


sys/kern/pty.c
```diff
@@ -0,0 +1,273 @@
...
+  SCOPED_MTX_LOCK(&tty->t_lock);
+
+  while (uio->uio_resid > 0) {
+    if ((error = uiomove(&c, 1, uio)))
+      break;
+    if ((error = pty_putc_sleep(tty, pty, c))) {
+      if (error == ENOTTY)
+        error = 0;
+      break;
+    }
```
pj1031999, line 109: In #910 that part of code looks different. Could you explain why?

pj1031999 (approved):
LGTM but please answer my questions.

include/sys/proc.h
```diff
@@ -13,7 +13,6 @@
 
 typedef struct thread thread_t;
 typedef struct proc proc_t;
-typedef struct cred cred_t;
```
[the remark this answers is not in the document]
j-piecuch: Because `sys/cred.h` already has it. It's an unnecessary change, so I have reverted it.

sys/kern/pty.c
j-piecuch: I forgot to merge the recent changes to #910. My bad.

franciscozdo (approved):
LGTM

include/util.h
```diff
@@ -2,6 +2,7 @@
 #define _UTIL_H_
 
 #include <sys/types.h>
+#include <sys/termios.h>
```
cahirwpz, line 5: The original header file performs forward declaration and does not include `termios.h`. I think it's better not to pollute visible symbol space.

cahirwpz (approved):
LGTM

PR #930, opened 2021-01-15
Fix gdb `kthread` command

j-piecuch:
Currently, `kthread <tid>` doesn't print the backtrace of the target thread, because the `Context` class was not updated when we simplified context handling.

- Update the `Context` class to reflect recent changes to context handling
- Handle case where regex doesn't match in `ProgramCounter::__str__()`


pj1031999 (approved):
LGTM

cahirwpz (approved):
LGTM

PR #934, opened 2021-01-18
pmap: fix bug with permissions

bkjg:
This pull request fix bug in pmap which allowed for example for writing to readonly memory.


sys/mips/pmap.c
```diff
@@ -307,6 +308,31 @@ static bool pmap_extract_nolock(pmap_t *pmap, vaddr_t va, paddr_t *pap) {
...
+  paddr_t old = *pap;
+
+  SCOPED_MTX_LOCK(&pmap->mtx);
+
+  bool result = pmap_extract_nolock(pmap, va, pap);
+
+  if (!result) {
+    *pap = old;
+    return false;
+  }
```
pj1031999, line 323: You don't need `result` here. You could call `pmap_extract_nolock` directly in `if` statement.

sys/mips/pmap.c
```diff
@@ -307,6 +308,31 @@ static bool pmap_extract_nolock(pmap_t *pmap, vaddr_t va, paddr_t *pap) {
   return true;
 }
 
+bool pmap_extract_and_hold(pmap_t *pmap, vaddr_t va, paddr_t *pap,
+                           vm_prot_t prot) {
+  assert(pap != NULL);
+  paddr_t old = *pap;
```
pj1031999, line 314: Do we really need to remember old value of `*pap`?
`pmap_extract_and_hold` is a kernel internal function so you don't give any information to user - if you could call `pmap_extract_and_hold` you can also call `pmap_extract`. We already give a information if `*pap` as valid by returning `true` or `false` from that function.

bin/utest/mmap.c
```diff
@@ -93,3 +96,38 @@ int test_mmap(void) {
...
+  }
+
+  assert(sigsegv_handled == 1);
+  assert(*(char *)addr == 0);
+
+  /* restore original behavior */
+  signal(SIGSEGV, SIG_DFL);
+
+  return 0;
+}
```
pj1031999, line 133: I think that you don't need non-local jumps here.

pj1031999 (changes requested):
I don't think that "Pmap fix" is a good name for that PR. Page table works before that PR. Only usage of _pmap_ was wrong in `tlb_exception_handler` and it's a reason for new interface. Please update name and write more useful description - it's not trivial change.

sys/mips/pmap.c
bkjg: ok. I changed it.

sys/mips/pmap.c
bkjg: Changed.

bin/utest/mmap.c
bkjg: I need it 'cause I have to handle sigsegv. Also I replaced setjmp and longjmp with sigsetjmp and siglongjmp

pj1031999 (changes requested):
Please fix errors on CLANG build.

`assertion "sigsegv_handled == 1" failed: file "/root/mimiker/bin/utest/mmap.c", line 126, function "test_mmap_readonly"`

bkjg:
> Please fix errors on CLANG build.
> 
> `assertion "sigsegv_handled == 1" failed: file "/root/mimiker/bin/utest/mmap.c", line 126, function "test_mmap_readonly"`

Fixed.

sys/mips/pmap.c
```diff
@@ -307,6 +308,26 @@ static bool pmap_extract_nolock(pmap_t *pmap, vaddr_t va, paddr_t *pap) {
...
+
+  pte_t pte = pmap_pte_read(pmap, va);
+
+  if ((prot & VM_PROT_READ && !(pte & PTE_READ)) ||
+      (prot & VM_PROT_WRITE && !(pte & PTE_WRITE))) {
+    return false;
+  }
+
+  return true;
+}
```
pj1031999, line 329: Do you really need to return pa? Maybe you can hide more details inside pmap module and don't expose them to other part of code. It should increase readability of `trap.c`.

bkjg: I think I need it. This physical address is read in `trap.c`, and then we look for the corresponding `vm_page_t`. Also returning pa allows to fix this bug only by changing as few as possible lines in `trap.c`.

pj1031999: Why you can't hide software bits emulation into pmap module?

bkjg: I don't understand what do you mean exactly. Could you explain?

bkjg: Changed.

PR #939, opened 2021-01-18
Introduce the mdelay function.

MichalBlk:
The `mdelay` function waits for a given amount of ms. It's devoted for use inside device drivers and any other field which requires accurate waiting at ms level.


sys/kern/time.c
```diff
@@ -47,4 +49,15 @@ time_t tm2sec(tm_t *t) {
...
+
+static void mdelay_timeout(__unused void *arg) {
+  /* Nothing to do here. */
+}
+
+void mdelay(useconds_t ms) {
+  callout_t handle;
+  bzero(&handle, sizeof(callout_t));
+  callout_setup_relative(&handle, ms, mdelay_timeout, NULL);
+  callout_drain(&handle);
```
wiklam, line 62: As far as I know, `mdelay` is busy waiting, and you implemented probably `msleep`.

sys/kern/time.c
```diff
@@ -47,4 +49,15 @@ time_t tm2sec(tm_t *t) {
...
+}
```
pj1031999, line 63: It's not clear for me. Argument for `mdelay` is ms (millisecond?) but `useconds_t` looks like microseconds. On the other hand second argument of `callout_setup_relative` is a `systime_t`. Could you explain?

mohrcore: I'm having similar thoughts on this one. I'm be waiting for an explanation. Other than that, the logic seems fine, but I can't accept it unless you clarify what's going on here.

cahirwpz:
@MichalBlk Since we don't have clear use case for this function, the code has rotten and is not eligible for merging anymore. Please close this PR.

Issue #932, opened 2021-01-17
Tests that don't work on AArch64.


pj1031999:
kernel tests:
- [x] pmap_kenter
- [x] rmbits
- [x] pmap_page_copy
- [x] producer_consumer - broken on MIPS
- [x] findspace
- [x] vmem

user tests:
- [x] mmap
- [x] munmap_sigsegv
- [x] sbrk
- [x] sbrk_sigsegv
- [x] signal_segfault
- [x] signal_cont_masked
- [x] signal_mask_nonmaskable
- [x] signal_sigsuspend
- [x] signal_sigsuspend_stop
- [x] signal_handler_mask
- [x] fork_sigchld_ignored
- [x] setjmp
- [x] sigaction_with_setjmp
- [x] sigaction_handler_returns
- [x] vfs_relative_dir
- [x] vfs_dot_dot_dir
- [x] vfs_dot_dot_across_fs
- [x] killpg_same_group
- [x] killpg_other_group
- [x] session_login_name
- [ ] gettimeofday
- [x] sharing_memory_simple
- [x] sharing_memory_child_and_grandchild
- [x] mmap_prot_none
- [x] mmap_prot_read

PR #941, opened 2021-01-19
AArch64: use virtual address instead of physical address in tlb_invalidate.

pj1031999:
Use virtual address instead of physical address in `tlb_invalidate`.
`TLBI VAE1IS` & `TLBI VAAE1IS` require virtual address as argument.

Fix kernel-space `pmap_page_copy` test for _AArch64_ - #932.

### DIFF
--- a/sys/aarch64/tlb.c
+++ b/sys/aarch64/tlb.c
@@ -9,6 +9,16 @@
 void tlb_invalidate(vaddr_t va, asid_t asid) {
   __dsb("ishst");
 
+  /*
+   * vae1is - Invalidate translation used at EL1 for the specified VA and
+   * Address Space Identifier (ASID) and the current VMID, Inner Shareable.
+   * vaae1is - Invalidate all translations used at EL1 for the specified
+   * address and current VMID and for all ASID values, Inner Shareable.
+   *
+   * Based on arm documentation
+   * [https://developer.arm.com/documentation/ddi0488/h/system-control/aarch64-register-summary/aarch64-tlb-maintenance-operations]
+   * and pmap_invalidate_page from FreeBSD - /sys/arm64/arm64/pmap.c.
+   */
   if (asid > 0) {
     __tlbi("vae1is", ASID_TO_PTE(asid) | (va >> PAGE_SHIFT));
   } else {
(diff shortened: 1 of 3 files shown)


sys/aarch64/tlb.c
```diff
@@ -6,13 +6,13 @@
 #define __dsb(x) __asm__ volatile("DSB " x)
 #define __isb() __asm__ volatile("ISB")
 
-void tlb_invalidate(pte_t pte, asid_t asid) {
+void tlb_invalidate(vaddr_t va, asid_t asid) {
   __dsb("ishst");
 
   if (asid > 0) {
-    __tlbi("vae1is", ASID_TO_PTE(asid) | (pte >> PAGE_SHIFT));
+    __tlbi("vae1is", ASID_TO_PTE(asid) | (va >> PAGE_SHIFT));
```
cahirwpz, line 13: Umm... can you explain what is the difference between `vae1is` and `vaae1is`, or at least point a reader to documentation.

pj1031999: `vae1is` - Invalidate translation used at EL1 for the specified VA and Address Space Identifier (ASID) and the current VMID, Inner Shareable.
`vaae1is` - Invalidate all translations used at EL1 for the specified address and current VMID and for all ASID values, Inner Shareable.

Based on [arm documentation](https://developer.arm.com/documentation/ddi0488/h/system-control/aarch64-register-summary/aarch64-tlb-maintenance-operations) and `pmap_invalidate_page` from FreeBSD - `/sys/arm64/arm64/pmap.c`

cahirwpz: Please put it in a comment it the source code.

pj1031999: done

cahirwpz: IIUC `asid = 0` only for kernel address space, and only kernel can use addresses that are translated with PT pointed by TTBR1. Why do we need to use `vaae1is` for kernel AS, instead of just `vae1s` ?

pj1031999: I don't know answer for your question. _NetBSD_, _FreeBSD_, _Linux_ and even _Fuchsia_ use `vaae1is` for kernel.
After removing `vaae1is` branch everything still works but I don't want to leave potential bug in that part of code for future.

cahirwpz: Ok. As usual - please put your findings into a comment in the code.

pj1031999: done

cahirwpz (approved):
LGTM

PR #944, opened 2021-01-20
Don't unwind the PCI bus interface.

MichalBlk:
There is no other device which can implement the PCI bus interface and a PCI request must be handled directly by device's parent PCI bus.


mohrcore:
I no longer know what is supposed to be dispatched.

The whole dispatching thing is a mess and there's nothing that would make dispatchable methods stand out. I "solved" this issue by making every method virtually dispatchable and moving the responsibility of handling non-sense scenarios to the to drivers. This way the entire infrastructure remains consistent. With that said, I don't like the changes you proposed, but I fully understand why do you propose them.

As for now, i'll let it with no explicit rejection or approval. This PR attempts to simpliofy some code and I'd happhappily approve such thing if there was a clear way of telling what methods are dispatchable (eg. a change their naming conventions).

pj1031999 (commented):
For me code looks good but I am not sure about motivation. Could you provide example where old behavior isn't good?

cahirwpz:
Since this PR does not solve any important problem we're facing I decided to close this PR. Perhaps we'll revisit it in the future.

PR #946, opened 2021-01-21
Sticky bit

franciscozdo:
This PR will add sticky bit feature to Mimiker.

When sticky bit is set on directory, removing files (and directories) in it is restricted to root, directory owner and file owner.

Also marked `/tmp` with this bit, as it is in other systems.


sys/kern/vfs_syscalls.c
```diff
@@ -237,6 +237,18 @@ int do_getdents(proc_t *p, int fd, uio_t *uio) {
   return error;
 }
 
+static int vfs_check_remove(vnode_t *dvp, vnode_t *vp, cred_t *cred) {
+  vattr_t dva;
+  accmode_t diracc = VWRITE;
+  int error;
+  VOP_GETATTR(dvp, &dva);
```
xThaid, line 244: You can consider checking if `VOP_GETATTR` fails.

include/sys/vnode.h
```diff
@@ -24,7 +24,7 @@ typedef struct componentname componentname_t;
 #define VNOVAL (-1)
 
 /* vnode access modes */
-typedef enum { VEXEC = 1, VWRITE = 2, VREAD = 4 } accmode_t;
+typedef enum { VEXEC = 1, VWRITE = 2, VREAD = 4, VADMIN = 8 } accmode_t;
```
xThaid, line 27: Could you add a short comment describing that new mode? Something simple like "being the file owner" should be more explanatory than `VADMIN` alone.

sys/kern/vfs_syscalls.c
franciscozdo: I forgot about that. Thanks

include/sys/vnode.h
franciscozdo: Done.

j-piecuch (approved):

xThaid (approved):

sys/kern/tmpfs.c
```diff
@@ -965,7 +965,7 @@ static int tmpfs_mount(mount_t *mp) {
   /* Allocate the root node. */
   vattr_t va;
   vattr_null(&va);
-  va.va_mode = S_IFDIR | ACCESSPERMS;
+  va.va_mode = S_IFDIR | ACCESSPERMS | S_ISTXT;
```
cahirwpz, line 968: Since `tmpfs` in theory could be mounted more than once, not every instance will be mounted on `/tmp`. Hence please do not add this attribute here, but rather `do_chmod` in `mount_fs`.

PR #945, opened 2021-01-21
Login program

franciscozdo:
This PR tries to port login from NetBSD.

I have marked every part of code that uses not implemented features with `NOT_IMPLEMENTED` macro.

(Changes needed by login were done here ->  #949)

PR #949, opened 2021-01-22
Login: changes in libc and headers

franciscozdo:
I added `sysexits.h` and `execlp()` functions.

I had to add macro `_KERNEL` in `file.h` to allow using it in userspace. (e.g `sys/mutex.h` cannot be included in userspace programs)

This changes are required by #945


cahirwpz (approved):
LGTM

PR #950, opened 2021-01-22
Allow callouts to be rescheduled from the callback

j-piecuch:
The major change is the addition of the `callout_reschedule()` function, which is used to reschedule a callout from its callback. It can be used to implement interval timers.


sys/kern/callout.c
```diff
@@ -62,9 +71,19 @@ static void callout_thread(void *arg) {
...
-    callout_clear_active(elem);
-    /* Wake threads that wait for execution of this callout in callout_drain. */
-    sleepq_broadcast(elem);
+
+    WITH_SPIN_LOCK (&ci.lock) {
+      callout_clear_active(elem);
+      if (elem->c_interval) {
+        /* Periodic callout: reschedule. */
+        elem->c_time += elem->c_interval;
+        _callout_schedule(elem);
```
wiklam, line 80: It shouldn't happen, but I have been thinking about a situation, when `elem->c_time + n * elem->c_interval < actual time` and we will schedule it n times knowing that the time already passed. Should we worry about such case?

j-piecuch: I don't think rescheduling to the nearest interval multiple in the future is inherently better than rescheduling to a missed deadline: to me it's a matter of choice.

wiklam (approved):

pj1031999 (approved):
LGTM

wiklam (changes requested):
Just checking GH Actions.

wiklam (approved):
LGTM, sorry for the struggle

wiklam (approved):
LGTM

cahirwpz (approved):

PR #951, opened 2021-01-22
Add various library routines required to port libterminfo

j-piecuch:
- Add `mi_vector_hash()` from NetBSD into libc
- Add endianness encoding/decoding functions from NetBSD to `sys/endian.h`
- Add `sys/bitops.h` from NetBSD
- Add `cdbr.c` and `cdbw.c` from NetBSD
They implement a simple persistent key-value database that `libterminfo` uses to store terminal capabilities


include/sys/types.h
```diff
@@ -2,9 +2,9 @@
 #define _SYS_TYPES_H_
 
 #include <sys/cdefs.h>
-#include <endian.h>
 #include <stdint.h> /* uint*_t, int*_t */
 #include <stddef.h> /* offsetof, NULL, ptrdiff_t, size_t, etc. */
+#include <endian.h>
```
franciscozdo, line 7: Moved by mistake?

franciscozdo (approved):
LGTM

Have you done any big changes to files imported from BSD?

include/sys/types.h
j-piecuch: Not really, it was because of a cyclic dependency: `sys/types.h` -> `endian.h` -> `sys/endian.h` -> `sys/types.h`. Now I've fixed it by not including `sys/endian.h` in `sys/endian.h`.

j-piecuch:
> Have you done any big changes to files imported from BSD?

I can't recall making any big changes, but I replaced a use of `arc4random()` (which we don't have) with `random()` in `cdbw.c` by adding `cdbw_random_seeder()`.

bkjg (approved):
LGTM

cahirwpz (approved):
LGTM

PR #952, opened 2021-01-22
test_sleepq_timed(): fix race on shared variable.

j-piecuch:
The variable `timed_received` is concurrently incremented by multiple threads, so we need to use atomics in order to avoid a race.


cahirwpz (approved):
LGTM

PR #953, opened 2021-01-22
Start kernel malloc earlier.

cahirwpz:
* Manage `vmem_t` and `pool_t` memory with `kmalloc`.
 * Create first kernel malloc arena statically.


pj1031999 (approved):
LGTM

PR #956, opened 2021-01-23
Fix lock order in pmap

bkjg:
Different order of acquiring the pv_list_lock and pmap_t::mtx can lead to deadlock. This PR choose one particular order and fix the order in all functions.


xThaid (approved):
Looks good.

cahirwpz (approved):
This is ok as a temporary solution. However we'll need to carefully review our locking strategy. We'll do it when a set of tools will be implemented by @xThaid .

PR #957, opened 2021-01-23
klog: fix race

xThaid:
There is a harmless race on the variable `klog.mask`. Although it is set under a spin lock, there is no synchronization when reading it.

Found by KCSAN.


pj1031999 (approved):
LGTM

xThaid:
I force pushed the branch to re-trigger the CI.

wiklam (approved):
LGTM

PR #958, opened 2021-01-24
proc: zero session structs before use

xThaid:
Sessions struct weren't zeroed on creation, so the field `tty_t *s_tty` could contain a garbage and thus every check like [this](https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/sys/kern/tty.c#1091) or [this](https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/sys/kern/proc.c?r=1a7dabd3#362) wasn't working as expected.


pj1031999 (approved):
LGTM

j-piecuch (approved):

PR #959, opened 2021-01-24
Increase KTEST_NAME_MAX to 50.

j-piecuch:
We have a test whose name (including the null terminator) is longer than 40: `user_sharing_memory_child_and_grandchild`.

```
>>> len("user_sharing_memory_child_and_grandchild")
40
```


bkjg (approved):
LGTM

cahirwpz (approved):
LGTM

PR #962, opened 2021-01-25
Allow fstat() on file descriptors without read access

franciscozdo:
`do_fstat` was using `fdtab_get_fd` with flag `FF_RED` for no reason.

needed by #945


pj1031999 (approved):
LGTM

cahirwpz:
Umm... can you actually provide a test that shows the change is necessary?

pj1031999:
@cahirwpz you need that for `stdout` flush in login. See `__smakebuf` function in libc.

j-piecuch (approved):

PR #963, opened 2021-01-25
Allow fstat() on file descriptors without read access.

j-piecuch:
Since we don't have read access to `stdout`, `fstat()` on `STDOUT_FILENO` would fail.
As a consequence, `__swhatbuf()` concludes that `stdout` isn't a tty, and we set full buffering instead of line buffering on `stdout`.
The intended behavior is that `stdout` is line buffered by default if it's a tty.


pj1031999:
Please see #962.

j-piecuch:
> Please see #962.

Oops, my bad :smile:

PR #964, opened 2021-01-25
vm_segment_destroy_range: fix offsets in pages

bkjg:
In vm_segment_destroy_range we change the start of the segment given as an argument. Unfortunately we don't change the offsets in pages in vm_segment_t::object. This PR fix it.


cahirwpz:
IIUC `pg->offset` means an offset from start of mapped object. In case of a memory-mapped file decreasing pointer would mean that we truncated some prefix of the file. Let's use analogy of a window. VM segment is a window through which we see some memory. If P pointer (located within a segment) references to M before truncation and P is still within the segment after segment, it must still point to M.

Am I missing something? Please note that [vm_map_entry](http://bxr.su/FreeBSD/sys/vm/vm_map.h#109) has `offset` field.

PR #965, opened 2021-01-26
Add per-test timeouts using callouts.

j-piecuch:
This PR adds a per-test timeout that makes sure that no single test takes more than a specified time. This will allow us to detect deadlocks faster, and increase the timeout for the whole test run, as we have recently been hitting the time limit even without apparent deadlocks, probably due to CircleCI throttling the CPU.
- Add a per test-timeout using callouts
- Increase the whole test run timeout to 60 seconds


franciscozdo (approved):
LGTM

cahirwpz:
Unfortunately this PR has rotten beyond repair. Moreover this functionality doesn't seem to be in high demand. We've been doing ok without it so far.

PR #966, opened 2021-01-26
AArch64: set user-space return address in setcontext

pj1031999:
* Copy correct data in `longjmp`.
* Set return address for user-space in `do_setcontext`.
* Fix most tests from #932.


lib/libc/gen/aarch64/longjmp.c
```diff
@@ -43,9 +43,9 @@ void longjmp(jmp_buf env, int val) {
   _REG(&uc, X28) = _REG(sc_uc, X28);
   _REG(&uc, X29) = _REG(sc_uc, X29);
 
-  _REG(&uc, SP) = _REG(&uc, SP);
-  _REG(&uc, LR) = _REG(&uc, LR);
-  _REG(&uc, PC) = _REG(&uc, PC);
+  _REG(&uc, SP) = _REG(sc_uc, SP);
```
cahirwpz, line 46: No wonder it did not work :D

cahirwpz (approved):
LGTM

PR #967, opened 2021-01-26
Migrate from CircleCI to GitHub Actions!

cahirwpz:
Since we're regularly running out of time on CircleCI it's high time to move to GH Actions but provide our own runner.


.github/workflows/default.yml
```diff
@@ -0,0 +1,34 @@
+name: Mimiker
+
+on: [push, pull_request]
```
cahirwpz, line 3: I think we need to disable `pull_request` otherwise _GH Actions_ starts whole run twice.

pj1031999: done

.github/workflows/default.yml
```diff
@@ -0,0 +1,88 @@
...
+      - uses: actions/checkout@v2
+      - run: make -j $(nproc) BOARD=rpi3 CLANG=1
+
+  kernel_tests_mips_gcc:
+    name: Tests MIPS - GCC with KASAN
+    runs-on: self-hosted
+    container: cahirwpz/mimiker-circleci:1.7.3
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/download-artifact@v2
```
cahirwpz, line 74: You have an implicit dependency that fails if a runner doesn't pick up `build_mips_gcc` earlier.

.github/workflows/default.yml
```diff
@@ -0,0 +1,88 @@
...
+          name: "mips gcc kernel+ramdisk"
+      - run: ./run_tests.py --board malta --thorough --non-interactive
+
+  kernel_tests_mips_clang:
+    name: Tests MIPS - Clang
+    runs-on: self-hosted
+    container: cahirwpz/mimiker-circleci:1.7.3
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/download-artifact@v2
```
cahirwpz, line 85: Ditto.

cahirwpz (commented):
I've launched four runners on `mimiker-ci`. That enabled job parallelism and caused problems with job dependencies that were not stated explicitly. Please fix that.

.github/workflows/default.yml
cahirwpz: Done.

.github/workflows/default.yml
cahirwpz: Done.

run_tests.py
```diff
@@ -37,11 +36,11 @@ def gdb_inspect(interactive):
     gdb_cmd = getvar('gdb.binary')
     if interactive:
         gdb_opts = ['-iex=set auto-load safe-path {}/'.format(os.getcwd()),
-                    '-ex=target remote localhost:%d' % gdb_port(),
+                    '-ex=target remote localhost:%d' % RandomPort(),
```
cahirwpz, line 39: I've a gut feeling it won't work for some reason... but we'll fix it later.

PR #969, opened 2021-01-27
fchdir: release old cwd vnode

xThaid:
We should drop the reference to the old cwd and acquire the reference to the new one while changing directory.

There was a bug causing that the new cwd was released twice.


cahirwpz (approved):
LGTM

xThaid:
That's right. It wasn't correct.

cahirwpz (approved):
LGTM

PR #970, opened 2021-01-28
tmpfs: Add file access, modify and change times

xThaid:
Since we have a reliable source of time, we can add support for access, modify, change times.


sys/kern/tmpfs.c
```diff
@@ -945,9 +992,38 @@ static int tmpfs_resize(tmpfs_mount_t *tfm, tmpfs_node_t *v, size_t newsize) {
   }
 
   v->tfn_size = newsize;
+  tmpfs_update_time(v, TMPFS_UPDATE_CTIME | TMPFS_UPDATE_MTIME);
   return 0;
 }
 
+static int tmpfs_chtimes(tmpfs_node_t *v, timespec_t *atime, timespec_t *mtime,
```
franciscozdo, line 999: Why don't you update ctime in this function?

franciscozdo (changes requested):
I think you've lost one time update.

sys/kern/tmpfs.c
xThaid: I'm not really sure what you are asking about, because there are two things. The first is that the ctime cannot be changed with this interface, but the second one is that you're right I forgot to update `ctime`.

franciscozdo: The second think is what I mean.

franciscozdo (approved):
LGTM

sys/kern/tmpfs.c
```diff
@@ -111,6 +115,12 @@ typedef struct tmpfs_node {
   };
 } tmpfs_node_t;
 
+enum tmpfs_time_type {
```
bkjg, line 118: I think the code style of `mimiker/sys/kern` directory would suggests here `typedef enum` instead of just `enum`.

bkjg (approved):
One little comment about code style but beside that, LGTM.

sys/kern/tmpfs.c
xThaid: good remark

cahirwpz (commented):
LGTM

PR #971, opened 2021-01-29
Fully control test-run using gdb.

cahirwpz:
It's high time for test run not to depend on output from kernel console but rather on gdb behaviour.


cahirwpz (commented):
It's ok for the time being. We'll improve on it later.

PR #972, opened 2021-01-30
syscalls: add lchown syscall

xThaid:
I saw it was waiting in the `missing.S` file, so I decided to implement it, since it is just a wrapper of the existing system call.


pj1031999 (approved):
LGTM

cahirwpz (approved):
LGTM

PR #974, opened 2021-01-31
Dev procstat

franciscozdo:
Simple utility to list running processes


sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,125 @@
...
+}
+
+static fileops_t dev_procstat_fileops = {
+  .fo_read = dev_procstat_read,
+  .fo_close = default_vnclose,
+  .fo_stat = default_vnstat,
+  .fo_ioctl = default_vnioctl,
+};
+
+static int write_to_buf(char *buf, size_t max_len, proc_t *p) {
```
bkjg, line 60: Do we make an assumption that this function will always be called with `buf` not equal to null? What will happen in some time in the future someone will use this function with null pointer by mistake?

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,125 @@
...
+    TAILQ_FOREACH (p, zombie_list, p_zombie) {
+      ret = write_to_buf(pstat->buf + offset, BUF_SIZE - offset, p);
+      if (ret < 0)
+        goto err;
+      offset += ret;
+    }
+  }
+  pstat->len = offset;
+
+  if ((error = vnode_open_generic(v, mode, fp)))
```
bkjg, line 99: I think this double parentheses aren't necessary and single parentheses would be enough.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,125 @@
...
+  fp->f_ops = &dev_procstat_fileops;
+  fp->f_data = pstat;
+  return 0;
+
+err:
+  kfree(M_TEMP, pstat);
+  return error;
+}
+
+static int dev_procstat_close(vnode_t *v, file_t *fp) {
```
bkjg, line 111: Why do we need to pass `vnode_t *v` as an argument if we won't use it?

bkjg (changes requested):

sys/kern/dev_procstat.c
pj1031999: They are necessary. If you use single parentheses you will get ` error: suggest parentheses around assignment used as truth value`.

sys/kern/dev_procstat.c
pj1031999: `vnodeops_t::v_close` is `typedef int vnode_close_t(vnode_t *v, file_t *fp);` so you need to set first argument as `vnode_t*`.

sys/kern/dev_procstat.c
pj1031999: It's ok. We do that in private interfaces.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,125 @@
...
+ * During read calls to /dev/procstat this info can be read.
+ *
+ * Example:
+ * euid   pid    ppid    pgrp    state    td_name  elfpath
+ * 0       1       0       1       R       init    /bin/ksh
+ *
+ */
+
+#define PSTAT_SIZE 1024
+#define BUF_SIZE (PSTAT_SIZE - sizeof(int))
```
pj1031999, line 25: `pstat_t::len` is `size_t` so if I understand correctly you should use `size_t` here instated of `int`.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,125 @@
...
+
+static int dev_procstat_open(vnode_t *v, int mode, file_t *fp) {
+  pstat_t *pstat = kmalloc(M_TEMP, PSTAT_SIZE, 0);
+  proc_list_t *zombie_list, *proc_list;
+  proc_t *p;
+  size_t offset = 0;
+  int ret, error = EFAULT;
+
+  if (pstat == NULL)
+    return EFAULT;
```
pj1031999, line 78: Why not `ENOMEM`?
- `ENOMEM` Cannot allocate memory.
- `EFAULT` Bad address.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,125 @@
...
+static fileops_t dev_procstat_fileops = {
+  .fo_read = dev_procstat_read,
+  .fo_close = default_vnclose,
+  .fo_stat = default_vnstat,
+  .fo_ioctl = default_vnioctl,
+};
+
+static int write_to_buf(char *buf, size_t max_len, proc_t *p) {
+  if (p->p_pid == 0)
+    return 0;
```
pj1031999, line 62: Is it possible?

franciscozdo: Yes, there exists `proc0`. I think we don't want to show it on list of processes.

sys/kern/dev_procstat.c
franciscozdo: yes, thanks

sys/kern/dev_procstat.c
franciscozdo: Yes, it is better

sys/kern/dev_procstat.c
pj1031999: Okay... Now I understand. It wasn't obvious for me to put that logic inside `write_to_buf`. From my point of view that function should do one simple thing - writing data into buffer, but do whatever is more clean solution for you.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,130 @@
...
+} pstat_t;
+
+static int dev_procstat_read(file_t *f, uio_t *uio) {
+  pstat_t *pst = f->f_data;
+
+  if ((size_t)f->f_offset >= pst->len)
+    return 0;
+
+  uio->uio_offset = f->f_offset;
+  size_t len = MIN(uio->uio_resid, pst->len);
```
xThaid, line 46: Shouldn't it be more like `MIN(uio->uio_resid, pst->len - uio->uio_offset);` ?

franciscozdo: Yes. Thanks!

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,137 @@
...
+ * XXX: Each instance of /dev/procstat has 4K bytes for process entries.
+ * It will be enough for about 40 processes (if we assume average size of entry
+ * to be 100 characters).
+ *
+ * Until we run more processes than ksh and some tests, this solution don't need
+ * to be more complex.
+ */
+
+#define PSTAT_SIZE 4096
+#define BUF_SIZE (PSTAT_SIZE - sizeof(size_t))
```
xThaid, line 31: not so important: maybe it would be better to subtract `sizeof(pstat_t)`, so the macro won't be broken if somebody will add any extra field in the struct.

xThaid (approved):
Despite one minor thing looks good. Thanks!

sys/kern/dev_procstat.c
franciscozdo: Changed

pj1031999 (approved):
LGTM

include/sys/proc.h
```diff
@@ -193,6 +193,8 @@ int do_fork(void (*start)(void *), void *arg, pid_t *cldpidp);
 /*! \brief Set login name associated with current session. */
 int do_setlogin(const char *name);
 
+void get_proc_lists(proc_list_t **ps, proc_list_t **zs);
```
cahirwpz, line 196: I'm in favor of making both list globally available. There's no need to introduce this weird getter.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,137 @@
...
+
+  SCOPED_MTX_LOCK(&p->p_lock);
+  return snprintf(buf, max_len, "%d\t%d\t%d\t%d\t%d\t%c\t%s\t%s\n",
+                  p->p_cred.cr_euid, p->p_pid, p->p_parent->p_pid,
+                  p->p_pgrp->pg_id, p->p_pgrp->pg_session->s_sid,
+                  state[p->p_state], p->p_thread->td_name, p->p_elfpath);
+}
+
+static int dev_procstat_open(vnode_t *v, int mode, file_t *fp) {
+  pstat_t *pstat = kmalloc(M_TEMP, PSTAT_SIZE, 0);
```
cahirwpz, line 83: This looks weird. Please move call to `kmalloc` below definition block.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,137 @@
...
+  size_t offset = 0;
+  int ret, error = ENOMEM;
+
+  if (pstat == NULL)
+    return ENOMEM;
+
+  get_proc_lists(&proc_list, &zombie_list);
+
+  WITH_MTX_LOCK (all_proc_mtx) {
+    TAILQ_FOREACH (p, proc_list, p_all) {
```
cahirwpz, line 95: This cries for a small subprocedure. You violated DRY principle ;-)

I'm somewhat surprised that @xThaid @pj1031999 @bkjg have not made a remark about it.

Makefile
```diff
@@ -3,7 +3,7 @@
 TOPDIR = $(CURDIR)
 
 # Directories which require calling make recursively
-SUBDIR = sys lib bin usr.bin etc include
+SUBDIR = sys lib bin usr.bin etc include scripts
```
cahirwpz, line 6: I don't think we need to introduce new subdirectory just for scripts. Please move our `ps` to `bin/ps` where it belongs to.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * We also have to handle 'proc0' here. It is on proc_list but we
+ * don't want to show it to user.
+ */
+static int write_to_buf(char *buf, size_t max_len, proc_t *p) {
+  if (p->p_pid == 0)
+    return 0;
+
+  SCOPED_MTX_LOCK(&p->p_lock);
+  return snprintf(buf, max_len, "%d\t%d\t%d\t%d\t%d\t%c\t%s\t%s\n",
```
cahirwpz, line 76: I wonder... why don't you:
 * devise a table of structures (let's say `procstat_t`),
 * fill it with data from `proc_t` (that's going to be really fast),
 * convert `procstat_t` to text when user reads the file.

It'll be an extra step, but you can easily estimate how much memory you need to store `procstat_t` array.

franciscozdo: It sounds good. There is one problem. When user calls read it can end inside one table entry. It will bring some weird logic, but I think it could be done.

sys/kern/dev_procstat.c
franciscozdo: There is small problem with iterating over this lists. They use different `TAILQ_ENTRY` (`p_all` and `p_zombie`). Loops must remain. Other things i will move to separate function.

include/sys/proc.h
franciscozdo: done

sys/kern/dev_procstat.c
franciscozdo: done

Makefile
franciscozdo: done

bin/ps/Makefile
```diff
@@ -0,0 +1,12 @@
...
+PROGRAM = ps
+BINMODE = 0755
+INSTALL-FILES = $(SYSROOT)/$(BINDIR)/$(PROGRAM)
+BINDIR = $(shell echo $(DIR) | cut -f 1 -d / | tr . /)
+
+include $(TOPDIR)/build/common.mk
+
+$(SYSROOT)/$(BINDIR)/$(PROGRAM): $(PROGRAM)
+	@echo "[INSTALL] $(DIR)$< -> /$(BINDIR)/$(PROGRAM)"
+	$(INSTALL) -D $(PROGRAM) --mode=$(BINMODE) $(SYSROOT)/$(BINDIR)/$(PROGRAM)
```
pj1031999, line 12: Could you use generic rules from `/build` like in other programs under `/bin`?

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * During read calls to /dev/procstat this info can be read.
+ *
+ * Example:
+ * euid   pid    ppid    pgrp   session  state    td_name  elfpath
+ * 0       1       0       1       1       R       init    /bin/ksh
+ */
+
+/* maximum size of output string of proc info */
+#define MAX_P_STRING (TD_NAME_MAX + PATH_MAX + 5 * 10 + 1 + 10)
```
pj1031999, line 23: `5 * 10 + 1 + 10` is not clear.

pj1031999 (approved):

bin/ps/Makefile
xThaid, line 12: Maybe this can be move to other, common file like `build.script.mk`?

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,175 @@
...
+
+typedef struct proc_info {
+  pid_t pid;
+  pid_t ppid;
+  uid_t uid;
+  pgid_t pgrp;
+  sid_t sid;
+  proc_state_t state;
+  char td_name[TD_NAME_MAX];
+  char elfpath[PATH_MAX];
```
xThaid, line 43: What about allocating the `elfpath` dynamically (using for example `kstrndup`)?

bin/ps/Makefile
franciscozdo: I made generic rules for scripts as @xThaid suggested.

sys/kern/dev_procstat.c
franciscozdo: I described that.

sys/kern/dev_procstat.c
franciscozdo: I was a bit scared about doing it dynamically because we hold lock here. I changed that and it works. And I think it should because proc lock does nothing with malloc. Could you confirm that?

pj1031999: It's safe to do that.

xThaid (approved):

cahirwpz (changes requested):
There's at least one memory leak in this change, that should have been identified by previous reviewers @pj1031999 @xThaid ;-) Correct me if I'm wrong.

There's no limit on opened instances of `procstat` file which could in theory cause _denial of service_ type of error. You should limit that in `dev_prostat_open` to some reasonable (but arbitrarily chosen) number (10?).

pj1031999:
@cahirwpz `kstrndup` was added after my approval ;)

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,175 @@
...
+/* must be called with proc_t::p_lock held */
+static void fill_proc_info(proc_t *p, proc_info_t *pi) {
+  pi->uid = p->p_cred.cr_euid;
+  pi->pid = p->p_pid;
+  pi->ppid = p->p_parent->p_pid;
+  pi->pgrp = p->p_pgrp->pg_id;
+  pi->sid = p->p_pgrp->pg_session->s_sid;
+  pi->state = p->p_state;
+  pi->elfpath = kstrndup(M_TEMP, p->p_elfpath, PATH_MAX);
+  strncpy(pi->td_name, p->p_thread->td_name, TD_NAME_MAX);
```
pj1031999, line 82: Memory leak.

pj1031999 (changes requested):

sys/kern/dev_procstat.c
franciscozdo: fixed

pj1031999:
Running `ps` multiple times causes kernel panic.

backtrace from KASAN build:

```
#0  halt () at /home/pjas/mimiker-procstat/sys/kern/klog.c:154
#1  0xc0146768 in klog_panic (origin=origin@entry=KL_VM, file=file@entry=0xc01693a0 "/home/pjas/mimiker-procstat/sys/mips/trap.c", line=line@entry=0x7b, format=format@entry=0xc01695e0 "KERNEL PANIC!!!", arg1=arg1@
entry=0x0, arg2=arg2@entry=0x0, arg3=arg3@entry=0x0, arg4=arg4@entry=0x0, arg5=arg5@entry=0x0, arg6=arg6@entry=0x0) at /home/pjas/mimiker-procstat/sys/kern/klog.c:166
#2  0xc010281c in kernel_oops (ctx=ctx@entry=0xc0035b7c) at /home/pjas/mimiker-procstat/sys/mips/trap.c:123
#3  0xc0102bcc in kern_trap_handler (ctx=ctx@entry=0xc0035b7c) at /home/pjas/mimiker-procstat/sys/mips/trap.c:251
#4  0xc0103174 in mips_exc_handler (ctx=0xc0035b7c) at /home/pjas/mimiker-procstat/sys/mips/trap.c:274
#5  0x801008e4 in kern_exc_enter () at /home/pjas/mimiker-procstat/sys/mips/ebase.S:429
#6  0xc01072ec in pmap_delete (pmap=0xc0019178) at /home/pjas/mimiker-procstat/sys/mips/pmap.c:534
#7  0xc0117284 in vm_map_delete (map=0xc0018140) at /home/pjas/mimiker-procstat/sys/kern/vm_map.c:194
#8  0xc013dda8 in proc_exit (exitstatus=exitstatus@entry=0x0) at /home/pjas/mimiker-procstat/sys/kern/proc.c:582
#9  0xc0132364 in sys_exit (p=0xc001b4f0, args=<optimized out>, res=<optimized out>) at /home/pjas/mimiker-procstat/sys/kern/syscalls.c:46
#10 0xc0102d2c in syscall_handler (ctx=ctx@entry=0xc0035ed8, result=result@entry=0xc0035e60) at /home/pjas/mimiker-procstat/sys/mips/trap.c:59
#11 0xc0102fa4 in user_trap_handler (ctx=ctx@entry=0xc0035ed8) at /home/pjas/mimiker-procstat/sys/mips/trap.c:202
#12 0xc01030ec in mips_exc_handler (ctx=0xc0035ed8) at /home/pjas/mimiker-procstat/sys/mips/trap.c:272
#13 <signal handler called>
#14 0x004331f0 in _exit () at /home/pjas/mimiker-procstat/lib/libc/sys/syscalls.S:3
#15 0x00437a00 in exit (status=0x0) at /home/pjas/mimiker-procstat/lib/libc/stdlib/exit.c:44
#16 0x0041e804 in quitenv () at /home/pjas/mimiker-procstat/bin/ksh/main.c:747
#17 0x0041e92c in unwind (i=i@entry=0x4) at /home/pjas/mimiker-procstat/bin/ksh/main.c:681
#18 0x0041eb00 in shell (s=s@entry=0x8001088, toplevel=toplevel@entry=0x1) at /home/pjas/mimiker-procstat/bin/ksh/main.c:586
#19 0x0041f824 in main (argc=<optimized out>, argc@entry=0x2, argv=argv@entry=0x7f7fffb4) at /home/pjas/mimiker-procstat/bin/ksh/main.c:447
#20 0x004331e0 in ___start (argc=0x2, argv=0x7f7fffb4, envp=<optimized out>) at /home/pjas/mimiker-procstat/lib/csu/crt0-common.c:18
#21 0x00000000 in ?? ()
```

kthread:
```
+-------+--------------------+--------------+----------+-------------------+
|  Id   |        Name        |    State     | Priority |   Waiting Point   |
+=======+====================+==============+==========+===================+
|     0 |        idle-thread |    TDS_READY |      255 | null              |
|     1 |            callout | TDS_SLEEPING |       32 | kern/callout.c:57 |
|     2 |               init | TDS_SLEEPING |      255 | 0xc0174dc0        |
|     3 | ns16550-tty-worker | TDS_SLEEPING |        3 | drv/ns16550.c:167 |
|     4 |               init | TDS_SLEEPING |      255 | 0xc0174dc0        |
| (*) 5 |               init |  TDS_RUNNING |      255 | null              |
|     6 |               init | TDS_SLEEPING |      255 | 0xc0174dc0        |
|     7 |               init | TDS_SLEEPING |      255 | 0xc0174dc0        |
|     8 |               init | TDS_SLEEPING |      255 | 0xc0174dc0        |
|     9 |               init | TDS_SLEEPING |      255 | 0xc0174dc0        |
|    10 |               init |    TDS_READY |      255 | null              |
|    11 |               init |    TDS_READY |      255 | null              |
|    12 |               init |    TDS_READY |      255 | null              |
|    13 |               init |    TDS_READY |      255 | null              |
|    14 |               init |    TDS_READY |      255 | null              |
|    15 |               init |     TDS_DEAD |      255 | null              |
|    16 |               init |    TDS_READY |      255 | null              |
|    17 |               init |    TDS_READY |      255 | null              |
|    18 |               init |    TDS_READY |      255 | null              |
|    19 |               init |    TDS_READY |      255 | null              |
|    20 |               init |    TDS_READY |      255 | null              |
+-------+--------------------+--------------+----------+-------------------+
```

kproc:
```
+-----+----------------------------------------+-----------+
| Pid |                  Tid                   |   State   |
+=====+========================================+===========+
|   0 | thread{idle-thread/0}                  | PS_NORMAL |
|   1 | thread{init/2}                         | PS_NORMAL |
|   2 | thread{init/4}                         | PS_NORMAL |
|   3 | thread{[bad-ptr 0x38]/[bad-ptr: 0x3c]} | PS_DYING  |
|   4 | thread{init/6}                         | PS_NORMAL |
|   5 | thread{init/7}                         | PS_NORMAL |
|   6 | thread{init/8}                         | PS_NORMAL |
|   7 | thread{init/9}                         | PS_NORMAL |
|   8 | thread{init/10}                        | PS_NORMAL |
|   9 | thread{init/11}                        | PS_NORMAL |
|  10 | thread{init/12}                        | PS_NORMAL |
|  11 | thread{init/13}                        | PS_NORMAL |
|  12 | thread{init/14}                        | PS_NORMAL |
|  14 | thread{init/16}                        | PS_NORMAL |
|  15 | thread{init/17}                        | PS_NORMAL |
|  16 | thread{init/18}                        | PS_NORMAL |
|  17 | thread{init/19}                        | PS_NORMAL |
|  18 | thread{init/20}                        | PS_NORMAL |
+-----+----------------------------------------+-----------+
```

franciscozdo:
@pj1031999 I got almost identical backtrace in master for command `ls & ls & ...`

j-piecuch:
The crashes are most likely caused by a kernel stack overflow. I cannot reproduce these on a KASAN build with kernel stack size = `2*PAGESIZE`.

pj1031999 (approved):

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,194 @@
...
+    TAILQ_FOREACH (p, &zombie_list, p_zombie) {
+      if (ps->nproc >= MAX_PROC)
+        goto out;
+      ps_entry_fill(&ps->ps[ps->nproc++], p);
+    }
+  }
+
+out:
+  if ((error = vnode_open_generic(v, mode, fp)))
+    return error;
```
j-piecuch, line 135: You also need to free all allocated memory (`ps` and each entry's `elfpath`) before returning the error.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,194 @@
...
+  ps->psize = ps_entry_tostring(ps->buf, &ps->ps[0]);
+  ps->offset = 0;
+  ps->cur_proc = 0;
+
+  fp->f_ops = &dev_procstat_fileops;
+  fp->f_data = ps;
+  return 0;
+}
+
+static int dev_procstat_read(file_t *f, uio_t *uio) {
```
j-piecuch, line 147: Since this is a file operation and not a vnode operation, there's no implicit synchronization between `read()` calls on the same open file. If a process forks while holding an open file descriptor to `/dev/procstat`, you can have two different processes calling `read()` concurrently on the same `file_t`. You need to synchronize them.

j-piecuch (changes requested):

sys/kern/dev_procstat.c
franciscozdo: Don't need that. I moved `vnode_open_generic` before any memory is allocated.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,192 @@
...
+static int dev_procstat_open(vnode_t *v, int mode, file_t *fp) {
+  WITH_MTX_LOCK (&procstat_lock) {
+    if (ps_opencnt >= MAX_PROCSTAT)
+      return EMFILE;
+    ps_opencnt++;
+    assert(ps_opencnt <= MAX_PROCSTAT);
+  }
+
+  int error;
+  if ((error = vnode_open_generic(v, mode, fp)))
```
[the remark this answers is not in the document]
franciscozdo: Thanks. Now i drop this reference after fail during memory allocation.

sys/kern/dev_procstat.c
franciscozdo: Added `ps_buf_t::lock` to synchornize reads.

sys/kern/dev_procstat.c
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  int error = 0;
+  proc_t *p;
+  ps_buf_t *ps;
+
+  if ((error = vnode_open_generic(v, mode, fp)))
+    goto err;
+
+  ps = kmalloc(M_TEMP, MAX_PROC * sizeof(ps_entry_t) + sizeof(ps_buf_t), 0);
```
j-piecuch, line 116: Actually, I looked at `kmalloc()` and it looks like it can't return `NULL` if you don't pass `M_NOWAIT`, so there's no need to check for `NULL` here. This should simplify things a bit.

j-piecuch (changes requested):

sys/kern/dev_procstat.c
franciscozdo: I don't think it is a good idea to drop this check. I think that in future it could be such a case and it would be hard to find this line if something braked.

j-piecuch: I think most if not all uses of `kmalloc()` without `M_NOWAIT` assume that the return value is not `NULL`, so I think the chances of that changing are near-zero.

franciscozdo: Okay, I changed it.

j-piecuch (approved):

cahirwpz (approved):

PR #935, opened 2021-01-18
[usb] Malta USB stack.

MichalBlk:
Implement USB on Malta platform. This involves:
- UHCI host controller driver
- USB bus driver (implement as a set of routines, not an actual bus driver)
- HUB support
- HID mouse driver
- HID keyboard driver
- USB mass storage device driver (Bulk-Only protocol)

PR #975, opened 2021-01-31
Introduce basic USB related header files.

MichalBlk:
The presented header files are required by #935.
Presented changes (regarding original files):
- `scsi.h`:
    - `length` and `addr` fields aren't expressed in byte units (i.e. instead of `uint8_t addr[4]` we have `uint32_t addr`). This simplifies little endian <-> big endian conversions.
    - I've added the `SID_FORMAT()` macro (used in drive identification).
- `uhci.h`: this is simply a subset of the original file (with a few stylistic changes).
- `uhcireg.h`:
    - I've added `PCI_USB_CLASSCODE` and `PCI_USB_SUBCLASSCODE`.
    - I've added ` UHCI_PORTSC()` instead of `UHCI_PORTSC{1,2}` (The UHCI specification allows more than two root hub ports).
    - I've added `UHCI_PORTSC_ONE` which is used in root hub port detection (I don't assume any specific number of root hub ports).
- `umass.h`: this file doesn't have a strict counterpart in the FreeBSD source tree. Contained definitions were extracted form the FreeBSD's `sys/dev/usb/storage/umass.c` file. This is in order to make the `umass.c` driver thinner and to contain some definitions required by the USB module (`UR_BBB_RESET` and `UR_BBB_GET_MAX_LUN`).
- `usbhid.h`: this is simply a subset of the original file (without the HID descriptor parsing definitions (we assume the boot protocol is supported)).
- `usb.h`:
    - I've added the `UV_MAKE()` macro (they use some counterpart defined elsewhere).
    - I've added the `US_DATASIZE` constant (they use a magic number).
    - I've added the `US_ENG_LID` and `US_ENG_STR` constants.
    - I've added the `US_TRANSFER_TYPE()` macro.

### DIFF
--- a/include/dev/scsi.h
+++ b/include/dev/scsi.h
@@ -1,8 +1,8 @@
 /*
- * Based on FreeBSD `scsi_all.h`.
+ * Based on FreeBSD `sys/cam/scsi/scsi_all.h`.
  */
-#ifndef _SCSI_H_
-#define _SCSI_H_
+#ifndef _DEV_SCSI_H_
+#define _DEV_SCSI_H_
 
 /* Opcodes. */
 #define REQUEST_SENSE 0x03
@@ -95,4 +95,4 @@ typedef struct scsi_read_capacity_data {
 
 #define REQUEST_TIMEOUT 4
 
-#endif /* _SCSI_H_ */
+#endif /* _DEV_SCSI_H_ */
--- a/include/dev/scsi.h
+++ b/include/dev/scsi.h
@@ -1,0 +1,98 @@
+/*
+ * Based on FreeBSD `scsi_all.h`.
+ */
+#ifndef _SCSI_H_
+#define _SCSI_H_
+
+/* Opcodes. */
+#define REQUEST_SENSE 0x03
+#define INQUIRY 0x12
+#define READ_CAPACITY 0x25
+#define READ_10 0x28
+#define WRITE_10 0x2A
+
+typedef struct scsi_request_sense {
+  uint8_t opcode; /* REQUEST_SENSE */
+  uint8_t byte2;
+  uint8_t unused[2];
+  uint8_t length;
+  uint8_t control;
+} __packed scsi_request_sense_t;
+
+typedef struct {
+  uint8_t opcode; /* INQUIRY */
+  uint8_t byte2;
+  uint8_t page_code;
+  uint16_t length;
+  uint8_t control;
+} __packed scsi_inquiry_t;
+
+typedef struct scsi_read_capacity {
+  uint8_t opcode; /* READ_CAPACITY */
+  uint8_t byte2;
+  uint8_t addr[4];
+  uint8_t unused[2];
+  uint8_t pmi;
+  uint8_t control;
+} __packed scsi_read_capacity_t;
+
+typedef struct scsi_rw_10 {
+  uint8_t opcode;
+  uint8_t byte2;
+  uint32_t addr;
+  uint8_t reserved;
+  uint16_t length;
+  uint8_t control;
+} __packed scsi_rw_10_t;
+
+typedef struct scsi_sense_data {
+  uint8_t error_code;
+  uint8_t segment;
+  uint8_t flags;
+  uint8_t info[4];
+  uint8_t extra_len;
+  uint8_t cmd_spec_info[4];
+  uint8_t add_sense_code;
+  uint8_t add_sense_code_qual;
+  uint8_t fru;
+  uint8_t sense_key_spec[3];
+} __packed scsi_sense_data_t;
+
+#define SHORT_INQUIRY_LENGTH 36
+
+#define SID_VENDOR_SIZE 8
+#define SID_PRODUCT_SIZE 16
+#define SID_REVISION_SIZE 4
+
+typedef struct {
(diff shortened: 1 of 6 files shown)


include/dev/scsi.h
```diff
@@ -0,0 +1,98 @@
+/*
+ * Based on FreeBSD `scsi_all.h`.
+ */
+#ifndef _SCSI_H_
+#define _SCSI_H_
+
+/* Opcodes. */
+#define REQUEST_SENSE 0x03
```
cahirwpz, line 8: We prefer enums since `gdb` can print their values as symbols. In this case maybe `typedef enum scsi_opcode { ... } scsi_opcode_t` ?

include/dev/scsi.h
```diff
@@ -0,0 +1,98 @@
+/*
+ * Based on FreeBSD `scsi_all.h`.
```
cahirwpz, line 2: Please use full path, i.e. `sys/cam/scsi/scsi_all.h`

include/dev/scsi.h
```diff
@@ -0,0 +1,98 @@
...
+ */
+#ifndef _SCSI_H_
```
cahirwpz, line 4: Definition name should reflect full path in mimiker's source tree starting from `include/`. In this case `_DEV_SCSI_H_`.

cahirwpz (commented):
All remarks for `scsi.h` file apply to other header files as well.

include/dev/scsi.h
MichalBlk: I've decided to stick to what they do. If I changed these definitions, I'd have to change many definitions form other header files to stick to this convention (e.g. see `usb.h`).

include/dev/scsi.h
MichalBlk: Done.

include/dev/scsi.h
MichalBlk: Done.

MichalBlk:
I've described the presented changes to the original files (see the PR's description).

include/dev/scsi.h
```diff
@@ -0,0 +1,98 @@
...
+
+#define SID_TYPE(inq_data) ((inq_data)->device & 0x1f)
+
+#define SID_QUAL(inq_data) (((inq_data)->device & 0xE0) >> 5)
+#define SID_QUAL_LU_CONNECTED 0x00
+
+#define SID_RMB 0x80
+#define SID_IS_REMOVABLE(inq_data) (((inq_data)->dev_qual2 & SID_RMB) != 0)
+
+#define SID_FORMAT(inq_data) ((inq_data)->response_format & 0x0f)
```
pj1031999, line 89: Could you add something like `SID_xxx_MASK` instead of constants in `SID_xxx`?

include/dev/uhci.h
```diff
@@ -0,0 +1,82 @@
...
+#define UHCI_TD_PID_SETUP 0x0000002d
+#define UHCI_TD_GET_PID(s) ((s)&0xff)
+#define UHCI_TD_SET_DEVADDR(a) ((a) << 8)
+#define UHCI_TD_GET_DEVADDR(s) (((s) >> 8) & 0x7f)
+#define UHCI_TD_SET_ENDPT(e) (((e)&0xf) << 15)
+#define UHCI_TD_GET_ENDPT(s) (((s) >> 15) & 0xf)
+#define UHCI_TD_SET_DT(t) ((t) << 19)
+#define UHCI_TD_GET_DT(s) (((s) >> 19) & 1)
+#define UHCI_TD_SET_MAXLEN(l) (((l)-1U) << 21)
+#define UHCI_TD_GET_MAXLEN(s) ((((s) >> 21) + 1) & 0x7ff)
```
pj1031999, line 57: Could you add something like `UHCI_TD_xxx_SHIFT` here?

include/dev/umass.h
```diff
@@ -0,0 +1,43 @@
...
+#define CBWCDBLENGTH 16
+
+typedef struct {
+  uint32_t dCBWSignature;
+  uint32_t dCBWTag;
+  uint32_t dCBWDataTransferLength;
+  uint8_t bCBWFlags;
+  uint8_t bCBWLUN;
+  uint8_t bCDBLength;
+  uint8_t CBWCDB[CBWCDBLENGTH];
```
pj1031999, line 22: We prefer lowercase with underscore.

include/dev/scsi.h
MichalBlk: I've decided to stick to the original sources (as the description states, in this file I've only added the `SID_FORMAT` macro).

include/dev/umass.h
MichalBlk: It was consulted with @cahirwpz, and we've decided to conform to their convention throughout the presented headers.

include/dev/uhci.h
MichalBlk: I can, but these would only be used inside this header, within the definitions of the macros above. I am trying to make only necessary changes to the original files.

include/dev/scsi.h
cahirwpz: Ok.

cahirwpz (approved):

PR #979, opened 2021-02-01
Statclock and gprof structures initialization

wiklam:
For [#863](https://github.com/cahirwpz/mimiker/pull/863) 
Kgmon and KGPROF is just for mips at the moment.
gmon.h is entirely copied from NetBSD with adjustments for our needs,


sys/kern/clock.c
```diff
@@ -4,21 +4,42 @@
...
+    if (instr < g->textsize) {
+      instr /= HISTFRACTION * sizeof(*g->kcount);
+      g->kcount[instr]++;
+    }
+  }
+}
+#else
+static void statclock(void) {
+}
+#endif
```
pj1031999, line 34: I think that you could use `__nothing` macro here like KASAN.

sys/kern/prof.c
```diff
@@ -0,0 +1,50 @@
...
+  p->tos = (tostruct_t *)profptr;
+  profptr += p->tossize;
+  p->kcount = (u_short *)profptr;
+  profptr += p->kcountsize;
+  p->froms = (u_short *)profptr;
+  p->state = GMON_PROF_ON;
+}
+#else
+void init_prof(void) {
+}
```
pj1031999, line 49: I think that you could use `__nothing` macro here like KASAN.

sys/kern/prof.c
```diff
@@ -0,0 +1,50 @@
...
+  else if (p->tolimit > MAXARCS)
+    p->tolimit = MAXARCS;
+  p->tossize = p->tolimit * sizeof(tostruct_t);
+
+  int size = p->kcountsize + p->tossize + p->fromssize;
+  int aligned_size = align(size, PAGESIZE);
+  profptr = kmem_alloc(aligned_size, M_NOWAIT | M_ZERO);
+  if (profptr == NULL) {
+    kprintf("Not enough memory for profiling!\n");
+    return;
```
pj1031999, line 38: Why not `klog`?

include/sys/gmon.h
```diff
@@ -0,0 +1,122 @@
...
+  u_long kcountsize;
+  u_short *froms;
+  u_long fromssize;
+  tostruct_t *tos;
+  u_long tossize;
+  long tolimit;
+  u_long lowpc;
+  u_long highpc;
+  u_long textsize;
+  u_long hashfraction;
```
pj1031999, line 110: As far as I know we prefer `uint16_t` & `unsigned long`.

include/sys/gmon.h
```diff
@@ -0,0 +1,122 @@
...
+};
+extern gmonparam_t _gmonparam;
+
+/*
+ * Possible states of profiling.
+ */
+#define GMON_PROF_ON 0
+#define GMON_PROF_BUSY 1
+#define GMON_PROF_ERROR 2
+#define GMON_PROF_OFF 3
```
pj1031999, line 120: Could you change that to enum?

include/sys/gmon.h
```diff
@@ -0,0 +1,122 @@
...
+ *	@(#)gmon.h	8.2 (Berkeley) 1/4/94
+ */
+
+#ifndef _SYS_GMON_H_
+#define _SYS_GMON_H_
+
+/*
+ * histogram counters are unsigned shorts (according to the kernel).
+ */
+#define HISTCOUNTER unsigned short
```
pj1031999, line 40: Why not `typedef`?

sys/debug/kgmon.py
```diff
@@ -0,0 +1,106 @@
...
+    return int(nomin * bar / denom) < int((nomin + 1) * bar / denom)
+
+
+class RawArc(Structure):
+    _fields_ = [('raw_frompc', c_uint), ('raw_selfpc', c_uint),
+                ('raw_count', c_int)]
+
+
+class ToStruct(Structure):
+    _fields_ = [('selfpc', c_uint), ('count', c_int),
```
pj1031999, line 22: `selfpc` is `u_long`, `count` is `long` in `gmon.h`. Is it ok?

include/sys/gmon.h
wiklam: Done

include/sys/gmon.h
wiklam: Done.

sys/kern/prof.c
wiklam: My bad.

sys/kern/prof.c
```diff
@@ -0,0 +1,50 @@
...
+  }
+  p->tos = (tostruct_t *)profptr;
+  profptr += p->tossize;
+  p->kcount = (u_short *)profptr;
+  profptr += p->kcountsize;
+  p->froms = (u_short *)profptr;
```
j-piecuch, line 46: We need to be careful about alignment here: `p->tos` must be aligned to `long`, and `p->kcount` and `p->froms` must be aligned to `u_short`. I think this works now, but it might not if we change the alignment of `p->textsize` or the value of `HASHFRACTION` or `HISTFRACTION`. At least add assertions here that check proper alignment.

sys/kern/clock.c
```diff
@@ -4,21 +4,42 @@
...
   return now;
 }
 
+#if KPROF
+static void statclock(void) {
+  gmonparam_t *g = &_gmonparam;
+  thread_t *td = thread_self();
+  if (td->td_kframe == NULL)
+    return;
+  uintptr_t pc = ctx_get_pc(td->td_kframe), instr;
```
j-piecuch, line 22: Why the comma? Doesn't this throw away the value of `ctx_get_pc(td->td_kframe)`?

sys/kern/prof.c
```diff
@@ -0,0 +1,50 @@
...
+
+gmonparam_t _gmonparam = {.state = GMON_PROF_OFF};
+
+#if KPROF
+void init_prof(void) {
+  void *profptr;
+  gmonparam_t *p = &_gmonparam;
+
+  p->lowpc = rounddown((unsigned long)__kernel_start,
+                       HISTFRACTION * sizeof(HISTFRACTION));
```
j-piecuch, line 20: Why `sizeof(HISTFRACTION)`? `HISTFRACTION` is defined as `2`, so it's implicitly an `int`... but why?

sys/debug/kgmon.py
```diff
@@ -0,0 +1,106 @@
...
+            frompc = self.lowpc
+            frompc += fromindex * self.hashfraction * sizeoffroms_p
+            toindex = froms
+
+            while toindex != 0:
+                prefix = f'_gmonparam.tos[{toindex}].'
+                rawarc.raw_frompc = frompc
+                rawarc.raw_selfpc = gdb.parse_and_eval(prefix + 'selfpc')
+                rawarc.raw_count = gdb.parse_and_eval(prefix + 'count')
+                ofile = "gmon.out"
```
j-piecuch, line 87: This is duplicated in every function, it's ugly. Why not make this a parameter?

sys/kern/clock.c
wiklam: Done.

sys/kern/prof.c
wiklam: Done.

j-piecuch (changes requested):

sys/kern/clock.c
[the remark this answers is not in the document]
j-piecuch: Nevermind, I got confused by the syntax. This is equivalent to:
```c
uintptr_t pc = ctx_get_pc(td->td_kframe);
uintptr_t instr;
```

sys/debug/kgmon.py
wiklam: It is ok, at this moment KPROF works for mips and we need just a 32 bit type, using `c_ulong` use 64 bits.

sys/kern/clock.c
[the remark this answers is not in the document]
pj1031999: If `instr` is not declared before that line it is ok. You declare `pc` with value returned by `ctx_get_pc(td->td_kframe)` and new uninitialized variable `instr`.

sys/debug/kgmon.py
wiklam: Done.

sys/kern/prof.c
wiklam: My bad it should be `HISTCOUNTER`, thanks!

sys/kern/clock.c
wiklam: I cleaned it up.

sys/kern/prof.c
wiklam: I am not sure if assertions are necessary, cause the order of assignments ensure the alignment.

sys/kern/prof.c
```diff
@@ -0,0 +1,50 @@
...
+  p->highpc =
+    roundup((unsigned long)__etext, HISTFRACTION * sizeof(HISTFRACTION));
+  p->textsize = p->highpc - p->lowpc;
+  p->kcountsize = p->textsize / HISTFRACTION;
```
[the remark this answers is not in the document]
wiklam: It is in bytes, I will add a comment which will redirect to `gmon.h`, which describe the meaning of the macros.

include/sys/gmon.h
wiklam: I will ask about it during the meeting.

sys/kern/prof.c
j-piecuch: Yes, but that depends on the value of `HISTFRACTION`. If it was something like `16`, `p->froms` would no longer be guaranteed to be properly aligned.

cahirwpz:
Please investigate interaction between `kgmon` and behavior of `MIPS (Clang)` test run.

sys/kern/prof.c
wiklam: Done.

sys/debug/kgmon.py
```diff
@@ -0,0 +1,97 @@
...
+            while toindex != 0:
+                tos = gdb.parse_and_eval('_gmonparam.tos[0]').address
+                memory = inferior.read_memory(tos, sizeof(ToStruct))
+                read = unpack('IiHH', memory)
+                rawarc.raw_frompc = frompc
+                rawarc.raw_selfpc = read[0]
+                rawarc.raw_count = read[1]
+                with open(OFILE, "ab") as of:
+                    of.write(rawarc)
+                toindex = read[2]
```
j-piecuch, line 82: I don't even know what these arcs mean (because you're not creating them anywhere), so I won't try to understand this code.

j-piecuch (changes requested):

sys/debug/kgmon.py
```diff
@@ -0,0 +1,97 @@
...
+        with open(OFILE, "ab") as of:
+            for i in array:
+                of.write(c_ushort(i))
+
+    def writearcinfo(self, inferior):
+        rawarc = RawArc()
+        sizeoffroms_p = gdb.parse_and_eval('sizeof(*_gmonparam.froms)')
+        endfrom = int(self.fromssize / sizeoffroms_p)
+
+        froms = gdb.parse_and_eval(f'_gmonparam.froms[0]').address
```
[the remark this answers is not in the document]
wiklam: It is :(

sys/kern/clock.c
```diff
@@ -12,9 +14,32 @@ systime_t getsystime(void) {
...
+  thread_t *td = thread_self();
+
+  if (td->td_kframe == NULL)
+    return;
+
+  pc = ctx_get_pc(td->td_kframe);
+  if (g->state == GMON_PROF_ON && pc >= g->lowpc) {
+    instr = pc - g->lowpc;
+    if (instr < g->textsize) {
+      instr /= HISTFRACTION * sizeof(*g->kcount);
```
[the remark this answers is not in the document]
wiklam: Thanks, it is a good idea.

sys/kern/gmon.c
```diff
@@ -0,0 +1,51 @@
...
+  p->tossize = p->tolimit * sizeof(tostruct_t);
+
+  int size = p->kcountsize + p->tossize + p->fromssize;
+  int aligned_size = align(size, PAGESIZE);
+  profptr = kmem_alloc(aligned_size, M_NOWAIT | M_ZERO);
+  if (profptr == NULL) {
+    klog("Not enough memory for profiling!\n");
+    return;
+  }
+  assert(is_aligned(profptr, sizeof(tostruct_t)));
```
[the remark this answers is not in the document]
wiklam: My bad

include/sys/gmon.h
```diff
@@ -0,0 +1,129 @@
...
+  u_short pad;
+} tostruct_t;
+
+/*
+ * The profiling data structures are housed in this structure.
+ */
+
+typedef struct gmonparam {
+  int state;
+  u_short *kcount;
```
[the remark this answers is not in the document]
wiklam: You are right. Didn't noticed that in different BSD's they use correctly HISTCOUNTER.

sys/debug/kgmon.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(self):
+        super().__init__('kgmon')
+
+    def __call__(self, args):
+        infer = gdb.inferiors()[0]
+        gmonparam = GmonParam(gdb.parse_and_eval('_gmonparam'))
+        gmonparam.writeheader()
+        gmonparam.writetickbuffer(infer)
+        gmonparam.writearcinfo(infer)
```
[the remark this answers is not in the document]
wiklam: I tried to simplify kgmon as far as I could.

sys/debug/kgmon.py
wiklam: I deleted this piece of code, it will be introduced when it will be used. I am really sorry for the struggle.

sys/debug/kgmon.py
```diff
@@ -0,0 +1,52 @@
...
+class Kgmon(SimpleCommand):
+    """Dump the gprof data to gmon.out"""
+
+    def __init__(self):
+        super().__init__('kgmon')
+
+    def __call__(self, args):
+        state = gdb.parse_and_eval('_gmonparam.state')
+        if state == gdb.parse_and_eval('GMON_PROF_NOT_INIT'):
+            print("Compile program with KGPROF=1 or gmon not innitalized yet")
```
j-piecuch, line 49:
```suggestion
            print("Compile program with KGPROF=1 or gmon not initialized yet")
```

j-piecuch (changes requested):

j-piecuch:
I found that if I don't gather profiling data from the idle thread, the data is a bit less skewed towards `cpu_intr_enable()`:
```
index % time    self  children    called     name
                                                 <spontaneous>
[1]     22.6    1.38    0.00                 cpu_intr_enable [1]
-----------------------------------------------
                                                 <spontaneous>
[2]     21.3    1.30    0.00                 bzero [2]
-----------------------------------------------
                                                 <spontaneous>
[3]      7.8    0.48    0.00                 bcopy [3]
-----------------------------------------------
                                                 <spontaneous>
[4]      4.9    0.30    0.00                 test_kmem [4]
-----------------------------------------------
                                                 <spontaneous>
[5]      4.0    0.25    0.00                 mtx_owned [5]
-----------------------------------------------
                                                 <spontaneous>
[6]      3.4    0.21    0.00                 mtx_unlock [6]
-----------------------------------------------
                                                 <spontaneous>
[7]      3.4    0.21    0.00                 _mtx_lock [7]
-----------------------------------------------
                                                 <spontaneous>
[8]      3.3    0.20    0.00                 memcpy [8]
-----------------------------------------------
```

sys/debug/kgmon.py
```diff
@@ -7,91 +7,46 @@
...
+        for i in array:
+            file.write(c_int(i))
 
-    def writetickbuffer(self, inferior):
-        kcountsize = self.kcountsize
-        kcount = gdb.parse_and_eval('_gmonparam.kcount[0]').address
+    def write_tick_buffer(self, inferior, file):
+        kcountsize = int(gdb.parse_and_eval('_gmonparam.kcountsize'))
+        kcount = gdb.parse_and_eval('_gmonparam.kcount')
         memory = inferior.read_memory(kcount, kcountsize * sizeof(c_ushort))
```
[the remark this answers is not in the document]
wiklam: IMO it is correct. kcountsize is etext size divided by 2 - tells us how many counters do we need to allocate for each 
> subroutine call points.
Description of `HASHFRACTION` in gmon.h.
In `init_prof()` we are adding kcountsize to a `u_short *` pointer (the pointer is moving by `kcountsize * sizeof(c_ushort)` - the space allocated for counters which store how many times the function was 'caught' by statclock)

include/sys/gmon.h
wiklam: Missed that during gmon.h port2.

sys/debug/kgmon.py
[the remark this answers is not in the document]
j-piecuch: In `init_prof()`, `profptr` has type `void *`, not `u_short *`, so adding `kcountsize` moves it by `kcountsize` bytes.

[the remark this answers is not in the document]
wiklam: You are right!

sys/debug/kgmon.py
```diff
@@ -0,0 +1,52 @@
...
+        for i in array:
+            file.write(c_int(i))
+
+    def write_tick_buffer(self, inferior, file):
+        kcountsize = int(gdb.parse_and_eval('_gmonparam.kcountsize'))
+        kcount = gdb.parse_and_eval('_gmonparam.kcount')
+        memory = inferior.read_memory(kcount, kcountsize)
+        array = unpack(kcountsize * 'c', memory)
+        for i in array:
+            file.write(c_char(i))
```
j-piecuch, line 34: Might as well do it like this at this point:
```suggestion
        file.write(inferior.read_memory(kcount, kcountsize))
```

j-piecuch (changes requested):

include/sys/gmon.h
```diff
@@ -0,0 +1,160 @@
...
+/*
+ * Possible states of profiling.
+ */
+typedef enum {
+  GMON_PROF_ON = 0,
+  GMON_PROF_BUSY = 1,
+  GMON_PROF_ERROR = 2,
+  GMON_PROF_OFF = 3,
+  GMON_PROF_NOT_INIT = 4,
+} gmon_flags_t;
```
j-piecuch, line 150: These aren't flags (i.e. they're not meant to be bitwise-ORed), so `gmon_state_t` is a better name IMO.

include/sys/gmon.h
```diff
@@ -0,0 +1,160 @@
...
+
+/*
+ * Sysctl definitions for extracting profiling information from the kernel.
+ */
+#define GPROF_STATE 0     /* int: profiling enabling variable */
+#define GPROF_COUNT 1     /* struct: profile tick count buffer */
+#define GPROF_FROMS 2     /* struct: from location hash bucket */
+#define GPROF_TOS 3       /* struct: destination/count structure */
+#define GPROF_GMONPARAM 4 /* struct: profiling parameters (see above) */
```
j-piecuch, line 159: Will these be used in the future?

j-piecuch (changes requested):

sys/kern/gmon.c
```diff
@@ -0,0 +1,59 @@
...
+#include <machine/vm_param.h>
+#include <sys/types.h>
+
+gmonparam_t _gmonparam = {.state = GMON_PROF_NOT_INIT};
+gmonhdr_t _gmonhdr = {.profrate = CLK_TCK};
+/* The macros description are provided in gmon.h */
+void init_kgprof(void) {
+#if KGPROF == 0
+  return;
+#endif
```
pj1031999, line 19: Is it possible to move that conditional compilation into header file like we do for KASAN?

```C
#if KGPROF
void init_kgprof(void);
#else
#define init_kgprof() __nothing
#endif
```

sys/kern/gmon.c
```diff
@@ -0,0 +1,59 @@
...
+  p->highpc = roundup((unsigned long)__etext, INSTR_GRANULARITY);
+  p->textsize = p->highpc - p->lowpc;
+  p->kcountsize = p->textsize / HISTFRACTION;
+  p->hashfraction = HASHFRACTION;
+  p->fromssize = p->textsize / HASHFRACTION;
+  p->tolimit = (p->textsize * ARCDENSITY) / 100;
+  if (p->tolimit < MINARCS)
+    p->tolimit = MINARCS;
+  else if (p->tolimit > MAXARCS)
+    p->tolimit = MAXARCS;
```
pj1031999, line 33: What do you think about?
```C
p->tolimit = min(max(p->tolimit, MINARCS), MAXARCS);
````
If we are using that pattern in more places we could create global macro for that.

File: sys/kern/gmon.c [hunk 26]
```diff
@@ -0,0 +1,59 @@
...
+  p->tossize = p->tolimit * sizeof(tostruct_t);
+
+  int size = p->kcountsize + p->tossize + p->fromssize;
+  int aligned_size = align(size, PAGESIZE);
+  profptr = kmem_alloc(aligned_size, M_NOWAIT | M_ZERO);
+  if (profptr == NULL) {
+    klog("Not enough memory for profiling!\n");
+    return;
+  }
```
pj1031999, line 42: From my point of view if you build kernel with profiling flags you probably know what you are doing and that function shouldn't fail.

wiklam: IMO we call this function just one time (it is not expensive) and it is better to be cautious.

sys/debug/kgmon.py
```diff
@@ -0,0 +1,49 @@
...
+        for i in array:
+            file.write(c_int(i))
+
+    def write_tick_buffer(self, inferior, file):
+        kcountsize = int(gdb.parse_and_eval('_gmonparam.kcountsize'))
+        kcount = gdb.parse_and_eval('_gmonparam.kcount')
+        file.write(inferior.read_memory(kcount, kcountsize))
+
+    def write_arc_info(self, inferior, file):
+        return
```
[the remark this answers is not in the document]
wiklam: Thanks, didn't know that.

include/sys/gmon.h
wiklam: I don't think so

sys/debug/kgmon.py
```diff
@@ -0,0 +1,46 @@
+import gdb
+import sys
+
+
+from .cmd import SimpleCommand
+
+
+class GmonOut():
+
+    def write_to_file(self, path="gmon.out"):
```
j-piecuch, line 10: You could also make this a static method, then you don't have to create an instance to call it, you can just do `GmonOut.write_to_file()`

j-piecuch (approved):

pj1031999 (approved):

PR #980, opened 2021-02-01
Port libterminfo from NetBSD

j-piecuch:
Summary of changes made to the original NetBSD sources:
- In `setupterm.c`, replace `#include <curses.h>` with `#include <termios.h>`
- In `termcap.c`, remove unnecessary casts of the first argument in calls to `_t_{flag,num,str}hash()`
- The `genhash` and `genthash` scripts now use `gperf` instead of `nbperf`. Since `gperf` and `nbperf` differ quite a bit in terms of options and output, the scripts had to be changed significantly.


bkjg (approved):
LGTM

franciscozdo (approved):
LGTM

cahirwpz (approved):

PR #985, opened 2021-02-02
Improve kernel crash post-mortem debugging experience.

cahirwpz:
* [x] fix missing `sleepq_lookup`
* [x] dump backtraces from all threads
* [x] fix `klog` table width issues
* [x] report failed test name
* [x] make sure `klog` output is sane
* [x] backtrace through userspace binaries as well
* [ ] maybe use record/replay function and dump replay log into artifacts ?

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -11,6 +11,7 @@ python import debug
 
 # favorite set of breakpoints
 break kernel_init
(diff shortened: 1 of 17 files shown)


pj1031999 (approved):
LGTM

PR #987, opened 2021-02-04
Removing modulo operation from PIT.

wiklam:
Modulo operation is too expensive, this is why we want to get rid of it. 
I also renamed some variables to make them more descriptive.


sys/drv/pit.c
```diff
@@ -11,56 +11,73 @@ typedef struct pit_state {
...
-  if (pit->cntr16_prev_read < cntr16_now)
-    ticks += pit->period_cntr;
+static void pit_update_time(pit_state_t *pit) {
+  assert(intr_disabled());
+  uint32_t last_ticks = pit->ticks;
+  uint32_t last_sec = pit->sec;
+  uint16_t now_ticks16 = pit_get_counter(pit);
+  uint16_t ticks_passed = now_ticks16 - pit->prev_ticks16;
+  if (pit->prev_ticks16 >= now_ticks16) {
+    pit->overflowed = false;
```
franciscozdo, line 49: Do you need to clear the `overflow` flag here or set? Condition says "current value is less than previous value" and this values can only get bigger, so I think it is case where overflow has occured.

sys/drv/pit.c
```diff
@@ -11,56 +11,73 @@ typedef struct pit_state {
...
 
 static intr_filter_t pit_intr(void *data) {
   pit_state_t *pit = data;
 
-  /* XXX: It's still possible for a tick to be lost. */
-  (void)pit_get_counter64(pit);
+  /* XXX: It's still possible for periods to be lost. */
+  pit_update_time(pit);
+  if (pit->overflowed) {
+    pit->ticks += pit->overflowed;
```
franciscozdo, line 73: I suggest explicitly adding one. It looks like `pit->overflowed` could have value other than 0 or 1.

sys/drv/pit.c
```diff
@@ -11,56 +11,73 @@ typedef struct pit_state {
...
+    if (pit->ticks >= TIMER_FREQ) {
+      pit->ticks -= TIMER_FREQ;
+      pit->sec++;
+    }
+  }
   tm_trigger(&pit->timer);
+  pit->overflowed = true;
```
franciscozdo, line 80: You handled overflow few lines up. Shouldn't you clear this flag instead of keep it set?

sys/drv/pit.c
```diff
@@ -100,11 +118,14 @@ static int timer_pit_stop(timer_t *tm) {
...
   device_t *dev = device_of(tm);
   pit_state_t *pit = dev->state;
-  uint64_t count = pit_get_counter64(pit);
   uint32_t freq = pit->timer.tm_frequency;
-  uint32_t sec = count / freq;
-  uint32_t frac = count % freq;
-  bintime_t bt = bintime_mul(HZ2BT(freq), frac);
+  uint32_t sec, ticks;
+  WITH_INTR_DISABLED {
+    pit_update_time(pit);
```
franciscozdo, line 124: Why don't you handle overflow here as you do in `pit_intr`? It can't occur here or is there some other explanation?

franciscozdo (changes requested):
I think you messed up `pit::overflow` semantics. See my comments.

Also why you handle overflow out of `pit_update_time` (and not always after calling this function)? I think code will be easier and cleaner (you can completely remove `overflowed` flag from `pit_state`). If I am wrong please show me some example which explains why you do it in that way.

cahirwpz:
Since we cannot read status of PIT interrupt (i.e. is it pending) we cannot reliably detect overflow.

Please refer to FreeBSD's [i8254_get_timecount](https://grok.dragonflybsd.org/xref/freebsd/sys/x86/isa/clock.c?r=339c1042#490).

sys/drv/pit.c
wiklam: It is a typo here should I add the period value.

sys/drv/pit.c
wiklam: I am setting this flag to check in the next interrupt if I already handled the overflow.

sys/drv/pit.c
wiklam: We clear it cause the counter already overflowed and we added the counter period.

sys/drv/pit.c
wiklam: Cause in pit_intr we are sure that the count_period already past (we just unsure how many) and we want to check if it is already added to ticks.

wiklam:
> I think you messed up `pit::overflow` semantics. See my comments.
> 
> Also why you handle overflow out of `pit_update_time` (and not always after calling this function)? I think the code will be easier and cleaner (you can completely remove `overflowed` flag from `pit_state`). If I am wrong please show me some example which explains why you do it in that way.

Maybe it will explain something, the main problem with PIT is that its counter count from N to 0 and cause an interrupt, N is the number of ticks of an interrupt period, which we set at the beginning for the timer. Sometimes we turn off interrupts and we can miss almost the whole period and this is why we introduce the overflowed flag, which helps us to keep track of such situations (sometimes using gettime allow us to catch that situation). 

Ok, maybe you are asking yourself why we just don't check if the value read by the counter is smaller than the previous one to detect overflow, it is ok but we can have a situation when we access the timer in different periods but with ascending value and we will not detect it with such way.

Maybe it is not obvious, but it is probably the best option at the moment. Please ask me if you still have some doubts, cause understanding it took me a while.

sys/drv/pit.c
```diff
@@ -11,56 +11,75 @@ typedef struct pit_state {
...
 
-  /* XXX: It's still possible for a tick to be lost. */
-  (void)pit_get_counter64(pit);
+  /* XXX: It's still possible for periods to be lost. */
+  pit_update_time(pit);
+  if (pit->overflowed) {
+    pit->ticks += pit->period_ticks;
+    if (pit->ticks >= TIMER_FREQ) {
+      pit->ticks -= TIMER_FREQ;
+      pit->sec++;
```
j-piecuch, line 76: Consider the following scenario:
1. A PIT interrupt happens, and we observe the return value of `pit_get_counter()` as `1`
2. When the next interrupt happens, interrupts are disabled by another thread.
3. That same thread, while still having interrupts disabled, calls `timer_pit_gettime()`, which observes the return value of  `pit_get_counter()` as `2`, but doesn't detect the overflow, so it appears as though one tick elapsed, but in reality `period_ticks + 1` elapsed.
4. When the thread enables interrupts, we detect the overflow and jump time forward by `period_ticks`.

This situation results in a very sudden time jump. [18254_get_timecount()](https://grok.dragonflybsd.org/xref/freebsd/sys/x86/isa/clock.c?r=339c1042#i8254_get_timecount) solves this by checking the status of the interrupt to detect when the counter overflows. Could we do the same?

sys/drv/pit.c
```diff
@@ -11,56 +11,75 @@ typedef struct pit_state {
   resource_t *regs;
   resource_t *irq_res;
   timer_t timer;
-  uint16_t period_cntr;      /* period as PIT counter value */
-  uint16_t cntr16_prev_read; /* last read counter value */
-  timercntr_t cntr64;        /* counter value since timer initialization */
+  bool overflowed;       /* should we add the counter period */
```
j-piecuch, line 14: I think `overflowed` is a poor name.
As I understand it, `overflowed == true` iff a timer overflow **hasn't** been detected since the last PIT interrupt, which makes it very unintuitive.

j-piecuch (changes requested):

sys/drv/pit.c
wiklam: At this moment we do not have such an interface allowing us to check pending interrupts, this the best option, which we can afford right now.

sys/drv/pit.c
wiklam: Renamed it, to `noticed_overflow`, hope now it is intuitive.

sys/drv/pit.c
```diff
@@ -11,56 +11,75 @@ typedef struct pit_state {
...
-  /* XXX: It's still possible for a tick to be lost. */
-  (void)pit_get_counter64(pit);
+  /* XXX: It's still possible for periods to be lost. */
+  pit_update_time(pit);
+  if (!pit->noticed_overflow) {
+    pit->ticks += pit->period_ticks;
+    if (pit->ticks >= TIMER_FREQ) {
+      pit->ticks -= TIMER_FREQ;
+      pit->sec++;
+    }
```
j-piecuch, line 77: This duplicates some logic from `pit_update_time()`, can you factor it out into another function, e.g. `pit_incr_ticks(uint32_t ticks)`?

j-piecuch (changes requested):

sys/drv/pit.c
```diff
@@ -11,56 +11,76 @@ typedef struct pit_state {
   resource_t *regs;
   resource_t *irq_res;
   timer_t timer;
-  uint16_t period_cntr;      /* period as PIT counter value */
-  uint16_t cntr16_prev_read; /* last read counter value */
-  timercntr_t cntr64;        /* counter value since timer initialization */
+  bool noticed_overflow; /* should we add the counter period */
```
franciscozdo, line 14: The comment here is misleading. If I understand correctly, this field tells us if "we have we noticed and handled overflow" (you use it in that way).

franciscozdo (changes requested):

sys/drv/pit.c
```diff
@@ -80,9 +99,11 @@ static int timer_pit_start(timer_t *tm, unsigned flags, const bintime_t start,
...
   assert(counter <= 0xFFFF);
 
-  pit->cntr64.val = 0;
-  pit->cntr16_prev_read = 0;
-  pit->period_cntr = counter;
+  pit->sec = 0;
+  pit->ticks = 0;
+  pit->prev_ticks16 = 0;
+  pit->period_ticks = counter;
+  pit->noticed_overflow = true;
```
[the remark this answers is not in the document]
wiklam: It should be false, didn't noticed it during the logic change.

franciscozdo (approved):

j-piecuch (approved):

cahirwpz (approved):

PR #993, opened 2021-02-12
AArch64: disable interrupts in user_exc_leave.

pj1031999:
Disable interrupts in `user_exc_leave`.

### DIFF
--- a/sys/aarch64/evec.S
+++ b/sys/aarch64/evec.S
@@ -165,6 +165,9 @@ ENTRY(handle_user_trap)
         mov     x0, sp
         bl      user_trap_handler
 user_exc_leave:
+        /* disable interrupts */
+        msr     daifset, (DAIF_I_MASKED >> 6)
+
         /* load thread_t::tdp_flags */
         /* thread_t::tdp_flags is a volatile unsigned - use 32-bit */
         load_pcpu x1
--- a/sys/aarch64/evec.S
+++ b/sys/aarch64/evec.S
@@ -166,7 +166,7 @@ ENTRY(handle_user_trap)
         bl      user_trap_handler
 user_exc_leave:
         /* disable interrupts */
-        msr     daifset, (DAIF_I_MASKED >> 6)
+        msr     daifset, #DAIF_I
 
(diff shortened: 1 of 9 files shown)


sys/aarch64/genassym.cf
```diff
@@ -5,6 +5,8 @@ include <sys/proc.h>
 include <aarch64/mcontext.h>
 include <aarch64/armreg.h>
 
+define DAIF_I_MASKED DAIF_I_MASKED
```
cahirwpz, line 8: Please add `-Werror` to `ASFLAGS` in `build/flags.mk` and clean up warnings.

sys/aarch64/evec.S
```diff
@@ -165,6 +165,9 @@ ENTRY(handle_user_trap)
         mov     x0, sp
         bl      user_trap_handler
 user_exc_leave:
+        /* disable interrupts */
+        msr     daifset, (DAIF_I_MASKED >> 6)
```
cahirwpz, line 169: Well... this is confusing.

Some observations:
 * `msr daifset, #x` and `msr daifclr, #x` use `DAIF_I_MASKED` shifted right by `6`,
 * `mrs %r, daif` mirrors interrupt flags from current `PSR`,
 *  in our code `DAIF_I_MASKED` is the same as `PSR_I`.

Maybe we should clearly separate use of `PSR_*` and `DAIF_*` ? I would much prefer that `DAIF_I` (without `MASKED` suffix if possible) was used only with `daifclr` and `daifset` registers and `PSR_I` with `daif`. I'd add a comment in `armreg.h` about this oddity.

cahirwpz (commented):
This is an important change but some cleanups would be welcome.

sys/aarch64/genassym.cf
pj1031999: done

sys/aarch64/evec.S
pj1031999: done

cahirwpz (approved):
LGTM

PR #994, opened 2021-02-12
AArch64: map kernel pages with VM_PROT_NONE without read permissions.

pj1031999:
Map kernel pages with `VM_PROT_NONE` without read permissions.

It fixes pmap_kenter and rmbits from #932.

### DIFF
--- a/sys/aarch64/pmap.c
+++ b/sys/aarch64/pmap.c
@@ -46,16 +46,16 @@ static const pte_t vm_prot_map[] = {
   [VM_PROT_READ] =
     ATTR_AP_RO | ATTR_SW_READ | ATTR_AF | pte_noexec | pte_common,
   [VM_PROT_WRITE] =
-    ATTR_AP_RW | ATTR_SW_WRITE | ATTR_DBM | ATTR_AF | pte_noexec | pte_common,
+    ATTR_AP_RW | ATTR_SW_WRITE | ATTR_AF | pte_noexec | pte_common,
   [VM_PROT_READ | VM_PROT_WRITE] = ATTR_AP_RW | ATTR_SW_READ | ATTR_SW_WRITE |
-                                   ATTR_DBM | ATTR_AF | pte_noexec | pte_common,
+                                   ATTR_AF | pte_noexec | pte_common,
   [VM_PROT_EXEC] = ATTR_AF | pte_common,
   [VM_PROT_READ | VM_PROT_EXEC] =
     ATTR_AP_RO | ATTR_SW_READ | ATTR_AF | pte_common,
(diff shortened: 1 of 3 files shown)


sys/aarch64/pmap.c
```diff
@@ -43,7 +43,7 @@ static const pte_t pte_default =
 static const pte_t pte_noexec = ATTR_XN | ATTR_SW_NOEXEC;
 
 static const pte_t vm_prot_map[] = {
-  [VM_PROT_NONE] = ATTR_XN | pte_noexec | pte_default,
+  [VM_PROT_NONE] = L3_PAGE | ATTR_SH(ATTR_SH_IS) | pte_noexec,
```
cahirwpz, line 46: Please remove `ATTR_SH`, `ATTR_AP` and `ATTR_S2_S2AP*`. These macro does not introduce any important concept and are cluttering our code.

If `pte_common` is `L3_PAGE | ATTR_SH(ATTR_SH_IS)` then should we set `ATTR_DBM` & `ATTR_AF` for all other cases? 

Could you add docstrings for `ATTR_DBM` & `ATTR_AF`? I do not remember what the mean...

pj1031999: I removed some macros and added comments about `DBM` and `AF` bits - please check if it's enough.

`DBM` means that page is in dirty state, `AF` means that page is valid.
I think that we should set `AF` flag for every usable kernel page - we already do that for _mips_.
For `DBM` flag you are right. I removed that flag from RO pages but I'm not sure about kernel pages. For _mips_ we set that flag for every writable kernel page but - please correct me if I'm wrong - dirty flag on _mips_ also means write access for page. From my point of view we can safely remove `DBM` flag from `vm_prot_map` and set it only from `pmap_emulate_bits` but still we should set it manually for initial page table in `boot.c`.

Both of them _could_ be managed by hardware - see `HD` and `HA` bits in TCR_EL (Translation Control Register) but we don't use that feature now.

cahirwpz: I updated DBM & AF description – I think its clearer now.

To follow _MIPS_ behaviour `DBM` and `AF` flags **must not** be set by default for user pages. For read-only kernel pages we need to set `AF` for read-write `DBM` as well.

pj1031999: As we discussed offline:
- I reverted DBM description change
- `pte_default` was renamed into `pte_common`
- `AF` now is explicit
- `AF` and `DBM` wasn't set for user pages exactly like in _mips_ - see `pmap_enter`

cahirwpz (approved):
LGTM

PR #995, opened 2021-02-12
AArch64: validate status register in setcontext syscall.

pj1031999:
Validate status register in setcontext syscall.

It fixes sigaction_with_setjmp, mmap_prot_none, mmap_prot_read from #932.

### DIFF
--- a/sys/aarch64/context.c
+++ b/sys/aarch64/context.c
@@ -58,8 +58,18 @@ int do_setcontext(thread_t *td, ucontext_t *uc) {
   mcontext_t *from = &uc->uc_mcontext;
   mcontext_t *to = td->td_uctx;
 
-  if (uc->uc_flags & _UC_CPU)
-    memcpy(&to->__gregs, &from->__gregs, sizeof(register_t) * (_REG_ELR + 1));
+  if (uc->uc_flags & _UC_CPU) {
+    register_t spsr = _REG(from, SPSR);
+    /* Validate CPU context. */
+    if (spsr & ~PSR_NZCV)
+      return EINVAL;
+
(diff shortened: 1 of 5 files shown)


sys/aarch64/context.c
```diff
@@ -59,7 +59,7 @@ int do_setcontext(thread_t *td, ucontext_t *uc) {
   mcontext_t *to = td->td_uctx;
 
   if (uc->uc_flags & _UC_CPU)
-    memcpy(&to->__gregs, &from->__gregs, sizeof(__gregset_t));
+    memcpy(&to->__gregs, &from->__gregs, sizeof(register_t) * (_REG_ELR + 1));
```
cahirwpz, line 62: IIUC you're skipping `SPSR` and `TPIDR`. 

`SPSR` stores both system flags (e.g `DAIF` bits) and user flags (e.g `VCNZ` bits used by comparison instructions). I guess you want to copy whole context but then make sure only user-settable bits can be affected.

cahirwpz (commented):
I find it difficult to believe the solution is correct.

sys/aarch64/context.c
cahirwpz: As you mentioned in private messages [cpu_mcontext_validate](https://grok.dragonflybsd.org/xref/netbsd/sys/arch/aarch64/aarch64/cpu_machdep.c) is what we want.

pj1031999: I added validation.

cahirwpz (approved):
LGTM

sys/aarch64/context.c
cahirwpz: Ok.

PR #996, opened 2021-02-13
Lock dependency validator

xThaid:
This is my attempt to find and fight with deadlocks.

The basic object the validator operates upon is a 'class' of locks.
A class of locks is a group of locks that are logically the same with respect to locking rules, even if the locks may have multiple instantiations. For example a lock in the vnode struct is one class, while each vnode has its own instantiation of that lock class. Every statically allocated lock has its own class.

Every thread has a list of classes of currently held locks, so when another lock is about to be locked, in the graph of lock dependency I add an edge from the last held lock to the current one.

And all the validator does right now is tracking dependencies between different lock-classes. The dependency graph can be easily read at any moment using gdb and `call lockdep_print_graph()`.

There are a few things that could be done:
1. Take into account the rest of locks (right now I only care about mutexes)
2. ~~Search for cycles in runtime~~ - done
3. Better reporting when a potential deadlock occurs


cahirwpz:
Please fix compilation errors under `GCC`.

sys/kern/lockdep.c
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  link = alloc_link();
+  link->from = hprev->lock_class;
+  link->to = hcur->lock_class;
+
+  TAILQ_INSERT_HEAD(&(hprev->lock_class->locked_after), link, entry);
+}
+
+void lock_acquire(lock_class_mapping_t *lock) {
```
j-piecuch, line 116: I think `lockdep_{acquire,release}()` are better names, since they convey that their only concern is dependency tracking.

j-piecuch (changes requested):
It would also be nice to be able to enable/disable the validator with a macro, in the same way as KASAN.

sys/kern/lockdep.c
```diff
@@ -0,0 +1,254 @@
...
+  spin_lock(&main_lock);
+}
+
+static inline void lockdep_unlock(void) {
+  spin_unlock(&main_lock);
+}
+
+static void bfs_init(void) {
+  for (int i = 0; i < class_cnt; i++) {
+    lock_classes[i].bfs_visited = 0;
```
j-piecuch, line 93: Could we optimize this a bit? For example, let's use a generation number that is incremented with each call to `check_path()`, and a node is considered visited if its `bfs_visited` value is equal to the generation number. Then we won't need to iterate over all lock classes here.

j-piecuch (changes requested):
The code could use some more explanatory comments.

sys/kern/lockdep.c
xThaid: Actually, I thought of this, but since we won't traverse the graph very often I decided not to do that for the sake of simplicity.

But you have noted that, so I will change it.

sys/kern/lockdep.c
```diff
@@ -0,0 +1,254 @@
...
+  const char *name;
+  TAILQ_HEAD(, lock_class_link) locked_after;
+
+  int bfs_visited;
+} lock_class_t;
+
+/* lock_class_link represents an edge in the graph of lock depedency. */
+typedef struct lock_class_link {
+  TAILQ_ENTRY(lock_class_link) entry;
+  lock_class_t *from, *to;
```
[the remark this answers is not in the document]
xThaid: This was left from the previous versions of the BFS. Can be deleted.

sys/kern/lockdep.c
```diff
@@ -0,0 +1,254 @@
...
+} lock_class_link_t;
+
+static spin_t main_lock = SPIN_INITIALIZER(0);
+
+#define CLASSHASH_SIZE 64
+#define CLASSHASH(key) ((uintptr_t)(key) % CLASSHASH_SIZE)
```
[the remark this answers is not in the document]
xThaid: Prove me if I'm wrong, but `uintptr_t` can be also 8 bytes (on aarch64). Beside that, could you describe the problem? There are already `CLASSHASH_SIZE` buckets and every of them is a head of a list.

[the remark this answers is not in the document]
j-piecuch: `lock_class_key_t` is an `int`, which is 4 bytes on both mips and aarch64, so all pointers to `lock_class_key_t` are aligned to 4 bytes, i.e. their value is a multiple of 4. If `key` is a multiple of 4, then `CLASSHASH(key)` will also be a multiple of 4, so `lock_hashtbl` will always be indexed using a multiple of 4, which reduces by 4x the effective number of buckets.

sys/kern/lockdep.c
```diff
@@ -6,21 +6,27 @@
...
+
+  /* Used in the graph traversal to decide if the graph node was already
+   * visited. */
+  int bfs_gen_id;
 } lock_class_t;
 
-/* lock_class_link represents an edge in the graph of lock depedency. */
+/*
+ * lock_class_link represents an edge in the graph of lock depedency. An edge
+ * C1-> C2 is created if the lock class C1 acquired while holding the class C2.
```
j-piecuch, line 25: I think it's the other way around: An edge C1 -> C2 is created if a lock inside lock class C2 is acquired while holding a lock belonging to lock class C1.

sys/kern/lockdep.c
```diff
@@ -0,0 +1,257 @@
...
+  bfs_q_enqueue(queue, SIMPLEQ_FIRST(&src->locked_after));
+  bfs_mark_visited(src);
+
+  while ((link = bfs_next_link(link)) || (link = bfs_q_dequeue(queue))) {
+    if (bfs_is_visited(link->to))
+      continue;
+    bfs_mark_visited(link->to);
+
+    if (link->to == target)
+      return 1;
```
j-piecuch, line 132:
```suggestion
      return true;
```

sys/kern/lockdep.c
```diff
@@ -0,0 +1,257 @@
...
+
+    if ((child = SIMPLEQ_FIRST(&link->to->locked_after))) {
+      if (!bfs_q_enqueue(queue, child))
+        panic("lockdep: bfs queue overflow");
+    }
+  }
+
+  return 0;
```
j-piecuch, line 140:
```suggestion
  return false;
```

sys/kern/lockdep.c
```diff
@@ -0,0 +1,257 @@
...
+ * <target> or not.
+ */
+static bool check_path(lock_class_t *src, lock_class_t *target) {
+  lock_class_link_t *child, *link = NULL;
+  bfs_queue_t *queue = &bfs_queue;
+
+  bfs_q_init(queue);
+  bfs_cur_gen++;
+
+  bfs_q_enqueue(queue, SIMPLEQ_FIRST(&src->locked_after));
```
j-piecuch, line 123: Is it guaranteed that the `src->locked_after` list is not empty? It's not obvious to me.

sys/kern/lockdep.c
```diff
@@ -0,0 +1,257 @@
...
+    if (hcur->lock_class == link->to)
+      return;
+  }
+
+  link = alloc_link();
+  link->to = hcur->lock_class;
+
+  SIMPLEQ_INSERT_HEAD(&(hprev->lock_class->locked_after), link, entry);
+
+  if (check_path(link->to, hprev->lock_class)) {
```
j-piecuch, line 214:
```suggestion
  if (check_path(hcur->lock_class, hprev->lock_class)) {
```
I think it's more readable that way

sys/kern/lockdep.c
```diff
@@ -0,0 +1,257 @@
...
+    panic("lockdep: cycle between locks %s and %s", hprev->lock_class->name,
+          link->to->name);
```
j-piecuch, line 216:
```suggestion
          hcur->lock_class->name);
```

j-piecuch (changes requested):

sys/kern/lockdep.c
xThaid: There's no such guarantee, but it doesn't affect the algorithm. In this case we simply won't enter the loop
```
while ((link = bfs_next_link(link)) || (link = bfs_q_dequeue(queue))) {
```
because both conditions aren't fulfilled.

j-piecuch: I see it now. Thanks.

sys/kern/lockdep.c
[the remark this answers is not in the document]
xThaid: Okay, now I get it. Thanks for the clarification.

sys/kern/lockdep.c
```diff
@@ -32,7 +32,12 @@ typedef struct lock_class_link {
 static spin_t main_lock = SPIN_INITIALIZER(0);
 
 #define CLASSHASH_SIZE 64
-#define CLASSHASH(key) ((uintptr_t)(key) % CLASSHASH_SIZE)
+/* We have to divide the key by the alignment of lock_class_key_t to prevent the
+   periodicity.
+   A key which is a multiple of 4 results in a hash which is also a multiple of
+   that number. So in the end we would only use every 4th bucket. */
+#define CLASSHASH(key)                                                         \
+  (((uintptr_t)(key) / alignof(lock_class_key_t)) % CLASSHASH_SIZE)
```
j-piecuch, line 40: I would also add a static assertion that `alignof(lock_class_mapping_t) >= alignof(lock_class_key_t)`, since `key` can also be a pointer to `lock_class_mapping_t`.

j-piecuch (changes requested):

sys/kern/lockdep.c
j-piecuch: ...but `lock_class_mapping_t` is aligned to 8 bytes on aarch64, so lock classes of statically allocated locks will have a `key` that's a multiple of 8, not 4... we could fix this by requiring `lock_class_key_t` to have the same alignment as `lock_class_mapping_t`, but that will waste more space for keys...

xThaid: I wouldn't care much about that extra memory. I've changed the type of `lock_class_key_t` to `uintptr_t` so the alignment will be always the length of the word.

j-piecuch: Ok, good enough.

j-piecuch (approved):

franciscozdo (changes requested):
Can you provide some information about using lockdep? Is it supposed to panic when something is wrong or it will show some additional information? I tried to use `call lockdep_print_graph()`, but it seems that there is no such function.

xThaid:
Yeah, the description of the PR is outdated. I've added short comment about usage in the code.

franciscozdo (approved):
LGTM except one typo in comment

sys/kern/lockdep.c
```diff
@@ -233,30 +233,43 @@ void lockdep_init(void) {
...
+  class = look_up_lock_class(lock);
+  assert(class);
+
+  /* Delete the class of the released lock and move the rest shift the rest */
+  for (int i = 0; i < thread->td_lock_depth; i++) {
+    if (class == thread->td_held_locks[i])
+      shift = true;
+    else if (shift)
+      thread->td_held_locks[i - 1] = thread->td_held_locks[i];
+  }
```
j-piecuch, line 268: What if we're holding two different locks of the same class? That could lead to the wrong class being removed when releasing one of them. I think we should restore the `held_lock_t` type and hold a pointer in it to the class as well as the exact lock that was acquired (or store just the lock, and access the class through it).

j-piecuch (changes requested):

sys/kern/lockdep.c
franciscozdo: But then we remember this class twice, so when we remove one we still have one left.

j-piecuch: Yes, but the exact pointer to the class being removed might not be the one that was added when the lock was acquired.

j-piecuch: Ok, I think I have convinced myself that this isn't an issue. This would be a problem if we could hold two locks belonging to the same class, with one lock from another class in between, e.g. `(A::lock1) (B::lock2) (A::lock3)`, but then acquiring `lock3` would create a cycle, so it's impossible.

sys/kern/lockdep.c
```diff
@@ -0,0 +1,278 @@
...
```
j-piecuch, line 268: I think this would be more readable if we split it into two loops: one that searches for the class being released, and another one that shifts all classes above it. We could also throw in an assertion that makes sure the class of the lock being release is actually held.
```suggestion
  int i = 0;
  /* Search for the class being released. */
  while (i < thread->td_lock_depth && class != thread->td_held_locks[i])
    i++;
  assert(i < thread->td_lock_depth);
  /* Move all classes above it downward in the stack. */
  for (i = i + 1; i < thread->td_lock_depth; i++)
    thread->td_held_locks[i - 1] = thread->td_held_locks[i];
```

j-piecuch (changes requested):

sys/kern/lockdep.c
xThaid: I also can't come up with a scenario where removing "not our" class could be considered harmful. In fact, when a new lock is being acquired the ordering of currently held locks doesn't matter.

sys/kern/lockdep.c
```diff
@@ -0,0 +1,280 @@
...
+
+  thread->td_held_locks[thread->td_lock_depth++] = class;
+
+  add_prev_link();
+
+  lockdep_unlock();
+}
+
+void lockdep_release(lock_class_mapping_t *lock) {
+  bool shift = false;
```
j-piecuch, line 254:
```suggestion
```

sys/kern/lockdep.c
```diff
@@ -0,0 +1,280 @@
...
+
+  class = look_up_lock_class(lock);
+  assert(class);
+
+  /* Delete the class of the released lock and shift the rest */
+  int i = 0;
+  /* Search for the class being released. */
+  while (i < thread->td_lock_depth && class != thread->td_held_locks[i])
+    i++;
+  assert(i < thread->td_lock_depth);
```
j-piecuch, line 267: Searching from the top of the stack should be faster than from the bottom.
```suggestion
  int i = thread->td_lock_depth - 1;
  /* Search for the class being released. */
  while (i >= 0 && class != thread->td_held_locks[i])
    i--;
  assert(i >= 0);
```

j-piecuch (changes requested):

sys/kern/lockdep.c
```diff
@@ -0,0 +1,278 @@
...
+
+  lockdep_unlock();
+}
+
+void lockdep_release(lock_class_mapping_t *lock) {
+  lock_class_t *class = lock->lock_class;
+  thread_t *thread = thread_self();
+  assert(class);
+
+  lockdep_lock();
```
j-piecuch, line 258: Do we even need to acquire the lock here? We're only accessing data that's private to the calling thread.

xThaid: Yeah, we only access fields from the thread and no one else can modify it beside that calling one.

j-piecuch: Actually, if we don't acquire the lock then we can receive an interrupt, but interrupt handlers can only use spinlocks, which are not (currently) checked by lockdep, so I don't think there is an issue, but there should be a comment mentioning it.

sys/kern/lockdep.c
```diff
@@ -250,6 +250,7 @@ void lockdep_acquire(lock_class_mapping_t *lock) {
   lockdep_unlock();
 }
 
+/* This function is called with the `lock` held. */
```
j-piecuch, line 253: If this comment is supposed to address my remarks about interrupts, this is not what I meant -- I wanted the comment to say that acquiring the lockdep lock here is unnecessary, but only because interrupt handlers can only use spinlocks, and spinlocks are not tracked by lockdep. If spinlocks were tracked by lockdep, we could get an interrupt in the middle of `lockdep_release()` and the interrupt handler could see an inconsistent state.

xThaid: OK, I get it now. Spinlocks will be added in the near future, not in this PR, so I will just put the locks back.

j-piecuch (approved):

cahirwpz (approved):

PR #998, opened 2021-02-14
Make it possible to roll back uiomove() operations

j-piecuch:
This PR adds the `uio_rollback(uio_t *uio, size_t nbytes)` function, which rolls back the state of `uio` by the specified number of bytes. Right now, the only place where we need it is in `pty_write()`, which can fail to write a character __after__ reading from the caller-supplied `uio` structure via `uiomove()`.
It required adding two new fields to the `uio_t` structure: the total number of bytes (done + remaining) and the sum of the lengths of processed segments. Now `uiomove()` doesn't need to modify elements of the `uio_iov` array.
`uio_rollback()` return `EINVAL` if the requested number of bytes to roll back is larger than the number of bytes processed.


sys/kern/uio.c
```diff
@@ -81,6 +82,20 @@ int uiomove(void *buf, size_t n, uio_t *uio) {
   return error;
 }
 
+void uio_rollback(uio_t *uio, size_t nbytes) {
+  /* We assume that the `nbytes` supplied by the caller won't cause us to go
+   * past the beginning of the original uio->uio_iov.  */
```
franciscozdo, line 87: I think you should check this. If you leave it like it is someone will have trouble with finding this in the future. (Or maybe it  will be used only in `pty_write()` so we don't have to worry about it?)

franciscozdo (changes requested):
I am not sure we don't have to worry about `nbytes`.

wiklam (commented):
I am not sure but I am assuming that the sum of all `iov_len` for an `iov` is smaller than the capacity of `size_t`; instead of holding the offset of the current segment, it would be better to hold the sum of the previous segments and calculate the segment offset from `uio_offset`?

j-piecuch:
> I am not sure but I am assuming that the sum of all `iov_len` for an `iov` is smaller than the capacity of `size_t`; instead of holding the offset of the current segment, it would be better to hold the sum of the previous segments and calculate the segment offset from `uio_offset`? This approach allows us to add some checks for free.

@wiklam `uio_offset` isn't an offset into the `uio` (i.e. it doesn't tell us how many bytes have been written). It is an offset into the target object, so it can have a non-zero value at the beginning.

sys/kern/uio.c
j-piecuch: Checking this would require adding an additional field to `uio_t`. For instance, we could keep the number of bytes written so far.

franciscozdo: Yes. Why can't we add it?

j-piecuch: We can, I just wasn't sure if it's worth it. I'll add the extra field.

franciscozdo (approved):

sys/kern/pty.c
```diff
@@ -108,7 +108,7 @@ static int pty_write(file_t *f, uio_t *uio) {
       break;
     if ((error = pty_putc_sleep(tty, pty, c))) {
       /* Undo the last uiomove(). */
-      uio->uio_resid++;
+      uio_rollback(uio, 1);
```
wiklam, line 111: To be consistent I would check the value returned by `uio_rollback`?

sys/tests/uiomove.c
```diff
@@ -50,13 +52,24 @@ static int test_uiomove(void) {
...
 
   res = uiomove((char *)text, strlen(text), &uio);
   assert(res == 0);
   buffer2[37] = 0; /* Manually null-terminate */
   res = strcmp(buffer2, "Example ====string ========with data ");
   assert(res == 0);
 
+  /* Roll back and redo the read and check that the result is the same. */
+  memset(buffer2, '=', sizeof(buffer2));
+  uio_rollback(&uio, 8 + 7 + 10);
```
wiklam, line 66: Same as above.

wiklam (commented):
Despite the two comments, the rest is LGTM.

wiklam (approved):
LGTM

j-piecuch:
Closing, as #1053 has been chosen as the best solution

PR #999, opened 2021-02-14
Make Mimiker compile under GCC11

xThaid:
Apparently GCC11 shows new warnings when compiling the kernel:
1. Function declarations are inconsistent with definitions (e.g. `int pipe(int *);` vs `int pipe(int fildes[2])`)
2. `__progname` in `getprogname.c` must be explicitly set as a weak symbol (strong symbol is defined in `crt0-common.c`)


pj1031999 (approved):
LGTM

cahirwpz:
Please first introduce GCC 11 into our build system.

jpszczolowski:
@xThaid have you tried compiling Mimiker with the GCC's Static Analyzer? It's `-fanalyzer` flag available from GCC10 onwards.

xThaid:
I wasn't even aware of that feature. Thanks @jpszczolowski !

So I have compiled the code with that flag and there were of plenty of warnings. Sadly, none of them was useful at all. Some were false positives, because GCC doesn't understand meaning of macros. For example this was detected as potential NULL deference of `data`:
```
  unsigned dir = cmd & IOC_DIRMASK;
  void *data = NULL;
  if (dir)
    data = kmalloc(M_TEMP, len, 0);
  if ((dir & IOC_IN) && (error = copyin(u_data, data, len)))
    goto fail;
```
but `dir & IOC_IN != 0` implies `cmd & IOC_DIRMASK != 0`, so `copyin` won't be never called with `data == NULL`.

What is even worse, the static analyzer is bugged (I have a non-official version, so it might be the cause). Here's an excerpt from GCC output:
```
                         |  276 |   if (pgrp) {
                         |      |   ^~ ~
                         |      |   |  |
                         |      |   |  (29) following 'true' branch (when 'pgrp' is non-NULL)...
                         |      |   (28) ...to here
                         |  277 |     tty_t *tty = pgrp->pg_session->s_tty;
                         |      |     ~~~~~        ~~~~~~~~~~~~~~~~
                         |      |     |                |
                         |      |     |                (31) dereference of NULL 'pgrp'
                         |      |     (30) ...to here
                         |

```
It firstly assumes that `pgrp` is NULL and then states there will be NULL deference on `pgrp`. It looks like a contradiction.

j-piecuch (approved):

cahirwpz (approved):

PR #1000, opened 2021-02-16
Fix deadlock caused by call to pmap_page_remove in pm_free_from_seg

### LINKED ISSUE #968: Deadlock between physmem_lock and pv_list_lock
There is no strict order of lock acquisition between `physmem_lock` and `pv_list_lock`. This results in a deadlock. 
Consider the scenario with two threads:
Thread 1:
```
#5  vm_page_alloc (npages=npages@entry=0x1) at /root/mimiker/sys/kern/vm_physmem.c:200
#6  pmap_pagealloc () at /root/mimiker/sys/mips/pmap.c:175
#7  pmap_add_pde (pmap=pmap@entry=0xc0019230, vaddr=vaddr@entry=0x400000) at /root/mimiker/sys/mips/pmap.c:184
#8  pmap_pte_write (pmap=pmap@entry=0xc0019230, vaddr=vaddr@entry=0x400000, pte=0xf8d8, pte@entry=0xf8c0, flags=flags@entry=0x0) at /root/mimiker/sys/mips/pmap.c:223
#9  pmap_enter (pmap=0xc0019230, va=va@entry=0x400000, pg=0xc01b18cc, prot=<optimized out>, flags=flags@entry=0x0) at /root/mimiker/sys/mips/pmap.c:336
#10 vm_page_fault (map=0xc00181c8, fault_addr=fault_addr@entry=0x40038c, fault_type=<optimized out>) at /root/mimiker/sys/kern/vm_map.c:433
```
`pv_list_lock` is acquired in `pmap_enter` and then in `vm_page_alloc` thread 1 tries to lock `physmem_lock`.

Thread 2:
```
#5  pmap_page_remove (pg=pg@entry=0xc01ace44) at /root/mimiker/sys/mips/pmap.c:387
#6  pm_free_from_seg (seg=<optimized out>, page=page@entry=0xc01ace44) at /root/mimiker/sys/kern/vm_physmem.c:242
#7  vm_page_free (page=0xc01ace44) at /root/mimiker/sys/kern/vm_physmem.c:257
#8  vm_object_remove_page_nolock (obj=obj@entry=0xc00325e8, page=<optimized out>) at /root/mimiker/sys/kern/vm_object.c:63
#9  vm_object_free (obj=0xc00325e8) at /root/mimiker/sys/kern/vm_object.c:92
#10 vm_segment_free (seg=seg@entry=0xc00335c8) at /root/mimiker/sys/kern/vm_map.c:132
#11 vm_segment_destroy (map=map@entry=0xc0018148, seg=0xc00335c8) at /root/mimiker/sys/kern/vm_map.c:161
#12 vm_map_delete (map=0xc0018148) at /root/mimiker/sys/kern/vm_map.c:197
```
`physmem_lock` is locked in `vm_page_free` and in `pv_list_lock` thread 2 wants to lock `pv_list_lock`.

I couldn't find obvious solution, so for now I leave the problem here.

cahirwpz:
It should fix #968 issue. I investigated every call to `vm_page_free` and made sure there's corresponding call to `pmap_remove` or `pmap_kremove`.

### DIFF
--- a/sys/kern/kmem.c
+++ b/sys/kern/kmem.c
@@ -74,20 +74,20 @@ vm_page_t *kva_find_page(vaddr_t ptr) {
   return NULL;
 }
 
-void kva_unmap(vaddr_t va, size_t size) {
-  assert(page_aligned_p(va) && page_aligned_p(size));
+void kva_unmap(vaddr_t ptr, size_t size) {
+  assert(page_aligned_p(ptr) && page_aligned_p(size));
 
-  kasan_mark_invalid((void *)va, size, KASAN_CODE_KMEM_FREED);
+  kasan_mark_invalid((void *)ptr, size, KASAN_CODE_KMEM_FREED);
 
-  size_t end = va + size;
-  size_t pgsz;
-  for (; va < end; va += pgsz) {
+  vaddr_t va = ptr;
+  while (va < ptr + size) {
     vm_page_t *pg = kva_find_page(va);
     assert(pg != NULL);
-    pgsz = pg->size * PAGESIZE;
-    pmap_kremove(va, pgsz);
+    va += pg->size * PAGESIZE;
     vm_page_free(pg);
   }
+
+  pmap_kremove(ptr, size);
 }
 
 void *kmem_alloc(size_t size, kmem_flags_t flags) {
--- a/sys/kern/kmem.c
+++ b/sys/kern/kmem.c
@@ -79,8 +79,9 @@ void kva_unmap(vaddr_t va, size_t size) {
 
   kasan_mark_invalid((void *)va, size, KASAN_CODE_KMEM_FREED);
 
+  size_t end = va + size;
   size_t pgsz;
-  for (; va < va + size; va += pgsz) {
+  for (; va < end; va += pgsz) {
     vm_page_t *pg = kva_find_page(va);
     assert(pg != NULL);
     pgsz = pg->size * PAGESIZE;
--- a/sys/kern/kmem.c
+++ b/sys/kern/kmem.c
@@ -74,21 +74,19 @@ vm_page_t *kva_find_page(vaddr_t ptr) {
   return NULL;
 }
 
-void kva_unmap(vaddr_t ptr, size_t size) {
-  assert(page_aligned_p(ptr) && page_aligned_p(size));
+void kva_unmap(vaddr_t va, size_t size) {
+  assert(page_aligned_p(va) && page_aligned_p(size));
 
-  kasan_mark_invalid((void *)ptr, size, KASAN_CODE_KMEM_FREED);
+  kasan_mark_invalid((void *)va, size, KASAN_CODE_KMEM_FREED);
 
-  vaddr_t va = (vaddr_t)ptr;
-  vaddr_t end = va + size;
-  while (va < end) {
+  size_t pgsz;
+  for (; va < va + size; va += pgsz) {
     vm_page_t *pg = kva_find_page(va);
     assert(pg != NULL);
-    va += pg->size * PAGESIZE;
+    pgsz = pg->size * PAGESIZE;
+    pmap_kremove(va, pgsz);
     vm_page_free(pg);
   }
-
-  pmap_kremove((vaddr_t)ptr, size);
 }
 
 void *kmem_alloc(size_t size, kmem_flags_t flags) {
--- a/sys/kern/kmem.c
+++ b/sys/kern/kmem.c
@@ -79,15 +79,19 @@ void kva_unmap(vaddr_t ptr, size_t size) {
 
   kasan_mark_invalid((void *)ptr, size, KASAN_CODE_KMEM_FREED);
 
+  vm_pagelist_t pglist;
+  TAILQ_INIT(&pglist);
+
   vaddr_t va = ptr;
   while (va < ptr + size) {
     vm_page_t *pg = kva_find_page(va);
     assert(pg != NULL);
     va += pg->size * PAGESIZE;
-    vm_page_free(pg);
+    TAILQ_INSERT_TAIL(&pglist, pg, pageq);
   }
 
   pmap_kremove(ptr, size);
+  vm_pagelist_free(&pglist);
 }
 
 void *kmem_alloc(size_t size, kmem_flags_t flags) {
--- a/sys/kern/vm_map.c
+++ b/sys/kern/vm_map.c
@@ -172,7 +172,7 @@ void vm_segment_destroy_range(vm_map_t *map, vm_segment_t *seg, vaddr_t start,
   }
 
   size_t length = end - start;
-  vm_object_remove_range(seg->object, start - seg->start, length);
+  vm_object_remove_pages(seg->object, start - seg->start, length);
   pmap_remove(map->pmap, start, end);
 
   if (seg->start == start) {
@@ -181,8 +181,8 @@ void vm_segment_destroy_range(vm_map_t *map, vm_segment_t *seg, vaddr_t start,
     seg->end = start;
   } else { /* a hole inside the segment */
     vm_object_t *obj = vm_object_clone(seg->object);
-    vm_object_remove_range(obj, 0, start - seg->start);
-    vm_object_remove_range(seg->object, end - seg->start, seg->end - end);
+    vm_object_remove_pages(obj, 0, start - seg->start);
+    vm_object_remove_pages(seg->object, end - seg->start, seg->end - end);
     vm_segment_t *new_seg =
       vm_segment_alloc(obj, end, seg->end, seg->prot, seg->flags);
     seg->end = start;
@@ -331,7 +331,7 @@ int vm_segment_resize(vm_map_t *map, vm_segment_t *seg, vaddr_t new_end) {
     /* Shrinking entry */
     off_t offset = new_end - seg->start;
     size_t length = seg->end - new_end;
-    vm_object_remove_range(seg->object, offset, length);
+    vm_object_remove_pages(seg->object, offset, length);
     /* TODO there's no reference to pmap in page, so we have to do it here */
     pmap_remove(map->pmap, new_end, seg->end);
   }
--- a/sys/kern/vm_map.c
+++ b/sys/kern/vm_map.c
@@ -166,6 +166,7 @@ void vm_segment_destroy_range(vm_map_t *map, vm_segment_t *seg, vaddr_t start,
   assert(mtx_owned(&map->mtx));
   assert(start >= vm_map_start(map) && end <= vm_map_end(map));
 
+  pmap_remove(map->pmap, start, end);
   if (seg->start == start && seg->end == end) {
     vm_segment_destroy(map, seg);
     return;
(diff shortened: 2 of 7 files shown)


sys/kern/vm_object.c
```diff
@@ -55,43 +55,36 @@ void vm_object_add_page(vm_object_t *obj, off_t offset, vm_page_t *pg) {
...
-  SCOPED_MTX_LOCK(&obj->mtx);
-
-  vm_object_remove_page_nolock(obj, page);
-}
-
-void vm_object_remove_range(vm_object_t *object, off_t offset, size_t length) {
-  SCOPED_MTX_LOCK(&object->mtx);
+static void vmo_remove_pages_nolock(vm_object_t *obj, vm_offset_t offset,
+                                    size_t length) {
+  assert(mtx_owned(&obj->mtx));
```
pj1031999, line 60: `vmo` vs `vm_object` - could you stick to one?

sys/kern/vm_object.c
```diff
@@ -55,43 +55,36 @@ void vm_object_add_page(vm_object_t *obj, off_t offset, vm_page_t *pg) {
...
   if (!refcnt_release(&obj->ref_counter))
     return;
 
-  WITH_MTX_LOCK (&obj->mtx) {
-    vm_page_t *pg, *next;
-    TAILQ_FOREACH_SAFE (pg, &obj->list, obj.list, next)
-      vm_object_remove_page_nolock(obj, pg);
-  }
-
+  vm_object_remove_pages(obj, 0, -PAGESIZE);
```
pj1031999, line 87: `len` argument in `vm_object_remove_pages` is `size_t`. Here you pass negative value. It's hard to understand without any description.

sys/kern/vm_object.c
cahirwpz: Done.

sys/kern/vm_object.c
cahirwpz: I could use `rounddown(~0, PAGESIZE)` or `~0 & ~(PAGESIZE-1)` – each one of them is somewhat ugly.

Maybe a macro `vm_object_remove_all_pages(obj)` that expands to `vm_object_remove_pages(obj, 0, -PAGESIZE)` would do ?

pj1031999: Macro is good enough.

cahirwpz: Done.

pj1031999 (approved):
LGTM

sys/kern/kmem.c
```diff
@@ -74,21 +74,20 @@ vm_page_t *kva_find_page(vaddr_t ptr) {
...
-  vaddr_t end = va + size;
-  while (va < end) {
+  size_t end = va + size;
+  size_t pgsz;
+  for (; va < end; va += pgsz) {
     vm_page_t *pg = kva_find_page(va);
     assert(pg != NULL);
-    va += pg->size * PAGESIZE;
+    pgsz = pg->size * PAGESIZE;
+    pmap_kremove(va, pgsz);
```
j-piecuch, line 88: It would be more efficient to unmap all the pages at once before the loop. Can we do that here?

sys/kern/vm_physmem.c
```diff
@@ -293,7 +293,6 @@ static void pm_free_from_seg(vm_physseg_t *seg, vm_page_t *page) {
   PAGECOUNT(page)++;
   page->flags |= PG_MANAGED;
   for (unsigned i = 0; i < page->size; i++) {
-    pmap_page_remove(&page[i]);
```
j-piecuch, line 296: I think it would be a good idea to add an assertion to `vm_page_free()` which checks that there are no mappings of that page, i.e. that `page->pv_list` is empty.

j-piecuch (changes requested):

sys/kern/vm_physmem.c
cahirwpz: I agree it would be nice to check here that page is not referenced anymore. If I used `&page[i].pv_list` that would be half-correct since a page can be mapped by `pmap_kenter`, which does not maintain `pv_entry` list. Unfortunately we don't maintain reference count on a page. If we did, I'd definitely check it here.

I added a comment that the check is missing here. Is that ok?

cahirwpz: Oh... one more thing... If I checked `&page[i].pv_list` here I would probably need to lock `pv_list_lock` which is exactly what I'm trying to avoid :)

sys/kern/kmem.c
cahirwpz: Ok. I reverted to the original structure of this function.

However... there's something fishy here. I cannot perform `pmap_kremove` before the loop, since I use `va -> pa` mapping to extract `vm_page_t` addresses. So there's a window of time when formally pages are free, but still visible through kernel virtual address space.

Ideally I would have to:
 * collect all pages to free on a list,
 * unmap the region with `pmap_kremove`,
 * and free the pages with `vm_pagelist_free`. 

The only problem is that the last function does not exists. Should I add it ?

sys/kern/vm_physmem.c
j-piecuch: IIUC, there must be no mappings of a page in order to call `vm_page_free()` on it. Furthermore, no one else will be allocated the page until `pm_free_from_seg()` does its thing, so at the beginning of `vm_page_free()` we can be sure that we're the only users of the page. I think it's safe to check `pv_list` without acquiring `pv_list_lock`.

sys/kern/kmem.c
```diff
@@ -79,16 +79,15 @@ void kva_unmap(vaddr_t ptr, size_t size) {
...
+  vaddr_t va = ptr;
+  while (va < ptr + size) {
     vm_page_t *pg = kva_find_page(va);
     assert(pg != NULL);
     va += pg->size * PAGESIZE;
     vm_page_free(pg);
   }
 
-  pmap_kremove((vaddr_t)ptr, size);
+  pmap_kremove(ptr, size);
```
j-piecuch, line 90: This is incorrect: someone can allocate and map the freed page before our call to `pmap_kremove()`, and our call to `pmap_kremove()` will unmap a page that someone else allocated and mapped. I think the call to `pmap_kremove()` should be before the loop.

j-piecuch (changes requested):

sys/kern/kmem.c
j-piecuch: Correction, we won't unmap someone else's mapping, since that someone will map the page at a different virtual address... still, I think we should first unmap the pages, and then free them.

sys/kern/vm_physmem.c
cahirwpz: Ok. It's still a useful temporary solution capable of spotting an error :)

sys/kern/kmem.c
cahirwpz: Ok. So `vm_pagelist_free` is necessary then.

sys/kern/kmem.c
cahirwpz: Seems like the answer is yes.

cahirwpz:
I need to fix a bug spotted by `assert`, add `vm_pagelist_free` and fix `kva_unmap`.

j-piecuch:
> I need to fix a bug spotted by `assert`, add `vm_pagelist_free` and fix `kva_unmap`.

I don't know if this is the bug you had in mind, but we need to move the `pmap_delete()` to the beginning of `vm_map_delete()`.

cahirwpz:
@j-piecuch If you have time to fix it right now, please do `commit` directly to this branch.

sys/kern/vm_object.c
```diff
@@ -55,43 +55,39 @@ void vm_object_add_page(vm_object_t *obj, off_t offset, vm_page_t *pg) {
...
-  vm_page_t *pg, *next;
-  TAILQ_FOREACH_SAFE (pg, &object->list, obj.list, next) {
-    if (pg->offset >= (off_t)(offset + length))
-      break;
-    if (pg->offset >= offset)
-      vm_object_remove_page_nolock(object, pg);
-  }
-}
+#define vm_object_remove_all_pages(obj)                                        \
+  vm_object_remove_pages((obj), 0, -PAGESIZE)
```
xThaid, line 84: I find this macro confusing. I presume that `-PAGESIZE` will be positive, but it doesn't look so obvious. And is there a reason to use macros instead of `static inline`?

xThaid (commented):
In general, looks good to me.

sys/kern/vm_object.c
cahirwpz: I can cast `-PAGESIZE` to `size_t` if that helps. This is a way to express succinctly the largest number aligned to `PAGESIZE` boundary.

sys/kern/vm_map.c
```diff
@@ -191,12 +191,12 @@ void vm_segment_destroy_range(vm_map_t *map, vm_segment_t *seg, vaddr_t start,
 }
 
 void vm_map_delete(vm_map_t *map) {
+  pmap_delete(map->pmap);
```
cahirwpz, line 194: I think you should consider moving this under `map->mtx` lock.

Remember that `vm_page_fault` can repopulate `pmap` with pages as long as there're segments in `vm_map` and nobody holds `map->mtx` lock. This may be a problem when we consider introduction of user-space threads.

cahirwpz (commented):
Thanks for the fix.

sys/kern/vm_map.c
j-piecuch: Hmmm... I'm not sure this is a problem, since any sane implementation of `exit` would wait until there's only one thread left before freeing the memory map, so there shouldn't be any page faults once we get to `vm_map_delete()`.

cahirwpz: You're right... hopefully that won't be a problem. Nonetheless `pmap` operation order change have made me think of various highly improbable scenarios.

sys/kern/vm_object.c
cahirwpz: Added a cast to `size_t`.

sys/kern/vm_physmem.c
```diff
@@ -293,27 +293,36 @@ static void pm_free_from_seg(vm_physseg_t *seg, vm_page_t *page) {
...
+void vm_page_free(vm_page_t *page) {
+  SCOPED_MTX_LOCK(physmem_lock);
+  vm_page_free_nolock(page);
+}
+
+void vm_pagelist_free(vm_pagelist_t *pglist) {
+  SCOPED_MTX_LOCK(physmem_lock);
+
+  vm_page_t *pg, *pg_next;
+  TAILQ_FOREACH_SAFE (pg, pglist, pageq, pg_next) { vm_page_free_nolock(pg); }
```
j-piecuch, line 325: `vm_page_free_nolock(pg)` calls `pm_free_from_seg(pg)`, which overwrites the `pg->pageq` entry, effectively removing `pg` from `pglist`, but since `TAILQ_REMOVE()` isn't used, I think this will leave `pglist` in a bad state. We can either call `TAILQ_REMOVE()` before freeing a page, or at the very least do `TAILQ_INIT(pglist)` at the end of the function.

j-piecuch (changes requested):

sys/kern/vm_physmem.c
cahirwpz: Done.

j-piecuch (approved):

PR #1004, opened 2021-02-18
adjust ringbuf_produce and consume for dma requirements

komik0:
rtl8139 needs produce and consume to manage buffer updated by dma


sys/kern/ringbuf.c
```diff
@@ -28,6 +28,24 @@ static void consume(ringbuf_t *buf, unsigned bytes) {
...
+  if (buf->count > buf->size)
+    buf->count = buf->size;
+}
+
+void ringbuf_consume(ringbuf_t *buf, unsigned bytes) {
+  buf->count -= bytes;
+  buf->tail += bytes;
+  buf->tail %= buf->size;
+  if (buf->count > buf->size)
+    buf->count = 0;
```
wiklam, line 46: The same as above, but with underflow.

sys/kern/ringbuf.c
```diff
@@ -28,6 +28,24 @@ static void consume(ringbuf_t *buf, unsigned bytes) {
...
 }
 
+/* below two function are to help track ringbuf pointers in case when it is
+ * changed by something else. e.q. DMA */
+void ringbuf_produce(ringbuf_t *buf, unsigned bytes) {
+  buf->count += bytes;
+  buf->head += bytes;
+  buf->head %= buf->size;
+  if (buf->count > buf->size)
+    buf->count = buf->size;
```
wiklam, line 38: If the buf->size is quite big, the buf->count can overflow and we will not notice it.

wiklam (commented):
IMO this PR does not have to depend on #1003

sys/kern/ringbuf.c
komik0: In case of DMA and my usage of this function it is not a problem. I will add assertion.

sys/kern/ringbuf.c
```diff
@@ -28,6 +28,22 @@ static void consume(ringbuf_t *buf, unsigned bytes) {
     buf->tail = 0;
 }
 
+/* below two function are to help to track ringbuf pointers in case when it is
+ * changed by something else. e.q. DMA */
+void ringbuf_produce(ringbuf_t *buf, unsigned bytes) {
+  assert(buf->count + bytes <= buf->size);
```
wiklam, line 34: Maybe I am too paranoic.
```suggestion
  assert(buf->count  <= buf->size && bytes <= buf->size &&  buf->count  <= buf->size - bytes);
```

sys/kern/ringbuf.c
```diff
@@ -11,7 +11,8 @@ void ringbuf_init(ringbuf_t *rb, void *buf, size_t size) {
 }
 
 static void produce(ringbuf_t *buf, unsigned bytes) {
```
xThaid, line 13: Can't we just merge these two functions `produce` and `ringbuf_produce`? They do basically the same. Similar situation with `consume` and `ringbuf_consume`.

sys/kern/ringbuf.c
```diff
@@ -28,6 +29,23 @@ static void consume(ringbuf_t *buf, unsigned bytes) {
...
 }
 
+/* below two function are to help to track ringbuf pointers in case when it is
+ * changed by something else. e.q. DMA */
+void ringbuf_produce(ringbuf_t *buf, unsigned bytes) {
+  assert(buf->count <= buf->size && bytes <= buf->size &&
+         buf->count <= buf->size - bytes);
+  buf->count += bytes;
+  buf->head += bytes;
+  buf->head %= buf->size;
```
xThaid, line 39: Modulo operation can be slow here. Since `buf->head` won't be bigger than `2 * buf->size` it can be done with conditional subtraction:
```
if (buf->head >= buf->size)
    buf->head -= buf->size;
```

sys/kern/ringbuf.c
```diff
@@ -28,6 +29,23 @@ static void consume(ringbuf_t *buf, unsigned bytes) {
...
+}
+
+void ringbuf_consume(ringbuf_t *buf, unsigned bytes) {
+  assert(buf->count >= bytes);
+  buf->count -= bytes;
+  buf->tail += bytes;
+  buf->tail %= buf->size;
```
xThaid, line 46: same here

xThaid (changes requested):

sys/kern/ringbuf.c
komik0: I was thinking about it.  I did it that way because that assumption was weird to me in case of produce/consume function:
```
assert(buf->head + bytes <= buf->size);
...
if (buf->head == buf->size)
  buf->head = 0;
```
I deduced (perhaps wrongly) that there was any reasons for handle this function in that way therefore I did it next to that functions. Done - tests (ringbuf) passed.

sys/kern/ringbuf.c
komik0: Thanks

sys/kern/ringbuf.c
komik0: Thank you - Done

sys/kern/ringbuf.c
```diff
@@ -107,7 +92,7 @@ int ringbuf_read(ringbuf_t *buf, uio_t *uio) {
     int res = uiomove((char *)buf->data + buf->tail, size, uio);
     if (res)
       return res;
-    consume(buf, size);
+    ringbuf_consume(buf, size);
```
komik0, line 95: @xThaid: that previously mentioned assertion  was confusing to me when I was wondering this usage of `consume`

xThaid: Yeah, the assumption was incorrect.

sys/kern/ringbuf.c
```diff
@@ -10,29 +10,32 @@ void ringbuf_init(ringbuf_t *rb, void *buf, size_t size) {
...
 }
 
-static void produce(ringbuf_t *buf, unsigned bytes) {
-  assert(buf->count + bytes <= buf->size);
-  assert(buf->head + bytes <= buf->size);
+/* below two function are to help to track ringbuf pointers in case when it is
+ * changed by something else. e.q. DMA */
+void ringbuf_produce(ringbuf_t *buf, unsigned bytes) {
+  assert(buf->count <= buf->size && bytes <= buf->size &&
+         buf->count <= buf->size - bytes);
```
xThaid, line 17: The first two statements are redundant, because they are just special cases of the third one. So:
`assert(buf->count <= buf->size - bytes);`
is sufficient.

xThaid (changes requested):
One more thing

sys/kern/ringbuf.c
wiklam: The first one should be already ok, but the second check is preventing us from underflow (unsigned).

xThaid: So we can simply reorder the last condition:`assert(bytes <= buf->size - buf->count);`, because we know that `buf->size >= buf->count`.

komik0: Thanks, done

xThaid (approved):

sys/kern/ringbuf.c
```diff
@@ -10,29 +10,31 @@ void ringbuf_init(ringbuf_t *rb, void *buf, size_t size) {
   rb->data = buf;
 }
 
-static void produce(ringbuf_t *buf, unsigned bytes) {
-  assert(buf->count + bytes <= buf->size);
-  assert(buf->head + bytes <= buf->size);
+/* below two function are to help to track ringbuf pointers in case when it is
```
wiklam, line 13:
```suggestion
/* below two functions are to help to track ringbuf pointers in case when it is
```
```suggestion
/* below two function are to help to track ringbuf pointers in case when it is
```

wiklam (approved):

sys/kern/ringbuf.c
```diff
@@ -10,29 +10,31 @@ void ringbuf_init(ringbuf_t *rb, void *buf, size_t size) {
...
   buf->count -= bytes;
   buf->tail += bytes;
-  if (buf->tail == buf->size)
-    buf->tail = 0;
+  if (buf->tail >= buf->size)
+    buf->tail -= buf->size;
+  assert(buf->tail < buf->size);
 }
 
 bool ringbuf_putb(ringbuf_t *buf, uint8_t byte) {
```
hadarai, line 33: You should describe what that function does. Like with two above.

include/sys/ringbuf.h
```diff
@@ -23,6 +23,8 @@ static inline bool ringbuf_full(ringbuf_t *buf) {
 }
 
 void ringbuf_init(ringbuf_t *rb, void *buf, size_t size);
+void ringbuf_produce(ringbuf_t *buf, unsigned bytes);
+void ringbuf_consume(ringbuf_t *buf, unsigned bytes);
 bool ringbuf_putb(ringbuf_t *buf, uint8_t byte);
```
jpszczolowski, line 28: Minor: you could change `unsigned` to `size_t` to stay consistent with the rest of `ringbuf_*` interface

sys/kern/ringbuf.c
```diff
@@ -10,29 +10,31 @@ void ringbuf_init(ringbuf_t *rb, void *buf, size_t size) {
   rb->data = buf;
 }
 
-static void produce(ringbuf_t *buf, unsigned bytes) {
-  assert(buf->count + bytes <= buf->size);
-  assert(buf->head + bytes <= buf->size);
+/* below two functions are to help to track ringbuf pointers in case when it is
+ * changed by something else. e.q. DMA */
+void ringbuf_produce(ringbuf_t *buf, unsigned bytes) {
```
jpszczolowski, line 15: Typo: `e.q.` :arrow_right: `e.g.`

jpszczolowski (commented):
I'd suggest changing the title & description to something meaningful and related to the scope of this PR.

sys/kern/ringbuf.c
komik0: Sorry, I am not an author of this function but imo this name suits well to what this function does (_putb_ -> put byte)

include/sys/ringbuf.h
komik0: thanks, done

jpszczolowski (approved):

include/sys/ringbuf.h
```diff
@@ -23,6 +23,8 @@ static inline bool ringbuf_full(ringbuf_t *buf) {
 }
 
 void ringbuf_init(ringbuf_t *rb, void *buf, size_t size);
+void ringbuf_produce(ringbuf_t *buf, size_t bytes);
```
cahirwpz, line 26: Please document the functions appropriately. Give a hint that these are used when external entity (wrt. main processor) reads or write into the buffer and it must be allocated in physical memory.

cahirwpz (changes requested):
The PR description doesn't describe your intent. Please document new interface.

include/sys/ringbuf.h
komik0: Ok, thanks - done

hadarai (approved):

cahirwpz:
Since we're going to drop support for Malta in the near future this driver has little to no value for us. Hence I decided to close the PR.

PR #1006, opened 2021-02-18
Make vm_segment_destroy() and vm_segment_destroy_range() more consistent

j-piecuch:
`vm_segment_destroy()` currently doesn't unmap its address range, while `vm_segment_destroy_range()` does.
To fix this, we introduce `vm_segment_destroy_unmapped()`, which does exactly what `vm_segment_destroy()` used to do, and make `vm_segment_destroy()` unmap its address range before calling `vm_segment_destroy_unmapped()`. No one is actually using `vm_segment_destroy()` now, but I decided to include it in case it's needed in the future.


sys/kern/vm_map.c
```diff
@@ -153,22 +153,28 @@ static void vm_map_insert_after(vm_map_t *map, vm_segment_t *after,
...
+void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
+  assert(mtx_owned(&map->mtx));
+  pmap_remove(map->pmap, seg->start, seg->end);
+  vm_segment_destroy_unmapped(map, seg);
+}
+
 void vm_segment_destroy_range(vm_map_t *map, vm_segment_t *seg, vaddr_t start,
                               vaddr_t end) {
   assert(mtx_owned(&map->mtx));
   assert(start >= vm_map_start(map) && end <= vm_map_end(map));
```
wiklam, line 173: IMO here we should check also that the start and end are inside a segment.

wiklam (commented):
Your change is lgtm, but do we really need the `vm_segment_destroy` function if it is not used right now?

wiklam (approved):
LGTM

franciscozdo (changes requested):
I don't thing we need this change now. Our VM system will be changed in near future by sth like [UVM](https://man.netbsd.org/uvm.9). I think that no one will be using this new `vm_segment_destroy()` function before this change.

j-piecuch:
Closing, as the VM code in question will likely be replaced in the near future.

PR #1008, opened 2021-02-18
Remove modulo operation (MIPS timer) & prefer MIPS timer over PIT

wiklam:
Modulo operation is too expensive in this place.


sys/mips/timer.c
```diff
@@ -27,9 +29,22 @@ static bintime_t mips_timer_gettime(timer_t *tm);
...
-  if (state->count.lo < state->last_count_lo)
+  if (state->count.lo < state->last_count_lo) {
+    state->cntr_modulo +=
+      ((uint32_t)(-1) - state->last_count_lo) + state->count.lo;
     state->count.hi++;
+  } else {
+    state->cntr_modulo += state->count.lo - state->last_count_lo;
+  }
+
+  if (state->cntr_modulo >= state->timer.tm_frequency) {
```
komik0, line 42: could you please explain or comment this line?

sys/mips/timer.c
```diff
@@ -85,10 +100,13 @@ static int mips_timer_stop(timer_t *tm) {
 static bintime_t mips_timer_gettime(timer_t *tm) {
   device_t *dev = tm->tm_priv;
   mips_timer_state_t *state = dev->state;
-  uint64_t count = read_count(state);
-  uint32_t sec = count / tm->tm_frequency;
-  uint32_t frac = count % tm->tm_frequency;
-  bintime_t bt = bintime_mul(HZ2BT(tm->tm_frequency), frac);
+  uint32_t sec, ticks;
+  WITH_INTR_DISABLED {
```
komik0, line 104: the interrupts are disabled to avoid change of time?

sys/mips/timer.c
```diff
@@ -117,7 +135,6 @@ static int mips_timer_attach(device_t *dev) {
   };
 
   tm_register(&state->timer);
-  tm_select(&state->timer);
```
komik0, line 120: why now we don't need to set `time_source`? is the `time_source` used anywhere?

komik0 (commented):
generally it looks good to me - but I have some question.

sys/mips/timer.c
wiklam: The `cntr_modulo` shows how many ticks elapsed since timer initialization modulo `tm_frequency`, at each time update we increase `cntr_modulo` with number of ticks which elapsed since the last check and if the value reaches `tm_frequency` it means that one second passed and we want to 'normalize' the counter.

sys/mips/timer.c
wiklam: We want the time read to be atomic.

sys/mips/timer.c
wiklam: It is set later in `init_clock. It is used for time access.

komik0 (approved):

sys/mips/timer.c
```diff
@@ -27,9 +29,22 @@ static bintime_t mips_timer_gettime(timer_t *tm);
 static uint64_t read_count(mips_timer_state_t *state) {
   SCOPED_INTR_DISABLED();
   state->count.lo = mips32_get_c0(C0_COUNT);
+
   /* detect hardware counter overflow */
-  if (state->count.lo < state->last_count_lo)
+  if (state->count.lo < state->last_count_lo) {
+    state->cntr_modulo +=
+      ((uint32_t)(-1) - state->last_count_lo) + state->count.lo;
```
j-piecuch, line 36: If `last_count_lo` is 1 and `count.lo` is 0, then we want to increase `cntr_modulo` by `UINT32_MAX`, right? This code will increase it by `UINT32_MAX - 1`. I don't think we need to do different arithmetic in the two cases, I think `state->cntr_modulo += state->count.lo - state->last_count_lo` will work just fine in both cases.

sys/mips/timer.c
```diff
@@ -65,10 +80,10 @@ static int mips_timer_start(timer_t *tm, unsigned flags, const bintime_t start,
 
   device_t *dev = tm->tm_priv;
   mips_timer_state_t *state = dev->state;
-
+  state->sec = 0;
+  state->cntr_modulo = 0;
   state->period_cntr = bintime_mul(period, tm->tm_frequency).sec;
   state->compare.val = read_count(state);
-  state->last_count_lo = state->count.lo;
```
j-piecuch, line 71: Why did you remove this line? Where is `last_count_lo` initialized?

sys/mips/timer.c
```diff
@@ -85,10 +100,13 @@ static int mips_timer_stop(timer_t *tm) {
...
-  uint32_t sec = count / tm->tm_frequency;
-  uint32_t frac = count % tm->tm_frequency;
-  bintime_t bt = bintime_mul(HZ2BT(tm->tm_frequency), frac);
+  uint32_t sec, ticks;
+  WITH_INTR_DISABLED {
+    read_count(state);
+    sec = state->sec;
+    ticks = state->cntr_modulo;
+  }
+  bintime_t bt = bintime_mul(HZ2BT(tm->tm_frequency), ticks);
```
j-piecuch, line 109: `HZ2BT()` does 64-bit division, which is as expensive as modulo. One way to work around this is to compute and store it in the timer structure at initialization time.

sys/mips/timer.c
j-piecuch, line 71: Why did you remove this line? Where is `last_count_lo` initialized?

j-piecuch (changes requested):

sys/mips/timer.c
j-piecuch: I think you should explicitly set it to 0 before calling `read_count()`, since you initialize other fields to 0 explicitly too.

sys/mips/timer.c
```diff
@@ -27,9 +29,22 @@ static bintime_t mips_timer_gettime(timer_t *tm);
...
     state->count.hi++;
+  } else {
+    state->cntr_modulo += state->count.lo - state->last_count_lo;
+  }
+
+  if (state->cntr_modulo >= state->timer.tm_frequency) {
+    state->cntr_modulo -= state->timer.tm_frequency;
+    state->sec++;
+  }
+  assert(state->cntr_modulo < state->timer.tm_frequency);
```
j-piecuch, line 46: When testing on QEMU 5.2.0, the kernel immediately panics on this assertion, because the counter value is too high at the beginning. We can fix this either by changing the `if` above to a `while`, or setting `C0_COUNT` to 0 in `mips_timer_start()`. I personally prefer the latter.

j-piecuch (changes requested):

sys/mips/timer.c
wiklam: My bad I figured wrong, that between 1 and 0 you have `UINT32_MAX - 1`, thanks!

sys/mips/timer.c
wiklam: You are right, it was a dumb decision, which can cause some confusion, thanks!

sys/mips/timer.c
wiklam: Missed that, thanks!

sys/mips/timer.c
wiklam: Done.

sys/mips/timer.c
```diff
@@ -78,12 +75,16 @@ static int mips_timer_start(timer_t *tm, unsigned flags, const bintime_t start,
...
 
+  mips32_setcount(0);
+
   device_t *dev = tm->tm_priv;
   mips_timer_state_t *state = dev->state;
   state->sec = 0;
   state->cntr_modulo = 0;
   state->period_cntr = bintime_mul(period, tm->tm_frequency).sec;
   state->compare.val = read_count(state);
+  state->last_count_lo = state->count.lo;
```
j-piecuch, line 86: `read_count()` makes use of `last_count_lo`, so I think you should initialize it to 0 before calling `read_count()` in the line above.

j-piecuch (changes requested):

j-piecuch (approved):

cahirwpz (approved):

cahirwpz (approved):

PR #1012, opened 2021-02-19
Implement sigpend() syscall, make test_signal_mask() more resilient

j-piecuch:
Semantics: https://pubs.opengroup.org/onlinepubs/9699919799/functions/sigpending.html

This syscall allows us to make the `test_signal_mask()` test more resilient to frequency changes resulting from `-icount shift=auto`.


bin/utest/signal.c
```diff
@@ -199,9 +199,12 @@ int test_signal_mask(void) {
   sigcont_handled = 0;
   kill(pid, SIGUSR1);
 
-  /* Yield a couple times to make sure that the child sends us the signal. */
-  for (int i = 0; i < 3; i++)
+  /* Wait until we get a signal from the child. */
+  sigset_t set;
+  do {
     sched_yield();
```
cahirwpz, line 205: Could we use `sigsuspend` here instead?

cahirwpz (commented):
`sigpending` is a nice addition to the kernel though not critical (in a broad sense).

I guess we need to revisit all uses of `sched_yield`. If they're used for synchronization purposes we should replace them with a valid method.

I have only one question regarding code. Otherwise LGTM

bin/utest/signal.c
j-piecuch: We could, but then the handler would run immediately after `sigsuspend()`, and in this test I want to check that the handler runs immediately after unblocking the signal with `sigprocmask()`.

cahirwpz: Right, I see.

cahirwpz (approved):

PR #1015, opened 2021-02-21
Identify which tests do not tolerate time jumps.

cahirwpz:
Maybe we should rethink those tests instead of returning to `icount` option for _QEmu_. Linux seems to tolerate time jumps well.


j-piecuch:
> Linux seems to tolerate time jumps well.

What do you mean by that? Have you run Linux kernel tests in the same environment (QEMU, inside a virtual machine, with other processes competing with QEMU for CPU time)?

cahirwpz:
I've run Debian Wheezy and at first glance it did behave well. However I have not run LTP tests on it.
Since I believe #1011 delivers correct solution for the problem I'm closing this PR.

PR #1017, opened 2021-02-23
Speed up Kernel Address Sanitizer runtime

cahirwpz:
This PR's goal is to rewrite commonly used routines from `libkern.h` in similar fashion to `kasan_memcpy` an `kasan_strlen`. This should impact kernel performance in KASAN testruns and hopefully it will prevent timeouts.


cahirwpz:
This issue will be addressed by forthcoming PR.

PR #1001, opened 2021-02-16
Simplify control flow in sig_kill()

j-piecuch:
- Add `ksiginfo_unqueue()` and `sigpend_delete_set()`
- Make `SA_CONT` a signal flag, as opposed to a signal action
- Make all members of `sigprop_t` have a value with one set bit
- Add convenience macros for signal testing signal actions and properties
- Simplify control flow in `sig_kill()`

PR #1018, opened 2021-02-23
Assert that the thread is stopped or stopping in thread_continue()

j-piecuch:
`thread_continue()` is meant to continue stopped threads, but currently we allow it to be called on all kinds of threads. This PR checks that the thread is in the expected state (i.e. either stopping or stopped).


j-piecuch:
This depends on #1001, so I will just include this change in that PR.

PR #1019, opened 2021-02-23
uart: introduce generic uart interface to interact with tty.

pj1031999:
Introduce generic uart interface to interact with tty.


include/sys/uart.h
```diff
@@ -0,0 +1,85 @@
...
+
+bool uart_getb_lock(uart_state_t *uart, uint8_t *byte_p);
+void uart_fill_txbuf(device_t *dev);
+
+void uart_init(device_t *dev, const char *name, size_t buf_size, void *state,
+               tty_t *tty);
+
+void uart_notify_out(tty_t *tty);
+
+intr_filter_t uart_intr(void *data /* device_t* */);
```
j-piecuch, line 83: I think only `uart_init()` and `uart_intr()` should be publicly visible.

sys/kern/uart.c
```diff
@@ -0,0 +1,114 @@
...
+ * Called with `tty->t_lock` held.
+ */
+void uart_notify_out(tty_t *tty) {
+  device_t *dev = tty->t_data;
+
+  if (ringbuf_empty(&tty->t_outq))
+    return;
+
+  uart_fill_txbuf(dev);
+}
```
j-piecuch, line 114: The TTY thread implementation right now is awkwardly split between `tty.c` and `uart.c`... I think it should be in a single file `uart_tty.c`.

include/sys/uart.h
```diff
@@ -0,0 +1,85 @@
...
+
+static inline uart_methods_t *uart_methods(device_t *dev) {
+  return (uart_methods_t *)(dev)->driver->interfaces[DIF_UART];
+}
+
+typedef struct uart_state {
+  spin_t u_lock;
+  ringbuf_t u_rx_buf;
+  ringbuf_t u_tx_buf;
+  tty_t *u_tty;
```
j-piecuch, line 34: I'm wondering whether we really need to store the TTY as part of the UART's state... Couldn't we store it in `tty_thread_t` instead? I feel like that's where it belongs.

sys/kern/tty.c
```diff
@@ -1147,6 +1149,50 @@ int tty_makedev(devfs_node_t *parent, const char *name, tty_t *tty) {
...
+          if (!tty_input(tty, byte))
+            klog("dropped character %hhx", byte);
+      }
+      if (work & TTY_THREAD_TXRDY)
+        uart_fill_txbuf(dev);
+    }
+  }
+}
+
+void tty_thread_create(const char *name, device_t *dev) {
```
j-piecuch, line 1187: A TTY thread can't be created for any device, so I think it should take `uart_state_t *` instead of `device_t *` as an argument.

j-piecuch (changes requested):
The problem I see with the current structure is that there is no clean separation between UART code and TTY thread code, so putting the code in separate files makes it harder to grok. IMO the code should be in a single file called `uart_tty.c` or `uart.c`.

sys/kern/tty.c
pj1031999: I agree but then - please correct me if I'm wrong - we can't use driver interfaces. They live in `device_t::driver::interfaces` and I'm not sure if parent reference in `uart_state_t` is a good idea.

include/sys/uart.h
pj1031999: Ok, I will do that.

include/sys/uart.h
pj1031999: So should I move `uart_getb_lock`, `uart_fill_txbuf`, `uart_fill_txbuf` into `uart_tty.c`?

j-piecuch: Yes, preferably as `static` functions.

include/sys/uart.h
pj1031999: done

sys/kern/uart.c
pj1031999: done

include/sys/uart.h
pj1031999: done

sys/kern/tty.c
```diff
@@ -1147,6 +1149,50 @@ int tty_makedev(devfs_node_t *parent, const char *name, tty_t *tty) {
...
+  uart_state_t *uart = dev->state;
+  tty_thread_t *ttd = &uart->u_ttd;
+  cv_init(&ttd->ttd_cv, "TTY thread notification");
+  ttd->ttd_thread =
+    thread_create(name, tty_thread, dev, prio_ithread(PRIO_ITHRD_QTY - 1));
+  sched_add(ttd->ttd_thread);
+}
+
```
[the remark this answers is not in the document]
pj1031999: done

include/sys/tty.h
```diff
@@ -197,4 +209,8 @@ static inline bool tty_opened(tty_t *tty) {
   return (tty->t_opencount > 0);
 }
 
+void tty_set_outq_nonempty_flag(tty_thread_t *ttd, tty_t *tty);
+
+void tty_thread_create(const char *name, device_t *dev);
+
```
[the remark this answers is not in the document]
pj1031999: done

include/sys/uart.h
```diff
@@ -0,0 +1,80 @@
...
+  uart_rx_ready_t rx_ready;
+  uart_putc_t putc;
+  uart_tx_ready_t tx_ready;
+  uart_tx_enable_t tx_enable;
+  uart_tx_disable_t tx_disable;
+} uart_methods_t;
+
+static inline uart_methods_t *uart_methods(device_t *dev) {
+  return (uart_methods_t *)(dev)->driver->interfaces[DIF_UART];
+}
```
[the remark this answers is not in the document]
pj1031999: From my point of view that convention is wrong and you should change that.
I'm not sure if using macro here have __any__ advantage over function - it's harder to read and more error prone.
I have converted that function into macro but I'm really unhappy with that.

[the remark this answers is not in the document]
xThaid: As far as I know we also different convention for some object-like structures like `fileops_t`, `vnodeops_t`, `vfsops_t`. I don't get why we can't just use something like this:
```
static inline int VFS_MOUNT(mount_t *m) {
    return m->mnt_vfsops->vfs_mount(m);
}
```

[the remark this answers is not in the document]
mohrcore: I don't think anyone really considers those macros to be superior, certainly I don't. It's just a matter of introducing changes that are irrelevant to the rest of the PR and solve the problem only partially.

As for `fileops_t` and `vnodeops_t`, I'm not that familiar with the details of these, but I suppose that they are could tackled along with `*_METHODS` within a signle PR that'd be focused on cosmetic changes regarding method wrappers.

mohrcore:
LGTM

mohrcore (approved):
LGTM.

MichalBlk (approved):
LGTM

j-piecuch (approved):

j-piecuch (approved):

PR #1020, opened 2021-02-23
New toolchain building method

xThaid:
This isn't complete, but I would like to get some feedback. There are a few things left to do:
- packaging
- short documentation
- add an ability to use custom gcc (by cloning repo or something)


cahirwpz:
Some remarks:

1. I'd expect host libraries (`gmp`, `isl`, `cloog`, `mpfr`, `mpc`) to be built once for both targets.
2. All sources should be unpacked exactly once.
3. Each target should have separate build directories for `gcc`, `binutils` and `gdb`.
4. We need to integrate Debian package build into corresponding Makefiles.

xThaid:
Each remark has been addressed.

toolchain/gnu/Makefile
```diff
@@ -0,0 +1,232 @@
+TOPDIR = $(CURDIR)
```
MichalBlk, line 1: Please add `# vim: tabstop=8 shiftwidth=8 noexpandtab:`. This is a common scheme used throughout the mimiker build system (see mimiker/build/).

toolchain/gnu/Makefile
```diff
@@ -0,0 +1,232 @@
...
+
+include $(TOPDIR)/common.mk
+
+all: package
```
MichalBlk, line 5: `all: ...` should appear before any Makefile includes. Please note that the default target is the first target which make encounters while scanning the file.

toolchain/gnu/Makefile
```diff
@@ -0,0 +1,232 @@
...
+.PHONY: all build clean build-host clean-host
+
+### GNU Binutils
+
+$(SRCDIR)/$(BINUTILS).tar.xz:
+	$(CURL) -o $@ $(BINUTILS-URL)
+
+binutils-unpack: $(SRCDIR)/binutils/.unpack
+$(SRCDIR)/binutils/.unpack: $(SRCDIR)/$(BINUTILS).tar.xz
+	cd $(SRCDIR) && tar xJf $^ && mv $(BINUTILS) binutils
```
MichalBlk, line 38: You shouldn't use any explicit commands like `cd`, `tar`, or `mv`. Appropriate wrappers should be used instead (please check mimiker/build/tools.mk).

toolchain/gnu/Makefile
```diff
@@ -0,0 +1,232 @@
...
+
+.PHONY: all build clean build-host clean-host
+
+### GNU Binutils
+
+$(SRCDIR)/$(BINUTILS).tar.xz:
+	$(CURL) -o $@ $(BINUTILS-URL)
+
+binutils-unpack: $(SRCDIR)/binutils/.unpack
+$(SRCDIR)/binutils/.unpack: $(SRCDIR)/$(BINUTILS).tar.xz
```
MichalBlk, line 37: Why do we need the `.unpack` file? Can't we use `$(SRCDIR)/binutils` instead (with no `touch $@`)?

toolchain/gnu/config.mk
```diff
@@ -0,0 +1,21 @@
...
+ISL-URL = "http://isl.gforge.inria.fr/$(ISL).tar.xz"
+MPFR-URL = "ftp://ftp.gnu.org/gnu/mpfr/$(MPFR).tar.xz"
+GMP-URL = "https://gmplib.org/download/gmp/$(GMP).tar.xz"
+MPC-URL = "https://ftp.gnu.org/gnu/mpc/$(MPC).tar.gz"
+CLOOG-URL = "http://www.bastoul.net/cloog/pages/download/$(CLOOG).tar.gz"
+BINUTILS-URL = "https://ftp.gnu.org/gnu/binutils/$(BINUTILS).tar.xz"
+GCC-URL = "https://ftp.gnu.org/gnu/gcc/$(GCC)/$(GCC).tar.xz"
+GDB-URL = "https://ftp.gnu.org/gnu/gdb/$(GDB).tar.xz"
+
+TARGETS = mipsel aarch64
```
MichalBlk, line 21: Instead of pasting lots of `.PHONY: ...` in `Makefile`, I'd suggest to do the following herein:
```
UTILS = isl mpfr gmp mpc cloog binutils gcc gdb
OPS   = unpack configure build install clean

PHONY: $(foreach util, $(UTILS), $(foreach op, $(OPS), $(util)-$(op)))
```

MichalBlk (changes requested):

toolchain/gnu/Makefile
xThaid: Without that `.unpack` file Make rebuilds the targets twice. I'm not sure why it happens, but modify time of the directory is constantly changed. So in the result, Make re-triggers the target. 

I don't think that these `.unpack` files is a huge problem.

MichalBlk (approved):
LGTM

cahirwpz (approved):

PR #1021, opened 2021-02-24
Timeout tests based on CPU time, not wall clock time

j-piecuch:
`util/cpulimit.py` is a script that executes a given command and sends a SIGUSR1 signal to the specified process once the CPU time limit is reached. We use it to run QEMU with a CPU time limit.

The script needs the `python-ptrace` package, so the CI environment needs to be updated to include it.


cahirwpz:
This solution seems to be quite convoluted. I wonder if QEmu patch wouldn't be well justified in here. I.e. add timeout to options, call setrlimit, handle SIGXCPU. If gdbserver is active then stop emulation, otherwise quit. What do you think?

j-piecuch:
@cahirwpz I gave it a shot today, and I wasn't able to get it working.

cahirwpz (approved):

PR #1022, opened 2021-02-25
pgrp_maybe_orphan(): send SIGHUP + SIGCONT to all members of pgrp if any member is stopped

j-piecuch:
From the specification of `_Exit()` in POSIX (https://pubs.opengroup.org/onlinepubs/9699919799/functions/_Exit.html):
> If the exit of the process causes a process group to become orphaned, and if any member of the newly-orphaned process group is stopped, then a `SIGHUP` signal followed by a `SIGCONT` signal shall be sent to each process in the newly-orphaned process group.


pj1031999 (approved):
LGTM

cahirwpz (approved):
LGTM

PR #1024, opened 2021-02-27
Port tetris from NetBSD

j-piecuch:
`tetris/input.c` needed to be modified so that it doesn't use `poll()`.


bin/tetris/input.c
```diff
@@ -0,0 +1,190 @@
...
+  signal(SIGALRM, handle_sigalrm);
+}
+
+static int read_one(char *c) {
+  if (have_char) {
+    have_char = 0;
+    *c = buffered_char;
+    return 1;
+  }
+  return read(STDIN_FILENO, c, 1);
```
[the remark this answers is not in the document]
j-piecuch: You're right, we should always read the buffered char after calling `rwait()`. I have fixed it.

franciscozdo (approved):

wiklam (approved):
LGTM, review focused on file bin/tetris/input.c.

sys/tests/callout.c
```diff
@@ -1,6 +1,7 @@
 #include <sys/klog.h>
 #include <sys/libkern.h>
 #include <sys/callout.h>
+#include <sys/time.h>
```
cahirwpz, line 4: A leftover after some change?

j-piecuch: Yes.

cahirwpz (approved):

PR #1028, opened 2021-03-03
[PoC] Detect kernel stack overflows using guard pages

j-piecuch:
This is a proof-of-concept implementation of kernel stack overflow detection using guard pages.
Every kernel stack (except `thread0`'s stack) has a guard page that will trap any accesses. In `kern_exc_enter`, we check for accesses to the guard page and switch to a panic stack if an illegal access is detected. This has to be done before we save the context on the stack (otherwise we would trigger the same exception again), so unfortunately we have to do it in assembly.


cahirwpz:
So far I haven't found this PR something I would like to be integrated into Mimiker in the long run. That's a pity but this tool isn't of much value, hence I decided to close the PR.

PR #1030, opened 2021-03-07
Simplify locking rules around pgrp_t::pg_lock

j-piecuch:
Right now there's only one place where we're acquiring multiple `pg_locks`, and we already happen to hold `all_proc_mtx` there, so we can simplify the locking rules by requiring that `all_proc_mtx` be held in order to acquire multiple `pg_locks`.


pj1031999 (approved):
LGTM

xThaid (approved):

cahirwpz (approved):
LGTM after looking into FreeBSD and NetBSD sources.

PR #1031, opened 2021-03-08
Assert for vm_segment_destroy_range and simplyfing do_munmap

wiklam:
Asserting that we are destroying the range between a segment can prevent us from some future bugs.
The max for `do_munmap` is unnecessary, cause `vm_map_find_segment` already guarantee that `addr >= start`.

### DIFF
--- a/sys/kern/vm_map.c
+++ b/sys/kern/vm_map.c
@@ -164,7 +164,7 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
 void vm_segment_destroy_range(vm_map_t *map, vm_segment_t *seg, vaddr_t start,
                               vaddr_t end) {
   assert(mtx_owned(&map->mtx));
-  assert(start >= vm_map_start(map) && end <= vm_map_end(map));
+  assert(seg->start >= vm_map_start(map) && seg->end <= vm_map_end(map));
   assert(start >= seg->start && end <= seg->end);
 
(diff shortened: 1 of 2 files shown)


pj1031999 (approved):
LGTM

sys/kern/vm_map.c
```diff
@@ -164,7 +164,8 @@ void vm_segment_destroy(vm_map_t *map, vm_segment_t *seg) {
 void vm_segment_destroy_range(vm_map_t *map, vm_segment_t *seg, vaddr_t start,
                               vaddr_t end) {
   assert(mtx_owned(&map->mtx));
-  assert(start >= vm_map_start(map) && end <= vm_map_end(map));
+  assert(seg->start >= vm_map_start(map) && seg->end <= vm_map_end(map));
```
j-piecuch, line 167: Is this assertion needed? Can't we assume that all segments in the `vm_map` are valid (i.e. they fall within the boundaries of the `vm_map`)?

wiklam: It is pretty obvious, idk why I had some doubts about it.

j-piecuch (approved):
LGTM. It looks like the semantics of `do_munmap()` are way too strict compared to [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/munmap.html), but I will fix that in another PR.

cahirwpz (approved):

PR #1033, opened 2021-03-08
Removing the expensive HZ2BT operation in pit gettime

wiklam:
The HZ2BT operation use a 64-bit division, which can be expensive, use the already calculated value for `HZ2BT(freq)`.


pj1031999 (approved):
LGTM

j-piecuch (approved):

cahirwpz (approved):

PR #1036, opened 2021-03-12
Added example change on the beginning of script.

hadarai:
bc i like


cahirwpz:
I guess we can close this PR?

hadarai:
That's right

PR #1037, opened 2021-03-12
Remove td_kframe field from thread_t

j-piecuch:
We don't use it anywhere.
Also, fix buggy FPU context save logic in `ctx_switch` on MIPS.


pj1031999 (approved):

franciscozdo (approved):

j-piecuch:
Kernel profiling (#979) makes use of `td_kframe`, so closing.

PR #1038, opened 2021-03-12
Preserve value of x18 register across kernel interrupts.

j-piecuch:
Currently, the value of the x18 register can be changed if an interrupt occurs in kernel mode. We fix this by not touching it in `ctx_load`.
Also, we simplify things by assuming in `user_exc_leave` that the user exception frame is at the top of the stack. Previously this was not always the case because `kern_execve()` directly called `user_exc_leave()`. Now, `kern_execve()` calls `switch_to_userspace()`, which sets the stack pointer to the user exception frame.


sys/aarch64/evec.S
```diff
@@ -60,28 +60,19 @@
...
 
+/* Load context from sp. */
 .macro  load_ctx el
         msr     daifset, #2
-.if \el == 0
-        load_pcpu x18
-        ldr     x18, [x18, #PCPU_CURTHREAD]
-        ldr     x18, [x18, #TD_UCTX]
-        mov     sp,  x18
-.endif
```
franciscozdo, line 80: Are you sure we don't need that code? I am not sure if userspace `sp` is the same as kernel `sp` in this situation (so we had to take it from some special place).

franciscozdo (commented):
Please answer my question.

sys/aarch64/evec.S
j-piecuch: See the comment in `mips/ebase.S`: `sp` points to the user exception frame if we came from `user_exc_enter` or `switch_to_userspace`, which are all the possible cases. The code I removed was needed because `kern_execve()` called `user_exc_leave()` directly, which would cause the stack pointer to point somewhere further down the stack. `switch_to_userspace()` adjusts the stack pointer so that it points to the user exception frame.

j-piecuch: Actually, there's another way to get to `user_exc_leave`: `thread_entry_setup()` sets up the kernel stack of a thread so that it returns to userspace immediately after being switched to. When that thread reaches `user_exc_leave`, its stack pointer will point at the user exception frame allocated at the bottom of the stack, so we're good.

franciscozdo: Thanks! Now I understand.

franciscozdo (approved):

pj1031999 (approved):
LGTM

cahirwpz (approved):

PR #1039, opened 2021-03-12
Fix build & test run with -O2 optimization flag

j-piecuch:
This PR fixes a couple of issues I encountered while trying to compile and run the kernel with `-O2`.

### DIFF
--- a/sys/aarch64/signal.c
+++ b/sys/aarch64/signal.c
@@ -35,7 +35,6 @@ int sig_send(signo_t sig, sigset_t *mask, sigaction_t *sa, ksiginfo_t *ksi) {
   _REG(uctx, X0) = sig;
   _REG(uctx, X1) = sc_info;
   _REG(uctx, X2) = sc_uctx;
-  _REG(uctx, SP) -= sizeof(intptr_t *);
   _REG(uctx, LR) = sc_code;
 
   return 0;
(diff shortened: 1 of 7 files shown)


pj1031999 (approved):

franciscozdo (approved):

sys/aarch64/signal.c
```diff
@@ -35,7 +35,6 @@ int sig_send(signo_t sig, sigset_t *mask, sigaction_t *sa, ksiginfo_t *ksi) {
   _REG(uctx, X0) = sig;
   _REG(uctx, X1) = sc_info;
   _REG(uctx, X2) = sc_uctx;
-  _REG(uctx, SP) -= sizeof(intptr_t *);
```
wiklam, line 38: Could you explain why we do not need it?

pj1031999:
@j-piecuch could you add O2 build and tests into CI in another PR?

j-piecuch:
> @j-piecuch could you add O2 build and tests into CI in another PR?

I could, but that's going to put more load on the server. I will do it if @cahirwpz approves.

sys/aarch64/signal.c
j-piecuch: We do it on MIPS because apparently the calling convention allows the callee to modify the top of the stack without allocating space itself. There is no such thing on ARM AFAIK.

pj1031999:
> I could, but that's going to put more load on the server. I will do it if @cahirwpz approves.

Build phase is relatively cheap. For tests we can limit `--times` parameter to something like 10.
I don't want your effort to be wasted by future changes.

wiklam (approved):

cahirwpz (approved):

PR #1040, opened 2021-03-13
KASAN: Dynamic shadow map size.

pj1031999:
Increase shadow map size in runtime.

Depends on #764.


j-piecuch (changes requested):

sys/mips/pmap.c
```diff
@@ -547,3 +552,42 @@ void pmap_delete(pmap_t *pmap) {
...
+    size = va - pmap_maxkvaddr;
+    pmap_maxkvaddr = va;
+  }
+
+  /*
+   * kasan_grow calls pmap_kenter which acquires pmap->mtx.
+   * But we are under maxkvaddr_lock from kmem so it's safe to call kasan_grow.
+   */
+  kasan_grow(size);
+  (void)size;
```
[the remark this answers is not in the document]
pj1031999: If `KASAN` is not defined then `kasan_grow` is `__nothing` and gcc will warn about unused variable.

sys/kern/kasan.c
```diff
@@ -217,10 +221,30 @@ static void call_ctors(void) {
   }
 }
 
+void kasan_grow(size_t size) {
+  assert(size % SUPERPAGESIZE == 0);
```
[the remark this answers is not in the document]
pj1031999: done

sys/kern/kasan.c
```diff
@@ -217,10 +221,30 @@ static void call_ctors(void) {
...
+  size_t num_pde = (size >> KASAN_SHADOW_SCALE_SHIFT) / SUPERPAGESIZE;
+  vaddr_t va = KASAN_MD_SHADOW_START + _kasan_shadow_size;
+
+  for (size_t i = 0; i < num_pde; ++i) {
+    for (size_t j = 0; j < PT_ENTRIES; ++j) {
```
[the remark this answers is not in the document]
pj1031999: I based on original KASAN - see `boot.c`. `SUPERPAGESIZE` is used here for calculate number of page directories entries that are need to allocate. Here I do only final mapping for shadow map but `pmap_kenter` allocate one page for each `i` in outer loop.
I didn't think about switching to superpages but yes it should be easy to do that in the future.

include/sys/kasan.h
```diff
@@ -53,6 +53,9 @@ typedef struct {
  * memory is usable. */
 void init_kasan(void);
 
+/* Increase address space covered by KASAN by size. */
+void kasan_grow(size_t size);
```
j-piecuch, line 57: Perhaps passing a virtual address instead of size would yield a cleaner interface? It would be more consistent with `pmap_growkernel()`, and we would no longer need for the extra `size` variable in `pmap_growkernel()`.

sys/mips/boot.c
```diff
@@ -100,11 +102,20 @@ __boot_text void *mips_init(void) {
...
 #if KASAN /* Prepare KASAN shadow mappings */
+  size_t kasan_sanitized_size =
+    roundup(va - MIPS_PHYS_TO_KSEG2(text), SUPERPAGESIZE);
+  /*
+   * TODO(pj): I believe that it's a correct value here... but thread0 stack
+   * shadow check triggers BIG error in that case so let's create bigger shadow
+   * for now.
+   * kasan_sanitized_size >> KASAN_SHADOW_SCALE_SHIFT;
+   */
+  size_t kasan_shadow_size = kasan_sanitized_size;
```
j-piecuch, line 113: I think you're getting an error because `kasan_shadow_size` isn't aligned to `SUPERPAGESIZE`, so `num_pde = kasan_shadow_size / SUPERPAGESIZE` will round down, and we will map less pages than required. The way to fix this is `kasan_sanitized_size = roundup(va - MIPS_PHYS_TO_KSEG2(text), SUPERPAGESIZE << KASAN_SHADOW_SCALE_SHIFT)`

sys/mips/pmap.c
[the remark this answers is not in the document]
j-piecuch: Maybe we wouldn't need it if we changed the interface of `kasan_grow()`?

j-piecuch (changes requested):

sys/mips/boot.c
pj1031999: Oh, thanks!

sys/mips/pmap.c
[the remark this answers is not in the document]
pj1031999: done

include/sys/kasan.h
pj1031999: done

sys/kern/kasan.c
```diff
@@ -217,10 +221,32 @@ static void call_ctors(void) {
...
+      va += PAGESIZE;
+      _kasan_shadow_size += PAGESIZE;
+    }
+  }
+
+  kasan_mark_valid(
+    (const void *)(KASAN_MD_SANITIZED_START + _kasan_sanitized_size), size);
+  _kasan_sanitized_size += size;
+}
+
```
j-piecuch, line 245: I don't like how this function looks for several reasons:
- The alignment requirement for `size` seems arbitrary (and pretty large). I don't see any particular reason why we would want to grow the sanitized space only in chunks of `SUPERPAGESIZE << 3`.
- The nested loop seems pointless at this point, and only makes the code harder to read.

I tried to write a simpler version and this is what I came up with:
```c
static inline vaddr_t kasan_va_to_shad_va(vaddr_t va) {
  return KASAN_MD_SHADOW_START +
         (va - KASAN_MD_SANITIZED_START) / KASAN_SHADOW_SCALE_SIZE;
}

void kasan_grow(vaddr_t maxkvaddr) {
  assert(mtx_owned(&maxkvaddr_lock))
  maxkvaddr = roundup2(maxkvaddr, PAGESIZE * KASAN_SHADOW_SCALE_SIZE);
  vaddr_t va = kasan_va_to_shad_va(_kasan_sanitized_end);
  vaddr_t end = kasan_va_to_shad_va(maxkvaddr);

  /* Allocate and map shadow pages to cover the new KVA space. */
  for (; va < end; va += PAGESIZE) {
    vm_page_t *pg = vm_page_alloc(1);
    pmap_kenter(va, pg->paddr, VM_PROT_READ | VM_PROT_WRITE, 0);
  }

  if (maxkvaddr > _kasan_sanitized_end) {
    kasan_mark_valid(
      (const void *)(_kasan_sanitized_end), maxkvaddr - _kasan_sanitized_end);
    _kasan_sanitized_end = maxkvaddr;
  }
}
```

Note that this assumes that we replace `_kasan_sanitized_size` and `_kasan_shadow_size` with `_kasan_sanitized_end`.

j-piecuch (changes requested):

sys/kern/kasan.c
```diff
@@ -217,10 +221,32 @@ static void call_ctors(void) {
   }
 }
 
+void kasan_grow(vaddr_t maxkvaddr) {
```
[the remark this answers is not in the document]
pj1031999: done

sys/kern/kasan.c
pj1031999: You are right. We decided with to do that for simplicity but now I see that removing that assumptions make code much cleaner.
I have applied your suggestion.
Aggressive rounding has been removed from `mips/pmap.c`.

j-piecuch (changes requested):

sys/kern/kasan.c
```diff
@@ -35,6 +38,7 @@ struct __asan_global {
   const void *odr_indicator; /* The address of the ODR indicator symbol */
 };
 
+size_t _kasan_sanitized_end;
```
j-piecuch, line 41:
```suggestion
vaddr_t _kasan_sanitized_end;
```

sys/mips/boot.c
```diff
@@ -128,8 +134,11 @@ __boot_text void *mips_init(void) {
   mips32_setindex(0);
   mips32_tlbwi();
 
-  /* Since the variable is in kseg2 we cannot initialize it earlier. */
+  /* Since variables are in kseg2 we cannot initialize them earlier. */
   _kernel_pmap_pde = pde;
+#if KASAN
+  _kasan_sanitized_end = kasan_sanitized_end;
```
j-piecuch, line 140:
```suggestion
  _kasan_sanitized_end = KASAN_MD_SANITIZED_START + kasan_sanitized_size;
```

sys/mips/boot.c
```diff
@@ -100,11 +101,16 @@ __boot_text void *mips_init(void) {
     pte[PTE_INDEX(va)] = PTE_PFN(pa) | PTE_KERNEL;
 
 #if KASAN /* Prepare KASAN shadow mappings */
+  vaddr_t kasan_sanitized_end =
+    roundup(va, SUPERPAGESIZE << KASAN_SHADOW_SCALE_SHIFT);
+  size_t kasan_sanitized_size = kasan_sanitized_end - KASAN_MD_SANITIZED_START;
+  size_t kasan_shadow_size =
+    roundup(kasan_sanitized_size >> KASAN_SHADOW_SCALE_SHIFT, SUPERPAGESIZE);
```
j-piecuch, line 108:
```suggestion
  /* The loop below where we map the shadow pages depends on
   * kasan_shadow_size % SUPERPAGESIZE == 0 for correctness. */
  size_t kasan_sanitized_size = roundup2(
    va - KASAN_MD_SANITIZED_START, SUPERPAGESIZE * KASAN_SHADOW_SCALE_SIZE);
  size_t kasan_shadow_size = kasan_sanitized_size / KASAN_SHADOW_SCALE_SIZE;
```

j-piecuch (changes requested):

sys/kern/kasan.c
```diff
@@ -239,8 +239,8 @@ void kasan_grow(vaddr_t maxkvaddr) {
   }
 
   if (maxkvaddr > _kasan_sanitized_end) {
-    kasan_mark_valid((const void *)(_kasan_sanitized_end),
-                     maxkvaddr - _kasan_sanitized_end);
+    kasan_mark_invalid((const void *)(_kasan_sanitized_end),
+                       maxkvaddr - _kasan_sanitized_end, 0);
```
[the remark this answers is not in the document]
pj1031999: Which error code is the best from your point of view? - see `sys/kasan.h`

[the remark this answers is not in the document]
j-piecuch: `KASAN_CODE_KMEM_FREED` seems closest, but we could introduce our own code, e.g. `KASAN_CODE_NEW` or `KASAN_CODE_NOT_ALLOCED`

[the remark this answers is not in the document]
pj1031999: done

include/sys/kasan.h
```diff
@@ -19,6 +19,7 @@
 #define KASAN_CODE_POOL_FREED 0xFD
 #define KASAN_CODE_KMALLOC_OVERFLOW 0xFE
 #define KASAN_CODE_KMALLOC_FREED 0xFF
+#define KASAN_CODE_NEW 0xDF
```
j-piecuch, line 22: Invalid codes must have values in the range `0xf1-0xff`, see comment above `kasan_mark()`.

j-piecuch (changes requested):

include/sys/kasan.h
pj1031999: done

j-piecuch (approved):

cahirwpz (approved):

PR #1043, opened 2021-03-15
Construct commandline during exec

franciscozdo:
During exec we construct commandline which next can be got from process (e.g. by procstat).
Also updated `/dev/procstat` to use this feature.


sys/kern/exec.c
```diff
@@ -323,6 +323,24 @@ static bool check_setid(vnode_t *vn, uid_t *uid, gid_t *gid) {
   return (*uid != (uid_t)-1) || (*gid != (gid_t)-1);
 }
 
+static char *prepare_pargs(exec_args_t *args) {
+  char *pargs = kmalloc(M_STR, MAX_PARGS_LEN, M_ZERO);
+  size_t it = 1, used = 0, max = MAX_PARGS_LEN;
+
+  while (used < max && it < args->argc) {
+    strncpy(pargs + used, args->argv[it], max - used);
```
j-piecuch, line 331: Use `strlcpy()`.

j-piecuch (changes requested):

sys/kern/dev_procstat.c
```diff
@@ -15,7 +15,7 @@
  * During read calls to /dev/procstat this info can be read.
  *
  * Example:
- * euid   pid    ppid    pgrp   session  state    elfpath
+ * euid   pid    ppid    pgrp   session  state    name
```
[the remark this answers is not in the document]
franciscozdo: Yes, you're right.

sys/kern/dev_procstat.c
```diff
@@ -76,6 +79,24 @@ static vnodeops_t dev_procstat_vnodeops = {
...
 
+static char *get_name(proc_t *p) {
+  char *buf = kmalloc(M_TEMP, PROC_NAME_MAX, M_ZERO);
+
+  /* copy program's name from elfpath */
+  char *start = strrchr(p->p_elfpath, '/');
+  if (start == NULL)
+    start = p->p_elfpath;
+
+  strncpy(buf, start + 1, PROC_NAME_MAX);
```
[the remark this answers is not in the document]
franciscozdo: Now I do it in other way. I omit `/` only if it is found by `strrchr()`.

j-piecuch (changes requested):

sys/kern/dev_procstat.c
```diff
@@ -181,7 +183,7 @@ static int dev_procstat_read(file_t *f, uio_t *uio) {
 
   uio->uio_offset = f->f_offset;
   while (uio->uio_resid > 0) {
-    size_t len = min(uio->uio_resid, ps->psize - ps->offset);
+    ssize_t len = min(uio->uio_resid, ps->psize - ps->offset);
```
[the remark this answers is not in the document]
franciscozdo: I changed too much.

sys/kern/dev_procstat.c
```diff
@@ -79,20 +79,22 @@ static vnodeops_t dev_procstat_vnodeops = {
...
-  char *buf = kmalloc(M_TEMP, PROC_NAME_MAX, M_ZERO);
+static char *get_command(proc_t *p) {
+  char *buf = kmalloc(M_TEMP, PROC_COMM_MAX, M_ZERO);
 
-  /* copy program's name from elfpath */
-  char *start = strrchr(p->p_elfpath, '/');
+  /* copy program's name from elfpath (without starting slash) */
+  char *start = strrchr(p->p_elfpath, '/') + 1;
   if (start == NULL)
     start = p->p_elfpath;
```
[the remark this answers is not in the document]
franciscozdo: Thanks

sys/kern/exec.c
```diff
@@ -323,6 +323,25 @@ static bool check_setid(vnode_t *vn, uid_t *uid, gid_t *gid) {
...
+
+    /* calculate how much we have copied */
+    len = min(len, max - used);
+    used += len;
+
+    /* add space between args */
+    if (used < max)
+      pargs[used++] = ' ';
+    ++it;
+  }
```
j-piecuch, line 341: You need to be careful about `used` and `max`: you need to decide if `used` includes the terminating null byte or not.
Assuming it does not include it, the loop should look like this:
```suggestion
  while (used + 1 < max && it < args->argc) {
    size_t len = strlcpy(pargs + used, args->argv[it], max - used);

    /* calculate how much we have copied */
    len = min(len, max - used - 1);
    used += len;

    /* add space between args */
    if (used + 1 < max)
      pargs[used++] = ' ';
    ++it;
  }
```

j-piecuch (changes requested):

sys/kern/exec.c
franciscozdo: Thanks

include/sys/proc.h
```diff
@@ -119,6 +121,8 @@ struct proc {
   /* program segments */
   vm_segment_t *p_sbrk; /* ($) The entry where brk segment resides in. */
   vaddr_t p_sbrk_end;   /* ($) Current end of brk segment. */
+
+  char *p_args; /* (!) first 100 bytes of arguments of process */
```
j-piecuch, line 125: You're not copying `p_args` on fork.

j-piecuch (changes requested):

include/sys/proc.h
franciscozdo: Fixed that.

sys/kern/fork.c
```diff
@@ -82,6 +83,12 @@ int do_fork(void (*start)(void *), void *arg, pid_t *cldpidp) {
   child->p_cwd = parent->p_cwd;
   child->p_cmask = parent->p_cmask;
 
+  /* Copy process arguments */
+  child->p_args = kmalloc(M_STR, MAX_PARGS_LEN, M_ZERO);
+  if (parent->p_args)
+    /* we can't copy p_args from proc0 */
+    memcpy(child->p_args, parent->p_args, MAX_PARGS_LEN);
+
```
j-piecuch, line 91: Maybe `proc_create()` is a better place for this? That's where `p_elfpath` is copied.

j-piecuch (changes requested):

sys/kern/fork.c
franciscozdo: You are right. They are treated in exactly the same way so they should be created in the same places.

sys/kern/exec.c
```diff
@@ -405,6 +424,9 @@ static int _do_execve(exec_args_t *args) {
   kfree(M_STR, p->p_elfpath);
   p->p_elfpath = kstrndup(M_STR, prog, PATH_MAX);
 
+  kfree(M_STR, p->p_args);
+  p->p_args = prepare_pargs(args);
+
   klog("Enter userspace with: pc=%p, sp=%p", eh.e_entry, stack_top);
   return EJUSTRETURN;
```
franciscozdo, line 432: @j-piecuch I am not concerned about this change. I wanted to have it near creating arguments for process.

include/sys/proc.h
```diff
@@ -183,6 +187,8 @@ int proc_getsid(pid_t pid, sid_t *sidp);
  * The SID is returned in `*sidp`. */
 int proc_getsid(pid_t pid, sid_t *sidp);
 
+int proc_getargv(proc_t *p, char *buf, size_t maxlen);
+
```
j-piecuch, line 191: Is this needed?

j-piecuch (changes requested):

sys/kern/exec.c
j-piecuch: Fair enough.

include/sys/proc.h
franciscozdo: No. It is trash left from previous changes. Removed

j-piecuch (approved):

xThaid (approved):

sys/kern/exec.c
```diff
@@ -323,22 +323,27 @@ static bool check_setid(vnode_t *vn, uid_t *uid, gid_t *gid) {
...
+    size_t wanted = strlcpy(pargs + used, args->argv[i], left);
 
     /* calculate how much we have copied (without terminating null byte) */
-    len = min(len, max - used - 1);
-    used += len;
+    used += min(wanted, left - 1);
 
     /* add space between args */
-    if (used + 1 < max)
+    if (left > 1)
```
j-piecuch, line 340: This is wrong: `left` is not updated after the call to `strlcpy()`, so this test doesn't check whether there's space available after the copy.

j-piecuch (changes requested):

sys/kern/exec.c
franciscozdo: Fixed. Thanks!

sys/kern/dev_procstat.c
```diff
@@ -90,15 +91,17 @@ static char *get_command(proc_t *p) {
     /* omit slash */
     start += 1;
 
-  ssize_t len = strlcpy(buf, start, PROC_COMM_MAX);
-  len = min(len, PROC_COMM_MAX - 1);
+  ssize_t wanted = strlcpy(buf, start, left);
+  used += min(wanted, left);
```
j-piecuch, line 95: Assuming `used` doesn't include the null terminator, this line should look like this:
```suggestion
  used += min(wanted, left - 1);
```

j-piecuch (changes requested):

sys/kern/dev_procstat.c
franciscozdo: Yes. `used` does not include terminating null byte. In this case it doesn't matter because when `strlcpy` used whole buffer. I will include this changes, because it could be more readable.

j-piecuch (approved):

cahirwpz (approved):

PR #1047, opened 2021-03-18
KASAN: Implementation for AArch64

pj1031999:
KASAN implementation based on mips version.

Depends on #1040.


sys/aarch64/start.S
```diff
@@ -34,6 +34,11 @@ _ENTRY(_start)
 
         BL      aarch64_init
 
+        MOV     x19, x0
+        BL      aarch64_init_boot_stack
+        mov     sp, x0
+        mov     x0, x19
+
```
j-piecuch, line 41: Why do we need to switch to another stack?

j-piecuch (changes requested):

sys/aarch64/pmap.c
```diff
@@ -602,3 +603,31 @@ void pmap_delete(pmap_t *pmap) {
...
+  maxkvaddr = roundup(maxkvaddr, L2_SIZE);
+
+  WITH_MTX_LOCK (&pmap->mtx) {
+    for (va = vm_kernel_end; va < maxkvaddr; va += L2_SIZE)
+      pmap_ensure_pte(pmap, va);
+  }
+
+  /*
+   * kasan_grow calls pmap_kenter which acquires pmap->mtx.
+   * But we are under va_kernel_end_lock from kmem so it's safe to call
```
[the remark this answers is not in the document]
pj1031999: done

sys/aarch64/start.S
pj1031999: First stack is located in low addresses because MMU is disabled. But in `rpi3.c` MMU is enabled and KASAN is working (partially). Stack access is sanitized even if KASAN is not ready. It's hard to disable that because `rpi3.c` calls external functions and we really want to sanitize those functions. Low addresses can't be sanitized in current design so I move kernel into temporary stack in va.
We did it before for mips - #636.

sys/aarch64/boot.c
```diff
@@ -22,9 +23,14 @@ __boot_data void *_bootmem_end;
 /* Kernel page directory entries. */
 paddr_t _kernel_pmap_pde;
 alignas(PAGESIZE) uint8_t _atags[PAGESIZE];
+alignas(PAGESIZE) uint8_t _boot_stack[PAGESIZE];
 
 extern char exception_vectors[];
 extern char hypervisor_vectors[];
+extern size_t _kasan_sanitized_end;
+#if KASAN
+__boot_data static vaddr_t kasan_sanitized_end;
```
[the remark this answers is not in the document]
pj1031999: It is used in `enable_mmu`.

sys/aarch64/boot.c
```diff
@@ -22,9 +23,14 @@ __boot_data void *_bootmem_end;
 /* Kernel page directory entries. */
 paddr_t _kernel_pmap_pde;
 alignas(PAGESIZE) uint8_t _atags[PAGESIZE];
+alignas(PAGESIZE) uint8_t _boot_stack[PAGESIZE];
 
 extern char exception_vectors[];
 extern char hypervisor_vectors[];
+extern size_t _kasan_sanitized_end;
```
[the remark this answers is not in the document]
pj1031999: done

include/aarch64/kasan.h
```diff
@@ -0,0 +1,36 @@
...
+#define KASAN_MD_OFFSET                                                        \
+  (KASAN_MD_SHADOW_START -                                                     \
+   (KASAN_MD_SANITIZED_START >> KASAN_SHADOW_SCALE_SHIFT))
+
+__always_inline static inline int8_t *kasan_md_addr_to_shad(uintptr_t addr) {
+  return (int8_t *)(KASAN_MD_OFFSET + (addr >> KASAN_SHADOW_SCALE_SHIFT));
+}
+
+__always_inline static inline bool kasan_md_addr_supported(uintptr_t addr) {
+  return addr >= KASAN_MD_SANITIZED_START && addr < KASAN_MD_SANITIZED_END;
```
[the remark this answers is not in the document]
pj1031999: Yes and no.
We can't sanitize address greater than `_kasan_sanitized_end` but we should't make any access to that addresses. So in case of KASAN error we will get kernel panic.
For me it was useful during writing that code but if you want I can change that for AArch64 and for mips.

include/aarch64/kasan.h
```diff
@@ -0,0 +1,36 @@
...
+static_assert(KASAN_MD_SHADOW_SIZE % SUPERPAGESIZE == 0,
+              "Shadow memory is described by several full PDEs");
+#define KASAN_MD_SHADOW_END (KASAN_MD_SHADOW_START + KASAN_MD_SHADOW_SIZE)
+
+/* Sanitized memory (accesses within this range are checked) */
+#define KASAN_MD_SANITIZED_START KERNEL_SPACE_BEGIN
+#define KASAN_MD_SANITIZED_SIZE                                                \
+  (KASAN_MD_SHADOW_SIZE << KASAN_SHADOW_SCALE_SHIFT)
+#define KASAN_MD_SANITIZED_END                                                 \
+  (KASAN_MD_SANITIZED_START + KASAN_MD_SANITIZED_SIZE)
```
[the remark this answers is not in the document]
pj1031999: It's related to next comment.

include/aarch64/kasan.h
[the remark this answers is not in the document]
j-piecuch: I don't get it. If we use `KASAN_MD_SANITIZED_END` instead of `_kasan_sanitized_end`, then once `_kasan_sanitized_end` grows beyond `KASAN_MD_SANITIZED_END`, we won't check addresses between the two even though we should.

sys/aarch64/boot.c
```diff
@@ -199,6 +208,39 @@ __boot_text static paddr_t build_page_table(void) {
 
   l0[L0_INDEX(DMAP_BASE)] = (pde_t)l1d | L0_TABLE;
 
+#if KASAN /* Prepare KASAN shadow mappings */
+  kasan_sanitized_end =
+    SUPERPAGESIZE + roundup(va, SUPERPAGESIZE << KASAN_SHADOW_SCALE_SHIFT);
```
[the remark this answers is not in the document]
pj1031999: `init_vm_page` does some weird assumptions.  Without that additional pages it's try to access not mapped pages in shadow map. This is also a reason why I am using `KASAN_MD_SANITIZED_END` instead of `_kasan_sanitized_end`.

sys/aarch64/boot.c
```diff
@@ -199,6 +208,39 @@ __boot_text static paddr_t build_page_table(void) {
...
+  size_t kasan_sanitized_size = kasan_sanitized_end - KASAN_MD_SANITIZED_START;
+  size_t kasan_shadow_size =
+    roundup(kasan_sanitized_size >> KASAN_SHADOW_SCALE_SHIFT, SUPERPAGESIZE);
+  vaddr_t kasan_shadow_end = KASAN_MD_SHADOW_START + kasan_shadow_size;
```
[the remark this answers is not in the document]
pj1031999: I have changed it a bit. Please look if it's enough.

sys/aarch64/boot.c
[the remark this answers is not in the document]
j-piecuch: Hmm... `init_vm_page()` calls `vm_boot_alloc()`, which manually adjusts `vm_kernel_end` without calling `pmap_growkernel()` (which I guess it can't, since page allocation doesn't work yet), so I think it's possible for  `vm_kernel_end` to go past `_kasan_sanitized_end`. Maybe we're enabling KASAN a bit too early?

sys/aarch64/start.S
j-piecuch: Could we return the address of the stack from `aarch64_init()`, in the same way it's done for MIPS? Currently `aarch64_init()` returns `_atags`, but it's a global buffer, so you could declare it as `extern` in `rpi3.c` and use it that way. Then, you could return the new stack pointer from `aarch64_init()`.

sys/aarch64/boot.c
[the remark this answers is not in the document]
j-piecuch: Ok... in `enable_mmu()`, how can you access `kasan_sanitized_end` at the end? The MMU is enabled, and according to `rpi3.ld`, `.boot` sections are at low addresses, but I don't see you mapping those addresses in `build_page_table()`.

sys/aarch64/start.S
pj1031999: done

sys/aarch64/boot.c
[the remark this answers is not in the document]
j-piecuch: Solving this problem is outside the scope of this PR.

include/aarch64/kasan.h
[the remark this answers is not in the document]
pj1031999: KASAN shouldn't grow beyond `KASAN_MD_SANITIZED_END`.
Seriously, before you reach `KASAN_MD_SANITIZED_END` you should be out of physical memory.
And it's useful to sanitize access to memory with va higher than `_kasan_sanitized_end` because it's invalid memory. If you modify `boot.c` in wrong way and set `_kasan_sanitized_end` to incorrect value you still want to get an error "invalid reference to shadow map".
I pushed change with limit for `kasan_grow`. Is it enough for you?

include/aarch64/kasan.h
```diff
@@ -0,0 +1,36 @@
...
+
+/* Part of internal compiler interface */
+#define KASAN_SHADOW_SCALE_SHIFT 3
+
+/* Shadow memory */
+#define KASAN_MD_SHADOW_START 0xffffff0000000000
+#define KASAN_MD_SHADOW_SIZE (1ULL << 30) /* 1 GB */
+static_assert(KASAN_MD_SHADOW_SIZE % SUPERPAGESIZE == 0,
+              "Shadow memory is described by several full PDEs");
+#define KASAN_MD_SHADOW_END (KASAN_MD_SHADOW_START + KASAN_MD_SHADOW_SIZE)
```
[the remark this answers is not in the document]
pj1031999: Partially removed.

include/aarch64/kasan.h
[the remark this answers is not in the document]
j-piecuch: Ok, I see. I was misinterpreting the meaning of `KASAN_MD_SANITIZED_END`: it's the upper limit for the value of `_kasan_sanitized_end`. The name is terrible. Same for `KASAN_MD_SHADOW_SIZE`: It's the __maximum__ shadow size, not the actual shadow size.

sys/aarch64/boot.c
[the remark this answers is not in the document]
pj1031999: It's mapped in first loop - boot sections.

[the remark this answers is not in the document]
j-piecuch: Yes, but the VA that is mapped there starts with `0xffff`, while the address of the boot section in `rpi3.ld` is `0x200000`. How does that work?

sys/aarch64/boot.c
```diff
@@ -199,6 +208,39 @@ __boot_text static paddr_t build_page_table(void) {
...
+    SUPERPAGESIZE + roundup2(va - KASAN_MD_SANITIZED_START,
+                             SUPERPAGESIZE * KASAN_SHADOW_SCALE_SIZE);
+  size_t kasan_shadow_size = kasan_sanitzied_size / KASAN_SHADOW_SCALE_SIZE;
+  vaddr_t kasan_shadow_end = KASAN_MD_SHADOW_START + kasan_shadow_size;
+  va = KASAN_MD_SHADOW_START;
+  kasan_sanitized_end = KASAN_MD_SANITIZED_START + kasan_sanitzied_size;
+  /* Allocate physical memory for shadow area */
+  pa = (paddr_t)bootmem_alloc(kasan_shadow_size);
+
+  while (va < kasan_shadow_end) {
```
hadarai, line 222: Maybe you should add more comments to this loop and checks inside. At least what whole loop is doing.
Variable name `va` hardly suggests that it's about "start" of some kasan_shadow, you should rename it.

build/arch.aarch64.mk
```diff
@@ -5,7 +5,17 @@ ELFTYPE := elf64-littleaarch64
 ELFARCH := aarch64
 
 ifeq ($(KERNEL), 1)
+KASAN ?= 0
 CFLAGS += -mcpu=cortex-a53+nofp -march=armv8-a+nofp -mgeneral-regs-only
+ifeq ($(KASAN), 1)
+  # Added to files that are sanitized
+  CFLAGS_KASAN = -fsanitize=kernel-address \
+				 -fasan-shadow-offset=0xe0001f0000000000 \
+                 --param asan-globals=1 \
```
jpszczolowski, line 14: Inconsistent indentation of `-fasan-shadow-offset`

sys/aarch64/start.S
```diff
@@ -33,6 +33,8 @@ _ENTRY(_start)
         MOV     sp, x3
 
         BL      aarch64_init
+        /* Temporary stack in VA. */
+        mov     sp, x0
```
jpszczolowski, line 38: Could you please add a comment somewhere (maybe here) explaining why we need a boot stack in proper virtual addresses for KASAN? I should've left such comment myself and you wouldn't fall into the trap of the kernel exploding while trying to access stack variables.

sys/aarch64/boot.c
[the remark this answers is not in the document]
pj1031999: Ok, let's assume that `kasan_sanitized_end` is located at 0x201xxx - it's true for many builds.
It's also true that we create mapping for 0xffff000000201000 in that loop.

Let's do page table walk
kernel virtual address: `0b1111111111111111000000000000000000000000001000000001` (without last 12 bits)
`kasan_sanitized_end`: `0b1000000001` (without last 12 bits)
L3 step: `0b111111111`
L2 step: `0b111111111000000000`
L1 step: `0b111111111000000000000000000`
L0 step: `0b111111111000000000000000000000000000`

So
`L0_INDEX(va) = 0 = L0_INDEX(kasan_sanitized_end)`
`L1_INDEX(va) = 0 = L1_INDEX(kasan_sanitized_end)`
`L2_INDEX(va) = 1 = L2_INDEX(kasan_sanitized_end)`
`L3_INDEX(va) = 1 = L3_INDEX(kasan_sanitized_end)`

Is it clear for you?

include/aarch64/kasan.h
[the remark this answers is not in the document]
pj1031999: Maybe `KASAN_MD_MAX_SHADOW_SIZE`?
I'm open to your suggestions.

build/arch.aarch64.mk
pj1031999: I have fixed that but I think that something is still broken and git create wrong diff.
Please look at [sources](https://github.com/pj1031999/mimiker/blob/aarch64-kasan/build/arch.aarch64.mk) -- I have added extra spaces.

sys/aarch64/start.S
pj1031999: Done.
Please check if it's enough.

sys/aarch64/boot.c
pj1031999: No I shouldn't - it's standard walk through page table and function name is `build_page_table`. There is noting more to say about that.
`va` is standard name for virtual address and it __is__ a virtual address. It's common pattern to use `va` in boot and pmap code.

build/arch.aarch64.mk
jpszczolowski: I mean, it's not a big deal :smile: , but I think the problem is that other lines have spaces only, and -`fasan-shadow-offset` has both tabs and spaces
image

sys/aarch64/start.S
jpszczolowski: :+1:

sys/aarch64/boot.c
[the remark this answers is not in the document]
j-piecuch: I understand how page table walking works, and I understand that only the low 48 bits of the virtual address are used to index the page table (9 + 9 + 9 + 9 + 12 = 48). I was therefore curious why kernel addresses have the highest bits set to `0xffff` - these bits shouldn't matter for address translation. I had to read up on ARM documentation to find out that there are two separate page table base registers: `TTBR_EL0` for the range `0x0000_0000_0000_0000`-`0x0000_ffff_ffff_ffff`, and `TTBR_EL1` for the range `0xffff_0000_0000_0000` to `0xffff_ffff_ffff_ffff`. In `enable_mmu()` you set these two registers to the same value, and that's how you're able to access variables in `.boot`.

include/aarch64/kasan.h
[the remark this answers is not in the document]
j-piecuch: Yes, `KASAN_MD_MAX_SHADOW_SIZE` and `KASAN_MD_MAX_SANITIZED_END` are ok.

sys/aarch64/boot.c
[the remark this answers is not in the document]
pj1031999: Yes, you are right.
I misunderstand your question and I thought that you asked about where it is mapped.
Sorry for that.

Ok, so is it enough? Or do you still want to do something with `kasan_sanitized_end`?

include/aarch64/kasan.h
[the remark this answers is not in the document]
pj1031999: done

build/arch.aarch64.mk
pj1031999: Ok, I retabed that file, but other files in `build` are still broken.
We have `vim: tabstop=8 shiftwidth=8 noexpandtab:` in that files but we are using 2 spaces.
@cahirwpz  could you look at that?

sys/aarch64/boot.c
```diff
@@ -199,6 +208,39 @@ __boot_text static paddr_t build_page_table(void) {
 
   l0[L0_INDEX(DMAP_BASE)] = (pde_t)l1d | L0_TABLE;
 
+#if KASAN /* Prepare KASAN shadow mappings */
+  size_t kasan_sanitzied_size =
```
j-piecuch, line 212: `sanitzied -> sanitized`

sys/aarch64/boot.c
```diff
@@ -199,6 +208,39 @@ __boot_text static paddr_t build_page_table(void) {
...
+    SUPERPAGESIZE + roundup2(va - KASAN_MD_SANITIZED_START,
+                             SUPERPAGESIZE * KASAN_SHADOW_SCALE_SIZE);
+  size_t kasan_shadow_size = kasan_sanitzied_size / KASAN_SHADOW_SCALE_SIZE;
+  vaddr_t kasan_shadow_end = KASAN_MD_SHADOW_START + kasan_shadow_size;
+  va = KASAN_MD_SHADOW_START;
+  kasan_sanitized_end = KASAN_MD_SANITIZED_START + kasan_sanitzied_size;
```
j-piecuch, line 218: We could get rid of the `kasan_sanitized_end` intermediate global variable if we do this:
```suggestion
  /* XXX _kasan_sanitized_end is at a high address which is not mapped yet,
   * so we access it using its physical address instead. */
  *(vaddr_t *)AARCH64_PHYSADDR(&_kasan_sanitized_end) =
    KASAN_MD_SANITIZED_START + kasan_sanitized_size;
```

sys/aarch64/boot.c
```diff
@@ -199,6 +208,39 @@ __boot_text static paddr_t build_page_table(void) {
...
+    if (l1k[L1_INDEX(va)] == 0)
+      l1k[L1_INDEX(va)] = (pde_t)bootmem_alloc(PAGESIZE) | L1_TABLE;
+
+    pde_t *l2k = (pde_t *)PTE_FRAME_ADDR(l1k[L1_INDEX(va)]);
+    if (l2k[L2_INDEX(va)] == 0)
+      l2k[L2_INDEX(va)] = (pde_t)bootmem_alloc(PAGESIZE) | L2_TABLE;
+
+    pde_t *l3k = (pde_t *)PTE_FRAME_ADDR(l2k[L2_INDEX(va)]);
+
+    for (int j = 0; j < PT_ENTRIES; j++) {
```
j-piecuch, line 236:
```suggestion
    for (int j = 0; va < kasan_shadow_end && j < PT_ENTRIES; j++) {
```

j-piecuch (changes requested):

sys/aarch64/boot.c
pj1031999: fixed

sys/aarch64/boot.c
```diff
@@ -199,6 +207,42 @@ __boot_text static paddr_t build_page_table(void) {
...
+
+    pde_t *l3k = (pde_t *)PTE_FRAME_ADDR(l2k[L2_INDEX(va)]);
+
+    for (int j = 0; va < kasan_shadow_end && j < PT_ENTRIES; j++) {
+      l3k[L3_INDEX(va)] = pa | ATTR_AP_RW | ATTR_XN | pte_default;
+      va += PAGESIZE;
+      pa += PAGESIZE;
+    }
+  }
+#endif /* !KASAN */
```
j-piecuch, line 244: I believe the convention is for the comment to state the condition of the corresponding `#if`, not its negation.
```suggestion
#endif /* KASAN */
```

sys/aarch64/start.S
```diff
@@ -33,6 +33,15 @@ _ENTRY(_start)
         MOV     sp, x3
 
         BL      aarch64_init
+        /* 
+         * Temporary stack in VA.
+         * It's needed because in rpi3.c accesses to stack are instrumented
+         * and KASAN works only for virtuall addresses.
```
j-piecuch, line 39:
```suggestion
         * and KASAN works only for virtual addresses.
```

sys/aarch64/start.S
```diff
@@ -33,6 +33,15 @@ _ENTRY(_start)
...
+         * Stack instrumentation is done directly by GCC not by our code so
+         * we can't disable KASAN for that file in simply way because we call
```
j-piecuch, line 41:
```suggestion
         * we can't disable KASAN for that file in a simple way because we call
```

j-piecuch (changes requested):

sys/aarch64/boot.c
```diff
@@ -22,9 +23,13 @@ __boot_data void *_bootmem_end;
 /* Kernel page directory entries. */
 extern paddr_t _kernel_pmap_pde;
 alignas(PAGESIZE) uint8_t _atags[PAGESIZE];
+static alignas(PAGESIZE) uint8_t _boot_stack[PAGESIZE];
 
 extern char exception_vectors[];
 extern char hypervisor_vectors[];
+#if KASAN
+extern size_t _kasan_sanitized_end;
+#endif
```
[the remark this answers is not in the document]
pj1031999: done

pj1031999:
After merge with master something is broken with `_pcpu_data`.

sys/aarch64/boot.c
pj1031999: I'm not sure.  It looks like we don't stick to any convention in `.c` files but I agree with you and I will commit your suggestion.

```
kern/vm_physmem.c
71:#endif

kern/mutex.c
22:#endif
35:#endif
77:#endif

kern/malloc.c
19:#endif /* !KASAN */
102:#endif
369:#endif
415:#endif /* !KASAN */

kern/vmem.c
205:#endif

kern/main.c
90:#endif

kern/Makefile
89:endif

kern/lockdep.c
284:#endif

kern/pool.c
25:#endif
45:#endif
83:#endif /* !KASAN */
219:#endif /* !KASAN */
267:#endif

mips/boot.c
124:#endif /* !KASAN */
141:#endif /* !KASAN */

aarch64/boot.c
55:#endif

aarch64/evec.S
66:.endif
80:.endif
84:.endif
110:.endif

tests/utest.c
132:#endif
141:#endif
149:#endif
166:#endif

tests/vm_map.c
14:#endif
18:#endif

tests/sched.c
43:#endif
97:#endif

tests/pool.c
39:#endif

libkern/stdio/scanf.c
404:#endif
```

sys/kern/kasan.c
```diff
@@ -8,6 +8,7 @@
 #include <sys/vm.h>
 #include <machine/vm_param.h>
 #include <machine/kasan.h>
+#include <machine/pmap.h>
```
franciscozdo, line 11: Why do you need this header here?

franciscozdo (changes requested):
I have question about one header. Beside that lgtm.

sys/kern/kasan.c
pj1031999: deleted

include/mips/pmap.h
```diff
@@ -1,10 +1,6 @@
 #ifndef _MIPS_PMAP_H_
 #define _MIPS_PMAP_H_
 
-#ifndef _MACHDEP
-#error "Do not use this header file outside kernel machine dependent code!"
-#endif
-
```
franciscozdo, line 7: A I right that you can now restore this lines?

franciscozdo (changes requested):

include/mips/pmap.h
pj1031999: restored

j-piecuch (approved):

pj1031999:
> After merge with master something is broken with `_pcpu_data`.

It's related to kernel stack overflow and vmem code.

franciscozdo (approved):

cahirwpz (approved):

PR #1048, opened 2021-03-18
malloc: fix pointers arithmetic

pj1031999:
Fix pointers arithmetic.

Bug found by KASAN.

### DIFF
--- a/sys/kern/malloc.c
+++ b/sys/kern/malloc.c
@@ -96,10 +96,8 @@ static inline __always_inline void bt_make(word_t *bt, size_t size,
   word_t val = size | flags;
   word_t *ft = (void *)bt + size - sizeof(word_t);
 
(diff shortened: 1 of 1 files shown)


sys/kern/malloc.c
```diff
@@ -96,10 +96,8 @@ static inline __always_inline void bt_make(word_t *bt, size_t size,
   word_t val = size | flags;
   word_t *ft = (void *)bt + size - sizeof(word_t);
 
-#ifndef _LP64
-  kasan_mark_valid(bt - 1, 2 * sizeof(word_t));
+  kasan_mark_valid((word_t *)bt - 1, 2 * sizeof(word_t));
```
j-piecuch, line 99: Is the cast to `word_t *` necessary? `bt` already has that type.

PR #1052, opened 2021-03-20
Introduce uio_peek() and uio_advance()

j-piecuch:
This PR attempts to solve the same issue that #998 attempts to solve, but in a different way.
The issue is that we might read some data from an `uio` in order to do some processing on it, and then find out that we can't actually do the processing, so we need to return an error and make it appear as though no data was transferred from the `uio`.
With `uio_peek()`, we can read data without modifying the `uio` structure, so if we encounter an error later, we can just return the error without having to do anything. If we don't encounter an error, we need to "commit" the read we did earlier using `uio_advance()`.


j-piecuch:
Closing, as #1053 has been chosen as the best solution

PR #1053, opened 2021-03-20
Implement saving & restoring uio state

j-piecuch:
Yet another attempt to solve the issue of rolling back after an `uiomove()`. 
It introduces a single extra field to `uio_t` (compared to 0 for #1052 and 2 for #998).
The `uio` state must be saved using `uio_save()` before calling `uiomove()`.


franciscozdo (approved):

sys/kern/uio.c
```diff
@@ -81,6 +81,21 @@ int uiomove(void *buf, size_t n, uio_t *uio) {
...
+}
+
+void uio_restore(uio_t *uio, const uiosave_t *save) {
+  size_t nbytes = save->us_resid - uio->uio_resid;
+  uio->uio_resid += nbytes;
+  uio->uio_offset -= nbytes;
+  uio->uio_iov_off = save->us_iov_off;
+  uio->uio_iov -= save->us_iovcnt - uio->uio_iovcnt;
+  uio->uio_iovcnt = save->us_iovcnt;
+}
```
wiklam, line 97: What do you think about `inline` for these functions?

wiklam (approved):
LGTM

sys/kern/uio.c
j-piecuch: In order to be inlined, these functions would need to be defined in the header `uio.h`. I don't think it's worth it.

include/sys/uio.h
```diff
@@ -17,11 +17,20 @@ ssize_t writev(int fd, const struct iovec *iov, int iovcnt);
 
 typedef struct vm_map vm_map_t;
 
+/* This structure stores just enough information about an uio
+ * to be able to restore it after the uio is modified by uiomove(). */
+typedef struct {
+  size_t us_resid;   /* Saved value of uio_resid */
+  int us_iovcnt;     /* Saved value of uio_iovcnt */
+  size_t us_iov_off; /* Saved value of uio_iov_off */
+} uiosave_t;
```
cahirwpz, line 26: I guess `uiostate_t` is a better name for it cause you want to save & restore `uio` internal state.

Also remember that we define structure types as follows: `typedef struct typename { ... } typename_t`.

include/sys/uio.h
```diff
@@ -17,11 +17,20 @@ ssize_t writev(int fd, const struct iovec *iov, int iovcnt);
...
+
 typedef enum { UIO_READ, UIO_WRITE } uio_op_t;
 
 typedef struct uio {
   iovec_t *uio_iov;      /* scatter/gather list */
   int uio_iovcnt;        /* length of scatter/gather list */
+  size_t uio_iov_off;    /* offset in current iov segment */
```
cahirwpz, line 33: To follow local coding style I'd rename `uio_iov_off` to `uio_iovoff` (look at `uio_iovcnt`). Same for `us_iov_off`.

cahirwpz (commented):
I added some stylistic remarks. Generally speaking looks ok.

include/sys/uio.h
j-piecuch: I chose `uiosave_t` because `uio_t` _is_ the `uio` state. It's not a big deal though, so I have changed it.

include/sys/uio.h
j-piecuch: Done

cahirwpz (approved):

PR #1054, opened 2021-03-21
Extend the toolchain of the AMD64 release.

MichalBlk:
Introduce AMD64 to mimiker's toolchain.


xThaid (approved):

pj1031999 (approved):

MichalBlk:
Unfortunately debian packages mustn't contain the '_' character, which imposes some problems because "x86_64" is the required target-triplet prefix in case of AMD64. Therefore, a distinction between appropriate package fields and target name is needed. The presented solution introduces a new variable `ARCH` to solve this problem.

toolchain/gnu/config.mk
```diff
@@ -2,11 +2,14 @@
...
 VERSION = 1.5
 
+# Libraries required to build the toolchain.
 ISL = isl-0.18
 MPFR = mpfr-4.0.2
 GMP = gmp-6.1.2
 MPC = mpc-1.1.0
 CLOOG = cloog-0.18.4
+
+# Cross tools composing the toolchain.
```
cahirwpz, line 12: `The toolchain is comprised of following packages:`

cahirwpz (commented):
Looks ok.

PR #1055, opened 2021-03-22
Set window size on startup

j-piecuch:
The `setwinsize` program moves the cursor to the bottom-right corner of the screen, reads the position of the cursor, and sets the terminal window size based on that.
The program is run only on startup, so if the actual window size changes at runtime, it won't be detected, but the user can update the screen size by invoking the `setwinsize` command manually.


bin/setwinsize/setwinsize.c
```diff
@@ -0,0 +1,51 @@
...
+  if (tcgetattr(term_fd, &old))
+    return err;
+  t = old;
+  cfmakeraw(&t);
+  if (tcsetattr(term_fd, TCSAFLUSH, &t))
+    return err;
+  /* Save current cursor position. */
+  fputs(save_cursor, termfile);
+  /* Move the cursor to the bottom right corner of the screen. */
+  const char *buf = tiparm(cursor_address, 1000, 1000);
```
pj1031999, line 32: What does 1000 mean?

j-piecuch: They are the row and column numbers that we're moving the cursor to. I have added more explanatory comments.

bin/setwinsize/setwinsize.c
```diff
@@ -0,0 +1,62 @@
...
+
+int main(void) {
+  int err = 1;
+  struct termios old, t;
+  FILE *termfile;
+  /* Open controlling terminal. */
+  if ((termfile = fopen("/dev/tty", "r+")) == NULL)
+    return err;
+  int term_fd = fileno(termfile);
+  if (setupterm(NULL, term_fd, NULL))
```
hadarai, line 18: I think you should add a comment about that that check also

xThaid (approved):

bin/ksh/config.h
```diff
@@ -229,7 +229,7 @@
 #define POSIXLY_CORRECT 1
 
 /* Specify default $ENV? */
-#define DEFAULT_ENV	"$HOME/.kshrc"
+#define DEFAULT_ENV	"/etc/.kshrc"
```
cahirwpz, line 232: Please mark `DEFAULT_ENV` with comment that we should restore it to original state when user home directories are available for write.

cahirwpz (approved):

PR #1057, opened 2021-03-22
e.MMC interface

mohrcore:
An interface for an e.MMC controller, designed to support e.MMC and e.MMC-derived buses.

This interface allows for
* sending standard commands
* sending application-specific commands (yes, we need this)
* receiving various responses
* reading datastreams from the controller
* writing datastream to the controller
* reading some control values
* writing some control values

This interface does not support:
* sending general commands
* boot operation mode
* legacy multi-card systems
* probably a lot of other stuff


include/sys/emmc.h
```diff
@@ -0,0 +1,279 @@
+#ifndef _SYS_EMMC_H_
```
MichalBlk, line 1: Should be `_DEV_EMMC_H_`.

include/sys/emmc.h
```diff
@@ -0,0 +1,279 @@
...
+} emmc_op_mode_t;
+
+typedef enum emmc_result {
+  EMMC_OK,
+  EMMC_ERR_IERR,
+  EMMC_ERR_TIMEOUT,
+  EMMC_ERR_NOT_RESPONDING,
+  EMMC_ERR_INTR,
+  EMMC_ERR_BUSY,
+} emmc_result_t;
```
MichalBlk, line 24: Note that this scheme of defining enum constants (not consecutive powers of 2) makes it harder to use binary operations like `&` and `|`.

include/sys/emmc.h
```diff
@@ -0,0 +1,279 @@
...
+  EMMC_CMD_LOCK_UNLOCK = 42,
+  EMMC_CMD_APP_CMD = 55,
+  EMMC_CMD_GEN_CMD = 56,
+} emmc_command_t;
+
+#define EMMC_I_DATA_DONE 0x01
+#define EMMC_I_WRITE_READY 0x02
+#define EMMC_I_READ_READY 0x04
+
+typedef uint64_t emmc_wait_flags_t;
```
MichalBlk, line 106: Why not enum? And in case it can't be, why not `uint8_t`?

include/sys/emmc.h
```diff
@@ -0,0 +1,279 @@
...
+#define EMMC_APP_CHKIDX 0x08
+#define EMMC_APP_CHKCRC 0x10
+/* At most one of these */
+#define EMMC_APP_RESP136 0x20
+#define EMMC_APP_RESP48 0x40
+
+#define EMMC_APP_NEEDS_CMDTYPE 0x03
+#define EMMC_APP_NEEDS_RESPTYPE 0x60
+
+typedef uint32_t emmc_app_flags_t;
```
MichalBlk, line 123: See `mmc_wait_flags_t` comment.

include/sys/emmc.h
```diff
@@ -0,0 +1,279 @@
...
+  emmc_send_app_cmd_t send_app_cmd;
+  emmc_wait_t wait;
+  emmc_read_dat_t read;
+  emmc_write_dat_t write;
+  emmc_get_prop_t get_prop;
+  emmc_set_prop_t set_prop;
+} emmc_methods_t;
+
+typedef struct emmc_device {
+  const uint64_t cid[2];
```
MichalBlk, line 159: Maybe a macro defining (and describing) these array sizes?

MichalBlk (approved):
LGTM

include/sys/emmc.h
mohrcore: Ideally yes. We should move some of the headers from `sys` to `dev`. However there's a small detail that makes it a bit awkward - I think that stuff like `sys/emmc.c` and `sys/isa.c` should be in a separate catalogue as well, but we don't have a `dev` catalogue for `.c` files, only for headers.

I'm keeping it this way until this gets cleared up.

include/sys/emmc.h
mohrcore: Those are meant to be mutually exclusive. No binary operation should be performed on these.

include/sys/emmc.h
mohrcore: It was declared as `uint64_t` because I wanted to make it a property that could be accessed using `emmc_get_prop` method. This was a design that resulted from an idea of giving an interrupt resource to the child of e.MMC bus. But this idea is gone now.

I'll change it to an enum. I would've done so already, but looks like I missed it in the last update.

include/sys/emmc.h
mohrcore: It's an enum now.

include/sys/emmc.h
mohrcore: It would be quite unnecessary. These were just high and low bytes of an 128-bit value. Anyway, the whole response thing is different now, and hopefully the new macros defined for accessing response data make it clear what's going on here.

include/sys/emmc.h
mohrcore: Ok, it's fixed now.

include/sys/emmc.h
```diff
@@ -0,0 +1,303 @@
...
+ * \param var pointer to where the associated value should be written to
+ * \return -1 if value was fetched successfully, 0 if it wasn't
+ */
+static inline int emmc_set_prop(device_t *dev, emmc_prop_id_t id,
+                                emmc_prop_val_t val) {
+  device_t *idev = EMMC_METHOD_PROVIDER(dev, set_prop);
+  return emmc_methods(idev->parent)->set_prop(dev, id, val);
+}
+
+static inline emmc_device_t *emmc_device_of(device_t *device) {
```
hadarai, line 296: Why this method doesn't have comment?
Intentionally or by oversight?

mohrcore: Does it really need one? Look at the body. It's easy enough to tell what it does. Besides it's similar to `pci_device_of`. It's a helper function meant to be used within drivers code.

The other functions have long descriptions simply because they are just wrappers for the interface and whoever implements the interface should follow the specification given by these comments.

include/sys/emmc.h
```diff
@@ -0,0 +1,303 @@
...
+  EMMC_ERR_BUSY,
+} emmc_result_t;
+
+/* Response types R1-R5 are described in JESD86-A441, page 94.
+ * Response R1 and R1b are normally both represented as `EMMCRESP_R1`,
+ * because they are structurally identical. */
+typedef enum emmc_resp_type {
+  EMMCRESP_NONE,
+  EMMCRESP_R1,
+  EMMCRESP_R1B, /* Can be used only as an argument for `emmc_send_app_cmd` */
```
hadarai, line 35: Why does it have letter "B"? Is it supposed to note that it's "second R1"? 
If it's an R1 only for `emmc_send_app_cmd` (as it seems to be) then, I think, it should have it or part of it in name. So the name of response type tells me so, I don't have to read comment. It's easier then to not use that R1B for anything else than `emmc_send_app_cmd`

include/sys/emmc.h
```diff
@@ -0,0 +1,303 @@
...
+  EMMC_ERR_TIMEOUT,
+  EMMC_ERR_NOT_RESPONDING,
+  EMMC_ERR_INTR,
+  EMMC_ERR_BUSY,
+} emmc_result_t;
+
+/* Response types R1-R5 are described in JESD86-A441, page 94.
+ * Response R1 and R1b are normally both represented as `EMMCRESP_R1`,
+ * because they are structurally identical. */
+typedef enum emmc_resp_type {
```
hadarai, line 32: why is it `emmc_resp_type` and not `emmc_response_type`?
I like full names, but if it has to be short then let all be short/long/full
In line 42. you have full name.

include/sys/emmc.h
```diff
@@ -0,0 +1,303 @@
...
+
+#include <inttypes.h>
+#include <sys/mimiker.h>
+#include <sys/device.h>
+#include <sys/endian.h>
+#include <stddef.h>
+
+/* These operation modes are described in JESD86-A441, with a brief summary
+   at page 27 */
+typedef enum emmc_op_mode {
```
hadarai, line 12: Same comment like to line 32., about with wy is it "op" not "operation"

include/sys/emmc.h
```diff
@@ -0,0 +1,303 @@
...
+ * \return EMMC_OK on success
+ */
+static inline emmc_result_t emmc_write(device_t *dev, const void *buf,
+                                       size_t len, size_t *wrote) {
+  device_t *idev = EMMC_METHOD_PROVIDER(dev, write);
+  return emmc_methods(idev->parent)->write(dev, buf, len, wrote);
+}
+
+/**
+ * \brief Geta a value describing the e.MMC device
```
hadarai, line 271: "Geta" -> "Gets"

include/sys/emmc.h
```diff
@@ -0,0 +1,303 @@
...
+ * \return -1 if value was fetched successfully, 0 if it wasn't
+ */
+static inline int emmc_get_prop(device_t *dev, emmc_prop_id_t id,
+                                emmc_prop_val_t *val) {
+  device_t *idev = EMMC_METHOD_PROVIDER(dev, get_prop);
+  return emmc_methods(idev->parent)->get_prop(dev, id, val);
+}
+
+/**
+ * \brief Seta a value describing the e.MMC device
```
hadarai, line 284: "Seta" -> "Sets"

include/sys/emmc.h
```diff
@@ -0,0 +1,303 @@
...
+}
+
+/**
+ * \brief Geta a value describing the e.MMC device
+ * \param dev e.MMC controller device
+ * \param id value identifier
+ * \param var pointer to where the associated value should be written to
+ * \return -1 if value was fetched successfully, 0 if it wasn't
+ */
+static inline int emmc_get_prop(device_t *dev, emmc_prop_id_t id,
```
hadarai, line 277: Does "prop" stands for "property"? 
I guess you should put that word in comment above or change function name to feature whole word

include/sys/emmc.h
```diff
@@ -0,0 +1,303 @@
...
+}
+
+/**
+ * \brief Seta a value describing the e.MMC device
+ * \param dev e.MMC controller device
+ * \param id value identifier
+ * \param var pointer to where the associated value should be written to
+ * \return -1 if value was fetched successfully, 0 if it wasn't
+ */
+static inline int emmc_set_prop(device_t *dev, emmc_prop_id_t id,
```
hadarai, line 290: same with that "prop" and "property"

include/sys/emmc.h
hadarai: If every driver developer will know what it does it doesn't need a comment.
I don't know yet what `pci_device_of` is. But I guess if I had to develop something here I'd know?

hadarai (changes requested):

include/sys/emmc.h
mohrcore: These names (including "**R1b**") are literally how the specification calls these responses. It also states that **R1** and **R1b** are structurally the same and the difference comes to the way the controller received them, which is irrelevant for a developer, unless they want to instruct the controller on how to handle a command that is not in specs.

The comment should be changed this. It should not state an intended use, as there's only one or for `emmc_resp_type_t` as a function parameter in this interface, but instead it should say that commands which return either **R1** or **R1b** responses are all reported to drivers as returning **R1**. Perhaps this comment could be made above `emmc_resp_type` declaration.

include/sys/emmc.h
mohrcore: You are right, there's an inconsistency. I think I'm going to charge `emmc_response` to `emmc_resp`. Shorter names are preferred as long as given the context one can make out their meaning. Look at _FreeBSD_'s code. There's plenty of short names there.

include/sys/emmc.h
hadarai: Okay, understood.

include/sys/emmc.h
mohrcore: Same, as before. Shorter names, are preferred as long as they are reasonable.

include/sys/emmc.h
mohrcore: Will fix that.

include/sys/emmc.h
mohrcore: Yes, you would. It's a helper function and a comment would be longer than the body, which pretty much defeats its purpose. It's not like quake's inverse square root procedure, where there's some kind of an obscure magical trick that needs an in-depth explanation.

mohrcore:
Further changes are needed. The current interface is not sufficient for proper SD card support. Specificatiobns such as _SDA Physical Layer Specification_ can overload some of the default **eMMC** commands which is not taken into the account here. Fortunately this seems to be something that can be easily fixed by expanding the interface for sending application-specific commands to also support default command overloads.

include/sys/emmc.h
mohrcore: Fixed.

include/sys/emmc.h
mohrcore: Fixed.

include/sys/emmc.h
mohrcore: Yes. I;m keeping it short.

include/dev/emmc.h
```diff
@@ -0,0 +1,288 @@
...
+  device_t *idev = EMMC_METHOD_PROVIDER(dev, set_prop);
+  return emmc_methods(idev->parent)->set_prop(dev, id, val);
+}
+
+static inline emmc_device_t *emmc_device_of(device_t *device) {
+  return (emmc_device_t *)((device->bus == DEV_BUS_EMMC) ? device->instance
+                                                         : NULL);
+}
+
+/* Use if a command is expected to respond with R1b */
```
hadarai, line 276: Please check comment standards if there shouldn't be sth like `\brief`

include/dev/emmc.h
```diff
@@ -0,0 +1,288 @@
...
+/* Read a bitfield */
+#define EMMC_FMASK48(r, o, n) ((EMMC_RESPV48((r)) >> (o)) & ((1 << (n)) - 1))
+/* Write to a bitfield */
+#define EMMC_FMASK48_WR(r, o, n, v)                                            \
+  EMMC_RESPV48((r)) = (EMMC_RESPV48((r)) & ~(((1 << (n)) - 1) << (o))) |       \
+                      (((v) & ((1 << (n)) - 1)) << (o))
+
+/* Accessors for fields in R1-R5 response.
+ * End bit CRC7 and headers are ommited, because not every controller provides
+ * access tio them */
```
hadarai, line 42: typo: tio -> to

sys/drv/emmc.c
```diff
@@ -0,0 +1,60 @@
+#include <sys/devclass.h>
+#include <dev/emmc.h>
+
+/* An ugly fix, bu I don't know if we can do better with current build system */
```
hadarai, line 4: typo:
bu -> but

hadarai (changes requested):
Just a few small things.
 I love that you use comment standard with that `\brief` and `\param`!

include/dev/emmc.h
mohrcore: No. This is not required.

include/dev/emmc.h
mohrcore: Fixed.

sys/drv/emmc.c
mohrcore: Fized.

hadarai (approved):

cahirwpz (approved):

PR #1061, opened 2021-03-28
USB host controller interface.

MichalBlk:
Introduce a USB host controller interface.
The supplied approach provides the following services:
- obtain the number of the controlled root hub ports
- check whether any device is attached to the specified root hub port
- reset an attached device
- transfer to or from an attached USB device


include/dev/usbhc.h
```diff
@@ -0,0 +1,162 @@
...
+ * \param endp endpoint address within the device
+ * \param addr USB device address on the USB bus
+ * \param interval when to perform the transfer
+ * \param usbb USB buffer used for the transfer
+ */
+static void usbhc_read(device_t *dev, uint16_t size, uint16_t mps, uint8_t endp,
+                       uint8_t addr, uint8_t interval, usb_buf_t *usbb) {
+  device_t *idev = USBHC_METHOD_PROVIDER(dev, read);
+  usbhc_methods(idev->parent)->read(dev, size, mps, endp, addr, interval, usbb);
+}
```
mohrcore, line 99: What kind of transfer is this? _Bulk_? _Interrupt_? I guess the transfer type is defined by the `flags` field of the `usbb` parameter. Even so, it feels a bit inconsistent given how the transfer type is specified in the names of _control_ transfers procedures. I understand that this is because this is a more general function and _control_ transfers require different semantics than _block_ and _interrupt_ ones. The brief description however, is the pretty much the same as for _control_ transfers.

include/dev/usbhc.h
```diff
@@ -0,0 +1,162 @@
...
+ * \param interval when to perform the transfer
+ * \param usbb USB buffer used for the transfer
+ */
+static void usbhc_write(device_t *dev, uint16_t size, uint16_t mps,
+                        uint8_t endp, uint8_t addr, uint8_t interval,
+                        usb_buf_t *usbb) {
+  device_t *idev = USBHC_METHOD_PROVIDER(dev, read);
+  usbhc_methods(idev->parent)
+    ->write(dev, size, mps, endp, addr, interval, usbb);
+}
```
mohrcore, line 120: Same thing as with `usbhc_read`.

include/dev/usb.h
```diff
@@ -145,4 +149,24 @@ typedef struct usb_endpoint_descriptor {
...
+ * Implementation specific constructs.
+ */
+
+typedef enum {
+  TF_CONTROL = 1,   /* control transfer */
+  TF_BULK = 2,      /* bulk transfer */
+  TF_INTERRUPT = 4, /* interrupt transfer */
+  TF_STALLED = 8,   /* STALL condition encountered */
+  TF_ERROR = 16,    /* errors other than STALL */
+} usb_transfer_flags_t;
```
mohrcore, line 162: Aren't _control_, _bulk_ and _interrupt_ transfers mutually exclusive? Personally I would've made a comment for an enum which contains OR-able flags with a mutually exclusive subset but I'm leaving that decision up to you.

include/dev/usb.h
```diff
@@ -145,4 +149,24 @@ typedef struct usb_endpoint_descriptor {
...
+
+/* USB buffer used for USB transfers. */
+typedef struct usb_buf {
+  ringbuf_t buf;              /* write source or read destination */
+  condvar_t cv;               /* wait for the transfer to complete */
+  spin_t lock;                /* buffer guard */
+  usb_transfer_flags_t flags; /* transfer description */
+} usb_buf_t;
```
mohrcore, line 170: Do you need that `flags` field to be there? I have some mixed feelings about it because of the transfer type flags.

There are two procedures for performing transfers and as far as I understand the `usbhc_read` and `usbhc_write` are used for `TF_BULK` and `TF_INTERRUPT` transfers, while `usb_control_read` and `usb_control_write` are used for `TF_CONTROL` transfers, meaning that the flags need to match the used procedure. However, I've been wondering wether this is really necessary,. Perhaps you could get rid of `TF_CONTROL` and make it implicit when using `usbhc_control_read` and `usb_control_write`. `TF_BULK` and `TF_INTERRUPT` then could be thrown out of the `usb_transfer_flags_t` and used as an enum parameter for `usbhc_read` and `usbhc_write` procedures. This would remove the need to ensure setting the correct transfer type when constructing a usb transfer buffer.

mohrcore (commented):
I don't feel confident enough to give you an LGTM unless you address my concerns. I don't have a particularly strong stance against any piece of code here, so I'm choosing to leave it without "requesting changes", but I feel there might be some room for improvements.

include/dev/usbhc.h
MichalBlk: I made a distinction between interrupt and bulk transfers, therefore `usbhc_intr_{in,out}` and `usbhc_bulk_{in,out}` has been introduced.

include/dev/usbhc.h
MichalBlk: Done.

include/dev/usb.h
MichalBlk: The transfer flags are gone. I contracted it to the error flags and renamed to `usb_error_flags_t`.

include/dev/usb.h
MichalBlk: Transfer flags are gone. The only thing I need now is error flags.

include/dev/usb.h
```diff
@@ -145,4 +149,21 @@ typedef struct usb_endpoint_descriptor {
...
+  UEF_STALLED = 1, /* STALL condition encountered */
+  UEF_ERROR = 2,   /* errors other than STALL */
+} usb_error_flags_t;
+
+/* USB buffer used for USB transfers. */
+typedef struct usb_buf {
+  ringbuf_t buf;           /* write source or read destination */
+  condvar_t cv;            /* wait for the transfer to complete */
+  spin_t lock;             /* buffer guard */
+  usb_error_flags_t flags; /* errors encountered during transfer */
```
mohrcore, line 166: `eflags` or `errors` would be better imo, but this is fine as well , especially if there's a possiblity that this field may get expanded to include more information than wether particular errors were encountered.

mohrcore (approved):
LGTM.

MichalBlk:
These changes have been merged in #935 (and it still works).

include/dev/usb.h
MichalBlk: There are some other flags now and thus I have applied your suggestion.

cahirwpz (approved):

PR #1063, opened 2021-03-29
Build and run test for KGPROF flag - MIPS

wiklam:


pj1031999 (commented):
Could you add `gmon.out` to artifacts?

PR #1066, opened 2021-03-29
Make AArch64 a first-class citizen

pj1031999:
Finally we have almost full port of Mimiker for AArch64.

We have working all system calls (except `gettimeofday`), KASAN, TTY and a lot of other things.
We can boot on physical board with success.
All tests are passing in CI.

It's a good time to make AArch64 a first-class citizen.

This PR improve debug experience for clang build and add more tests to CI.

#1065 is required.


pj1031999:
Clang build is deprecated now.

PR #1068, opened 2021-04-01
[PoC] AArch64: use dtb instead of atags

pj1031999:
Use dtb instead of atags for AArch64 bootstrap.


cahirwpz:
I'm closing this PR since #1094 has been opened.

PR #1069, opened 2021-04-02
Clean vm_object interface

franciscozdo:
* Added `vm_object_hold` function to hide reference counter and renamed `vm_object_free` to `..._drop`.
* Renamed (and removed anonymous struct) entry on `vm_object`'s page list in `vm_page`.

### DIFF
--- a/sys/kern/vm_object.c
+++ b/sys/kern/vm_object.c
@@ -21,7 +21,7 @@ vm_page_t *vm_object_find_page(vm_object_t *obj, vm_offset_t offset) {
   SCOPED_MTX_LOCK(&obj->mtx);
 
   vm_page_t *pg;
-  TAILQ_FOREACH (pg, &obj->list, obj.list) {
+  TAILQ_FOREACH (pg, &obj->list, objpages) {
     if (pg->offset == offset)
       return pg;
   }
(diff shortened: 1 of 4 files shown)


sys/kern/vm_object.c
```diff
@@ -81,13 +81,19 @@ void vm_object_remove_pages(vm_object_t *obj, vm_offset_t off, size_t len) {
...
 
-void vm_object_free(vm_object_t *obj) {
-  if (!refcnt_release(&obj->ref_counter))
-    return;
+void vm_object_hold(vm_object_t *obj) {
+  refcnt_acquire(&obj->ref_counter);
+}
+
+void vm_object_drop(vm_object_t *obj) {
+  WITH_MTX_LOCK (&obj->mtx) {
```
j-piecuch, line 91: Is taking the lock necessary? Dropping the refcount is atomic, and if it drops to 0, we're the only thread with a reference to the object, so is synchronization necessary here?

franciscozdo: Yes, you're right. I misunderstood this.

pj1031999 (approved):

wiklam (approved):

PR #1074, opened 2021-04-04
build: fix KASAN build

### LINKED ISSUE #1070: KASAN flag is not honored
Setting KASAN flag in `config.mk` doesn't work.

`/sys/kern/Makefile` includes `/config.mk` indirectly by `/build/build.kern.mk` after `ifeq ($(KASAN), 1)` test. As a result `/sys/kern/kasan.c` and `/sys/kern/kasan_quar.c` are missing in `SOURCES`.

pj1031999:
Fix #1070 

`sys/kern/Makefile` do operations on `KASAN` variable before it is defined. As a result following ifeq is always false:

```make
ifeq ($(KASAN), 1)
SOURCES += kasan.c kasan_quar.c
# KASAN's run-time library mustn't be sanitized!
kasan.o : CFLAGS_KASAN =
else
# Clean dependencies without having to pass `KASAN=1` to `make clean`.
CLEAN-FILES += .kasan.D .kasan_quar.D
endif
```

The only solution for make `KASAN` build is passing KASAN=1 directly from cmdline.
We can avoid that by removing ifeq from Makefile and compiling kasan*.c always.

To do that we need to modify these .c files because they require to have
`KASAN=1`.

It's not the best solution but works with minimal effort.

### DIFF
--- a/sys/kern/Makefile
+++ b/sys/kern/Makefile
@@ -32,6 +32,8 @@ SOURCES = \
 	initrd.c \
 	interrupt.c \
 	isa.c \
+	kasan.c \
+	kasan_quar.c \
 	kenv.c \
 	klog.c \
 	kmem.c \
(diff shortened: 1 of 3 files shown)


franciscozdo (approved):

hadarai:
Do we have to have that variable `KASAN`? 
I mean, can't we just behave like value of that variable is 1 without setting and/or declaring it?

pj1031999:
> Do we have to have that variable `KASAN`?
> I mean, can't we just behave like value of that variable is 1 without setting and/or declaring it?

You can't always set `KASAN` to 1 because we have builds without sanitizer.

wiklam (approved):
I would add a description for the hack - why we use it.

cahirwpz:
We don't need this hack anymore since #1090 was introduced.

PR #1079, opened 2021-04-06
Remove MD bus definitions.

MichalBlk:
`include/$(ARCH)/bus_defs.h` only includes `include/sys/common_bus_defs.h` regardless of the value of `$(ARCH)`, therefore I've deleted these files and rename `include/sys/common_bus_defs.h` to `include/sys/bus_defs.h`. This solution is less ambiguous.

### DIFF
--- a/include/aarch64/bus_defs.h
+++ b/include/aarch64/bus_defs.h
@@ -1,1 +1,0 @@
-#include <sys/common_bus_defs.h>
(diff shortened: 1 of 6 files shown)


pj1031999 (approved):

mohrcore (approved):
Not much to say here, it's an obvious LGTM.

cahirwpz (approved):

PR #1080, opened 2021-04-06
build: yet another attempt to fix make format

pj1031999:
Adapt `make format` for multiple architectures.

### DIFF
--- a/sys/mips/Makefile
+++ b/sys/mips/Makefile
@@ -33,6 +33,8 @@ CPPFLAGS += -D_MACHDEP
 
 include $(TOPDIR)/build/build.kern.mk
 
+ifeq ($(ARCH), mips)
 $(SOURCES): assym.h
+endif
 
 boot.o : CFLAGS_KASAN =
--- a/sys/mips/Makefile
+++ b/sys/mips/Makefile
@@ -12,7 +12,7 @@ SOURCES = \
 	ebase.S \
 	gt64120.c \
 	interrupt.c \
-	$(BOARD).c \
+	malta.c \
 	pmap.c \
 	rootdev.c \
 	sigcode.S \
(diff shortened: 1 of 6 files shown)


pj1031999:
Note: CI has different clang version than build server and as a result `delay` function is formatted in wrong way.

franciscozdo (approved):

sys/mips/Makefile
```diff
@@ -12,7 +12,7 @@ SOURCES = \
 	ebase.S \
 	gt64120.c \
 	interrupt.c \
-	$(BOARD).c \
+	malta.c \
```
hadarai, line 15: It's really strange thing you do in this commit. You add new variable `$ARCH`, but on the same time you want to remove usage of one - `BOARD`, but one the same time you don't remove declaration of that `BOARD` variable. 
Shouldn't you add more variable with their usage's or remove them by hardcoding more things?
BTW I don't think that hardcoding more things as you do it in this line is good.

pj1031999: First of all `ARCH` already exists. `ARCH` is __current__ architecture. I have added `ARCHS` which is set of all supported architectures. It has been added because when you build mips (default) build system doesn't known about aarch64 subdirectory. As a result `make` doesn't call any recursion on that directory. Here I append every architecture for generic format target.
`BOARD` needs to be removed in that approach. Let's assume that you compile mips build. Then `BOARD=malta` and you call `sys/aarch64` with `BOARD=malta` and it's wrong because `malta.c` doesn't exist in `sys/aarch64`.

To be honest I would like to remove `BOARD`. We should build generic image for each architecture and initialize devices using information provided by dtb but it's not a target of that PR.

wiklam (approved):

cahirwpz:
Since handling source code formatting is handled by #1090 this PR is not needed anymore.

PR #1082, opened 2021-04-11
UVM: WIP

franciscozdo:
Here are all changes to make uvm working in mimiker stacked together.

PR #1081, opened 2021-04-11
Rename some structures and fields in VM subsystem

franciscozdo:
I think it is time to clean names in our VM subsystem. It will be also easier to compare our system with BSD because structures will have similar names.

Changes:
*  renamed fields of `vm_object`
* `vm_segment` -> `vm_map_entry`

PR #1083, opened 2021-04-11
MInor fixes and maintenance around VM.

MichalBlk:
Changes:
- delete unnecessary header inclusions.
- fix some typos.
- change the `vm_pgr_type_t` flags prefix form `VM_*` to `VM_PGR_*` (`VM_*` is already used by `vm_flags_t`).
- make `vm_segment_destroy` static as it's only called within the `vm_map` module and it's tricky to use due to the absence of any pmap calls inside the function.
- fix the `vm_map_insert_after` invocation in `vm_segment_destroy_range`.
- pass `0` instead of `VM_SEG_SHARED` in the call to ` vm_segment_alloc` inside ` vm_map_alloc_segment` (note that the following call to `vm_map_insert` will reset the segment's flags based on the `flags` argument anyway).
- rename `vm_map_object_dump` to `vm_object_dump` as it's part of the `vm_object` namespace.
- add counter in `vm_pgr_type_t` to control references to `pagers`.
- add and fix assertions.
- fill the `pgr_type` fields in `pagres`.
- use `vm_offset_t` instead of `off_t`.

### DIFF
--- a/include/sys/vm.h
+++ b/include/sys/vm.h
@@ -26,7 +26,7 @@ typedef enum {
   VM_PROT_NONE = 0,
   VM_PROT_READ = 1,  /* can read page */
   VM_PROT_WRITE = 2, /* can write page */
-  VM_PROT_EXEC = 4   /* can execute page */
+  VM_PROT_EXEC = 4,  /* can execute page */
 } vm_prot_t;
 
 #define VM_PROT_MASK (VM_PROT_READ | VM_PROT_WRITE | VM_PROT_EXEC)
--- a/include/sys/vm_map.h
+++ b/include/sys/vm_map.h
@@ -1,10 +1,8 @@
 #ifndef _SYS_VM_MAP_H_
 #define _SYS_VM_MAP_H_
 
-#include <sys/queue.h>
-#include <sys/pmap.h>
 #include <sys/vm.h>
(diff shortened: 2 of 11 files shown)


pj1031999:
I don't see any reason for that change.
You have fixed random typos, added random comas, renamed random constants/enums/functions. Maybe now it's more readable but it's not a real change.

I see only one minor change to something that was introduced in #831.

It's even worse from my perspective. We are already during rewriting of VM subsystem and your change makes it harder because now we need to merge that changes into something complicated. @franciscozdo has already started renaming of structures in VM subsystem in #1081 and it's a good place for that changes.

I believe that everything that you have introduced here will be removed in #1082.

@franciscozdo do you agree? Do you have something to add?

franciscozdo:
I think you can push changes that rename functions and pagers to my PR #1081 (but please follow convention used in that pr).

IMO bug fixes like this in `vm_segment_destroy_range` should be separated into small PR that is easy to review. And I am not sure if this function will exist after our changes to VM system.

Other changes are not crucial so I suggest you to wait with them until we finish implementing UVM.

MichalBlk:
Since the introduced changes are not crucial and they interfere with work done by @franciscozdo and @pj1031999, I've decided to close this PR.

PR #1062, opened 2021-03-29
Function instrumentation for kgprof - MIPS

wiklam:
Introducing instrumentation, which allows us to create a call graph of functions.
Functions like `intr_disable` or files `spinlock.c` are not instrumented cause they are used for instrumentation.
Function logic in mcount.c didn't change - just simplified.

PR #1084, opened 2021-04-12
Kgprof profclock

wiklam:
Introducing choosing a different clock for profiling that will not be synchronized with the main (hardclock) clock.
Depends on #1062


build/arch.mips.mk
```diff
@@ -24,4 +24,9 @@ endif
 # Added to all files
 GCC_ABIFLAGS += -msoft-float
 CLANG_ABIFLAGS += -msoft-float
+ifeq ($(KGPROF), 1)
+  CFLAGS_KGPROF = -finstrument-functions \
+                  -finstrument-functions-exclude-file-list=spinlock.c  \
+                  -finstrument-functions-exclude-function-list=intr_disable,intr_enable
```
pj1031999, line 30: Why you don't use `__no_instrument_function` for these functions?

wiklam: I left it for discussion cause:
1. `intr_disable` and  `intr_enable` are the only substrings, which cover more than one function definition
2. I am not sure if we want to use `__no_instrument_function` only for `copystr` otherwise, if we would add in the near future another program, which will use `finstrument-functions` there is a high probability we will move the definitions back to `-finstrument-functions-exclude-function-list=`.

sys/kern/kgprof.c
```diff
@@ -2,11 +2,12 @@
 #include <sys/context.h>
 #include <sys/gmon.h>
 #include <sys/kmem.h>
+#include <sys/timer.h>
 #include <sys/thread.h>
 #include <machine/vm_param.h>
 
-static gmonparam_t _gmonparam = {.state = GMON_PROF_NOT_INIT};
-static gmonhdr_t _gmonhdr = {.profrate = CLK_TCK};
+gmonparam_t _gmonparam = {.state = GMON_PROF_NOT_INIT};
```
franciscozdo, line 9: If you made it global please add `extern` definition in header.

sys/kern/clock.c
```diff
@@ -8,30 +8,47 @@
...
   if (tm_start(clock, TMF_PERIODIC | TMF_TIMESOURCE, (bintime_t){},
                HZ2BT(CLK_TCK)))
     panic("Failed to start system clock!");
   klog("System clock uses \'%s\' hardware timer.", clock->tm_name);
+
+  if (profclock != NULL) {
+    tm_init(profclock, prof_clock, NULL);
+    set_kgprof_profrate(PROF_TCK);
+    if (tm_start(profclock, TMF_PERIODIC, (bintime_t){}, HZ2BT(PROF_TCK))) {
+      klog("Failed to star profclock!");
```
franciscozdo, line 47:
```suggestion
      klog("Failed to start profclock!");
```

franciscozdo (approved):

sys/kern/kgprof.c
franciscozdo, line 9: If you made it global please add `extern` definition in header.

sys/kern/clock.c
franciscozdo, line 47:
```suggestion
      klog("Failed to start profclock!");
```

franciscozdo (changes requested):

sys/kern/kgprof.c
wiklam: Already done.

franciscozdo (approved):

pj1031999 (approved):
LGTM

cahirwpz:
Please resolve conflicts!

cahirwpz (changes requested):
I'd like kernel profile timer to be of higher frequency than system clock. Moreover the timer should be enabled only when processor executes kernel code.

cahirwpz:
@wiklam What's the status of this PR ?

wiklam:
> @wiklam What's the status of this PR ?

Please take a look

cahirwpz (approved):
LGTM

PR #1086, opened 2021-04-16
Rewrite munmap

franciscozdo:
Current procedure of unmaping is a bit complicated.

* Implement destroying range of vm_map_entry with splitting entry into parts.
* Simplify `do_munmap`
* Add test (and adjust one old) to check map entry splitting


sys/kern/mmap.c
```diff
@@ -63,22 +63,31 @@ int do_munmap(vaddr_t addr, size_t length) {
...
 
   WITH_VM_MAP_LOCK (uspace) {
-    while (addr < right_boundary) {
-      vm_map_entry_t *ent = vm_map_find_entry(uspace, addr);
-      if (!ent)
-        return EINVAL;
+    /* Find first entry affected by unmapping memory. */
+    vm_map_entry_t *ent = vm_map_find_entry(uspace, start);
+    if (!ent)
+      return 0;
```
wiklam, line 75: > Addresses in the range [addr,addr+len) are outside the valid range for the address space of a process.

sys/kern/mmap.c
```diff
@@ -63,22 +63,31 @@ int do_munmap(vaddr_t addr, size_t length) {
...
 
-      vaddr_t end = vm_map_entry_end(ent);
+    while (vm_map_entry_start(ent) <= start && vm_map_entry_end(ent) >= end) {
+      vaddr_t rm_start = max(start, vm_map_entry_start(ent));
+      vaddr_t rm_end = min(end, vm_map_entry_end(ent));
 
-      vm_map_entry_destroy_range(uspace, ent, addr, min(right_boundary, end));
+      /* Next entry that could be affected is right after current one.
+       * Since we can delate it entirely, we have to take next entry now. */
```
wiklam, line 82:
```suggestion
       * Since we can delete it entirely, we have to take the next entry now. */
```

sys/kern/vm_map.c
```diff
@@ -161,33 +165,49 @@ void vm_map_entry_destroy(vm_map_t *map, vm_map_entry_t *ent) {
   vm_map_entry_free(ent);
 }
 
+static inline vm_map_entry_t *vm_mapent_copy(vm_map_entry_t *src) {
```
wiklam, line 168:
```suggestion
static inline vm_map_entry_t *vm_map_entry_copy(vm_map_entry_t *src) {
```

sys/kern/vm_map.c
```diff
@@ -161,33 +165,49 @@ void vm_map_entry_destroy(vm_map_t *map, vm_map_entry_t *ent) {
...
+}
+
+/* Split vm_map_entry into two not empty entries.
+ * Returns entry which is after base entry. */
+static vm_map_entry_t *vm_map_entry_split(vm_map_t *map, vm_map_entry_t *ent,
+                                          vaddr_t splitat) {
+  assert(mtx_owned(&map->mtx));
+  assert(ent->start < splitat && splitat + 1 < ent->end);
+
+  vm_map_entry_t *new_ent = vm_mapent_copy(ent);
```
wiklam, line 182:
```suggestion
  vm_map_entry_t *new_ent = vm_map_entry_copy(ent);
```

wiklam (commented):
The rest seems good

sys/kern/mmap.c
franciscozdo: You are right. It should be `EINVAL`.

sys/kern/vm_map.c
```diff
@@ -161,33 +165,49 @@ void vm_map_entry_destroy(vm_map_t *map, vm_map_entry_t *ent) {
...
+                                           src->prot, src->flags);
+  return new;
+}
+
+/* Split vm_map_entry into two not empty entries.
+ * Returns entry which is after base entry. */
+static vm_map_entry_t *vm_map_entry_split(vm_map_t *map, vm_map_entry_t *ent,
+                                          vaddr_t splitat) {
+  assert(mtx_owned(&map->mtx));
+  assert(ent->start < splitat && splitat + 1 < ent->end);
```
wiklam, line 180: Why `splitat + 1 < ent->end` not `splitat < ent->end`?

franciscozdo: Hmm, it should be even stronger. We don't want to have empty `vm_map_entries` and smallest `vm_map_entry` describes at leas one page of memory. `splitat` should be page aligned and smaller than `ent->end`. I will fix that.

sys/kern/vm_map.c
```diff
@@ -80,6 +80,10 @@ vaddr_t vm_map_entry_end(vm_map_entry_t *ent) {
   return ent->end;
 }
 
+vm_map_entry_t *vm_map_entry_next(vm_map_entry_t *ent) {
+  return TAILQ_NEXT(ent, link);
+}
+
```
wiklam, line 86: I would add inline and replaced TAILQ_NEXT in line 233 and 325 vm_map.c with this function.

wiklam (changes requested):

wiklam (approved):

sys/kern/vm_map.c
```diff
@@ -161,33 +165,53 @@ void vm_map_entry_destroy(vm_map_t *map, vm_map_entry_t *ent) {
   vm_map_entry_free(ent);
 }
 
+static inline vm_map_entry_t *vm_map_entry_copy(vm_map_entry_t *src) {
+  if (src->object)
+    vm_object_hold(src->object);
```
pj1031999, line 170: This case is not tested. `src->object` is always a pointer to something else than null.

pj1031999 (commented):
Overall experience is good but I can't find a reason why tests failed.

#90112cf didn't change anything, control flow is the same so only kernel size was changed.

sys/kern/vm_map.c
franciscozdo: I will remove it and add when we will have this case in Mimiker.

franciscozdo: Yes, but I will leave it. We do exactly the same in `vm_map_entry_free`. I think it should be here, because if we will have that case we wouldn't have to look for code which assumes that `vm_object` is present.

pj1031999 (approved):

sys/kern/mmap.c
```diff
@@ -63,22 +63,31 @@ int do_munmap(vaddr_t addr, size_t length) {
...
-      vm_map_entry_t *ent = vm_map_find_entry(uspace, addr);
-      if (!ent)
-        return EINVAL;
+    /* Find first entry affected by unmapping memory. */
+    vm_map_entry_t *ent = vm_map_find_entry(uspace, start);
+    if (!ent)
+      return EINVAL;
 
-      vaddr_t end = vm_map_entry_end(ent);
+    while (vm_map_entry_start(ent) <= start && vm_map_entry_end(ent) >= end) {
```
xThaid, line 77: I don't understand this condition. What if the end of the first `vm_map_entry_t` is smaller than `end`? Then we won't delete anything and I think we should.

xThaid: As I understand it we are iterating as long as the intersection of the range of `vm_map_entry_t` and `[start, end]` is nonempty. Wouldn't something like `while (vm_map_entry_start(ent) < end)` suffice?

franciscozdo: Yes, you are right. I repaired it as you suggested.

xThaid (approved):

sys/kern/vm_map.c
```diff
@@ -161,33 +165,53 @@ void vm_map_entry_destroy(vm_map_t *map, vm_map_entry_t *ent) {
...
+ * Returns entry which is after base entry. */
+static vm_map_entry_t *vm_map_entry_split(vm_map_t *map, vm_map_entry_t *ent,
+                                          vaddr_t splitat) {
+  assert(mtx_owned(&map->mtx));
+  assert(page_aligned_p(splitat));
+  assert(ent->start < splitat && splitat < ent->end);
+
+  vm_map_entry_t *new_ent = vm_map_entry_copy(ent);
+
+  /* clip both entries */
```
cahirwpz, line 188: I wonder... shouldn't we have `vm_map_entry::offset` that helps to determine which part of vm object is actually visible through this map entry?

Let's say I have a map entry `A(start=va+0,size=16384,vmobj=VO)`. I split it into `B(start=va+0,size=4096,vmobj=VO)` and `C(start=va+8192,size=8192,vmobj=VO)`. How page fault routine will know for `va+8192` that it should look up a page in `VO` at offset 8192, instead of 0?

cahirwpz (commented):
Wrt coding style it looks fine.

However I suspect you missed a major bug, since you haven't tested the change throughly.

sys/kern/vm_map.c
franciscozdo: Yes, we need that field. I'va added it and test to check if it works.

cahirwpz (approved):

PR #1090, opened 2021-04-19
build: adjust build system for multiple config options

pj1031999:
Create separate SOURCES-$(OPTION) for each $(OPTION) listed in config.mk
* if $(OPTION) is set to 1 then SOURCES-$(OPTION) will be built and linked into
  kernel
* if $(OPTION) is set to 0 then SOURCES-$(OPTION) will be excluded from
  building
* values other than 0 and 1 are undefined

SOURCES_ALL contains all SOURCES-$(OPTION) even if $(OPTION) is set to 0.
SOURCES_ALL will be always formatted.


franciscozdo (approved):

wiklam (approved):

MichalBlk (approved):
LGTM

PR #1096, opened 2021-04-24
[evdev 3/3] Import evtest program

xThaid:
The last part of the #1089 PR. This one imports the utility tool used for evdev - evtest.

`versionsort.c` isn't in the libc, because we don't want code there other than imported from BSDs.

Blocked by: #1124 

To test the program, you need to compile `evtest` separately (my running `make` in evtest directory). Then launch Mimiker in graphical mode, run `evtest` and press some keys with the window focused.


pj1031999 (approved):
LGTM

PR #1104, opened 2021-05-02
Fix authorization during change of file timestamps

franciscozdo:
This check was too simple. According to manual semantics of utimesat is a bit different [1].
Times can be change to any value only by owner or root. Any other user can set times only to current time.

Also changed return type of `cred_can_utime` to match other functions used to set file attributes.

[1] [utimensat(2)](https://man.netbsd.org/utimensat.2)


include/sys/vnode.h
```diff
@@ -24,6 +24,8 @@ typedef struct cred cred_t;
  * vnodeops should not modify attributes set to VNOVAL. */
 #define VNOVAL (-1)
 
+#define VA_UTIMES_NULL 1
```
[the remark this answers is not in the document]
franciscozdo: Thanks. I've added a comment.

xThaid (approved):

include/sys/vnode.h
```diff
@@ -24,6 +24,10 @@ typedef struct cred cred_t;
  * vnodeops should not modify attributes set to VNOVAL. */
 #define VNOVAL (-1)
 
+/* Flag for vattr:va_flags.
+ * Utimes argument was null. */
+#define VA_UTIMES_NULL 1
```
pj1031999, line 29: I think that we prefer `enum` for flags.

pj1031999 (approved):
Beside of one minor comment this change looks good to me.

include/sys/vnode.h
franciscozdo: Changed.

cahirwpz:
**For the record:**

1. Linux [futimens](https://linux.die.net/man/3/futimens) says:
```
Permissions requirements
       To  set  both file timestamps to the current time (i.e., times is NULL,
       or both tv_nsec fields specify UTIME_NOW), either:

       1. the caller must have write access to the file;
       2. the caller's effective user ID must match the owner of the file; or
       3. the caller must have appropriate privileges.

       To make any change other than setting both timestamps  to  the  current
       time  (i.e.,  times is not NULL, and neither tv_nsec field is UTIME_NOW
       and neither tv_nsec field is UTIME_OMIT), either condition 2 or 3 above
       must apply.

       If both tv_nsec fields are specified as UTIME_OMIT, then no file owner‐
       ship or permission checks are performed, and the  file  timestamps  are
       not modified, but other error conditions may still be detected.
```

2. FreeBSD [futimens](https://www.freebsd.org/cgi/man.cgi?query=futimens&sektion=2) says:

```
     If both tv_nsec fields are UTIME_OMIT, the timestamps remain unchanged
     and no permissions are needed for the file itself, although search per‐
     missions may be required for the path prefix.  The call may or may not
     succeed if the named file does not exist.

     If both tv_nsec fields are UTIME_NOW, the caller must be the owner of the
     file, have permission to write the file, or be the super-user.

     For all other values of the timestamps, the caller must be the owner of
     the file or be the super-user.
```

3. POSIX standard [futimens](https://pubs.opengroup.org/onlinepubs/9699919799/functions/futimens.html) says:

```
Only a process with the effective user ID equal to the user ID of the file, or with write access to the file, or with appropriate privileges may use futimens() or utimensat() with a null pointer as the times argument or with both tv_nsec fields set to the special value UTIME_NOW. Only a process with the effective user ID equal to the user ID of the file or with appropriate privileges may use futimens() or utimensat() with a non-null times argument that does not have both tv_nsec fields set to UTIME_NOW and does not have both tv_nsec fields set to UTIME_OMIT. If both tv_nsec fields are set to UTIME_OMIT, no ownership or permissions check shall be performed for the file, but other error conditions may still be detected (including [EACCES] errors related to the path prefix).
```

cahirwpz:
Since you have not provided any tests for `futimens` I'm not sure which cases described by the manual pages are covered. For sure you don't handle `UTIME_NOW`. Please re-read manual pages I posted above.

sys/kern/tmpfs.c
```diff
@@ -997,10 +998,9 @@ static int tmpfs_resize(tmpfs_mount_t *tfm, tmpfs_node_t *v, size_t newsize) {
 }
 
 static int tmpfs_chtimes(tmpfs_node_t *v, timespec_t *atime, timespec_t *mtime,
-                         cred_t *cred) {
-  int err;
-  if ((err = cred_can_utime(v->tfn_vnode, v->tfn_uid, cred)))
-    return err;
+                         cred_t *cred, va_flags_t vaflags) {
+  if (!cred_can_utime(v->tfn_vnode, v->tfn_uid, cred, vaflags))
```
cahirwpz, line 1002: Is there any good reason for this check to be performed by a filesystem instead of code in one layer above, i.e `vfs_*`? I don't expect the code to differ at all between various filesystems...

franciscozdo: I think it was done to be consistent with previous operations (chmod, and chown). They are also handled on filesystem layer. We can move all these checks to layer above, but I think it sould be done in one separate PR.

cahirwpz:
Let's do several things in separate PRs:
 * handle `UTIME_*` values correctly
 * implement tests
 * evaluate if permission checks can be move up into VFS layer

cahirwpz (approved):

PR #1105, opened 2021-05-02
Move ioflags from VOP_READ/WRITE arguments into uio::uio_ioflags

cahirwpz:
* Clean up file flags: they express allowed operations and I/O operation flags (nonblocking, append, ...)

Changes extracted from #1103.


MichalBlk (approved):
LGTM

PR #1106, opened 2021-05-03
bin/dd: import dd from NetBSD

pj1031999:
* alt_io and nosigpipe were commented from ioflags in dd
* sccsid and __RCSID were commented
* raise_default_signal was added to libutil
* strsuftoll and strsuftollx were added to libc
* posix_memalign was exposed


franciscozdo (changes requested):
Please remove unnecessary code from all files.
Could you also provide some diff that shows your changes made to sources taken from NetBSD?

This change is also really big. I am wondering if we need so complicated `dd`. Is there any implementation of `dd` that is smaller and simpler then that? Maybe you should look in other *BSD systems and check if it will be compatible with our system.

pj1031999:
@mohrcore any update?

mohrcore:
@pj1031999 I'll make an update this week. Sorry for keeping you all waiting. I need to organize time in some decent way now that I have an internship, thesis, mimiker and a non-cs side-project to take care of at the same time. I admit that since my focus shifted towards `vcache`, when it comes to mimiker, `bin/dd` ended up being rather low on my priority list.

mohrcore:
There's also the question whether we should close this PR and merge a simpler implementation, like the one from OpenBSD instead. I agree with @franciscozdo, we probably don't need such complex implementation of `dd` anyway. @cahirwpz what do you think? Last time we spoke, you or @pj1031999 told me that the reason for using NetBSD's implementation is that taking the userspace from NetBSD is a sort of a convention in Mimiker. Is it still the case or should I not bother with that?

cahirwpz:
I've looked into suckless [dd](https://git.suckless.org/sbase/file/dd.c.html) implementation which is only 233 lines of code. I cannot understand why NetBSD's counterpart require 10 times more lines to do basically the same. I'm strongly leaning towards removing all NetBSD tools and replace them with suckless tools. Mimiker doesn't need to be so bloated.

PR #1111, opened 2021-05-08
Tags standard wiki page

hadarai:
It's a page with the proposition of new standard for tags, that are used to mark more important comments. The goal is to ease up finding specific comments
Information:
- new convection with description for new ones
- (proposition) paragraph explaining what is being changed


cahirwpz:
Please provide a description for the PR.

cahirwpz:
Please rewrite this document so that it tells a reader what tags are used for and when to use them.

PR #1116, opened 2021-05-10
Introduce ringbuf_contains.

MichalBlk:
This function is required by #1114.


MichalBlk:
FTTB, this change is no longer needed.

PR #1117, opened 2021-05-13
Add the alignment feature to the pool allocator.

MichalBlk:
We should be able to tell a pool what alignment should be ensured.

This PR contains changes presented in #1141.


include/sys/pool.h
```diff
@@ -19,8 +19,12 @@ typedef struct pool pool_t;
 /*! \brief Called during kernel initialization. */
 void init_pool(void);
 
-/*! \brief Creates a pool of objects of given size. */
-pool_t *pool_create(const char *desc, size_t size);
+/* Default pool item alignment. */
+#define P_DEF_ALIGN sizeof(uint64_t)
+
+/*! \brief Creates a pool of objects of given size
```
hadarai, line 25: You can remove that `\brief`,
Actually everywhere in that file

sys/kern/pool.c
```diff
@@ -242,16 +247,20 @@ static void pool_dtor(pool_t *pool) {
   klog("destroyed pool '%s' at %p", pool->pp_desc, pool);
 }
 
-static void pool_init(pool_t *pool, const char *desc, size_t size) {
+static void pool_init(pool_t *pool, const char *desc, size_t size,
```
hadarai, line 250: Consider adding a comment explaining each of the parameters.
I mean I'm not sure if it's needed, but there are 4 of them which is a quite

MichalBlk: This is a static function, hence is used only within the file which defines it.

include/sys/pool.h
MichalBlk: While this is a good idea, it's definitely isn't a good idea to do this in this PR because it would cloud the real changes.

hadarai (approved):

PR #1119, opened 2021-05-16
`SIGPIPE` tests

hadarai:
Introduction of `int pipe2(int pipefd[2], int flags)` tests. This PR tests delivery of `SIGPIPE` and return of `EPIPE`. Because these things happen simultaneously, I check them both at once. 
There are two test cases:
 - Parent writing to child who closed reading end, resulting in `SIGPIPE` delivery and return of `EPIPE` for parent.
 - Child writing to parent who closed reading end, resulting in `SIGPIPE` delivery and return of `EPIPE` for child.

On the same time it's a test of symmetry pipe implementation.

By the way there is a fix for a test fail cause and some additions to "Test infrastructure" readme.


bin/utest/pipe.c
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /* parent */
+  close(pipe_fd[0]); /* closing read end of pipe */
+
+  // Sync with end of child execution
+  wait_for_child_exit(child_pid, EXIT_SUCCESS);
+
+  // This is supposed to trigger SIGPIPE and return EPIPE
+  assert(write(pipe_fd[1], "hello world\n", 12) == EPIPE);
```
franciscozdo, line 50: Does `write` return -1 or `EPIPE`? Manual says that it will return -1 and set `errno` to appropriate value.

bin/utest/pipe.c
```diff
@@ -0,0 +1,91 @@
...
+
+void parent_sigpipe_handler(int signo) {
+  parent_signal_delivered = 1;
+  return;
+}
+
+void child_sigpipe_handler(int signo) {
+  child_signal_delivered = 1;
+  return;
+}
```
franciscozdo, line 27: You can use one function `sigpipe_handler` and reset value of variable `signal_delivered` when you want to use it.

bin/utest/pipe.c
```diff
@@ -0,0 +1,91 @@
...
+  if (child_pid == 0) { /* child */
+    close(pipe_fd[1]);  /* closing write end of pipe */
+    close(pipe_fd[0]);  /* closing read end of pipe */
+    exit(EXIT_SUCCESS);
+  }
+
+  /* parent */
+  close(pipe_fd[0]); /* closing read end of pipe */
+
+  // Sync with end of child execution
```
franciscozdo, line 46: Please be consistent with comments style. Use `/* ... */` style.

bin/utest/pipe.c
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /* parent */
+  close(pipe_fd[1]); /* closing write end of pipe */
+  close(pipe_fd[0]); /* closing read end of pipe */
+
+  /* send SIGUSR1 informing that parent closed both ends of pipe */
+  kill(child_pid, SIGUSR1);
+
+  wait_for_child_exit(child_pid, EXIT_SUCCESS);
```
franciscozdo, line 88: When I run this tests on linux it doesn't work here. `waitpid` waits for any change of child status (e.g. being signalled) so you have to do it in some different way (maybe put more synchronization between parent and child).

franciscozdo (changes requested):
Please run this tests on linux, because I can't make them run properly.

bin/utest/pipe.c
hadarai: You're right, I forgot to assure myself with manual

bin/utest/pipe.c
hadarai: Okay, changed to `waitpid`  `while` change of state is not to `exit`

bin/utest/pipe.c
```diff
@@ -0,0 +1,93 @@
...
+  close(pipe_fd[0]); /* closing read end of pipe */
+
+  /* send SIGUSR1 informing that parent closed both ends of pipe */
+  kill(child_pid, SIGUSR1);
+
+  /* I really want child to finish, not just change it's state.
+   * so i don't use wait_for_child_exit
+   */
+  int wstatus = 0;
+  while (!WIFEXITED(wstatus)) {
```
hadarai, line 87: I'd change it to `do-while` but I'm not sure if I should use that construction at all here

bin/utest/pipe.c
```diff
@@ -0,0 +1,93 @@
...
+    assert(waitpid(child_pid, &wstatus, 0) == child_pid);
+  }
+  assert(WEXITSTATUS(wstatus) == EXIT_SUCCESS);
```
franciscozdo, line 90: This code never runs. When `wstatus == 0` condition in while is false and `EXIT_SUCESS` is equal 0 so assert after loop never fails (even if child exited with wrong status).

franciscozdo: You should change initial value of `wstatus`.

franciscozdo (changes requested):

bin/utest/pipe.c
hadarai: I thought about it... but as we see I thought wrong 😅

hadarai: Maybe I should change code of `wait_for_child_exit` in `util.c`?

hadarai:
Tests showed that we were 
- not sending `SIGPIPE` to process which tries to write to pipe where no one reads.
- returning wrong error. When no one reads then "pipe's broken"(`EPIPE`) and not "Illegal seek" (`ESPIPE`) (why is it even called like that?)

bin/utest/pipe.c
```diff
@@ -0,0 +1,102 @@
...
+
+#include "utest.h"
+#include "util.h"
+
+static sig_atomic_t signal_delivered;
+
+void sigpipe_handler(int signo) {
+  signal_delivered = 1;
+  return;
+}
```
pj1031999, line 21: Why it is exposed?

bin/utest/pipe.c
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /* parent */
+  close(pipe_fd[0]); /* closing read end of pipe */
+
+  /* Sync with end of child execution */
+  wait_for_child_exit(child_pid, EXIT_SUCCESS);
+
+  /* This is supposed to trigger SIGPIPE and return EPIPE */
+  int write_ret = write(pipe_fd[1], "hello world\n", 12);
```
pj1031999, line 48: `write` returns `ssize_t` which is `intptr_t` which should have 8 bytes on AArch64.

bin/utest/Makefile
```diff
@@ -12,6 +12,7 @@ SOURCES = \
 	main.c \
 	misbehave.c \
 	mmap.c \
+	pipe.c \
 	pgrp.c \
```
pj1031999, line 16: 'g' < 'i'

sys/kern/file_syscalls.c
```diff
@@ -33,6 +33,13 @@ int do_write(proc_t *p, int fd, uio_t *uio) {
 
   uio->uio_ioflags |= f->f_flags & IO_MASK;
   error = f->f_ops->fo_write(f, uio);
+  if (error) {
+    if (error == EPIPE) {
+      mtx_lock(&p->p_lock);
+      sig_kill(p, &DEF_KSI_RAW(SIGPIPE));
+      mtx_unlock(&p->p_lock);
+    }
+  }
```
pj1031999, line 42: It's not a new test.
I suggest to do one change in one pr.

sys/kern/pipe.c
```diff
@@ -108,7 +108,7 @@ static int pipe_write(file_t *f, uio_t *uio) {
 
   /* Reading end is closed, no use in sending data there. */
   if (consumer->closed)
-    return ESPIPE;
+    return EPIPE;
```
pj1031999, line 111: Description of PR doesn't contain that change.

sys/tests/README.md
```diff
@@ -1,7 +1,6 @@
-Test infrastructure
----
+# Test infrastructure
```
pj1031999, line 1: Why tests for `SIGPIPE` require changes in README?

sys/kern/file_syscalls.c
cahirwpz: That's ok. The change is so small I decided not to burden @hadarai with creating a new PR.

sys/tests/README.md
cahirwpz: Right. Please move it to another PR. As you write new tests I guess you'll be in position to improve `tests/README.md` several times.

bin/utest/pipe.c
hadarai: I forgot to add that `static`

sys/kern/pipe.c
hadarai: Because it turned up later, so I decided to add a comment extending main description

sys/tests/README.md
hadarai: Done, #1143 was born

cahirwpz (changes requested):
Since `assert` is a macro, you should avoid passing expressions with side effects to it, since in theory the argument can be evaluated multiple times.

sys/kern/file_syscalls.c
```diff
@@ -33,6 +33,13 @@ int do_write(proc_t *p, int fd, uio_t *uio) {
 
   uio->uio_ioflags |= f->f_flags & IO_MASK;
   error = f->f_ops->fo_write(f, uio);
+  if (error) {
+    if (error == EPIPE) {
+      mtx_lock(&p->p_lock);
```
pj1031999, line 38: I prefer `proc_lock` or `WITH_PROC_LOCK`.

hadarai: thx 4 suggestion, It was really hard for me to find correct way of using mutexes in kernel

bin/utest/pipe.c
```diff
@@ -0,0 +1,106 @@
...
+
+#include "utest.h"
+#include "util.h"
+
+static sig_atomic_t signal_delivered;
+
+static void sigpipe_handler(int signo) {
+  signal_delivered = 1;
+  return;
+}
```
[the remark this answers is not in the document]
cahirwpz: `return` is redundant, you can remove it.

pj1031999 (approved):

franciscozdo (approved):

cahirwpz (approved):
LGTM

PR #1122, opened 2021-05-16
[kevent 3/5] kevent implementation

xThaid:
This PR is the main part of the [kevent](https://man.netbsd.org/NetBSD-6.0/kevent.2) subsystem. It implements the base functionalities of kqueues, knotes and kevents.

~~WIP: I need to add some comments.~~ Ready for a review.


sys/kern/event.c
```diff
@@ -1,15 +1,452 @@
...
+    kev = &changelist[i];
+
+    error = kqueue_get_obj(p, kev, &obj);
+
+    if (error == 0 && (error = kqueue_register(kq, kev, obj)) == 0)
+      continue;
+
+    /* No more space left in the the eventlist, so simply return the error. */
+    if (nevents == nerrors)
+      return error;
```
mohrcore, line 347: Shouldn't `retval` be set here, just like it is in line 356?

sys/kern/event.c
```diff
@@ -1,15 +1,452 @@
...
+                              .fo_close = kqueue_close,
+                              .fo_stat = kqueue_stat,
+                              .fo_seek = kqueue_seek,
+                              .fo_ioctl = kqueue_ioctl};
+
+static int kqueue_get_obj(proc_t *p, kevent_t *kev, void **obj) {
+  if (kev->filter == EVFILT_READ || kev->filter == EVFILT_WRITE)
+    return fdtab_get_file(p->p_fdtable, kev->ident, FF_READ, (file_t **)obj);
+
+  return EINVAL;
```
mohrcore, line 140: According to NetBSD and OpenBSD manuals, a bad file descriptor should cause `kevent` to return `EBADF`.

```suggestion
  return EBADF;
```

mohrcore (changes requested):

sys/kern/event.c
```diff
@@ -1,15 +1,452 @@
...
+  int kq_count;          /* number of pending events */
+  knote_tailq_t kq_head; /* list of pending events */
+  mtx_t kq_lock;         /* mutex for queue access */
+  condvar_t kq_cv;
+  knlist_t kq_knhash[KN_HASHSIZE]; /* hash table for knotes */
+} kqueue_t;
+
+/* Returns a hash bucket for a given object */
+static inline knlist_t *kq_gethash(kqueue_t *kq, void *obj) {
+  return &kq->kq_knhash[((unsigned long)obj) % KN_HASHSIZE];
```
pj1031999, line 45: `uintptr_t`

sys/kern/event.c
```diff
@@ -1,15 +1,452 @@
...
+    while ((kn = SLIST_FIRST(&kq->kq_knhash[i])) != NULL) {
+      knote_drop(kn);
+    }
+  }
+}
+
+static void kqueue_destroy(kqueue_t *kq) {
+  kqueue_drain(kq);
+  mtx_destroy(&kq->kq_lock);
+  cv_destroy(&kq->kq_cv);
```
pj1031999, line 74: NIT: I'm not sure if mtx should be destroyed before condvar that uses this lock.

sys/kern/event.c
```diff
@@ -1,15 +1,452 @@
...
+static int kqueue_ioctl(file_t *f, u_long cmd, void *data) {
+  return EOPNOTSUPP;
+}
+
+static fileops_t kqueueops = {.fo_read = kqueue_read,
+                              .fo_write = kqueue_write,
+                              .fo_close = kqueue_close,
+                              .fo_stat = kqueue_stat,
+                              .fo_seek = kqueue_seek,
+                              .fo_ioctl = kqueue_ioctl};
```
pj1031999, line 134: If you add `,` after last element then clang-format will be more happy to make it read-friendly ;)

sys/kern/event.c
```diff
@@ -1,15 +1,452 @@
...
+   * filter which have already been triggered.
+   */
+  kn->kn_kevent.udata = kev->udata;
+
+  event = kn->kn_filtops->filt_event(kn, 0);
+  WITH_MTX_LOCK (&kq->kq_lock) {
+    if (event && (kn->kn_status & KN_QUEUED) == 0)
+      knote_enqueue(kn);
+  }
+  mtx_unlock(kn->kn_objlock);
```
pj1031999, line 234: `WITH_MTX...`?

pj1031999 (commented):
tests?

sys/kern/event.c
```diff
@@ -1,15 +1,452 @@
...
+  }
+
+  /* To ensure the correctness of the iteration over pending events,
+   * we need to move already processed knotes to another list. `kq_head`
+   * can change in between iterations, because we are releasing the lock
+   * for a moment.
+   */
+
+  while (count < nevents) {
+    assert(mtx_owned(&kq->kq_lock));
```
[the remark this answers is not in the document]
xThaid: The locking strategy in the loop is non-trivial, so I thought that it is good to check important invariant at the beginning of the loop. 

So to answer the question, no, it is not necessary, but may be helpful in the future when changes will be made and can prevent some bugs.

sys/kern/event.c
xThaid: No, this was a deliberate decision. 

From the kevent(2) manual:
```
If an error occurs while
     processing	an element of the changelist and there is enough room in the
     eventlist,	then the event will be placed in the eventlist with EV_ERROR
     set in flags and the system error in data.	 Otherwise, -1 will be re-
     turned, and errno will be set to indicate the error condition.
```

sys/kern/event.c
xThaid: This error code is returned when a filter is invalid, not a file description. And such a case we should return `EINVAL`.

An excerpt from the manual:
`   [EINVAL]		The specified time limit or filter is invalid.`

xThaid:
We decided not to write tests for this feature.

pj1031999 (approved):

sys/kern/event.c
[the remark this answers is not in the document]
kboronski-ant: If you find it to be helpful then I think it's alright. Otherwise, I think there's no point in leaving it.
Imo leaving asserts like that is usually useful only when a lock is expected to be in a certain state when entering a procedure.

sys/kern/event.c
kboronski-ant: Ok, thanks for clarifying that.

sys/kern/event.c
```diff
@@ -1,15 +1,454 @@
...
+
+    /* To call `filt_event` we need to lock `kn_objlock`, but firstly
+     * we need to release `kq_lock` since holding it would violate the locking
+     * order. `kn_objlock` must be taken before `kq_lock`.
+     */
+    mtx_unlock(&kq->kq_lock);
+    mtx_lock(kn->kn_objlock);
+    event = kn->kn_filtops->filt_event(kn, 0);
+    mtx_lock(&kq->kq_lock);
+    mtx_unlock(kn->kn_objlock);
```
kboronski-ant, line 307: This is weird. Is there any particular reason why the lock/unlock don't follow a reverse order? I don't think there's anything wrong with this code, but seeing it written like that makes me think that maybe I'm missing something.

kboronski-ant (approved):
LGTM

mohrcore (approved):
Looks good for me, but not the CI, silly me approved from alt account :')

sys/kern/event.c
xThaid: I believe that there was not a particular reason and I agree that it may looks confusing.

Changed.

PR #1123, opened 2021-05-16
Implement sync and fsync syscalls as NOPs

mohrcore:
The rationale is that we don't have a buffering system yet, so doing nothing inside these syscalls satisfies all of their exit conditions for a successfull execution.

This allows programs such as `dd` (#1106) to operate without exiting early with error messages resulting from missing syscalls.


mohrcore:
The code builds fine in my local workspace, yet it fails to build in CI. Looks like the required code generation doesn't happen here.

xThaid:
Try running `make -C sys/kern sysent.h` in the main directory. It should rebuild syscalls table and so on.

cahirwpz (approved):

PR #1124, opened 2021-05-16
[kevent 4/5] Implement select syscall as a wrapper of kevent

xThaid:
As it turned out, [select](https://man.netbsd.org/NetBSD-6.0/select.2) system call can be implemented as a libc wrapper for kevent. Maybe it isn't the most efficient solution, but will do.


lib/libc/sys/select.c
```diff
@@ -0,0 +1,56 @@
...
+    int nevents = 0;
+
+    if (timeout != NULL)
+        tv2ts(timeout, &timeout_ts);
+
+    kq = kqueue();
+    if (kq < 0)
+        return -1;
+
+    events = malloc(2 * nfds * sizeof(struct kevent));
```
wiklam, line 23: Handle situation when malloc return NULL.
You have to check if nfds is a correct argument ([EINVAL](https://man7.org/linux/man-pages/man2/select.2.html))

lib/libc/sys/select.c
```diff
@@ -0,0 +1,56 @@
...
+
+int select(int nfds, fd_set *restrict readfds, fd_set *restrict writefds,
+           fd_set *restrict exceptfds, struct timeval *restrict timeout) {
+    int kq;
+    int ret;
+    struct timespec timeout_ts;
+    struct kevent *events;
+    int nevents = 0;
+
+    if (timeout != NULL)
```
wiklam, line 16: Check timeout too.

wiklam (changes requested):
What about incompatible kevent/kqueue errors?

lib/libc/sys/select.c
xThaid: Could you elaborate on that? What should I check? The specification of `select` syscall allows all values of `timeout`.

wiklam: I was thinking about returning `EINVAL` for negative and large values of timeout.

xThaid: I see. The specification however, doesn't say anything about big nor negative values.

cahirwpz: NetBSD's [select](https://man.netbsd.org/select.2) manual states:

> [EINVAL] The specified time limit is invalid.  One of its components is negative or too large.

xThaid: This is interesting. I have no idea where I found that information a few months ago.

cahirwpz (approved):

PR #1126, opened 2021-05-19
Port evdev to devnode interface

xThaid:


sys/drv/evdev.c
```diff
@@ -516,7 +516,7 @@ static int evdev_dev_create(evdev_dev_t *evdev) {
    * FIXME Not the best solution, but will do for now. */
```
hadarai, line 516: is it still valid comment?

sys/drv/evdev.c
```diff
@@ -469,10 +475,9 @@ static int evdev_open(vnode_t *v, int mode, file_t *fp) {
   WITH_MTX_LOCK (&evdev->ev_lock)
     LIST_INSERT_HEAD(&evdev->ev_clients, client, ec_link);
 
-  fp->f_ops = &evdev_fileops;
-  fp->f_data = client;
+  fp->f_data = dev;
 
-  return error;
+  return 0;
```
hadarai, line 480: why there isn't any kind of error returned?

sys/drv/evdev.c
```diff
@@ -440,26 +454,18 @@ static int evdev_ioctl(file_t *f, u_long cmd, void *data) {
...
-  .fo_ioctl = evdev_ioctl,
-};
-
-static int evdev_open(vnode_t *v, int mode, file_t *fp) {
-  evdev_dev_t *evdev = devfs_node_data(v);
-  int error = 0;
-
-  if ((error = vnode_open_generic(v, mode, fp)))
-    return error;
+static int evdev_open(devnode_t *master_dev, file_t *fp, int oflags) {
```
cahirwpz, line 457: You should reject attempts to open the device node when user passes `O_WRONLY` or `O_RDWR` to `open(2)`.

sys/drv/evdev.c
cahirwpz: In the current implementation actually nothing can fail, including memory allocation due to `M_WAITOK` flag.

sys/drv/evdev.c
cahirwpz: Yes, it is. Unfortunately.

cahirwpz (changes requested):

sys/drv/evdev.c
cahirwpz: Done!

cahirwpz (approved):

PR #1127, opened 2021-05-19
Port stdvga to devnode interface

xThaid:


sys/drv/stdvga.c
```diff
@@ -122,47 +122,38 @@ static int stdvga_set_fbinfo(stdvga_state_t *vga, fb_info_t *fb_info) {
...
-    goto fail;
-
-  if ((error = stdvga_set_fbinfo(vga, &vga->fb_info)))
-    goto fail;
+  if ((error = stdvga_set_fbinfo(vga, &vga->fb_info))) {
+    atomic_store(&vga->usecnt, 0);
+    return error;
+  }
 
   return 0;
```
hadarai, line 139: if you return `error` above, maybe you should do the same here?

cahirwpz: It can be easily proven that the function may only return 0 here. We prefer to exit early with an error and reach final return only when the function performed its action successfully.

sys/drv/stdvga.c
```diff
@@ -122,47 +122,38 @@ static int stdvga_set_fbinfo(stdvga_state_t *vga, fb_info_t *fb_info) {
...
 }
 
-static int stdvga_open(vnode_t *v, int mode, file_t *fp) {
-  stdvga_state_t *vga = devfs_node_data(v);
+static int stdvga_open(devnode_t *dev, file_t *fp, int oflags) {
```
cahirwpz, line 125: `stdvga_open` should fail if the user wants to open it for reading since `stdvga_read` is not provided.

sys/drv/stdvga.c
```diff
@@ -122,47 +122,38 @@ static int stdvga_set_fbinfo(stdvga_state_t *vga, fb_info_t *fb_info) {
...
   return 0;
-
-fail:
-  atomic_store(&vga->usecnt, 0);
-  return error;
 }
 
-static int stdvga_close(vnode_t *v, file_t *fp) {
-  stdvga_state_t *vga = devfs_node_data(v);
+static int stdvga_close(devnode_t *dev, file_t *fp) {
```
cahirwpz, line 142: Perhaps we should switch to the default resolution and clean the screen on close?

cahirwpz (changes requested):

sys/drv/stdvga.c
cahirwpz: stdvga switches to default resolution and clears the screen after the device is closed.

sys/drv/stdvga.c
cahirwpz: Done!

cahirwpz (approved):

PR #1128, opened 2021-05-19
Port dev cons & null & zero to devnode interface

xThaid:


sys/kern/devfs.c
```diff
@@ -149,9 +149,12 @@ static int devfs_fop_write(file_t *fp, uio_t *uio) {
 }
 
 static int devfs_fop_close(file_t *fp) {
+  int error;
   devnode_t *dev = fp->f_data;
   refcnt_release(&dev->refcnt);
-  return dev->ops->d_close(dev, fp);
+  error = dev->ops->d_close(dev, fp);
+  vnode_drop(fp->f_vnode);
+  return error;
```
hadarai, line 157: Why not return `dev->ops->d_close(dev, fp)` since it is a value of `error`?
Truth be said, now I know that the value of sophisticated expression with two arrows has a value of its error.

sys/drv/dev_cons.c
```diff
@@ -8,7 +8,7 @@
 
 #define UART_BUF_MAX 100
 
-static int dev_cons_write(vnode_t *t, uio_t *uio) {
+static int dev_cons_write(devnode_t *dev, uio_t *uio) {
```
hadarai, line 11: What does `cons` stand for?

cahirwpz: cons = console

sys/tests/vfs.c
```diff
@@ -54,35 +54,6 @@ static int test_vfs(void) {
   vnode_drop(dev_zero);
   assert(dev_zero->v_usecnt == 1);
 
-  uio_t uio;
```
cahirwpz, line 57: I'd expect you to rewrite following tests instead of just removing the code.

cahirwpz (changes requested):
Please do not remove tests, but improve them.

sys/kern/devfs.c
cahirwpz: We need to drop a reference before we return `error`.

sys/tests/vfs.c
cahirwpz: Ok. I verified that `/dev/zero` and `/dev/null` are throughly tested by `utest/fd.c`. OTOH testing `/dev/cons` has little use.

cahirwpz (approved):

PR #1132, opened 2021-05-23
Rewrite the keyboard test using evdev.

MichalBlk:
Adjust the keyboard test to use the evdev mechanism.

### DIFF
--- a/bin/test_kbd/test_kbd.c
+++ b/bin/test_kbd/test_kbd.c
@@ -12,6 +12,8 @@
 typedef struct input_event input_event_t;
 
 int main(int argc, char **argv) {
+  /* TODO: there should be a better way to select the required event file.
+   * FTTB, let's just hardcode the path. */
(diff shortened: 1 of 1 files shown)


bin/test_kbd/test_kbd.c
```diff
@@ -1,18 +1,40 @@
...
+      input_event_t *evp = &buf[i];
+
+      /* We only care about keycodes. */
+      if (evp->type != EV_KEY)
+        continue;
+
+      /* We shouldn't see any reserved keycodes. */
+      assert(evp->code != KEY_RESERVED);
+
+      const char *state = evp->value ? "pressed" : "released";
```
xThaid, line 33: There's another value of the event - 2. It means that the key is repeated

MichalBlk:
There's no point in introducing a simple evdev-based keyboard test as we'll have the `evtest` program (see #1096).

PR #1134, opened 2021-05-24
[usb] HID keyboard driver.

MichalBlk:
This PR contains changes introduced in #1137 and #1333.


sys/drv/hidkbd.c
```diff
@@ -0,0 +1,212 @@
...
+  for (size_t i = 0; i <= HIDKBD_NMODKEYS; i++) {
+    uint8_t prev = prev_modkeys & (1 << i);
+    uint8_t cur = modkeys & (1 << i);
+
+    if (prev == cur)
+      continue;
+
+    uint16_t keycode = evdev_mod2key(i);
+    int32_t value = cur ? 1 : 0;
+    evdev_push_event(hidkbd->evdev, EV_KEY, keycode, value);
```
xThaid, line 84: There's a wrapper for this function. It could be just `evdev_push_key(hidkbd->evdev, keycode, value);`. Now I have noticed that I did the same in the `atkbdc` driver, so it could be changed as well.

sys/drv/hidkbd.c
```diff
@@ -0,0 +1,212 @@
...
+
+  for (size_t i = 0; i <= HIDKBD_NMODKEYS; i++) {
+    uint8_t prev = prev_modkeys & (1 << i);
+    uint8_t cur = modkeys & (1 << i);
+
+    if (prev == cur)
+      continue;
+
+    uint16_t keycode = evdev_mod2key(i);
+    int32_t value = cur ? 1 : 0;
```
xThaid, line 83: Not very important, so feel free to ignore it: `cur != 0` instead of `cur ? 1 : 0`

sys/drv/hidkbd.c
```diff
@@ -0,0 +1,212 @@
...
+    value = 0;
+    goto prev_up;
+
+  both_up:
+    keycodes++;
+  prev_up:
+    prev_keycodes++;
+
+    if (keycode != KEY_RESERVED) {
+      evdev_push_event(hidkbd->evdev, EV_KEY, keycode, value);
```
xThaid, line 121: Here also `evdev_push_key` could be used.

sys/drv/hidkbd.c
```diff
@@ -0,0 +1,212 @@
...
+
+    uint16_t keycode = evdev_mod2key(i);
+    int32_t value = cur ? 1 : 0;
+    evdev_push_event(hidkbd->evdev, EV_KEY, keycode, value);
+    evdev_sync(hidkbd->evdev);
+  }
+}
+
+/* Process keycodes. */
+static void hidkbd_process_keycodes(hidkbd_state_t *hidkbd, uint8_t *keycodes) {
```
xThaid, line 90: I think a few sentences describing what this function does would be very useful to understand it. I find it non-trivial.

And do you assume anything about the order of keys in the report?

USB specification says:

> The order of keycodes in array fields has no significance. Order determination is done by the host software comparing the contents of the previous report to the current report. If two or more keys are reported in one report, their order is indeterminate

So if you get a report with keys A,B and then B,A, this code will send an event that A has just been released.

xThaid (changes requested):
I have reviewed only evdev-related code since I don't know many details about usb.

sys/drv/hidkbd.c
MichalBlk: Done!

sys/drv/hidkbd.c
MichalBlk: I changed the expression to use `evdev_key_events_t`.

sys/drv/hidkbd.c
MichalBlk: Done.

sys/drv/hidkbd.c
MichalBlk: Unfortunately, I made some assumptions regarding the ordering of keys in input reports. The logic has been changed.

MichalBlk:
This PR is once again ready for code review!

sys/drv/hidkbd.c
```diff
@@ -0,0 +1,208 @@
...
+
+  /* Obtain a USB buf. */
+  usb_buf_t *buf = usb_buf_alloc();
+
+  /* Register an interrupt input transfer. */
+  usb_data_transfer(dev, buf, &report, sizeof(hidkbd_in_report_t),
+                    USB_TFR_INTERRUPT, USB_DIR_INPUT);
+
+  for (;;) {
+    /* Wait for the data to arrvie. */
```
cahirwpz, line 145: Typo

sys/drv/hidkbd.c
```diff
@@ -0,0 +1,208 @@
...
+static void hidkbd_push_evdev_event(hidkbd_state_t *hidkbd,
+                                    uint8_t hidkbd_keycode, uint32_t value) {
+  uint16_t evdev_keycode = evdev_hid2key(hidkbd_keycode);
+  if (evdev_keycode == KEY_RESERVED)
+    return;
+  evdev_push_key(hidkbd->evdev, evdev_keycode, value);
+  evdev_sync(hidkbd->evdev);
+}
+
+static void hidkbd_process_modkeys(hidkbd_state_t *hidkbd, uint8_t modkeys) {
```
cahirwpz, line 80: Shouldn't you use output report to light up special keys like _caps lock_ and so on?

cahirwpz (commented):
Most of my comments for #1139 applies to this PR as well.

sys/drv/hidkbd.c
MichalBlk: Fixed.

cahirwpz (approved):
LGTM

PR #1136, opened 2021-05-27
Add filesystem source support for mount

mohrcore:
Up until now, mounting filestetms required only a filesystem type and a destination path. However, given the incoming block device  support, it will be also necessary to be able to specify the source of filesystem's underlying data in order to be able to mount filesystems that reside on external storage units.


pj1031999:
@mohrcore I'm not familiar with vfs subsystem. Please find somebody with more experience for review.

PR #1138, opened 2021-05-27
Add a user space mouse test.

MichalBlk:
Add a simple mouse test.


MichalBlk:
There's no point in introducing a simple evdev-based keyboard test as we'll have the `evtest` program (see #1096).

PR #1144, opened 2021-05-31
dtb: fix absolutely shitty behavior of bootloaders

pj1031999:
Bootloader can decide to load dtb into middle of random physical page and we need to deal with that inside kernel. For that purpose we can store offset of dtb for first phys page.

- store offset of dtb
- calculate mapping size for dtb including gap at the first page


MichalBlk (approved):
LGTM

cahirwpz (approved):

PR #1147, opened 2021-05-31
Add support for ext2 filesystem

mohrcore:
The goal is to allow ext2 filesystems (rev. 1) to be mounted from external mass media storage devices.
Depends on #1136.


pj1031999 (changes requested):
WTF?
```
#ifndef STUDENT
...
#endif /* STUDENT */
```

mohrcore:
@pj1031999 It's not ready yet and in case you are wondering - yes, I was granted permission to use that code and put it into this PR.

cahirwpz:
We're not planning to integrate this code with Mimiker. Hence I close this PR.

PR #1148, opened 2021-05-31
Pipe non-block tests

hadarai:
This PR tests usage of `pipe2` with flag `O_NONBLOCK`. Test cases:
1. Open pipe with `O_NONBLOCK` flag, test with `fnctl` if it's set, unset it, check if it's unset 
2. parent doesn't read, child fills `pipe`, tries again to `write` - sleep interrupted by `SIGALRM` (returns `EINTR` ) 
3. parent doesn't read, child has non-blocking writing end, fills `pipe`, tries again `write` - gets `EAGAIN`
4. parent closes it's writing end, child `read` - returns `0` 
5. parent doesn't write, keeps that end open, child `read`s - sleep interrupted by `SIGALRM`(returns `EINTR`)
6. parent doesn't write, child keeps non-blocking end open and does `read` - gets `EAGAIN`


hadarai:
Changes made to syscall handling:
- any `pipe` now revokes `do_pipe2`, eventually with empty flags.
- both flags (`O_NONBLOCK` and `O_CLOEXEC`) in `pipe2` are now supported, so when present are set for both pipe ends
- in non-blocking mode, read from empty and write to full return immediately with `errno` equal to `EAGAIN`
- in blocking mode, read from empty and write to full sleep with interruptible (by properly installed signal) sleep

hadarai:
In test 3 and 6 if `O_NONBLOCK` is set for only one `pipe` end, there is some bug with `fd` table. 
More details under `2c66946` CI check

cahirwpz:
Tests fail with:
```
| 5.802805 | 68 | kern/pipe.c:139      | KL_FILE    | Assertion "!pipe->writer_closed" failed!                                     |
| 5.802835 | 68 | kern/ktest.c:30      | KL_TEST    | Test "user_pipe_blocking_flag_manipulation" failed!  
```

bin/utest/pipe.c
```diff
@@ -15,15 +16,19 @@
...
-static void sigpipe_handler(int signo) {
-  assert(signo == SIGPIPE);
-  signal_delivered = 1;
+static void handler(int signo) {
+  if (signo == SIGPIPE) {
+    assert(signo == SIGPIPE);
+    signal_delivered = 1;
+  }
+  if (signo == SIGALRM) {
+  }
```
pj1031999, line 25: What with `SIGALRM`?

bin/utest/pipe.c
```diff
@@ -104,3 +109,226 @@ int test_pipe_child_signaled(void) {
...
+
+    struct sigaction sa = {
+      .sa_handler = handler,
+      .sa_flags = 0,
+    };
+
+    sigaction(SIGALRM, &sa, NULL);
+
+    int page_size = getpagesize();
+    char data[page_size];
```
pj1031999, line 166: VLA is a horrible idea.

bin/utest/pipe.c
```diff
@@ -104,3 +109,226 @@ int test_pipe_child_signaled(void) {
...
+
+    for (int i = 0; i < page_size; i++) {
+      data[i] = i + '0';
+    }
```
pj1031999, line 170: Potential overflow.

bin/utest/pipe.c
```diff
@@ -104,3 +109,226 @@ int test_pipe_child_signaled(void) {
...
+
+  if (child_pid == 0) {
+    close(pipe_fd[0]); /* closing read end of pipe */
+
+    int page_size = getpagesize();
+    /* prepare varying data */
+    char data[page_size];
+    for (int i = 0; i < page_size; i++) {
+      data[i] = i + '0';
+    }
```
pj1031999, line 211: ditto

sys/kern/syscalls.c
```diff
@@ -432,12 +432,10 @@ static int sys_pipe2(proc_t *p, pipe2_args_t *args, register_t *res) {
   int error;
 
   klog("pipe2(%x, %x)", u_fdp, flags);
-
-  if (flags & ~O_CLOEXEC) {
+  if (flags & ~O_CLOEXEC & ~O_NONBLOCK) {
```
pj1031999, line 435: Please add `( )` - I can guarantee you that most of us don't know binop priorities.

pj1031999 (changes requested):

bin/utest/pipe.c
hadarai: What does "VLA" mean?

pj1031999: variable-length array

hadarai: What do you suggest then? Set it to a constant value? Do you suggest any? Maybe Linux page size...?

bin/utest/pipe.c
hadarai: What do you mean?
I use it to simply wake up from interruptible sleep.

pj1031999: What is a purpose of that if statement?

bin/utest/pipe.c
pj1031999: I prefer to allocate array on heap (if you don't know size). In our case page size is always 4096 - see implementation of `getpagesize` - so you can make it static or allocate whole array on stack.

bin/utest/pipe.c
hadarai: OK, I get what you mean

bin/utest/pipe.c
cahirwpz: Yup. Please avoid VLA. IIRC It's not written down anywhere, but it's forbidden inside the kernel. We don't want to repeat mistakes Linux went through at some point.

bin/utest/pipe.c
```diff
@@ -104,3 +108,226 @@ int test_pipe_child_signaled(void) {
...
+int test_pipe_blocking_flag_manipulation(void) {
+  int pipe_fd[2];
+
+  /* creating pipe */
+  int pipe2_ret = pipe2(pipe_fd, O_NONBLOCK);
+  assert(pipe2_ret == 0);
+
+  /* check if flag is set */
+  assert(fcntl(pipe_fd[0], F_GETFL) & O_NONBLOCK);
+  assert(fcntl(pipe_fd[1], F_GETFL) & O_NONBLOCK);
```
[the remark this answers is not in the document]
hadarai: you're right. additional variables have been added

pj1031999 (changes requested):
Test "user_pipe_blocking_flag_manipulation" failed! - please fix

bin/utest/pipe.c
```diff
@@ -104,3 +107,233 @@ int test_pipe_child_signaled(void) {
...
+
+    struct sigaction sa = {
+      .sa_handler = sigpipe_handler,
+      .sa_flags = 0,
+    };
+
+    sigaction(SIGALRM, &sa, NULL);
+
+    int page_size = 4096; // based on implementation of getpagesize()
+    char data[page_size];
```
pj1031999, line 171: You have inconsistency in your code. Here you assume that page size is 4096 (it's true for us), but latter you call `getpagesize`. Please do that in one way.

Also I'm not a fan of VAL and I suggest to avoid them.

hadarai: I also don't like VAL, but return value of `getpagesize()` may change at some point and that test should be ready for that.

bin/utest/pipe.c
```diff
@@ -104,3 +107,234 @@ int test_pipe_child_signaled(void) {
...
+
+    struct sigaction sa = {
+      .sa_handler = sigpipe_handler,
+      .sa_flags = 0,
+    };
+
+    sigaction(SIGALRM, &sa, NULL);
+
+    int page_size = getpagesize();
+    char *data = malloc(page_size * sizeof(char));
```
pj1031999, line 171: `free`

bin/utest/pipe.c
```diff
@@ -104,3 +107,234 @@ int test_pipe_child_signaled(void) {
...
+  /* forking */
+  child_pid = fork();
+  assert(child_pid >= 0);
+
+  if (child_pid == 0) {
+    close(pipe_fd[0]); /* closing read end of pipe */
+
+    int page_size = getpagesize();
+    /* prepare varying data */
+    char *data = malloc(page_size * sizeof(char));
```
pj1031999, line 213: ditto

pj1031999 (approved):

cahirwpz (approved):

PR #1153, opened 2021-06-11
vcache

mohrcore:
We need basic vcache functionality to support any external filesystem.

### DIFF
--- a/include/sys/vnode.h
+++ b/include/sys/vnode.h
@@ -85,8 +85,11 @@ typedef struct {
 } vnlock_t;
 
 typedef struct vnode {
-  vnodetype_t v_type;        /* Vnode type, see above */
-  TAILQ_ENTRY(vnode) v_list; /* Entry on the mount vnodes list */
+  vnodetype_t v_type;          /* Vnode type, see above */
+  TAILQ_ENTRY(vnode) v_list;   /* Entry on the mount vnodes list */
+  TAILQ_ENTRY(vnode) v_free;   /* Entry on vnode free list (vcache) */
+  TAILQ_ENTRY(vnode) v_cached; /* Entry on vnode list in vcache bucket */
+  ino_t ino;                   /* inode number (used by vcache) */
 
   vnodeops_t *v_ops; /* Vnode operations */
   void *v_data;      /* Filesystem-specific arbitrary data */
--- a/sys/kern/vfs_vcache.c
+++ b/sys/kern/vfs_vcache.c
@@ -1,0 +1,124 @@
+#define KL_LOG KL_VFS
+
+#include <sys/mimiker.h>
+#include <sys/klog.h>
+#include <sys/vfs.h>
+#include <sys/vnode.h>
+#include <sys/mount.h>
+#include <sys/pool.h>
+#include <sys/lock.h>
+
+static POOL_DEFINE(P_VCACHE, "vcache", sizeof(vnode_t));
+
+#define VCACHE_BUCKET_CNT 64
+#define VCACHE_INITIAL_FREE_CNT 128
+
+typedef TAILQ_HEAD(vnode_list, vnode) vnode_list_t;
+
+static vnode_list_t vcache_free;
+static vnode_list_t vcache_buckets[VCACHE_BUCKET_CNT];
+static lock_t vcache_giant_lock;
+
+typedef size_t vcache_t; 
+
+static vcache_t vcache_hash(mount_t *mp, ino_t ino) {
+  return ((vcache_t)mp->mnt_vnodecovered >> 3 + ino) % VCACHE_BUCKET_CNT;
+}
+
+/* XXX: this is taken from vfs_vnode.c */
+static void vnlock_init(vnlock_t *vl) {
+  spin_init(&vl->vl_interlock, 0);
+  cv_init(&vl->vl_cv, "vnode sleep cv");
+}
+
+static vnode_t *vcache_new_vnode(void) {
+  vnode_t *vn = pool_alloc(P_VCACHE, M_ZERO);
+  vn->v_type = V_NONE;
+  vnlock_init(&vn->v_lock);
+  TAILQ_INSERT_TAIL(&vcache_free, vn, v_free);
+  return vn;
+}
+
+void vfs_vcache_init(void) {
+  for (size_t i = 0; i < VCACHE_INITIAL_FREE_CNT; i++)
+   (void)vcache_new_vnode();
+
+  mtx_init(&vcache_giant_lock, 0);
+}
+
+/**
+ * \brief Get a free vnode from a hashlist.
+ * \param mp mountpoint of the inode's filesystem
+ * \param ino inode to be used for vnode lookup
+ */
+vnode_t *vfs_vcache_hashget(mount_t *mp, ino_t ino) {
+  vcache_t bucket = vcache_hash(mp, ino);
+
+  SCOPED_MTX_LOCK(&vcache_giant_lock);
+
+  vnode_t *vn = NULL;
+  TAILQ_FOREACH(vn, &vcache_buckets[bucket], v_cached) {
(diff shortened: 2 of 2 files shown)


pj1031999 (changes requested):
Please resolve conflicts.

mohrcore:
No idea on why the `vmem` test fails. It strikes me as very bizarre. Is the pool allocator broken, or am I just doing something wrong?

pj1031999:
@mohrcore value of `VF_CACHED` in your code is 0. The only place where you are using this flag calls pool allocator (and is always called).  When you change `VF_CACHED` to 1 then your code will be working better.

mohrcore:
@pj1031999 thank you for you comment. That was certainly causing some issues. I've fixed that now.

mohrcore:
Allow me to explain my reasoning behind adding a new _VOP_:

A vnode use count as I get it represents the number of entities that need direct access to data held by a vnode. A use count greater than zero, therefore implies that the vnode is not recycleable at the moment (ie. can't be on vcache freelist). When the use count drops to zero, the vnode should be moved to a vcache freelist and thus become recycleable. This will be done through a `vfs_vcache_return` procedure called when a vnode with `VF_CACHED` flag set gets its use counter dropped to zero. Right before that happens it's be crucial for a filesytem code to take action if a vnode is no longer valid, ie. it's corresponding inode is to be removed from a filesystem (eg. user has deleted a file). The filesystem code should _detach_ any of its data from a vnode to prevent vcache from returning an outdated vnode on request that uses a new inode with the same number as the old one. This action can be done using `vfs_vcache_detach` procedure, which will mark the node _"detached"_ and will prevent it from being put into any bucket within vcache in `vfs_vcache_return`. Such action could've been performed inside `VOP_RECLAIM` that used to be called when vnode usecount dropped to zero.

The problem is that it was the last point at which a filesystem code could do anything in regards to a vnode, before it gets recycled. That meant that any inodes that were allocated by a filesystem needed to be freed in that moment, which in regards to filesystems present on block devices makes the vcache impractical. (I guess alternatively they could be cached internally by a filesystem) A filesystem would need to flush an inode to a block device every time the use count drops to zero and load it back the moment the vnode goes back into use. Thus `VOP_RECLAIM` is now called only when the vcache system *needs* to recycle the vnode for use with another filesystem, or when the filesystem chooses to detach its data from a vnode. A vnode is expected to have its usecnt equal to 0 at this point. In `VOP_RECLAIM` a filesystem is required to ensure that all the resources it has associated to the vnode will be in correct state once the vnode is gone, so they could be attached later to another vnode if needed. For _tmpfs_ it just means a little cleanup work (removing an inode if a file is expected to be removed). For something like _ext2_ it could mean flushing the inode to its block device (and given that block buffering will most likely require a vnode in order to identify a block, this could also mean flushing and de-buffering the entire data associated with that vnode).

Because it's still necessary to take action the moment vnode use count drops to zero, as described eariler (detaching filesystem data from invalid vnode), the old `VOP_RECLAIM`became `VOP_INACTIVE`.

According to NetBSD manual:

_"**VOP_INACTIVE()** is called when the kernel is no longer using the vnode."_
_"**VOP_RECLAIM()** is called when a vnode is being reused for a different file system."_

@cahirwpz what are your thoughts on this?

mohrcore:
`vmem` test still fails, but I'm unable to reproduce this failure locally. It seems to be some rare heizenbug. I'm unfamiliar with vmem at this point, so I have no clue what is a potential cause (other than a race condidtion I might be overlooking).

mohrcore:
Changing the `pool_alloc` to `kmalloc` seems to solve the issue. I re-ran the entire testing pipeline three times with no failures after that small change.

My take is that the pool allocator is broken, or there's some limitation to its use that I'm not aware of. Or the `vmem` test is broken (I don't get what it does tho) Feel free to call it bs, but I don't see another explanation atm.

include/sys/vnode.h
```diff
@@ -52,6 +52,7 @@ typedef int vnode_mkdir_t(vnode_t *dv, componentname_t *cn, vattr_t *va,
 typedef int vnode_rmdir_t(vnode_t *dv, vnode_t *v, componentname_t *cn);
 typedef int vnode_access_t(vnode_t *v, accmode_t mode, cred_t *cred);
 typedef int vnode_ioctl_t(vnode_t *v, u_long cmd, void *data, file_t *fp);
+typedef int vnode_inactive(vnode_t *v);
 typedef int vnode_reclaim_t(vnode_t *v);
```
pj1031999, line 56: Inconsistent naming schema.

sys/kern/tmpfs.c
```diff
@@ -628,33 +633,20 @@ static vnodeops_t tmpfs_vnodeops = {.v_lookup = tmpfs_vop_lookup,
                                     .v_mkdir = tmpfs_vop_mkdir,
                                     .v_rmdir = tmpfs_vop_rmdir,
                                     .v_access = vnode_access_generic,
+                                    .v_inactive = tmpfs_vop_inactive,
                                     .v_reclaim = tmpfs_vop_reclaim,
                                     .v_readlink = tmpfs_vop_readlink,
                                     .v_symlink = tmpfs_vop_symlink,
                                     .v_link = tmpfs_vop_link};
```
pj1031999, line 640: Please add `,` after last struct member - then shitty clang-format will be happy and it will produce human readable output :)

sys/kern/tmpfs.c
```diff
@@ -997,19 +994,20 @@ static int tmpfs_resize(tmpfs_mount_t *tfm, tmpfs_node_t *v, size_t newsize) {
...
-  mtx_lock(&v->tfn_timelock);
+  mtx_lock(&node->tfn_timelock);
   if (atime->tv_sec != VNOVAL)
-    v->tfn_atime = *atime;
+    node->tfn_atime = *atime;
   if (mtime->tv_sec != VNOVAL)
-    v->tfn_mtime = *mtime;
-  mtx_unlock(&v->tfn_timelock);
+    node->tfn_mtime = *mtime;
+  mtx_unlock(&node->tfn_timelock);
```
pj1031999, line 1008: `WITH_MTX_LOCK`

sys/kern/vfs_vcache.c
```diff
@@ -0,0 +1,157 @@
...
+
+static vnode_list_t vcache_free;
+static vnode_list_t vcache_buckets[VCACHE_BUCKET_CNT];
+static mtx_t vcache_giant_lock;
+
+typedef size_t vcache_t;
+
+static vcache_t vcache_hash(mount_t *mp, ino_t ino) {
+  return (((vcache_t)mp->mnt_vnodecovered >> 3) + ino) % VCACHE_BUCKET_CNT;
+}
```
pj1031999, line 32: Try to convince me that it's a good hash function

sys/kern/vfs_vcache.c
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  mtx_init(&vcache_giant_lock, 0);
+}
+
+vnode_t *vfs_vcache_reborrow(mount_t *mp, ino_t ino) {
+  SCOPED_MTX_LOCK(&vcache_giant_lock);
+
+  return vcache_hashget(mp, ino);
+}
```
pj1031999, line 111: It's the only place that uses `vcache_hashget` in code. Why they are separated functions?

sys/kern/vfs_vcache.c
```diff
@@ -0,0 +1,157 @@
...
+
+  /* Get the Least Recently Used vnode */
+  vn = TAILQ_FIRST(&vcache_free);
+
+  /* Remove it from free list */
+  TAILQ_REMOVE(&vcache_free, vn, v_free);
+
+  /* Remove it from its bucket (if present in any) */
+  error = vfs_vcache_detach(vn);
+  assert(error == 0);
```
pj1031999, line 136: What is a purpose of this assertion? Why you don't propagate error?

PR #1154, opened 2021-06-14
Timer stop & start functions fix

wiklam:
We want to set up the interrupts while attaching a timer and eventually stop it from invoking interrupts.
Removing unnecessary function `tm_select` (choosing time source is doing while setting up timers in clock.c).
Typo in rtc.c.

### DIFF
--- a/sys/mips/timer.c
+++ b/sys/mips/timer.c
@@ -52,7 +52,7 @@ static int set_next_tick(mips_timer_state_t *state) {
   /* calculate next value of compare register based on timer period */
   do {
     state->compare.val += state->period_cntr;
-    mips32_set_c0(C0_COMPARE, state->compare.lo);
+    mips32_setcompare(state->compare.lo);
     (void)read_count(state);
     ticks++;
   } while (state->compare.val <= state->count.val);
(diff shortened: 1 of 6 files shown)


sys/mips/timer.c
```diff
@@ -52,7 +52,7 @@ static int set_next_tick(mips_timer_state_t *state) {
   /* calculate next value of compare register based on timer period */
   do {
     state->compare.val += state->period_cntr;
-    mips32_set_c0(C0_COMPARE, state->compare.lo);
+    mips32_setcompare(state->compare.lo);
```
hadarai, line 55:
```suggestion
    mips32_set_compare(state->compare.lo);
```

pj1031999: `mips32_set_compare` - where it is defined? What is wrong with `mips32_setcompare`?

hadarai: Oh right, I didn't check the definition... 
Just noticed absolutely incoherent var/fun name, but it matches the naming scheme of definitions around

cahirwpz:
IIRC The intent of this PR is to change timer drivers so that resource allocation (including interrupt handler registration) is  performed only when the driver is attached. This involves masking and unmasking an interrupt in `*_start` and `*_stop` timer routines. There's no infrastructure that allows for a single `intr_handler_t` to be temporarily disabled without tearing it down. Hence we resort to configuring timers in such a way they're effectively disabled. Is that correct?

PR #1155, opened 2021-06-14
Kgmon fix

wiklam:
Fixing omitting incrementing index while the from_val value is 0.
Fromindex is a hash for a function (the caller) it invoked other functions (callee) stored in to_struct.
Not incrementing the fromindex for functions, which didn't call a function (`from_val == 0`) will assign the called functions to the previous hashes - the call graph is not reliable.

### DIFF
--- a/sys/debug/kgmon.py
+++ b/sys/debug/kgmon.py
@@ -32,9 +32,10 @@ def gmon_write(path):
         size = calcsize(tos_rep)
         tos_array = unpack(tos_rep * int(gparam.tossize/size), memory)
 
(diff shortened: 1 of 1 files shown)


pj1031999 (approved):

cahirwpz (approved):

cahirwpz:
Could you provide a description for the bug this PR fixes? How does this bug manifest itself?

wiklam:
> Could you provide a description for the bug this PR fixes? How does this bug manifest itself?

Done

PR #1158, opened 2021-06-21
Store temporary userspace vm_map in process in paging_on_demand_and_memory_protection_demo()

j-piecuch:
The correctness of the test depended on not switching to another process, which can't be guaranteed even if we disable preemption, since a thread in another process might be holding a mutex that we want to acquire. Therefore, it is necessary to store the temporary userspace `vm_map` in the current process, so that it is restored when we switch back.


pj1031999 (approved):

cahirwpz (approved):
LGTM

PR #1160, opened 2021-06-22
Make pipe I/O operations interruptible by signals.

j-piecuch:
Without this, processes sleeping on pipe I/O won't be interrupted by signals, which means e.g. that when a process in a background pipeline tries to read from the controlling terminal, processes in the pipeline that are sleeping on pipe I/O won't be stopped.

### DIFF
--- a/sys/kern/pipe.c
+++ b/sys/kern/pipe.c
@@ -87,8 +87,11 @@ static int pipe_read(file_t *f, uio_t *uio) {
       return 0;
 
     /* pipe empty, producer exists, wait for data */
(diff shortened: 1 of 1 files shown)


cahirwpz (approved):
LGTM

@hadarai I expect you to test it.

hadarai (approved):
Ok, so my PR #1148 tests that. It also was fixing it, but not as well as this change here. To test it as simply as possible I added that change to that #1148 . So checks proving this change works are [here](https://github.com/cahirwpz/mimiker/pull/1148/checks?check_run_id=2888537483)

PR #1161, opened 2021-06-23
`pipe`s simplified reimplementation 

hadarai:
With suggestions from @j-piecuch we've changed the current implementation of pipes, mostly fixes and simplifications.
Currently, our pipes are bidirectional, which is a feature implemented only by FreeBSD. On other BSDs and Linux, pipes are unidirectional. Making our pipes unidirectional will simplify the code at a small cost, since bidirectional pipes are not widely used.
Now there is only one structure with less fields shared by producers and consumers, and support for `O_CLOEXEC` has been added.


cahirwpz (approved):

cahirwpz:
I bet almost all software for Unix-like system depends on POSIX semantics for pipes. Hence it's a good idea to simplify our implementation and make a move against BSD semantics.

j-piecuch (approved):

PR #1163, opened 2021-09-03
Clean up and fixes for timers

wiklam:


cahirwpz (approved):

sys/drv/pit.c
```diff
@@ -115,24 +115,25 @@ static int timer_pit_start(timer_t *tm, unsigned flags, const bintime_t start,
...
-  uint32_t sec, ticks;
+  uint64_t sec, ticks;
   WITH_INTR_DISABLED {
     pit_update_time(pit);
     sec = pit->sec;
     ticks = pit->ticks;
   }
   bintime_t bt = bintime_mul(tm->tm_min_period, ticks);
-  bt.sec += sec;
+  assert(bt.sec == 0);
```
cahirwpz, line 135: Why such an assumption?

sys/mips/timer.c
```diff
@@ -100,14 +100,15 @@ static int mips_timer_stop(timer_t *tm) {
 static bintime_t mips_timer_gettime(timer_t *tm) {
   device_t *dev = tm->tm_priv;
   mips_timer_state_t *state = dev->state;
-  uint32_t sec, ticks;
+  uint64_t sec, ticks;
```
cahirwpz, line 103: Do we need to change ticks to `uint64_t` as well?

sys/mips/timer.c
```diff
@@ -100,14 +100,15 @@ static int mips_timer_stop(timer_t *tm) {
...
   WITH_INTR_DISABLED {
     read_count(state);
     sec = state->sec;
     ticks = state->cntr_modulo;
   }
   bintime_t bt = bintime_mul(tm->tm_min_period, ticks);
-  bt.sec += sec;
+  assert(bt.sec == 0);
```
cahirwpz, line 110: Same question as in PIT...

cahirwpz:
PR description is missing...

sys/drv/pit.c
wiklam: Cause the multiplied value is a modulo of timer frequency and the sum of time which the ticks took should be less than a second.

sys/mips/timer.c
wiklam: Fixed!

pj1031999 (approved):

cahirwpz (approved):

PR #1164, opened 2021-09-06
Clean up for time infrastructure

wiklam:
Correct the size in tm2sec.
First use of kgprof -- noticed that tm2sec have been called 3 times, even if it should be called once - corrected it by calculating the macro argument before.
Renaming ticks to cntr in PIT to stay by the convention and distinguish the difference between a counter tick and a timer tick.


pj1031999 (approved):

cahirwpz:
@wiklam Please resolve conflicts!

cahirwpz (approved):
LGTM

PR #1091, opened 2021-04-20
Adding kgprof instrumentation for AArch64

wiklam:
Depends on #1062

PR #1168, opened 2021-09-15
Make rpi3 as default board.

pj1031999:
MIPS architecture is dead. It still will be supported by Mimiker but it's time for changes for something modern like 64-bit ARM.

That commit changes default settings in build system and launcher to use rpi3 as default board.


cahirwpz:
I could accept it if all major features are in sync between MIPS and AArch64. ATM I'm not sure if profiling works on AArch64.

pj1031999:
> I could accept it if all major features are in sync between MIPS and AArch64. ATM I'm not sure if profiling works on AArch64.

@wiklam ?

cahirwpz:
Need to test and merge #1091 first.

cahirwpz (approved):

PR #1169, opened 2021-10-02
Install etc/shells in sysroot.

MichalBlk:
`mimiker/etc/shells` should be copied to `sysroot/etc` along with the other files from this directory.


cahirwpz (approved):
LGTM

PR #1170, opened 2021-10-02
Delete PMAP_* boundaries and use defines from vm_params instead.

MichalBlk:
pmap.h files define `PMAP_*` boundaries for user and kernel address spaces but these values are already defined (and used by the `vm_map` module) in vm_params.h.


cahirwpz (approved):
LGTM

PR #1171, opened 2021-10-02
[MIPS] Do not invalidate tlb entry for a kseg0 page table address.

MichalBlk:
In `pmap_add_pde` we invalidate a kseg0 page table address but kseg0 is an unmapped VM segment.
Besides, a klog invocation has been fixed.


cahirwpz:
This PR will become a part of #1243

PR #1173, opened 2021-10-02
Do not double timers on the timers list.

MichalBlk:
In `tm_release`, there's no need to insert a timer to `timers` as the timer hasn't been removed yet.
A timer is removed from `timers` only upon deregistration.


cahirwpz (approved):
LGTM

PR #1175, opened 2021-10-02
Save and restore the kframe pointer.

MichalBlk:
Currently, `kframe` is poorly handled on MIPS and not handled at all on AArch64. This PR properly saves the previous value of the kernel exception frame pointer upon entrance to a nested exception and restores it afterwards.


sys/aarch64/trap.c
```diff
@@ -167,3 +174,23 @@ void kern_trap_handler(ctx_t *ctx) {
       kernel_oops(ctx);
   }
 }
+
+__no_profile void kern_exc_handler(ctx_t *ctx, kexc_reason_t reason) {
```
wiklam, line 178: Is it possible to get the reason from `exc_code` as it is done in MIPS?

sys/mips/trap.c
```diff
@@ -277,4 +282,6 @@ __no_profile void mips_exc_handler(ctx_t *ctx) {
   } else {
     intr_root_handler(ctx);
   }
+
+  td->td_kframe = kframe_saved;
```
wiklam, line 286: As far as I understood, previously we did not restore the old context value, why we didn't notice that?

sys/aarch64/evec.S
```diff
@@ -145,7 +145,8 @@ ENTRY(handle_kern_trap)
         .cfi_signal_frame
         save_ctx 1
         mov     x0, sp
-        bl      kern_trap_handler
```
wiklam, line 148: Just as a notice - "magic constants", not sure if it would be better to define the values, from the enum below

MichalBlk: Yeah, I could move `kexc_reason_t` to a header and expose these constants through genassym, but I'm not sure if any of existing Aarch64 headers is suitable for this enum.

sys/mips/trap.c
MichalBlk: Because nobody relies on that (see e.g. how interrupts are handled - `intr_root_handler` takes the context as an argument although it could rely on the pointer in thread structure).

sys/aarch64/trap.c
MichalBlk: I believe it isn't. In Aarch64 traps are vectored and I would expect one to distinguish the cause based on the vector being used.

MichalBlk:
There is one flaw in this solution I need to fix (regarding MIPS).

PR #1183, opened 2021-11-05
Process kernel linker scripts using CPP.

MichalBlk:
Required by #1176.

We would like to support several (RISC-V, platform) setups having a single linker script for the kernel. To achieve this goal we can parametrize linker scripts whit constants exposed by genassym (see http://bxr.su/NetBSD/sys/arch/riscv/conf/kern.ldscript).

### DIFF
--- a/sys/Makefile
+++ b/sys/Makefile
@@ -12,22 +12,14 @@ CLEAN-FILES += mimiker.elf.map
 
 include $(TOPDIR)/build/build.kern.mk
 
-LDFLAGS += -T $(TOPDIR)/sys/kernel.ld
+LDSCRIPT = $(TOPDIR)/sys/$(ARCH)/$(ARCH).ld
+LDFLAGS += -T $(LDSCRIPT)
 LDLIBS	= -Wl,--start-group \
 	    -Wl,--whole-archive $(KLIBLIST) -Wl,--no-whole-archive \
             -lgcc \
           -Wl,--end-group
 
-kernel.ld: $(ARCH)/$(ARCH).ld
-	@echo "[CPP] $(DIR)$< -> $(DIR)$@"
-ifeq ($(CPPLDSCRIPT), 1)
-	$(CPP) $(CPPFLAGS) -I$(TOPDIR)/include $< | grep -v '^#' | \
-		grep -v '^$$' > $@
-else
-	$(CP) $< $@
-endif
-
-mimiker.elf: $(KLIBDEPS) kernel.ld
+mimiker.elf: $(KLIBDEPS) $(LDSCRIPT)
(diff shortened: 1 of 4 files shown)


sys/Makefile
```diff
@@ -12,13 +12,22 @@ CLEAN-FILES += mimiker.elf.map
 
 include $(TOPDIR)/build/build.kern.mk
 
-LDFLAGS += -T $(TOPDIR)/sys/$(ARCH)/$(ARCH).ld
+LDFLAGS += -T $(TOPDIR)/sys/kernel.ld
```
pj1031999, line 15: I don't like this change. It's architecture specific so from my point of view it should lives in architecture directory.

sys/Makefile
```diff
@@ -12,13 +12,22 @@ CLEAN-FILES += mimiker.elf.map
...
 
-mimiker.elf: $(KLIBDEPS)
+kernel.ld: $(ARCH)/$(ARCH).ld
+	@echo "[CPP] $(DIR)$< -> $(DIR)$@"
+ifeq ($(CPPLDSCRIPT), 1)
+	$(CPP) $(CPPFLAGS) -I$(TOPDIR)/include $< | grep -v '^#' | \
+		grep -v '^$$' > $@
+else
+	$(CP) $< $@
+endif
```
pj1031999, line 28: Please run *man gcc* and read it *carefully*.
You can remove this error prone magic and use
`$(CPP) $(CPPFLAGS) -I$(TOPDIR)/include -P -o $@ $<`
I also don't understand why do you need `CPPLDSCRIPT` - I removed that if statement and it works for AArch64.
I see that there is a problem with mips but you can fixed it by force undef & define (and it's a common solution for that).

sys/Makefile
MichalBlk: I agree. It's been moved to architecture directory.

sys/Makefile
MichalBlk: I took the formula directly form NetBSD, but I agree your solution is much better! Regarding conditional CPP processing (i.e. `CPPLDSCRIPT`), I agree it's a weird idea and it also has been removed. @cahirwpz proposed introducing optional `$(ARCH).ld.in` files generating `$ARCH).ld` files.

pj1031999 (approved):
LGTM

sys/Makefile
pj1031999: Looks much better now.

pj1031999:
One more thing (not related to your changes) - we should fix *all* makefiles. In headers we define how we should format makefiles and we don't follow these rules.

cahirwpz (approved):

PR #1184, opened 2021-11-05
Remove needless UC defines.

MichalBlk:
There is a handful of UC defines which are unused (in both architectures).


pj1031999 (approved):

cahirwpz (approved):
LGTM

PR #1186, opened 2021-11-05
Add Renode to launch.

MichalBlk:
Required by #1176.

Renode has built-in support for SOCs generated using LiteX. We need it to simulate the VexRiscv platform.

Includes changes from #1185.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 *.orig
 *.log
 *#
+core
 
 # Linker output
 *.map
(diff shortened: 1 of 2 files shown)


pj1031999:
@MichalBlk renode is not part of my area in Mimiker (and I don't want to change that). I guess that @cahirwpz is much better choice for being a reviewer.

cahirwpz (approved):
LGTM

PR #1189, opened 2021-11-25
[DEV] Remove redundant argument from root interrupt handler.

MichalBlk:
`intr_root_filter` routine contains a redundant argument `arg`. It's completely needless (and not used) as a rootdev driver can pass any needed data through its state (accessed through `driver_t` structure).


cahirwpz (approved):
LGTM

PR #1190, opened 2021-11-25
[DEV] Remove interrupt handler priority.

MichalBlk:
FTTB, all interrupt handlers associated with a given interrupt event are equally important (we always set the priority to 0).


cahirwpz (approved):
LGTM

PR #1191, opened 2021-11-25
[dev] Inhibit releasing resources that device owns.

MichalBlk:
Each device has statically assigned resources handed to it by the parent bus. Those resources can't be released using `bus_release_resource`, they can only be deactivated. `bus_release_resource` should only be called on resources allocated dynamically (i.e. by direct calls to `bus_alloc_resource`).

### DIFF
--- a/include/sys/bus.h
+++ b/include/sys/bus.h
@@ -199,10 +199,8 @@ int bus_activate_resource(device_t *dev, resource_t *r);
  */
 void bus_deactivate_resource(device_t *dev, resource_t *r);
 
-static inline void bus_release_resource(device_t *dev, resource_t *r) {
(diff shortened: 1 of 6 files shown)


pj1031999 (commented):
I don't understand these changes. You have moved unused function from header file into c sources. Why?

MichalBlk:
Although the described problem still occurs, it's probably not the best idea to deal with this now. The issue should be solved when adding a function to release a resource in the resource list interface. The resource list implementation may call the corresponding bus interface function (as is done for `device_{add,take}_resource`). Depending on the actual implementation some convention will have to be established. FTTB, the PR will be closed as the bus release function is not used.

PR #1192, opened 2022-01-08
[TESTS] Check pointer size instead of architecture in memory related tests

MichalBlk:


pj1031999 (approved):

wiklam (approved):

cahirwpz (changes requested):
I'm in favour of this change. However we should sort out other places that refer to macros that encode bitness of an architecture, i.e. `_LP64`.

cahirwpz (approved):

PR #1194, opened 2022-01-09
[libc] Introduce generic implementation of bcopy, memcpy, memmove and strlen

MichalBlk:
Some architectures don't supply specialized implementation of aforementioned functions in NetBSD's libc.
These architectures rely on generic implementation written in C. RISC-V is among them. With generic implementation in `gen`
and presented changes in `common.mk`, one can simply make a symlink named `$(ARCH)` to `gen`.


lib/libc/string/gen/strlen.c
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#include <sys/cdefs.h>
+
+#if !defined(_KERNEL) && !defined(_STANDALONE)
+#include <assert.h>
+#include <string.h>
+#else
+#include <sys/libkern.h>
+#endif
```
pj1031999, line 39: Do we need to include anything for that simple function?

lib/libc/string/gen/memcpy.c
```diff
@@ -0,0 +1,8 @@
+/*	$NetBSD: memcpy.c,v 1.2 2013/12/02 21:21:33 joerg Exp $	*/
+
+#define MEMCOPY
+#include "bcopy.c"
+
+#if defined(__ARM_EABI__)
+__strong_alias(__aeabi_memcpy, memcpy)
+#endif
```
pj1031999, line 8: Is it used?

lib/libc/string/gen/memmove.c
```diff
@@ -0,0 +1,10 @@
+/*	$NetBSD: memmove.c,v 1.2 2013/12/02 21:21:33 joerg Exp $	*/
+
+#define MEMMOVE
+#include "bcopy.c"
+
+#if defined(__ARM_EABI__)
+__strong_alias(__aeabi_memmove, memmove)
+  __strong_alias(__aeabi_memmove4, memmove)
+    __strong_alias(__aeabi_memmove8, memmove)
+#endif
```
pj1031999, line 10: ditto

pj1031999 (commented):
I would opt for removing unused implementation from bcopy.c.

lib/libc/string/gen/strlen.c
MichalBlk: You're right. Just `stddef.h` will do.

lib/libc/string/gen/memcpy.c
MichalBlk: It isn't. We won't use ARM EABI, so the code has been removed.

lib/libc/string/gen/memmove.c
MichalBlk: Removed.

MichalBlk:
The code has been cleaned. FTTB, I left the size optimized implementations and I will consult @cahirwpz if I should remove them.

build/common.mk
```diff
@@ -68,7 +68,7 @@ include $(TOPDIR)/build/tools.mk
 
 SRCDIR ?= .
 
-vpath %.c $(SRCDIR)
+vpath %.c $(SRCDIR) $(SRCDIR)/$(ARCH)
```
cahirwpz, line 71: What is the purpose of this change? Is it really needed?

cahirwpz (approved):

PR #1182, opened 2021-11-05
[RISCV] Bootstrap

MichalBlk:
Implement the bootstrap process (i.e. `_start` -> `kernel_init`).

This PR contains changes from #1228, #1216,  #1212, #1211, #1210, #1203, #1200, and #1194.

PR #1200, opened 2022-01-13
[dtb] Introduce FDT module based on FreeBSD's FDT and OF modules.

MichalBlk:
The current DTB module doesn't implement some crucial functions independent of the target architecture. Missing functionalities include memory description parsing and boot arguments extraction. The presented change introduces an FDT module based on the FDT and OF modules from FreeBSD.


sys/kern/dtb.c
```diff
@@ -16,27 +16,114 @@ static uintptr_t _dtb_offset = 0;
...
+uint64_t dtb_cell(int parent, int node, const char *name) {
+  int len;
+  const int addr_cells = dtb_addr_cells(parent);
+  const uint32_t *prop = fdt_getprop(_dtb_root, node, name, &len);
+  if (prop == NULL || ((size_t)len != addr_cells * sizeof(uint32_t)))
+    panic("Invalid %s property in node %d (parent=%d)!", name, node, parent);
+  return dtb_to_cpu(prop, addr_cells);
+}
+
+void dtb_mem(uint64_t *addr_p, uint64_t *size_p) {
```
wiklam, line 101: Not sure if it is importnat, but do we assure that the pointers in this function and bellow are not NULL?

sys/kern/dtb.c
```diff
@@ -16,27 +16,114 @@ static uintptr_t _dtb_offset = 0;
...
+}
+
+static int dtb_offset(const char *path) {
+  int offset = fdt_path_offset(_dtb_root, path);
+  if (offset < 0)
+    panic("Failed to find dtb offset for %s!");
+  return offset;
+}
+
+static int dtb_addr_cells(int node) {
```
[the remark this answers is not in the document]
MichalBlk: No. Our fdt library uses ints as node offsets.

sys/kern/dtb.c
```diff
@@ -16,27 +16,114 @@ static uintptr_t _dtb_offset = 0;
...
+  int len;
+  const uint32_t *prop = fdt_getprop(_dtb_root, node, "#address-cells", &len);
+  /* #address-cells is a 4-byte property */
+  if (prop == NULL || (size_t)len != sizeof(uint32_t))
+    panic("Failed to retreive #address-cells property for node %d!", node);
+  return fdt32_to_cpu(*prop);
+}
+
+static long dtb_to_cpu(const uint32_t *prop, int addr_cells) {
```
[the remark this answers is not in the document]
MichalBlk: Changed.

sys/aarch64/board.c
```diff
@@ -8,77 +8,37 @@
...
-  if (fdt_check_header(dtb))
-    panic("dtb incorrect header!");
-
+static void process_dtb(char **tokens, kstack_t *stk) {
   char buf[32];
-  snprintf(buf, sizeof(buf), "memsize=%d", process_dtb_memsize(dtb));
+  uint64_t start, size;
+
+  /* Memory boundaries. */
+  dtb_mem(&start, &size);
```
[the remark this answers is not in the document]
MichalBlk: There is an error in the following line. `start` should be unused. It is unused because our port assumes that memory starts at address 0. This is not portable to make such assumption (as memory boundaries should be read from dtb) and we'll fix it (RISC-V uses both `start` and `size`).

sys/kern/dtb.c
MichalBlk: Yes we assume the pointers are not NULL. It's rather a save assumption as it's a kernel module and any NULL dereference will cause a page fault and thereby kernel oops.

sys/kern/dtb.c
```diff
@@ -16,27 +16,114 @@ static uintptr_t _dtb_offset = 0;
...
+}
+
+/*
+ * XXX: FTTB, we consider only a single pair <addr size>.
+ */
+void dtb_reg(int parent, int node, uint64_t *addr_p, uint64_t *size_p) {
+  int len;
+  const int addr_cells = dtb_addr_cells(parent);
+  const uint32_t *prop = fdt_getprop(_dtb_root, node, "reg", &len);
+  if (prop == NULL || ((size_t)len % (2 * addr_cells * sizeof(uint32_t)) != 0))
```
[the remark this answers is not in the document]
wiklam: In this case it is a power of two and it would be better to check the bits with an and.
```suggestion
  if (prop == NULL || (((size_t)len & (2 * addr_cells * sizeof(uint32_t)) -1) != 0))
```

include/sys/dtb.h
```diff
@@ -16,10 +18,31 @@ void *dtb_root(void);
...
+uint64_t dtb_cell(int parent, int node, const char *name);
+
+/* Get physical memory boundaries. */
+void dtb_mem(uint64_t *addr_p, uint64_t *size_p);
+
+/* Get reserved memory boundaries. */
+void dtb_rsvdmem(uint64_t *addr_p, uint64_t *size_p);
+
+/* Get initrd memory boundaries. */
+void dtb_rd(uint64_t *addr_p, uint64_t *size_p);
```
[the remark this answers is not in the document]
wiklam: Is it worth to mention that somewhere?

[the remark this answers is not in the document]
MichalBlk: It is mentioned in `pmap_bootstrap` introduced in #1182 :) .

sys/kern/dtb.c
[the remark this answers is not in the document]
MichalBlk: You're right and we already have a macro for this purpose `is_aligned`. Fixed.

include/sys/dtb.h
```diff
@@ -3,6 +3,8 @@
 
 #include <sys/types.h>
 
+#define DTB_ROOT_NODE 0
```
pj1031999, line 6: Do you need to expose this outside?

sys/kern/dtb.c
```diff
@@ -16,27 +16,116 @@ static uintptr_t _dtb_offset = 0;
...
+}
+
+/*
+ * XXX: FTTB, we assume a single reserved memory range.
+ */
+void dtb_rsvdmem(unsigned long *addr_p, unsigned long *size_p) {
+  int node = dtb_offset("reserved-memory");
+  int subnode = fdt_first_subnode(_dtb_root, node);
+  if (subnode < 0)
+    panic("reserved-memory node doesn't contain any range!");
```
pj1031999, line 115: Inconsistent logs (others start with capital letters).

sys/kern/dtb.c
```diff
@@ -16,27 +16,116 @@ static uintptr_t _dtb_offset = 0;
...
+             unsigned long *size_p) {
+  int len;
+  const int addr_cells = dtb_addr_cells(parent);
+  const size_t pair_size = 2 * addr_cells * sizeof(uint32_t);
+  const uint32_t *prop = fdt_getprop(_dtb_root, node, "reg", &len);
+  if (prop == NULL || !is_aligned(len, pair_size))
+    panic("Invalid reg property in node %d (parent=%d)!", node, parent);
+
+  *addr_p = dtb_to_cpu(prop, addr_cells);
+  *size_p = dtb_to_cpu(prop + addr_cells, addr_cells);
```
pj1031999, line 91: `assert(addr_p)` & `assert(size_p)` or ignore if arguments are `NULL`. It's external interface so somebody will use it in wrong way.

sys/aarch64/board.c
```diff
@@ -8,77 +8,37 @@
...
+
+  /* Kernel cmdline. */
+  tokens = cmdline_extract_tokens(stk, dtb_cmdline(), tokens);
   *tokens = NULL;
 }
 
 void *board_stack(paddr_t dtb) {
-  dtb_early_init(dtb, fdt_totalsize(PHYS_TO_DMAP(dtb)));
+  init_kasan();
+  init_klog();
```
pj1031999, line 39: I'm not sure if it's a good place for that. I agree that we can do that here but it's not related to preparing stack.

sys/aarch64/board.c
```diff
@@ -88,10 +48,17 @@ void *board_stack(paddr_t dtb) {
...
+  process_dtb(kenvp, stk);
   kstack_fix_bottom(stk);
   init_kenv(kenvp);
 
+  /* If klog-mask argument has been supplied, let's update the mask. */
+  const char *klog_mask = kenv_get("klog-mask");
+  if (klog_mask) {
+    unsigned mask = strtol(klog_mask, NULL, 16);
+    klog_setmask(mask);
+  }
```
pj1031999, line 60: ditto

sys/aarch64/board.c
```diff
@@ -8,77 +8,37 @@
...
-    panic("dtb incorrect header!");
-
+static void process_dtb(char **tokens, kstack_t *stk) {
   char buf[32];
-  snprintf(buf, sizeof(buf), "memsize=%d", process_dtb_memsize(dtb));
+  uint64_t start, size;
+
+  /* Memory boundaries. */
+  dtb_mem(&start, &size);
+  snprintf(buf, sizeof(buf), "memsize=%lu", size);
```
pj1031999, line 22: Please add `TODO` or create issue that we have hidden assumption about start of memory.

include/sys/dtb.h
MichalBlk: FTTB, I don't. This may change when I will make PRs regarding RISC-V drivers, but for now I will move it to the source file.

sys/kern/dtb.c
MichalBlk: This is only because the log starts with a name of property. I can rearrange the message so the property isn't at the beginning. Should I do it?

sys/kern/dtb.c
MichalBlk: I have been told a few times by @cahirwpz that in kernel, instead of making an assertion that a pointer isn't NULL, we prefer to let the kernel crash and the user can see in klog's logs the faulting PC and address. @cahirwpz should I insert the proposed assertions?

sys/aarch64/board.c
MichalBlk: You're right. I proposed a new PR regarding this issue (#1203).

sys/aarch64/board.c
MichalBlk: #1203

sys/aarch64/board.c
MichalBlk: Done.

MichalBlk:
A lot has changed. In the previous solution I didn't consider the size-cells property and assumed it to be equal to address-cells property which is incorrect. Currently the PR is moving to blocked as in contains changes from #1203.

cahirwpz (approved):

cahirwpz:
@MichalBlk Please update PR description / title to match its' current state.

PR #1203, opened 2022-01-15
[boot] Initialize klog before building kernel environment.

MichalBlk:
We want to use dtb module inside `board_stack` (i.e. while building kernel environment), but to do so we need to initialize ksan and klog before moving to `board_stack`. This PR introduces a new function to accommodate the needed calls.

### DIFF
--- a/sys/kern/kenv.c
+++ b/sys/kern/kenv.c
@@ -1,7 +1,7 @@
 #include <sys/kenv.h>
 #include <sys/libkern.h>
 
-static char **_kenvp;
+static char **_kenvp = (char *[1]){NULL};
 static char **_kinit = (char *[2]){NULL, NULL};
 
 void init_kenv(char **kenvp) {
--- a/sys/kern/kenv.c
+++ b/sys/kern/kenv.c
@@ -1,7 +1,7 @@
 #include <sys/kenv.h>
 #include <sys/libkern.h>
 
-static char **_kenvp = (char *[1]){NULL};
+static char **_kenvp;
 static char **_kinit = (char *[2]){NULL, NULL};
 
 void init_kenv(char **kenvp) {
--- a/sys/kern/klog.c
+++ b/sys/kern/klog.c
@@ -127,7 +127,7 @@ unsigned klog_setmask(unsigned newmask) {
   return atomic_exchange(&klog.mask, newmask);
 }
 
-void klog_update_mask(void) {
+void klog_config(void) {
   const char *mask_str = kenv_get("klog-mask");
   if (mask_str) {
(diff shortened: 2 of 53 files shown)


pj1031999 (commented):
Hmm... I'm not sure what do you want to achieve. Let me express my thoughts.
- kasan is still disabled before `board_init` because `boot.c` can't be sanitized
- `init_klog` already parse `klog-mask` from kernel env, I understand that we don't have access to `klog-mask` before stack is created, but in previous model we don't have to do that - it's ugly from my point of view, and it's not a responsibility of MD code to do that... maybe something like `klog_update_mask` will be better
- I see that `panic` should not be used before `init_klog` and I didn't notice it when I was writing this code... I think that MD panic (which doesn't call klog) will be good enough since we already stop on `halt` in gdb session
- for me raw assembler is cleaner than embedding it into C code (also I noticed that we are inconsistent `__asm` vs `asm`)

so from my point of view you only enabled `klog` for `board_stack`. Am I missing something?

MichalBlk:
You're right. All I want to do is to initialize klog before calling `board_stack` as dtb module will use it (#1200). I don't want to introduce MD panic for usage before initializing klog as dtb module will also be called form drivers code. What would you say about that: I will move `init_klog` call to `_start` (after `aarch64_init`), `init_kasan` will stay in `board_init`.

pj1031999:
Oh, I see, I missed that now you are using generic code for dtb. In that case it sounds good. Ping me after update and I will be happy to see code.

sys/kern/kenv.c
```diff
@@ -1,7 +1,7 @@
 #include <sys/kenv.h>
 #include <sys/libkern.h>
 
-static char **_kenvp;
+static char **_kenvp = (char *[1]){NULL};
```
pj1031999, line 4: Short comment why it is needed?

pj1031999 (approved):
LGTM

sys/kern/kenv.c
MichalBlk: It was needed because of `_kenvp` dereference in `init_klog`, but ss I'm thinking about it now, the aforementioned dereference shouldn't be present. I have changed the code so that
1. klog is initialized with the default mask.
2. `klog_update_mask` set the appropriate mask (if supplied).
Now the pointed change is obsolete.

sys/kern/klog.c
```diff
@@ -43,8 +43,7 @@ static const char *subsystems[] = {
 };
 
 void init_klog(void) {
```
cahirwpz, line 45: Why not initialize klog structure statically? Do we really need code for doing that ?

MichalBlk: You're right! Done.

PR #1206, opened 2022-01-17
[thread] introduce dedicated functions for locks

pj1031999:
- `thread_lock(td)` - acquire `td::td_lock`
- `thread_unlock(td)` - release `td::td_lock`
- `thread_owned(td)` - return `spin_owned(td->td_lock)`


cahirwpz:
This PR introduces only cosmetic changes so it's not crucial in any way. Moreover it has largely rotten over the last 6 months. Hence I decided to close it.

PR #1207, opened 2022-01-20
[sched] introduce sched_remove

pj1031999:
Add ability to remove thread from scheduler.


cahirwpz (commented):
What is the rationale behind introducing this function? You know that I'm reluctant to merge code that will become handy at some undetermined moment in the future ;-)

PR #1212, opened 2022-03-11
[RISCV][libc] Employ the generic libc string functions.

MichalBlk:
Contains changes introduced in #1194.


MichalBlk:
This will be contained in #1194.

PR #1218, opened 2022-04-01
[RISCV] Introduce kernel SBI library.

MichalBlk:
A supervisor mode RISC-V kernel communicates with its execution environment via Supervisor Binary Interface (SBI). The presented change introduces a kernel library for issuing services defined by the SBI specification.

### DIFF
--- a/sys/riscv/sbi.c
+++ b/sys/riscv/sbi.c
@@ -25,7 +25,16 @@
  * SUCH DAMAGE.
  */
 
-/* MODIFIED */
+/*
+ * MODIFIED
+ *
+ * We make the following assumptions based on OpenSBI v1.0-44-g9cd95e1:
+ *  - the TIMER, IPI, RFENCE, HSM, and LEGACY extensions are always supported,
+ *  - the SRST extension support is platform-dependent.
+ *
+ * If the newest OpenSBI version introduces any changes to the aforementioned
+ * assumptions, the code should be appropriately modified.
+ */
 
 #include <sys/klog.h>
 #include <sys/mimiker.h>
@@ -45,20 +54,16 @@ static u_long sbi_spec_version;
 static u_long sbi_impl_id;
 static u_long sbi_impl_version;
 
-static bool has_time_extension = false;
-static bool has_ipi_extension = false;
-static bool has_rfnc_extension = false;
 static bool has_srst_extension = false;
-static bool has_console_putchar_extension = false;
-static bool has_console_getchar_extension = false;
 
 /* Hardware implementation info. */
 static u_long mvendorid; /* CPU's JEDEC vendor ID */
 static u_long marchid;
 static u_long mimpid;
 
-static sbi_ret_t sbi_call(u_long ext, u_long func, u_long arg0, u_long arg1,
-                          u_long arg2, u_long arg3, u_long arg4) {
+static inline sbi_ret_t sbi_call(u_long ext, u_long func, u_long arg0,
+                                 u_long arg1, u_long arg2, u_long arg3,
+                                 u_long arg4) {
   u_long error, value;
   __asm __volatile("mv a0, %2\n\t"
                    "mv a1, %3\n\t"
@@ -124,14 +129,7 @@ static sbi_ret_t sbi_get_mimpid(void) {
  */
 
 int sbi_set_timer(uint64_t val) {
-  sbi_ret_t ret;
-
-  /* Use the TIME legacy replacement extension, if available. */
-  if (has_time_extension)
-    ret = SBI_CALL1(SBI_EXT_ID_TIME, SBI_TIME_SET_TIMER, val);
-  else
-    ret = SBI_CALL1(SBI_SET_TIMER, 0, val);
-
+  sbi_ret_t ret = SBI_CALL1(SBI_EXT_ID_TIME, SBI_TIME_SET_TIMER, val);
   return (int)ret.error;
 }
 
@@ -141,11 +139,7 @@ int sbi_set_timer(uint64_t val) {
 
 /* NOTE: this request cannot fail. */
 void sbi_send_ipi(const u_long *hart_mask) {
-  /* Use the IPI legacy replacement extension, if available. */
-  if (has_ipi_extension)
-    (void)SBI_CALL2(SBI_EXT_ID_IPI, SBI_IPI_SEND_IPI, *hart_mask, 0);
-  else
-    (void)SBI_CALL1(SBI_SEND_IPI, 0, (u_long)hart_mask);
+  (void)SBI_CALL2(SBI_EXT_ID_IPI, SBI_IPI_SEND_IPI, *hart_mask, 0);
 }
 
 /*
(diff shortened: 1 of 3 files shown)


sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+  __asm __volatile("ecall"
+                   : "+r"(a0), "+r"(a1)
+                   : "r"(a2), "r"(a3), "r"(a4), "r"(a6), "r"(a7)
+                   : "memory");
+
+  return (sbi_ret_t){
+    .error = a0,
+    .value = a1,
+  };
+}
```
pj1031999, line 79: I'm not sure if we can trust compiler enough to use that tricky syntax - from my point of view it's better to write this function in pure assembler.

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+void sbi_set_timer(uint64_t val) {
+  sbi_ret_t ret;
+
+  /* Use the TIME legacy replacement extension, if available. */
+  if (has_time_extension)
+    ret = SBI_CALL1(SBI_EXT_ID_TIME, SBI_TIME_SET_TIMER, val);
+  else
+    ret = SBI_CALL1(SBI_SET_TIMER, 0, val);
+
+  assert(ret.error == SBI_SUCCESS);
```
pj1031999, line 133: Why we can't propagate error and give opportunity to decide how to interpret that?

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+void sbi_send_ipi(const u_long *hart_mask) {
+  sbi_ret_t ret;
+
+  /* Use the IPI legacy replacement extension, if available. */
+  if (has_ipi_extension)
+    ret = SBI_CALL2(SBI_EXT_ID_IPI, SBI_IPI_SEND_IPI, *hart_mask, 0);
+  else
+    ret = SBI_CALL1(SBI_SEND_IPI, 0, (u_long)hart_mask);
+
+  assert(ret.error == SBI_SUCCESS);
```
pj1031999, line 149: ditto

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+void sbi_remote_fence_i(const u_long *hart_mask) {
+  sbi_ret_t ret;
+
+  /* Use the RFENCE legacy replacement extension, if available. */
+  if (has_rfnc_extension)
+    ret = SBI_CALL2(SBI_EXT_ID_RFNC, SBI_RFNC_REMOTE_FENCE_I, *hart_mask, 0);
+  else
+    ret = SBI_CALL1(SBI_REMOTE_FENCE_I, 0, (u_long)hart_mask);
+
+  assert(ret.error == SBI_SUCCESS);
```
pj1031999, line 165: ditto

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+  sbi_ret_t ret;
+
+  /* Use the RFENCE legacy replacement extension, if available. */
+  if (has_rfnc_extension)
+    ret = SBI_CALL4(SBI_EXT_ID_RFNC, SBI_RFNC_REMOTE_SFENCE_VMA, *hart_mask, 0,
+                    start, size);
+  else
+    ret = SBI_CALL3(SBI_REMOTE_SFENCE_VMA, 0, (u_long)hart_mask, start, size);
+
+  assert(ret.error == SBI_SUCCESS);
```
pj1031999, line 178: ditto

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+
+  /* Use the RFENCE legacy replacement extension, if available. */
+  if (has_rfnc_extension)
+    ret = SBI_CALL5(SBI_EXT_ID_RFNC, SBI_RFNC_REMOTE_SFENCE_VMA_ASID,
+                    *hart_mask, 0, start, size, asid);
+  else
+    ret = SBI_CALL4(SBI_REMOTE_SFENCE_VMA_ASID, 0, (u_long)hart_mask, start,
+                    size, asid);
+
+  assert(ret.error == SBI_SUCCESS);
```
pj1031999, line 193: ditto

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+}
+
+/*
+ * Hart state management extension.
+ */
+
+int sbi_hsm_hart_start(u_long hart, u_long start_addr, u_long priv) {
+  sbi_ret_t ret =
+    SBI_CALL3(SBI_EXT_ID_HSM, SBI_HSM_HART_START, hart, start_addr, priv);
+  return (int)ret.error;
```
pj1031999, line 203: Why result is truncated to int?

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+}
+
+void sbi_hsm_hart_stop(void) {
+  (void)SBI_CALL0(SBI_EXT_ID_HSM, SBI_HSM_HART_STOP);
+}
+
+int sbi_hsm_hart_status(u_long hart) {
+  sbi_ret_t ret = SBI_CALL1(SBI_EXT_ID_HSM, SBI_HSM_HART_STATUS, hart);
+  return (ret.error != SBI_SUCCESS ? (int)ret.error : (int)ret.value);
```
pj1031999, line 212: ditto

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+    (void)SBI_CALL1(SBI_CONSOLE_PUTCHAR, 0, ch);
+}
+
+int sbi_console_getchar(void) {
+  if (has_console_getchar_extension) {
+    /*
+     * XXX: The "error" is returned here because legacy SBI functions
+     * continue to return their value in a0.
+     */
+    return (SBI_CALL0(SBI_CONSOLE_GETCHAR, 0).error);
```
pj1031999, line 243: ditto

sys/riscv/sbi.c
```diff
@@ -0,0 +1,346 @@
...
+static u_long sbi_spec_version;
+static u_long sbi_impl_id;
+static u_long sbi_impl_version;
+
+static bool has_time_extension = false;
+static bool has_ipi_extension = false;
+static bool has_rfnc_extension = false;
+static bool has_srst_extension = false;
+static bool has_console_putchar_extension = false;
+static bool has_console_getchar_extension = false;
```
pj1031999, line 53: Not strong opinion - you can consider creating enum with or of extensions.

pj1031999 (commented):
A bunch of questions. I don't have strong opinion here since this code is copied but I want to understand some quirks.

sys/riscv/sbi.c
MichalBlk: FreeBSD and Linux do it this way, however, I'm not sure whether it requires some additional compiler options to handle it properly in all scenarios (e.g. when the -O3 is used). I have changed it to relay on bare assembly.

sys/riscv/sbi.c
MichalBlk: I agree. Done.

sys/riscv/sbi.c
MichalBlk: Done.

sys/riscv/sbi.c
MichalBlk: Done.

sys/riscv/sbi.c
MichalBlk: as above

sys/riscv/sbi.c
MichalBlk: as above

sys/riscv/sbi.c
MichalBlk: We decided to rely on the current version of OpenSBI. The only extension whose support is platform-specific is the system reset extension, so now we have only a single flag.

MichalBlk: The assumptions are described at the top of `sys/riscv/sbi.c`.

cahirwpz (approved):

PR #1224, opened 2022-04-04
[dev] Interrupt controller interface.

MichalBlk:
With the arrival of RISC-V we finally have interrupt controllers that cannot be perceived as buses. Therefore, the current bus interface must be split into two separate interfaces:
- bus interface,
- interrupt controller interface.

### DIFF
--- a/include/sys/bus.h
+++ b/include/sys/bus.h
@@ -134,9 +134,6 @@ extern bus_space_t *generic_bus_space;
 #define bus_space_map(t, a, s, hp) (*(t)->bs_map)((a), (s), (hp))
 
 struct bus_methods {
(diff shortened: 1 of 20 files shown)


pj1031999 (approved):
LGTM

pj1031999:
@MichalBlk what is the reason for renaming?

MichalBlk:
@pj1031999 @cahirwpz doesn't like the "ic" abbreviation. I suggested ICU (Interrupt Control Unit) whereas @cahirwpz suggested PIC (Programmable Interrupt Controller). It seems that the latter term is used more often so I used PIC. The reason why the functions start with the `pic_` prefix is to be consistent with the remaining interfaces (e.g. `bus_*`, `uart_*`).

cahirwpz (approved):

PR #1188, opened 2021-11-19
[RISCV] Kenrel init.

MichalBlk:

PR #1226, opened 2022-04-20
[RISCV] HiFive Unleashed

MichalBlk:
This PR contains all changes needed to run Mimiker on the HiFive Unleashed QEmu emulated board.


cahirwpz:
Leftovers after RISC-V 64-bit port were merged into #1188

PR #1233, opened 2022-04-29
[dev] Simplebus

MichalBlk:


cahirwpz:
Since we've merged #1235 we don't need to invest more time into simplebus, however we'll tackle the topic in near future.

PR #1234, opened 2022-05-01
Add drive images through CLI

mohrcore:
Allows to specify drive image(s) to Qemu when launching mimiker through `launch` script.

Usage:
```
launch -s sd:image.img
launch --storage sd:image.img
```

The argument for `--storage` consists of two parts splitted by `:` delimiter. On the left of `:` there's a name of a drive bus associated with a board, on the right there's a path to an image file. The format of the file is decided as a part of drive bus config, see `qemu.boards.<board_name>.drives` config in `launch`.

### DIFF
--- a/launch
+++ b/launch
@@ -52,6 +52,7 @@ CONFIG = {
         'gdbport': RandomPort(),
         'graphics': False,
         'network': False,
+        'storage': [],
         'elf': 'sys/mimiker.elf',
         'initrd': 'initrd.cpio',
         'args': [],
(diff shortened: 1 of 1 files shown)


cahirwpz (approved):

pj1031999 (commented):
I'm not sure if it's working as intended.
`'sd': 'if=sd,index=0,format=qcow2,file={path}'`
IIUC devices need to have unique indexes (maybe I don't) so this will be working only if you give 0 or 1 device, on the other hand user interface for lunch script support multiple devices. I guess that somebody will introduce platform with *2* hard drives and then they will waste a few hours debugging that because we still doesn't support qemu error logs in user friendly way.

From my point of view let's simplify that - you want to support only one sd card now so do that instead of something that isn't testable in our environment.

PR #1230, opened 2022-04-25
[RISCV] Physical map management.

MichalBlk:
Physical address map management for the RISC-V target architecture. The pmap module relies on the provided tlb flushing module.

PR #1235, opened 2022-05-02
[dev] Basic simplebus feature.

MichalBlk:
This is an initial approach to implement the simplebus bus. The first step is to stop assigning resources explicitly and parse the FDT instead. We still need to implement the simplebus enumeration but that will be introduced in a separate PR after the RISC-V port is finally merged into the master branch.


cahirwpz:
@pj1031999 If you started working on the review of this PR, please do not continue your work. We need you to give feedback on more important PRs like #1230

include/dev/simplebus.h
```diff
@@ -0,0 +1,24 @@
+#ifndef _DEV_SIMPLEBUS_H_
+#define _DEV_SIMPLEBUS_H_
+
+typedef uint32_t phandle_t;
+typedef struct device device_t;
```
pj1031999, line 5: Let's include real source of these typedefs.

sys/drv/simplebus.c
```diff
@@ -0,0 +1,199 @@
...
+
+  int icells;
+  int err = FDT_intr_cells(iparent, &icells);
+  if (err)
+    return err;
+
+  pcell_t *tuples;
+  int ntuples =
+    FDT_getencprop_alloc_multi(node, "interrupts", icells, (void **)&tuples);
+  if (ntuples < 0)
```
pj1031999, line 58: Error is only on `-1`. Am I missing something?

sys/drv/simplebus.c
```diff
@@ -0,0 +1,199 @@
...
+}
+
+static int sb_get_interrupts_extended(device_t *dev, fdt_intr_t *intrs,
+                                      size_t *cntp) {
+  phandle_t node = dev->node;
+
+  pcell_t *cells;
+  int ncells = FDT_getencprop_alloc_multi(node, "interrupts-extended",
+                                          sizeof(pcell_t), (void **)&cells);
+  if (ncells < 0)
```
pj1031999, line 86: ditto

sys/kern/fdt.c
```diff
@@ -130,14 +144,58 @@ ssize_t FDT_getencprop(phandle_t node, const char *propname, pcell_t *buf,
   if (buflen % 4)
     return -1;
 
-  ssize_t ret = FDT_getprop(node, propname, buf, buflen);
-  if (ret < 0)
+  ssize_t rv = FDT_getprop(node, propname, buf, buflen);
+  if (rv < 0)
```
pj1031999, line 148: Please avoid name changes if you introduce new features and do that in separate commits.

sys/kern/fdt.c
```diff
@@ -130,14 +144,58 @@ ssize_t FDT_getencprop(phandle_t node, const char *propname, pcell_t *buf,
...
+      FDT_free(buf);
+      buf = NULL;
+      goto end;
+    }
+  }
+  rv = len / elsz;
+
+end:
+  *bufp = buf;
+  return rv;
```
pj1031999, line 177: You can just return `-1` and leave `*bufp` unchanged in case of error. For me it will be more readable.

include/dev/simplebus.h
MichalBlk: Done.

sys/drv/simplebus.c
MichalBlk: You're right. I changed it to ` == -1`.

sys/drv/simplebus.c
MichalBlk: Done.

sys/kern/fdt.c
MichalBlk: Good point. I removed these changes from this PR.

sys/kern/fdt.c
MichalBlk: I agree. It makes the code much more readable. Done.

cahirwpz (approved):

PR #1240, opened 2022-05-02
[RISCV] Trap handling module.

MichalBlk:
This code implements trap handling for the RISC-V target architecture.
NOTE: it requires the trap entry module to function properly.


sys/riscv/trap.c
```diff
@@ -0,0 +1,301 @@
...
+ *  - a0-5: args
+ *
+ * NOTE: the following code assumes all arguments to syscalls are passed
+ * via registers.
+ */
+static_assert(SYS_MAXSYSARGS <= FUNC_MAXREGARGS - 1,
+              "Syscall args don't fit in registers!");
+
+static void syscall_handler(mcontext_t *uctx, syscall_result_t *result) {
+  /* TODO: eventually we should have a platform-independent syscall handler. */
```
pj1031999, line 142: Could you explain that comment?

sys/riscv/trap.c
```diff
@@ -0,0 +1,301 @@
...
+
+  if (code > SYS_MAXSYSCALL) {
+    args[0] = code;
+    code = 0;
+  }
+
+  sysent_t *se = &sysent[code];
+  size_t nargs = se->nargs;
+
+  assert(nargs <= SYS_MAXSYSCALL);
```
pj1031999, line 156: I have mixed fillings about this assertion. It's ok to do that here but I think that it would be better to have separate place where we can validate number of arguments for each system call. Why? because here we only validate used syscalls so it's possible that we don't have full code coverage.

sys/riscv/trap.c
```diff
@@ -0,0 +1,301 @@
...
+  thread_t *td = thread_self();
+
+  if (td->td_pflags & TDP_FPUINUSE)
+    return false;
+
+  /*
+   * May be a FPE trap. Enable FPE usage
+   * for this thread and try again.
+   */
+  bzero(uctx->__fregs, sizeof(__fregset_t));
```
pj1031999, line 180: We are using `bzero, bcopy, memcpy, memset` in our code without any convention.
We have around:
- 120x `memset`
- 220x `memcpy`
-  30x `bzero`
- 40x `bcopy`

could we unify everything?

According to various manuals:
```
The bcopy() function conforms to IEEE Std 1003.1-2001 (``POSIX.1'').  The
     IEEE Std 1003.1-2004 (``POSIX.1'') revision marked it as legacy; the IEEE
     Std 1003.1-2008 (``POSIX.1'') revision removed it from the specification.
```

sys/riscv/trap.c
```diff
@@ -0,0 +1,301 @@
...
+    case SCAUSE_ILLEGAL_INSTRUCTION: {
+#if FPU
+      if (fpu_handler((mcontext_t *)ctx))
+        break;
+#endif
+      void *epc = (void *)_REG(ctx, PC);
+      klog("%s at %p!", exceptions[code], epc);
+      sig_trap(ctx, SIGILL);
+      break;
+    }
```
pj1031999, line 232: It can be not clear why `fpu_handler` is a solution for `ILLEGAL_INSTRUCTION`. Please add comment for future readers ;)

sys/riscv/trap.c
cahirwpz: Since we took large pieces of code from *BSD that we don't feel like modifying we need to keep those functions. However I reckon it'd be wise choice not to use `bzero` and `bcopy` in our code.

sys/riscv/trap.c
```diff
@@ -0,0 +1,301 @@
...
+  return _REG(ctx, CAUSE) & SCAUSE_INTR;
+}
+
+__no_profile static inline unsigned ctx_code(ctx_t *ctx) {
+  return _REG(ctx, CAUSE) & SCAUSE_CODE;
+}
+
+__no_profile static inline unsigned ctx_intr_enabled(ctx_t *ctx) {
+  return _REG(ctx, SR) & SSTATUS_SPIE;
+}
```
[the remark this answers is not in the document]
MichalBlk: You're right. I also changed the signature of `ctx_intr_enabled` to return `bool`.

sys/riscv/trap.c
```diff
@@ -0,0 +1,301 @@
...
+    case SCAUSE_STORE_ACCESS_FAULT:
+    case SCAUSE_INST_PAGE_FAULT:
+    case SCAUSE_LOAD_PAGE_FAULT:
+    case SCAUSE_STORE_PAGE_FAULT:
+      klog("Caused by reference to %lx!", _REG(ctx, TVAL));
+      break;
+
+    case SCAUSE_ILLEGAL_INSTRUCTION: {
+      uint32_t badinstr = *(uint32_t *)epc;
+      klog("Illegal instruction %08x in kernel mode!", badinstr);
```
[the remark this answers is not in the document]
MichalBlk: This is the encoding of the fault instruction (address is stored in `epc`). We assume that the compressed extension is not used.

sys/riscv/trap.c
```diff
@@ -0,0 +1,301 @@
...
+    case SCAUSE_STORE_MISALIGNED:
+    case SCAUSE_STORE_ACCESS_FAULT:
+    case SCAUSE_INST_PAGE_FAULT:
+    case SCAUSE_LOAD_PAGE_FAULT:
+    case SCAUSE_STORE_PAGE_FAULT:
+      klog("Caused by reference to %lx!", _REG(ctx, TVAL));
+      break;
+
+    case SCAUSE_ILLEGAL_INSTRUCTION: {
+      uint32_t badinstr = *(uint32_t *)epc;
```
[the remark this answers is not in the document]
MichalBlk: I used `{}` as I had some variables defined under switch cases. This has been removed thus `{}` are gone.

sys/riscv/trap.c
```diff
@@ -0,0 +1,301 @@
...
+    goto fault;
+  }
+
+  vm_prot_t access;
+  if (code == SCAUSE_INST_PAGE_FAULT)
+    access = VM_PROT_EXEC;
+  else if (code == SCAUSE_LOAD_PAGE_FAULT)
+    access = VM_PROT_READ;
+  else
+    access = VM_PROT_WRITE;
```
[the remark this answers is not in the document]
MichalBlk: I use the following convention for `if` statements with a single instruction inside:
- if there's a comment beside the statement then I include braces,
- otherwise, I don't use braces.
I applied this convention as it seems to me that it's used throughout the majority of the code (e.g. in corresponding trap modules for other architectures).
I will leave this decision to @cahirwpz.

sys/riscv/trap.c
MichalBlk: This comment is here because I saw it in the corresponding function for MIPS. I didn't ask @cahirwpz to verify if this is still meaningful but I understand that it's not. The comment has been removed.

sys/riscv/trap.c
MichalBlk: First of all, there should be `SYS_MAXSYSARGS` instead of `SYS_MAXSYSCALL` :). I agree with you. I will consult it with @cahirwpz to identify where this check should be performed (for all syscall arg structs).

sys/riscv/trap.c
MichalBlk: Done. The comment is placed before the definition of `fpu_handler`.

sys/riscv/trap.c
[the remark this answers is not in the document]
cahirwpz: If block is one liner, then it's ok to omit braces. However, if there're at least two lines (even if one of them is a comment) then I'd expect to enclose it with braces.

sys/riscv/trap.c
cahirwpz: While I understand your point I think we should firstly seek to move `syscall_handler` to `sys/gen` directory and make it common for all architectures. Then we'll solve all problems in one place.

PR #1241, opened 2022-05-02
[td] Generic thread entry setup.

MichalBlk:
FTTB, `thread_entry_setup` implementation is identical on each target architecture. Therefore, this PR extracts the generic implementation into a separate module.


cahirwpz:
I like this change. There're some parts in machine dependant code that looks the same for each architecture. I think that extracting the code into `gen` directory is a good move. Let's apply DRY (don't repeat yourself) principle whenever there's a possibility to do so.

cahirwpz (approved):

PR #1245, opened 2022-05-09
GPIO interface

mohrcore:
Interface for GPIO drivers to use.
The motivation is to allow different drivers to share GPIO resources.


cahirwpz:
Closing since the PR was superseded by #1277

PR #1246, opened 2022-05-12
[RISCV] Exception entry.

MichalBlk:
This PR provides exception entry points for the RISC-V target architecture.


sys/riscv/exception.S
```diff
@@ -1,13 +1,300 @@
...
+.endm
+
+.macro load_reg reg, offset
+	REG_L	\reg, (\offset)(sp)
+.endm
+
+.macro load_csr csr, offset, tmp
+	load_reg \tmp, \offset
+	csrw	\csr, \tmp
+.endm
```
pj1031999, line 27: It is used also in `switch.S`. Could you share this code, we are already using cpp for asm files so it should be simple.

sys/riscv/exception.S
```diff
@@ -1,13 +1,300 @@
...
+
+	save_reg_cfi t0, CTX_T0
+	save_reg_cfi t1, CTX_T1
+	save_reg_cfi t2, CTX_T2
+	save_reg_cfi t3, CTX_T3
+	save_reg_cfi t4, CTX_T4
+	save_reg_cfi t5, CTX_T5
+	/* XXX: we don't provide the CFI directive for $t6 as
+	 * GDB from toolchain 1.5.3 has faulty frame info implementation
+	 * which confuses this register with PC. */
```
pj1031999, line 59: What with latest gdb?

sys/riscv/exception.S
MichalBlk: I moved common macros to `include/riscv/asm.h`.

sys/riscv/exception.S
MichalBlk: I tired it with the following toolchain: (GCC 12.1.0, Binutils 2.38, GDB 12.1). The same issue still occurs. I updated the comment to explicitly specify what configuration was used.

cahirwpz (approved):
Please remove unnecessary references to `FPU`.

sys/riscv/exception.S
pj1031999: For my curiosity - could you check llvm build with lldb?

MichalBlk:
I know that this is irrelevant, but I don't want to open a separate PR for this. I removed rootdev driver from the `root` devclass (that was a workaround to enable kernel compilation, however with the arrival of RISC-V drivers this is no longer needed and must be removed).

PR #1249, opened 2022-05-13
[RISCV] Change ISA to IMA.

MichalBlk:
FTTB, the only supported hardware platform for the RISC-V 32 target architecture is LiteX VexRiscv. Our configuration of the VexRiscv core doesn't support any floating-point extension which makes it impossible to compile the system using the current toolchain as we rely on libgcc which expects hard floats.

Besides, more recent releases of the gnu tools distinguish zicsr and zifence extensions from the i extension.


cahirwpz:
This change will be merged into #1257

PR #1252, opened 2022-05-13
[cfg] Introduce TRAP_USER_ACCESS configuration flag.

MichalBlk:
Some architectures provide means to trap whenever kernel mode software attempts to access user virtual address space. Although the feature is useful, when enabled, it makes it impossible for GDB to produce the full backtrace for exceptions encountered in userspace. Thereby, its usage should be conditional and controlled by a configuration switch.
An example of such architecture is RISC-V which provides the SUM bit (enable Supervisor User Memory access).


MichalBlk:
The RISC-V-specific `SUM` flag has been changed to a system-wide `TRAP_USER_ACCESS` flag.

cahirwpz (approved):

PR #1256, opened 2022-05-20
[RISC-V] Physical address map implementation

MichalBlk:
Architecture-dependent implementation of the pmap module.


sys/riscv/pmap.c
```diff
@@ -1,48 +1,122 @@
...
 
-void pmap_md_setup(pmap_t *pmap) {
-  panic("Not implemented!");
+void pmap_md_activate(pmap_t *umap) {
+  /* Update kernel page directory entries within the pmap. */
+  pmap_t *kmap = pmap_kernel();
+  size_t halfpage = PAGESIZE / 2;
+  WITH_MTX_LOCK (&kmap->mtx) {
+    memcpy((void *)phys_to_dmap(umap->pde) + halfpage,
+           (void *)phys_to_dmap(kmap->pde) + halfpage, halfpage);
```
pj1031999, line 102: `phys_to_dmap` returns `void *` already.

pj1031999 (approved):
LGTM

sys/riscv/pmap.c
```diff
@@ -1,48 +1,122 @@
...
  */
 
-void pmap_md_setup(pmap_t *pmap) {
-  panic("Not implemented!");
+void pmap_md_activate(pmap_t *umap) {
+  /* Update kernel page directory entries within the pmap. */
+  pmap_t *kmap = pmap_kernel();
+  size_t halfpage = PAGESIZE / 2;
+  WITH_MTX_LOCK (&kmap->mtx) {
+    memcpy((void *)phys_to_dmap(umap->pde) + halfpage,
```
cahirwpz, line 101: While I suggested lazy solution for the problem, I'd like not to copy half of page directory each time we switch address spaces. Perhaps we could add generation counter for each `pmap`. Generation for a pmap would increase if somebody called `pmap_growkernel`. When we switch to another pmap and active pmap's generation counter is more recent than ours, we would do the copy. @MichalBlk @pj1031999 What do you think?

pj1031999: Is it observable performance issue? We have other critical paths in kernel that are slow and inefficient. As long as nobody will run fully profiled kernel under heavy multithread load I don't really care about single function. How much instruction it is? 256 moves? I guess that mutex lock in previous line is more expensive.

Anyway, I opt for leaving it as is now, merge support for RISC-V and then focus on fixing existing bad behavior issues. Next we should invest in performance issue since our human resources are super limited ;)

sys/riscv/pmap.c
MichalBlk: Good point. Done.

cahirwpz (approved):
Consider leaving a comment about possible optimization for `pmap_md_activate` or implement it if you wish.

sys/riscv/pmap.c
cahirwpz: @MichalBlk accepted the challenge and implemented it ;-)

PR #1259, opened 2022-05-20
[dts] LiteX VexRiscv

MichalBlk:
Device tree source for our LiteX VexRiscv hardware platform.


sys/dts/litex-riscv.dts
```diff
@@ -0,0 +1,124 @@
...
+/dts-v1/;
+
+/ {
+        #address-cells = <1>;
+        #size-cells = <1>;
+
+
+        chosen {
+            linux,initrd-start = <0x42000000>;
+            linux,initrd-end = <0x45000000>;
```
cahirwpz, line 11: This should depend on the actual size of ramdisk attached by Renode.

sys/dts/litex-riscv.dts
```diff
@@ -0,0 +1,124 @@
...
+            #size-cells = <1>;
+            ranges;
+
+            opensbi@40f00000 {
+                reg = <0x40f00000 0x80000>;
+            };
+        };
+
+        clocks {
+            sys_clk: litex_sys_clk {
```
cahirwpz, line 67: What is the function of this clock? How could it be used by other subsystems?

sys/dts/litex-riscv.dts
```diff
@@ -0,0 +1,124 @@
...
+            };
+
+            liteuart0: serial@f0001000 {
+                compatible = "litex,liteuart";
+                reg = <0xf0001000 0x100>;
+                interrupts = <1>;
+                status = "okay";
+            };
+
+            gpio@f0002800 {
```
cahirwpz, line 110: We don't have any driver for that. Could you link somewhere the list of peripherals available for Litex SoC?

sys/dts/litex-riscv.dts
MichalBlk: Although you're right that the size of the ram disk should be determined automatically, obtaining the value from RENODE's platform description may not be the best idea. We agreed that the size should be determined at launch time using device tree overlays (similar to #1258).

sys/dts/litex-riscv.dts
MichalBlk: FTTB, we don't have any explicit use cases of the clocks described by the `clocks` node. For now, the node will be removed.

sys/dts/litex-riscv.dts
MichalBlk: Unfortunately, it is not so easy to list all the peripherals. We have two basic device types when it comes to the LiteX VexRiscv hardware platform:
- peripherals provided by the FPGA board which implements the softcore
- devices implemented by LiteX
Although devices located on the FPGA board are usually well described by the vendor, when it comes to LiteX we don't have any list of implemented cores. Moreover, some HW components are hosted by the main LiteX repository, whereas others are maintained separately (e.g. https://github.com/enjoy-digital/litepcie).
Perhaps we should mention it in some documentation section.

cahirwpz (approved):
LGTM

PR #1261, opened 2022-05-23
[drv] Fix simplebus.

MichalBlk:
- It turns out that it's not that rare for a device to generate more 8 interrupts (e.g. plic and clint in HiFive Unleashed)
- Fix tuples allocation for interrupt cells
- Fix validation logic
- Fix tuple copying


cahirwpz (approved):
LGTM

PR #1264, opened 2022-05-26
[KASAN] Extract common MD code.

MichalBlk:
It turns out that architecture-dependent code consists of only two defines: start of the shadow memory and size of the sanitized address. The remaining values are computed using the aforementioned macros.

Besides, this PR cleans up the code. This includes removing the redundant `kasan_va_to_shadow` function that has exactly the same goal as `addr_to_shad` (previously `kasan_md_addr_to_shad`), i.e. compute the address of the shadow byte describing the provided sanitized memory area.


sys/kern/kasan.c
```diff
@@ -222,17 +236,12 @@ static void call_ctors(void) {
   }
 }
 
-static inline vaddr_t kasan_va_to_shadow(vaddr_t va) {
-  return KASAN_MD_SHADOW_START +
-         (va - KASAN_MD_SANITIZED_START) / KASAN_SHADOW_SCALE_SIZE;
-}
```
pj1031999, line 228: It's not obvious why it chan be replaced by `addr_to_shad` if you look into that first time. Could you add a comment to commit message with explanation?

pj1031999 (approved):

sys/kern/kasan.c
```diff
@@ -299,7 +308,6 @@ void __asan_unregister_globals(struct __asan_global *globals, size_t n) {
   /* never called */
 }
 
-/* Note: alloca is currently used in strntoul and test_sleepq_sync functions */
```
jpszczolowski, line 302: Is that no longer true?

jpszczolowski (approved):

sys/kern/kasan.c
MichalBlk: Yes, this message is obsolete.

jpszczolowski: Just FYI, this wasn’t about us explicitly using alloca in the implementation but about the compiler using alloca while compiling those functions

sys/kern/kasan.c
MichalBlk: Sure. I extended the description of the PR to explain the removal of `kasan_va_to_shadow`. I'll also tell @cahirwpz to mention it in the comment message while doing squash and merge.

PR #1267, opened 2022-06-03
[trap]  Extract common page fault handler

MichalBlk:
Every supported architecture uses exactly the same page fault handler logic, thereby the code should be extracted to a generic machine-dependent module.


cahirwpz (approved):
LGTM

cahirwpz:
@pj1031999 We're waiting for you review ;-)

sys/aarch64/trap.c
```diff
@@ -47,52 +42,13 @@ static void syscall_handler(register_t code, ctx_t *ctx,
...
-  if (td->td_onfault) {
-    _REG(ctx, PC) = td->td_onfault;
-    td->td_onfault = 0;
-  } else if (usermode) {
-    /* Send a segmentation fault signal to the user program. */
-    sig_trap(ctx, SIGSEGV);
-  } else {
-    kernel_oops(ctx);
-  }
+  return access;
```
pj1031999, line 51: I prefer brackets for more complex control flow like here.

sys/gen/pmap.c
```diff
@@ -442,6 +441,41 @@ int pmap_emulate_bits(pmap_t *pmap, vaddr_t va, vm_prot_t prot) {
...
+  assert(pmap);
+
+  if (!(error = pmap_emulate_bits(pmap, vaddr, access)))
+    return 0;
+
+  if (error == EACCES)
+    goto fault;
+
+  vm_map_t *vmap = vm_map_user();
+  assert(vmap);
```
pj1031999, line 461: It's already asserted in `vm_map_user` body.

sys/riscv/trap.c
```diff
@@ -195,6 +140,14 @@ static bool fpu_handler(mcontext_t *uctx) {
   return true;
 }
 
+static vm_prot_t exc_access(u_long exc_code) {
+  if (exc_code == SCAUSE_INST_PAGE_FAULT)
+    return VM_PROT_EXEC;
+  else if (exc_code == SCAUSE_LOAD_PAGE_FAULT)
+    return VM_PROT_READ;
+  return VM_PROT_WRITE;
```
pj1031999, line 148: `switch` statement is more readable imho.

sys/aarch64/trap.c
```diff
@@ -153,15 +111,19 @@ void user_trap_handler(mcontext_t *uctx) {
...
   if ((_REG(ctx, SPSR) & PSR_I) == 0)
     cpu_intr_enable();
 
-  switch (ESR_ELx_EXCEPTION(esr)) {
+  switch (exc_code) {
     case EXCP_INSN_ABORT:
     case EXCP_DATA_ABORT:
-      abort_handler(ctx, esr, far, false);
+      klog("%x at $%lx, caused by reference to $%lx!", exc_code, _REG(ctx, PC),
+           far);
```
pj1031999, line 124: This log is common for all architectures - maybe it's good idea to move that to `pmap_fault_handler`? WDYT?

pj1031999 (approved):

cahirwpz:
@MichalBlk Please resolve conflicts!

sys/aarch64/trap.c
MichalBlk: Done.

sys/gen/pmap.c
MichalBlk: Right. This line is obsolete. Done.

sys/riscv/trap.c
MichalBlk: I agree. `switch` is less verbose in this case.

sys/aarch64/trap.c
MichalBlk: I thought about this. The problem is that on RISC-V and MIPS we print an exception string instead of a bare exception code (`exc_code`). Each architecture would have to provide a function like `const char *exc_str(u_long exc_code)` and if the function was missing, a bare exception code would be printed. I discussed this solution with @cahirwpz and we came to conclusion that leaving the printing to the MD module is a better solution.

PR #1268, opened 2022-06-03
[vm] Remove kernel vm_map since all kernel mappings are wired

MichalBlk:
With the arrival of #1243, the kernel can only use wired memory mappings. This makes maintaining  `kspace` needles and the generic `vm_map` logic can be simplified to consider only user virtual memory maps.

### DIFF
--- a/sys/aarch64/trap.c
+++ b/sys/aarch64/trap.c
@@ -76,7 +76,6 @@ static void abort_handler(ctx_t *ctx, register_t esr, vaddr_t vaddr,
     goto fault;
 
   vm_map_t *vmap = vm_map_user();
-  assert(vmap);
 
   if (vm_page_fault(vmap, vaddr, access) == 0)
     return;
--- a/sys/aarch64/trap.c
+++ b/sys/aarch64/trap.c
@@ -75,7 +75,8 @@ static void abort_handler(ctx_t *ctx, register_t esr, vaddr_t vaddr,
   if (error == EACCES || error == EINVAL)
     goto fault;
 
-  vm_map_t *vmap = vm_map_cur();
+  vm_map_t *vmap = vm_map_user();
+  assert(vmap);
 
   if (vm_page_fault(vmap, vaddr, access) == 0)
     return;
--- a/sys/aarch64/trap.c
+++ b/sys/aarch64/trap.c
@@ -75,11 +75,8 @@ static void abort_handler(ctx_t *ctx, register_t esr, vaddr_t vaddr,
   if (error == EACCES || error == EINVAL)
     goto fault;
 
-  vm_map_t *vmap = vm_map_lookup(vaddr);
-  if (!vmap) {
-    klog("No virtual address space defined for %lx!", vaddr);
-    goto fault;
-  }
+  vm_map_t *vmap = vm_map_cur();
+
   if (vm_page_fault(vmap, vaddr, access) == 0)
     return;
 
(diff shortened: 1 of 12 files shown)


sys/aarch64/trap.c
```diff
@@ -75,11 +75,9 @@ static void abort_handler(ctx_t *ctx, register_t esr, vaddr_t vaddr,
   if (error == EACCES || error == EINVAL)
     goto fault;
 
-  vm_map_t *vmap = vm_map_lookup(vaddr);
-  if (!vmap) {
-    klog("No virtual address space defined for %lx!", vaddr);
-    goto fault;
-  }
+  vm_map_t *vmap = vm_map_user();
+  assert(vmap);
```
pj1031999, line 79: Now every invocation of `vm_map_user` expects non-null return value. Let's move assert to `vm_map_user`.

sys/kern/vm_map.c
```diff
@@ -85,32 +70,15 @@ inline vm_map_entry_t *vm_map_entry_next(vm_map_entry_t *ent) {
...
-
-vm_map_t *vm_map_lookup(vaddr_t addr) {
-  if (vm_map_address_p(vm_map_user(), addr))
-    return vm_map_user();
-  if (vm_map_address_p(vm_map_kernel(), addr))
-    return vm_map_kernel();
-  return NULL;
+static bool vm_map_contains_p(vaddr_t start, vaddr_t end) {
+  return USER_SPACE_BEGIN <= start && end <= USER_SPACE_END;
 }
```
pj1031999, line 75: I'm not sure if name of this function should be prefixed by `vm_map_` since it check if given range belongs to userspace. Maybe something like `userspace_p(...)?

pj1031999, line 75: I'm not sure if name of this function should be prefixed by `vm_map_` since it check if given range belongs to userspace. Maybe something like `userspace_p(...)?

sys/kern/vm_map.c
```diff
@@ -237,8 +205,8 @@ static int vm_map_findspace_nolock(vm_map_t *map, vaddr_t /*inout*/ *start_p,
   assert(page_aligned_p(start) && page_aligned_p(length));
 
   /* Bounds check */
-  start = max(start, vm_map_start(map));
-  if (start + length > vm_map_end(map))
+  start = max(start, (vaddr_t)USER_SPACE_BEGIN);
```
pj1031999, line 208: If gcc is annoying here you can change definition of `USER_SPACE_BEGIN` to `((vaddr_t) 0xmagic)` since it's always used in `vaddr_t` context.

pj1031999, line 208: If gcc is annoying here you can change definition of `USER_SPACE_BEGIN` to `((vaddr_t) 0xmagic)` since it's always used in `vaddr_t` context.

pj1031999: The same for `*_END`.

pj1031999 (approved):
LGTM

sys/kern/vm_map.c
MichalBlk: I'm wondering if it should be located in `sys/gen/pmap.c` as it already contains `user_addr_p` which used to be global. FTTB, I changed `vm_map_contains_p` to `userspace_p` as suggested.

sys/kern/vm_map.c
cahirwpz: Sounds ok to me.

cahirwpz: Let's do it later when we add documentation for kernel virtual address space layout.

PR #1269, opened 2022-06-03
[RISCV] Make the size of the boot memory allocator area dynamic.

MichalBlk:
The assumption of a constant size boot memory allocator area should be dropped as with KASAN is becomes tedious to maintain four different configurations (32/64 + KASAN on / KASAN off) that must be adjusted whenever the area size has to change.


pj1031999 (approved):

cahirwpz (approved):
LGTM

PR #1271, opened 2022-06-05
[dev] Support for the SOC ranges fdt property

MichalBlk:
Some buses utilize the `ranges` property to express mappings between bus-relative addresses and parent-relative addresses. This change adds support for the `ranges` property within the `soc` device node.


cahirwpz (approved):
LGTM

PR #1272, opened 2022-06-05
[aarch64][dev] Simplebus support

MichalBlk:
The BCM2835 driver should employ the simplebus add child feature instead of using obsolete manual resource assignment.


sys/drv/bcm2835_pic.c
```diff
@@ -0,0 +1,204 @@
...
+
+static void bcm2835_pic_disable_irq(intr_event_t *ie) {
+  bcm2835_pic_state_t *bcm2835_pic = ie->ie_source;
+  unsigned irq = ie->ie_irq;
+
+  bus_size_t reg = BCM2835_INTC_IRQBDISABLE;
+  if (irq < BCM2835_INT_GPU1BASE)
+    reg = BCM2835_INTC_IRQ1DISABLE;
+  else if (irq < BCM2835_INT_BASICBASE)
+    reg = BCM2835_INTC_IRQ2DISABLE;
```
mohrcore, line 46: Minor detail, I just found the ordering of these ifs to be confusing.
```suggestion
  bus_size_t reg = 0;
  if (irq < BCM2835_INT_GPU1BASE)
    reg = BCM2835_INTC_IRQ1DISABLE;
  else if (irq < BCM2835_INT_BASICBASE)
    reg = BCM2835_INTC_IRQ2DISABLE;
  else
    reg = BCM2835_INTC_IRQBDISABLE;
```

sys/drv/bcm2835_pic.c
```diff
@@ -0,0 +1,204 @@
...
+  unsigned irq = ie->ie_irq;
+
+  bus_size_t reg = BCM2835_INTC_IRQBENABLE;
+  if (irq < BCM2835_INT_GPU1BASE)
+    reg = BCM2835_INTC_IRQ1ENABLE;
+  else if (irq < BCM2835_INT_BASICBASE)
+    reg = BCM2835_INTC_IRQ2ENABLE;
+
+  uint32_t en = in4(reg);
+  out4(reg, en | (1 << irq));
```
pj1031999, line 64: Not for this PR but for future:
It's time to do bitops in right way - once for whole codebase. 
In lib/libc/gen/fts.c we have following macros: `CLR`, `ISSET`, `SET`. I would opt for creating functions / macros in mimiker.h for that purpose and use them everywhere in kernel code.

sys/drv/bcm2835_pic.c
pj1031999: I considered multiple solutions but all of them are equal to current code. It's short, clean, easy to understand. You can add brackets for better readability.

I don't like @mohrcore solution because it's more confusing to reader. You add another branch. `reg` is explicitly initialized for 0 without any reason - it's another level of confusion. As reviewer I can think that you forgot about something and your code is invalid because of that additional initialization.

sys/drv/bcm2835_pic.c
```diff
@@ -0,0 +1,204 @@
...
+
+static void bcm2835_pic_enable_irq(intr_event_t *ie) {
+  bcm2835_pic_state_t *bcm2835_pic = ie->ie_source;
+  unsigned irq = ie->ie_irq;
+
+  bus_size_t reg = BCM2835_INTC_IRQBENABLE;
+  if (irq < BCM2835_INT_GPU1BASE)
+    reg = BCM2835_INTC_IRQ1ENABLE;
+  else if (irq < BCM2835_INT_BASICBASE)
+    reg = BCM2835_INTC_IRQ2ENABLE;
```
pj1031999, line 61: It's the same code (I mean logic) as before. If you want you can consider adding helper function irq -> reg.

sys/drv/bcm2835_pic.c
```diff
@@ -0,0 +1,204 @@
...
+
+static void bcm2835_pic_teardown_intr(device_t *pic, device_t *dev,
+                                      resource_t *irq) {
+  intr_event_remove_handler(irq->r_handler);
+}
+
+static int bcm2835_pic_map_intr(device_t *pic, device_t *dev, phandle_t *intr,
+                                int icells) {
+  if (icells != 2)
+    return -1;
```
pj1031999, line 112: Magic number requires comment ;)

sys/drv/bcm2835_pic.c
```diff
@@ -0,0 +1,204 @@
...
+  if (type >= BCM2835_INTRTYPE_CNT)
+    return -1;
+  if (irq_off >= BCM2835_NIRQPERTYPE)
+    return -1;
+
+  unsigned irq = irq_off;
+  if (type == BCM2835_INTRTYPE_GPU1)
+    irq += BCM2835_NIRQPERTYPE;
+  else if (type == BCM2835_INTRTYPE_BASIC)
+    irq += BCM2835_NIRQPERTYPE * 2;
```
pj1031999, line 127: ditto

sys/drv/bcm2835_pic.c
```diff
@@ -0,0 +1,204 @@
...
+}
+
+static int bcm2835_pic_attach(device_t *pic) {
+  bcm2835_pic_state_t *bcm2835_pic = pic->state;
+
+  rman_init(&bcm2835_pic->rm, "BCM2835 PIC interrupt sources");
+  rman_manage_region(&bcm2835_pic->rm, 0, BCM2835_NIRQ);
+
+  bcm2835_pic->mem = device_take_memory(pic, 0, RF_ACTIVE);
+  assert(bcm2835_pic->mem);
```
pj1031999, line 173: For future - we don't handle errors of this function and always require that return value is valid. Maybe we should explicitly said that this function never fail and move assert to body.

sys/drv/bcm2835_rootdev.c
```diff
@@ -1,137 +1,45 @@
...
-
-static void enable_local_irq(int irq) {
-  assert(irq < BCM2836_INT_NLOCAL);
-  uint32_t reg = bus_space_read_4(rootdev_bus_space, rootdev_local_handle,
-                                  BCM2836_LOCAL_TIMER_IRQ_CONTROLN(0));
-  bus_space_write_4(rootdev_bus_space, rootdev_local_handle,
-                    BCM2836_LOCAL_TIMER_IRQ_CONTROLN(0), reg | (1 << irq));
-}
+#define in4(addr) bus_read_4(&rd->mem_local, (addr))
+#define out4(addr, val) bus_write_4(&rd->mem_local, (addr), (val))
```
pj1031999, line 25: Using local variables in macros is easy way to fuck-up something.

sys/drv/bcm2835_rootdev.c
```diff
@@ -1,137 +1,45 @@
...
-    enable_gpu_irq(irq - BCM2835_INT_GPU0BASE, BCM2835_INTC_IRQ1ENABLE);
-  } else if (irq < BCM2835_INT_BASICBASE) {
-    /* Enable GPU1 IRQ. */
-    enable_gpu_irq(irq - BCM2835_INT_GPU1BASE, BCM2835_INTC_IRQ2ENABLE);
-  } else {
-    /* Enable base IRQ. */
-    enable_gpu_irq(irq - BCM2835_INT_BASICBASE, BCM2835_INTC_IRQBENABLE);
-  }
+  uint32_t irqctrl = in4(BCM2836_LOCAL_TIMER_IRQ_CONTROLN(0));
+  out4(BCM2836_LOCAL_TIMER_IRQ_CONTROLN(0), irqctrl | (1 << irq));
```
pj1031999, line 33: Add comment:
> TODO: fix needed for SMP.

sys/drv/bcm2835_rootdev.c
```diff
@@ -165,43 +77,28 @@ static void rootdev_teardown_intr(device_t *pic, device_t *dev,
...
+  intr_event_t **events = rd->intr_event;
+  uint32_t pending = in4(BCM2836_LOCAL_INTC_IRQPENDINGN(0));
 
   while (pending) {
-    int irq = ffs(pending) - 1;
-    /* XXX: some pending bits are shared between BASIC and GPU0/1. */
+    unsigned irq = ffs(pending) - 1;
     if (events[irq])
       intr_event_run_handlers(events[irq]);
     pending &= ~(1 << irq);
```
pj1031999, line 89: IIRC comment about shared bits is important.

sys/drv/bcm2835_rootdev.c
```diff
@@ -213,52 +110,56 @@ DEVCLASS_DECLARE(emmc);
...
+         simplebus_add_child(bus, "/soc/serial", unit++, bcm2835_pic, NULL)))
+    return err;
 
-  /* EMMC */
-  dev = device_add_child(bus, 3);
-  dev->pic = bus;
-  dev->devclass = &DEVCLASS(emmc);
+  if ((err =
+         simplebus_add_child(bus, "/soc/sdhci", unit++, bcm2835_pic, &sdhci)))
+    return err;
```
pj1031999, line 156: Ok but be consistent. Previously in this function you have pattern:
```
error = foo(...);
if (error)
  return error_code;
```
here you have
```
if ((error = foo(...)))
  return error_code;
```
both of them are readable, but please be consistent.
@cahirwpz we need clean rule for what is allowed.

sys/drv/bcm2835_rootdev.c
```diff
@@ -213,52 +110,56 @@ DEVCLASS_DECLARE(emmc);
...
+  sdhci->devclass = &DEVCLASS(emmc);
   /* Due to the lack of proper GPIO routines, the driver uses a workaround that
    * requires the GPIO memory to be passed to it as a resource */
   /* TODO: Mak a proper GPIO interface and driver */
```
pj1031999, line 160: Could you fix that typo? :) You already modify this part of file.

sys/kern/bus.c
```diff
@@ -7,7 +7,12 @@
...
                    bus_space_handle_t *handle_p) {
-  *handle_p = kmem_map_contig(addr, size, PMAP_NOCACHE);
+  bus_addr_t pg_start = rounddown2(addr, PAGESIZE);
+  bus_addr_t pg_end = roundup2(addr + size, PAGESIZE);
+  bus_size_t off = addr & (PAGESIZE - 1);
+
+  vaddr_t va = kmem_map_contig(pg_start, pg_end - pg_start, PMAP_NOCACHE);
+  *handle_p = va + off;
   return 0;
 }
```
pj1031999, line 17: Finally somebody did that!

pj1031999 (approved):
Good job!

A few comments from my side. Nothing related to code logic.

mohrcore (approved):
LGTM

sys/drv/bcm2835_pic.c
MichalBlk: Great idea. We have a number of use cases throughout device drivers and each of them implements the operations on its own (in exactly the same way).

sys/drv/bcm2835_pic.c
MichalBlk: I'm not sure. @cahirwpz what do you think?

sys/drv/bcm2835_rootdev.c
MichalBlk: I agree. It's error-prone and looks pretty weird. However, it's local to this file and it's common for our drivers to use this convention (which is granted by @cahirwpz). @cahirwpz do we want to drop this convention?

sys/drv/bcm2835_pic.c
MichalBlk: I could introduce `intrtype_to_reg(intrtype, regtype)` where `regtype` is an enum (`enable`, `disable`, `pending`). On the other hand, FTTB, I only need to translate `intrtype` once to enable register and once to disable register so maybe that is too generic (and requires a new type definition). @cahirwpz what do you think?

sys/drv/bcm2835_pic.c
MichalBlk: Maybe we should introduce a define for this purpose, e.g. `PIC_INVALIRQ`. What do you think?

sys/drv/bcm2835_rootdev.c
MichalBlk: Yes but it regards the BCM2835 peripheral interrupt controller, not the BCM2836 interrupt controller. The comment has been moved to `bcm2835_intr.c`.

sys/drv/bcm2835_rootdev.c
MichalBlk: The first one (in line 132) is different because `err` is undefined at that point. But I agree. It's inconsistent and we need a clear convention for such constructs.

sys/drv/bcm2835_rootdev.c
MichalBlk: Done.

sys/drv/bcm2835_pic.c
MichalBlk: I added a comment to explain what happens here and where is the description of the layout.

sys/kern/bus.c
MichalBlk: I'm glad too :).

sys/dts/rpi3.dts
```diff
@@ -1,16 +1,71 @@
...
+			compatible = "arm,pl011", "arm,primecell";
+			reg = <0x7e201000 0x200>;
+			interrupts = <2 25>;
+		};
+
+		sdhci: sdhci@7e300000 {
+			compatible = "brcm,bcm2835-sdhci";
+			reg = <0x7e300000 0x100>;
+			interrupts = <2 30>;
+		};
```
mohrcore, line 61: One little caveat. In mimiker, we threat the sd controller as a generic emmc controller that provides a bus interface. Then we call methods from that interface using a child device that implements SD protocol. Thus I believe this should be renamed to `emmc` and `sdhci` should be used for a child node of `emmc`.
```suggestion
		emmc: emmc@7e300000 {
			compatible = "brcm,bcm2835-emmc";
			reg = <0x7e300000 0x100>;
			interrupts = <2 30>;
		};
```

mohrcore: Also, you forgot to update the probe procedure of the emmc driver.

cahirwpz: Your remark is valid. Please put it into a separate PR together with better `probe` function that references FDT.

sys/drv/bcm2835_pic.c
cahirwpz: Let's not complicate. This code is readable enough for us to track. Let's leave it simple.

sys/drv/bcm2835_pic.c
cahirwpz: I agree. Since device register access is a common use case we should provide easy to use set of macros to access them. But let's merge this PR quickly since already #1277 depends on it.

sys/drv/bcm2835_pic.c
cahirwpz: Return value of -1 is documented in: https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/include/sys/interrupt.h?r=157186b5#170 There're no other errors this function may return.

PR #1276, opened 2022-06-13
[RISCV] Fix copy routines.

MichalBlk:
- the `onfault` pointer is not cleared appropriately
- the 0 length case is broken as we jump in the middle of the `copycommon` block


cahirwpz (approved):
LGTM

PR #1279, opened 2022-06-20
[initrd] Access initrd using dmap

MichalBlk:
We don't have to map the large initrd to kernel virtual memory as the entire archive can be accessed using dmap.


cahirwpz (approved):
LGTM

PR #1280, opened 2022-06-20
[physmem] Access vm_page_t structs using dmap.

MichalBlk:
Currently, the target architecture must provide kernel page directories for the `vm_page_t` structs during the boot phase as physmem initialization cannot rely on the buddy system to allocate physical pages. This problem can be solved by relying on dmap to access the `vm_page_t` structs.

### DIFF
--- a/sys/kern/vm_physmem.c
+++ b/sys/kern/vm_physmem.c
@@ -71,7 +71,6 @@ static void *vm_boot_alloc(size_t n) {
 
   seg->start += n;
   seg->npages -= n / PAGESIZE;
-  assert(seg->npages);
 
   return va;
 }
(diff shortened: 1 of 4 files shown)


sys/kern/vm_physmem.c
```diff
@@ -77,20 +67,13 @@ static void *vm_boot_alloc(size_t n) {
...
-    }
-
-    pmap_kenter((vaddr_t)va, pa, VM_PROT_READ | VM_PROT_WRITE, 0);
-  }
+  void *va = phys_to_dmap(seg->start);
 
-  vm_kernel_end += n;
+  seg->start += n;
+  seg->npages -= n / PAGESIZE;
+  assert(seg->npages);
```
pj1031999, line 74: Could you explain this assertion?

MichalBlk: This assert is needles as the condition is ensured by the condition of the preceding `while` loop. I removed it.

PR #1284, opened 2022-06-20
emmc: Maintain internal error state and nop when not sane

mohrcore:
Used to simplify SD card driver code


include/dev/emmc.h
```diff
@@ -140,9 +140,20 @@ typedef struct emmc_cmd {
 #define EMMC_BUSWIDTH_4 0x02 /* 4-bit bus */
 #define EMMC_BUSWIDTH_8 0x04 /* 8-bit but */
 
-typedef uint64_t emmc_error_t;
-
-#define EMMC_ERROR_CMD_TIMEOUT 0x01
+typedef enum {
+  EMMC_ERROR_TIMEOUT = 0x01,          /* Controller timed-out on a request */
+  EMMC_ERROR_PROP_NOTSUP = 0x02,      /* Requested property is not supported*/
```
MichalBlk, line 145: Missing space before terminating `*/`.

sys/drv/bcm2835_emmc.c
```diff
@@ -28,8 +28,10 @@ typedef struct bcmemmc_state {
   spin_t lock;           /* Covers `pending`, `intr_recv` and `emmc`. */
   uint64_t rca;          /* Relative Card Address */
   uint64_t host_version; /* Host specification version */
-  volatile uint32_t pending; /* All interrupts received */
-  emmc_error_t errors;
+  volatile uint32_t pending;  /* All interrupts received */
+  emmc_error_t errors;        /* Error flags */
+  emmc_error_t ignore_errors; /* Error flags that do not cause invalidation of
```
MichalBlk, line 33: `ignore_errors` sounds like a flag that (globally) disables error handling. Maybe `ignored_errors`?

sys/drv/bcm2835_emmc.c
```diff
@@ -65,11 +67,25 @@ static emmc_error_t bcemmc_decode_errors(uint32_t interrupts) {
...
 
+static inline int bcmemmc_invalid_state(bcmemmc_state_t *state) {
+  return state->errors & (~state->ignore_errors | EMMC_ERROR_INTERNAL |
+                          EMMC_ERROR_INVALID_STATE);
+}
+
+static inline emmc_error_t bcmemmc_set_error(bcmemmc_state_t *state,
+                                             emmc_error_t error_flags) {
+  state->errors = error_flags;
+  if ((error_flags & !state->ignore_errors) && !bcmemmc_invalid_state(state)) {
```
MichalBlk, line 83: It doesn't make sense to me. First, I think you wanted to do bitwise negation instead of logical negation in the first expression (i.e. `!state->ignore_errors` -> `~state->ignore_errors`). If my suggestion is valid then the second expression is useless. It is equivalent to the following:
`(state->errors & ~state->ignore_errors) && !(state->errors & (~state->ignore_errors | <something>))`
which can be written as
`(error & ~ignore) && !((error & ~ignore) | (error & <something>))`
further
`(error & ~ignore) && !(error & ~ignore) && !(error & <something>)`
this yields
`false`

sys/drv/bcm2835_emmc.c
```diff
@@ -113,21 +129,24 @@ static inline int bcmemmc_check_intr(bcmemmc_state_t *state) {
...
  * \warning This procedure may put the thread to sleep.
  */
-static int32_t bcmemmc_intr_wait(device_t *dev, uint32_t mask) {
+static emmc_error_t bcmemmc_intr_wait(device_t *dev, uint32_t mask) {
   bcmemmc_state_t *state = (bcmemmc_state_t *)dev->state;
 
   assert(mask != 0);
 
+  if (bcmemmc_invalid_state(state))
+    return bcmemmc_set_error(state, EMMC_ERROR_INVALID_STATE);
```
MichalBlk, line 138: (Feel free to ignore.) I find the description of the API presented in `emmc.h` a little confusing. It describes the return value as if only a single code could be returned at once (i.e. as if error codes could be encoded using binary numbers instead of one hot). I think it would be helpful to mention in the header that the return code is actually a mask of error codes.

sys/drv/bcm2835_emmc.c
```diff
@@ -389,26 +419,36 @@ static int bcmemmc_set_prop(device_t *cdev, uint32_t id, uint64_t var) {
     case EMMC_PROP_RW_RCA:
       state->rca = var;
       break;
+    case EMMC_PROP_RW_ERRORS:
+      /* The only way to reset nternal error is to reset the entire controller
```
MichalBlk, line 423: Typo `nternal`

sys/drv/bcm2835_emmc.c
```diff
@@ -389,26 +419,36 @@ static int bcmemmc_set_prop(device_t *cdev, uint32_t id, uint64_t var) {
...
+       * using the `reset` method. */
+      state->errors = var & ~EMMC_ERROR_INTERNAL;
```
MichalBlk, line 425: Shouldn't be `state->errors = var | (state->errors & EMMC_ERROR_INTERNAL)`?

MichalBlk (changes requested):

sys/drv/bcm2835_emmc.c
mohrcore: No.
This is meant to be used to clear error flags. The exception is the `EMMC_ERROR_INTERNAL` flag, which can't be cleared by setting `ERRORS` property. The only way to clear it is to reset the controller.

sys/drv/bcm2835_emmc.c
mohrcore: Fixed and simplified this logic.

sys/drv/bcm2835_emmc.c
mohrcore: Fixed.

sys/drv/bcm2835_emmc.c
mohrcore: I've addede an extra note in emmc.h that explains it.

sys/drv/bcm2835_emmc.c
mohrcore: Good suggestion. I changed the field name to `ignored_errors`.

include/dev/emmc.h
mohrcore: Fixed.

sys/drv/bcm2835_emmc.c
MichalBlk: Sorry if it's obvious and I'm missing it. If I understand correctly, `var` is provided by the user, so e.g. if `var = 0` and `state->errors = EMMC_ERROR_INTERNAL`, then after the assignment `state->errors = 0`. Where is the problem in this reasoning?

mohrcore: Ok, thanks for pointing that out, that was indeed a mistake on my side. I fixed it:
```c
state->errors = var | (state->errors & EMMC_ERROR_INTERNAL);
```

PR #1285, opened 2022-06-23
[RISCV] Remove faulty CSR immediate macros.

MichalBlk:
It turns out that the `CSR_ZIMM` is faulty, e.g. it can classify an expression of the type `array_addr - imm` as a constant within the range [0, 31].


cahirwpz (approved):
LGTM

PR #1286, opened 2022-06-23
[RISCV] Fix virtual memory boot phase entry.

MichalBlk:
Whenever writing to the `satp` register we need to issue the `sfence.vma` instruction to ensure that the writing operation takes effect.

This fix causes the following `nop` instruction to always trigger an exception on both hardware platforms.


cahirwpz (approved):
LGTM

PR #1287, opened 2022-06-23
[RISCV][dev] PLIC support for SiFive Unleashed.

MichalBlk:
The HART hierarchy on our VexRiscv platform is very different from the one utilized in SiFive Unleashed. The PLIC driver must be adjusted to support the SiFive board.

### DIFF
--- a/sys/drv/plic.c
+++ b/sys/drv/plic.c
@@ -13,8 +13,30 @@
  *   https://github.com/riscv/riscv-plic-spec
  */
 
-/* PLIC memory map. */
+/*
+ * NOTE: FTTB, we designate a single HART to handle interrupts from all
+ * peripheral-level devices.
+ *
+ * For the 32-bit platform, the handling HART is the first one:
+ *   - ctx0 - machine mode
+ *   - ctx1 - supervisor mode
+ *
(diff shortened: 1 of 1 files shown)


sys/drv/plic.c
```diff
@@ -13,8 +13,30 @@
  *   https://github.com/riscv/riscv-plic-spec
  */
 
-/* PLIC memory map. */
+/*
+ * NOTE: FTTB, we designate a single HART to handle interrupts from all
+ * peripheral-level devices.
+ *
+ * For the 32-bit platform, the handling HART is the first one:
```
cahirwpz, line 20: This requires some more explanation. Could you refer a reader to the documentation?

MichalBlk: At the beginning of this file is the following note:
```
/*
 * Details of the operation of PLIC as well as the memory layout
 * can be found in the official specification which is maintained on github:
 *   https://github.com/riscv/riscv-plic-spec
 */
```
Isn't that sufficient?

MichalBlk: I changed the description to include a link to the original fu540 manual.

cahirwpz (approved):
LGTM

PR #1288, opened 2022-06-24
[dev] Simplebus enumeration.

MichalBlk:


MichalBlk:
After merging #1352, FDT bus enumeration will take just a few lines of code.

PR #1290, opened 2022-06-25
Replace libgcc with compiler_rt builtins from LLVM project

cahirwpz:
Our long term goal is to switch completely to LLVM ecosystem, since maintaining growing number of cross-compilers is too tedious. This is a major step since we switch from libgcc (GPLv2) to compiler-rt library (BSD). Thankfully the license is compatible with our code base (excerpt from `LICENSE.txt`):

> The compiler_rt library is dual licensed under both the University of Illinois
> "BSD-Like" license and the MIT license.  As a user of this code you may choose
> to use it under either license.  As a contributor, you agree to allow your code
> to be used under both.


pj1031999 (approved):
LGTM

sys/libkern/builtins/Makefile
```diff
@@ -0,0 +1,11 @@
+# vim: tabstop=8 shiftwidth=8 noexpandtab:
```
hadarai, line 1: Is that just a comment or a note to the reader?

cahirwpz: This is a standard Vim editor annotation for specifying extra commands to be executed at the time file is being open. Please read more here: https://vim.fandom.com/wiki/Modeline_magic

MichalBlk (approved):

PR #1292, opened 2022-07-03
[dev] Fix return value of simplebus interrupt processing function.

MichalBlk:
`sb_intr_to_rl` should return an error code.


cahirwpz (approved):
LGTM

PR #1294, opened 2022-07-04
Add simplebus procedure for adding child by phandle

mohrcore:
I can't find a way to get the absolute path back from a phandle. This makes it impossible to use simplebus for attaching children of bus other than rootdev. On the other hand, I can iterate over phandles of children nodes of a bus device node. This PR allows me to use those phandles to attach such children using simplebus (see the emmc controller driver)

### DIFF
--- a/include/dev/simplebus.h
+++ b/include/dev/simplebus.h
@@ -4,7 +4,7 @@
 #include <sys/device.h>
 
 /*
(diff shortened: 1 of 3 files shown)


cahirwpz:
Superseded by #1288 hence can be safely closed.

PR #1298, opened 2022-07-08
[RISCV][debug] Enable thread backtrace.

MichalBlk:
Enable thread backtrace for the `kthread` debug command.


sys/debug/ctx.py
```diff
@@ -25,6 +29,9 @@ def __init__(self):
         if arch_name.startswith('aarch64'):
             self.arch = 'aarch64'
             self.reg_size = 64
+        if arch_name.startswith('riscv'):
+            self.arch = 'riscv'
+            self.reg_size = 64
```
cahirwpz, line 34: What about `riscv32` ?

MichalBlk: Fixed.

PR #1299, opened 2022-07-08
[dts] SiFive Unleashed

MichalBlk:
Add a device tree source for the SiFive Unleashed target platform.


cahirwpz:
Please rename `bootargs.patch` to `cmdline-set-dts-bootargs.patch`, so that its name summarizes what the patch does.

MichalBlk:
The patch has been renamed. I also included the initrd load address for sifive_u in `launch`.

PR #1303, opened 2022-07-12
[dts] Set initrd boundaries dynamically.

MichalBlk:
The initrd boundaries provided to the kernel via FDT should be set dynamically by the launching script.


cahirwpz (approved):
LGTM

PR #1304, opened 2022-07-13
[include] Fix builtin defines.

MichalBlk:
The `ffs`, `clz`, and `ctz` automatically cast the input argument to a 32-bit integer, thereby a 64-bit argument is handled incorrectly.
E.g. in the following snippet from `init_vm_page`, `pa` is truncated to a 32-bit value regardless of the pointer size of the target architecture.
```
size_t size = 1 << min(PM_NQUEUES - 1, ctz(pa / PAGESIZE));
```

### DIFF
--- a/include/sys/mimiker.h
+++ b/include/sys/mimiker.h
@@ -16,9 +16,9 @@
 #include <sys/types.h>
 
 #define log2(x) (CHAR_BIT * sizeof(unsigned long) - __builtin_clzl(x) - 1)
-#define ffs(x) (size_t)(__builtin_ffs(x))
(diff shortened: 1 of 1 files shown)


cahirwpz:
Most likely you'd like to use `_Generic` feature of C11 for that. Please refer to: https://en.cppreference.com/w/c/language/generic

MichalBlk:
The PR has been changed to utilize the `_Generic` construct.

cahirwpz:
I know I asked about using `_Generic`, but looking at the code after the change, I'll never be tempted again to recommend its use. I'll restore original version of the file and will merge it.

cahirwpz (approved):

PR #1305, opened 2022-07-13
[usb] Fix UHCI pool size calculation.

MichalBlk:
Each UCHI transfer requires a single data buffer and a single transfer buffer, thereby the number of transfer buffers in the `P_TFR` pool must be equal to the number of data buffers in the `P_DATA` pool.


cahirwpz:
Let's discuss this during online meeting. I need to refresh my understanding of USB / UHCI.

hadarai (approved):

sys/drv/uhci.c
```diff
@@ -51,17 +51,26 @@ typedef struct uhci_state {
...
  */
 
-#define UHCI_TFR_BUF_SIZE 128
-#define UHCI_TFR_POOL_SIZE PAGESIZE
+#define UHCI_DATA_BUF_SIZE 512UL
 
-#define UHCI_DATA_BUF_SIZE (2 * UHCI_TD_MAXLEN)
-#define UHCI_DATA_POOL_SIZE (64 * PAGESIZE)
+#define UHCI_TFR_CTRL_MAX_TDS                                                  \
+  (2 + (UHCI_DATA_BUF_SIZE - sizeof(usb_dev_req_t)) / USB_MAX_IPACKET)
```
pj1031999, line 57: It's... magic :) Please add explanation.

sys/drv/uhci.c
```diff
@@ -464,15 +473,23 @@ static void uhci_init_frames(uhci_state_t *uhci) {
   out32(UHCI_FLBASEADDR, uhci_physaddr(uhci->frames));
 }
 
+static inline size_t uhci_align_pool_size(size_t size) {
+  size_t log = log2(size);
+  size_t asize = 1UL << log;
+  return (asize != size) ? (asize << 1) : asize;
+}
```
pj1031999, line 480: It's useful function, I thought that we already have something like that defined im mimiker.h (or in similar place). Could you make it generic? I'm pretty sure that other subsystems can borrow that implementation in the future.

sys/drv/uhci.c
```diff
@@ -464,15 +473,23 @@ static void uhci_init_frames(uhci_state_t *uhci) {
...
-  pool_add_page(P_TFR, tfr_pool, UHCI_TFR_POOL_SIZE);
+    (void *)kmem_alloc_contig(NULL, tfr_pool_asize, PMAP_NOCACHE);
+  pool_add_page(P_TFR, tfr_pool, roundup2(UHCI_TFR_POOL_SIZE, PAGESIZE));
 
+  size_t data_pool_asize = uhci_align_pool_size(UHCI_DATA_POOL_SIZE);
   void *data_pool =
-    (void *)kmem_alloc_contig(NULL, UHCI_DATA_POOL_SIZE, PMAP_NOCACHE);
-  pool_add_page(P_DATA, data_pool, UHCI_DATA_POOL_SIZE);
+    (void *)kmem_alloc_contig(NULL, data_pool_asize, PMAP_NOCACHE);
+  pool_add_page(P_DATA, data_pool, roundup2(UHCI_DATA_POOL_SIZE, PAGESIZE));
```
pj1031999, line 492: Why kmem doesn't do roundup2 size allocations here?

sys/drv/uhci.c
MichalBlk: Done (both for control and data transfers).

sys/drv/uhci.c
MichalBlk: Good point. It turns out that I already have a second use case for this function (#1323). I created a separate PR for this function (#1330).

cahirwpz (approved):
LGTM

PR #1308, opened 2022-07-14
[RISCV] Run tests in CI for SiFive Unleashed

cahirwpz:
All patches and workarounds required to run tests in CI for RISC-V SiFive Unleashed board.


sys/dts/sifive_u.dts
```diff
@@ -149,9 +149,18 @@
...
+                 * mmode_resv0@80000000 {
+                 *   no-map;
+                 *   reg = <0x00 0x80000000 0x00 0x40000>;
+                 * };
+                 */
+                /*
+                 * opensbi@80000000 {
+                 *   reg = <0x0 0x80000000 0x0 0x40000>;
+                 * };
+                 */
```
pj1031999, line 163: They are modified by qemu on top of this dts or what is the final state od dtb seen by kernel?

sys/kern/fdt.c
```diff
@@ -236,8 +240,10 @@ int FDT_get_reserved_mem(fdt_mem_reg_t *mrs, size_t *cntp) {
        child = FDT_peer(child)) {
     if (cnt == FDT_MAX_RSV_MEM_REGS)
       return ERANGE;
+#if 0
     if (FDT_hasprop(child, "no-map"))
       continue;
+#endif
```
pj1031999, line 246: dead code?

.github/workflows/default.yml
```diff
@@ -84,7 +93,7 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: "mips kernel+ramdisk"
-      - run: ./run_tests.py --board malta --timeout=80 --times=50
+      - run: ./run_tests.py --board malta --timeout=60 --times=50
```
pj1031999, line 96: Is it magically faster now?

sys/kern/fdt.c
cahirwpz: Possibly. I want @MichalBlk to comment on it.

.github/workflows/default.yml
cahirwpz: No. OTOH I don't know why it was so high compared to AArch64, so I lowered it a bit.

sys/dts/sifive_u.dts
cahirwpz: You need to dump machine memory to see actual dts used by the kernel. Please see a comment in `FDT_init` function.

.github/workflows/default.yml
pj1031999: See 4fe61bea98451a38898e323506841e1fe0e8338c

cahirwpz: If you look into tests runtime AArch64 & MIPS take roughly 30 minutes and 45 minutes for RISC-V. Timeouts seem to be set up properly.

In fact the only way to set up timeouts in more reasonable way would be to measure time taken by each test. So test framework should log for each test and each testrun how much time did it take. Then we could make well informed decision.

sys/kern/fdt.c
```diff
@@ -39,6 +39,10 @@ void FDT_init(void *va) {
   if (err < 0)
     FDT_panic(err);
   fdtp = va;
+  klog("FDT: %p - %p", fdtp, fdtp + fdt_totalsize(va));
+  /* XXX: To save DTB to file use GDB command:
+   * `memory dump memory board.dtb start_addr end_addr`,
```
MichalBlk, line 44: Shouldn't be `dump memory ..` instead of `memory dump memory ...`?

sys/dts/sifive_u.dts
```diff
@@ -149,9 +149,18 @@
...
+                /* QEmu patches this DTS and adds following on its own:
+                 *
+                 * mmode_resv0@80000000 {
+                 *   no-map;
+                 *   reg = <0x00 0x80000000 0x00 0x40000>;
+                 * };
+                 */
+                /*
+                 * opensbi@80000000 {
+                 *   reg = <0x0 0x80000000 0x0 0x40000>;
```
MichalBlk, line 161: If I perform the dump, this node also contains the `no-map` property.

MichalBlk (approved):

sys/kern/fdt.c
MichalBlk: This code should be removed.

sys/dts/sifive_u.dts
cahirwpz: What is the purpose of `no-map` property?

sys/kern/fdt.c
cahirwpz: Fixed!

sys/kern/fdt.c
cahirwpz: Done!

PR #1318, opened 2022-07-24
build/tools: s/cache/ccache/

pj1031999:
Fix which invocation - cache -> ccache.


cahirwpz (approved):
I'm astonished how many mistakes I did introducing ccache. Thanks for the fix!

PR #1320, opened 2022-07-26
Add /sbin/init based on suckless sinit

cahirwpz:
* [ ] implement sigtimedwait system call
 * [ ] see why second instance of shell cannot open `/dev/tty`


cahirwpz:
@panantoni01 Please extract `sinit` to another PR according to new method of importing external software. Please take into account `etc/` directory as well as stub of sigtimedwait syscall. Just leave implementation of `sys_sigtimedwait` and its userspace tests to @korbiniak.

sys/kern/signal.c
```diff
@@ -1,3 +1,5 @@
+#include "sys/siginfo.h"
```
cahirwpz, line 1: Please use convention for including header files that can be easily inferred by looking at source code of various kernel source code files.

cahirwpz:
Please resolve conflicts. Take into account recent changes to user test infrastructure introduced in #1374 and sleepq interface in #1378.

sys/kern/signal.c
```diff
@@ -201,6 +203,81 @@ int do_sigsuspend(proc_t *p, const sigset_t *mask) {
...
+      error = EAGAIN;
+      goto out;
+    }
+    if (tsp != NULL && !timevalid) {
+      error = EINVAL;
+      goto out;
+    }
+  }
+
+  error = sleepq_wait_timed(&td->td_sigmask, "sigtimedwait()", timeout);
```
cahirwpz, line 256: Please use here new interface for `sleepq_wait_timed` and pass `&p->p_lock` to it.

sys/kern/signal.c
```diff
@@ -201,6 +204,88 @@ int do_sigsuspend(proc_t *p, const sigset_t *mask) {
...
+  thread_t *td = p->p_thread;
+
+  if (tsp != NULL) {
+    if (tsp->tv_nsec < 0 || tsp->tv_nsec >= 1000000000) {
+      return EINVAL;
+    }
+    /* This is the same behaviour as in OpenBSD. */
+    if (tsp->tv_sec < 0) {
+      return EAGAIN;
+    }
```
franciscozdo, line 221: I think you should only set the `timevalid` variable. Timeout value is chcecked only if you have to wait for signal (according to [FreeBSD manual](https://www.freebsd.org/cgi/man.cgi?query=sigtimedwait&sektion=2&n=1#:~:text=%5BEINVAL%5D%09%09The%20timeout%20argument%20specified%20a%20tv_nsec%20value%20less%0A%09%09%09than%20zero%20or%20greater%20than%20or%20equal%20to%201000%20million.%0A%09%09%09Kernel%20only%20checks%20for%20this%20error%20if%20no%09signal%20is%0A%09%09%09pending%09in%20set%20and%20it%20is%20necessary%20to%20wait.)).

franciscozdo, line 221: Could you give some resource to that behavior in OpenBSD? I can't find it.

franciscozdo (changes requested):
Mostly minor changes or questions, because I am not an expert in signals.

I am also wondering what are the files `etc/rc.init` and `etc/rc.shutdown`. I think they are not needed by any of changes in this PR.

sys/kern/signal.c
korbiniak: https://mimiker.ii.uni.wroc.pl/source/xref/NetBSD/sys/kern/sys_sig.c?r=53d22107#752

korbiniak: I looked both on FreeBSD and OpenBSD implementations and I think I confused them here. However I think that freebsd's manual does not specify in which order it processes when the timeout is invalid:

`If the timespec structure pointed to by timeout is zero-valued and if none
 of the signals specified by set are pending, then sigtimedwait() returns im‐
 mediately with an error.  If timeout is the NULL pointer, the behavior is
 unspecified.` 

In OpenBSD it returns immediately if timeout is invalid or `tsp->tv_sec < 0`: https://mimiker.ii.uni.wroc.pl/source/xref/NetBSD/sys/kern/sys_sig.c?r=53d22107#752

The question now is which approach should we take? Wdyt?

sys/kern/sleepq.c
```diff
@@ -324,7 +324,8 @@ int sleepq_wait_timed(void *wchan, const void *waitpt, mtx_t *mtx,
     callout_schedule(&td->td_slpcallout, timeout);
   }
 
-  td->td_flags |= (timeout > 0) ? TDF_SLPTIMED : TDF_SLPINTR;
+  td->td_flags |= TDF_SLPINTR;
```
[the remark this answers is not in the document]
korbiniak: Yes, I did change the semantics as I think it was a bug that `TDF_SLPINTR` was not set. See discussion here: https://discord.com/channels/951841212267130950/1054133625731424337

sys/kern/syscalls.c
```diff
@@ -1330,5 +1330,34 @@ static int sys_kevent(proc_t *p, kevent_args_t *args, register_t *res) {
...
+      return error;
+  }
+
+  error = copyin_s(u_set, set);
+  if (error)
+    return error;
+
+  error = do_sigtimedwait(p, set, &kinfo, tsp);
+  if (error)
+    return error;
```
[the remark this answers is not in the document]
korbiniak: I need to allocate space for the copied `u_timeout`, thus I have this `timeout` variable. I use `tsp` pointer so I don't have to check if the timeout was copied in or not (as I checked that already before). If I didn't use it the call to `do_sigtimedwait` would have been like this or similar:
`
if ((error = do_sigtimedwait(p, set, &kinfo, u_timeout ? &timeout : NULL)))
  return error;
`
Which approach is nicer?

sys/kern/signal.c
franciscozdo: In NetBSD manual there is almost identical behavior described as in FreeBSD ([link](https://man.netbsd.org/sigtimedwait.2)). I haven't found this syscall in OpenBSD manual. I'd suggest to stick to the behavior described in manuals. @cahirwpz What do you think?

sys/kern/signal.c
```diff
@@ -201,6 +204,88 @@ int do_sigsuspend(proc_t *p, const sigset_t *mask) {
...
+
+    timevalid = 1;
+    timeout = ts2hz(tsp);
+    /* Timeout set to 0 means to not block, but for sleepq it means to
+     * wait indefinitely */
+    if (timeout == 0) {
+      if (tsp->tv_nsec == 0 && tsp->tv_sec == 0) {
+        timeout = -1;
```
franciscozdo, line 229: I also think that if you move the comment that means "do not block" here and remove the comment in lines 225-226 then it would be more clear.

franciscozdo: Eventually provide some comment in style that from NetBSD (https://mimiker.ii.uni.wroc.pl/source/xref/NetBSD/sys/kern/sys_sig.c?r=53d22107#741). I really like that because it is immidiately clear what given values mean.

bin/utest/signal.c
```diff
@@ -295,6 +296,93 @@ TEST_ADD(signal_sigsuspend) {
   return 0;
 }
 
+/* ======= signal_sigtimedwait ======= */
+int sigtimedwait_child(void *arg) {
+  pid_t ppid = getppid();
+  sched_yield();
+  syscall_ok(kill(ppid, SIGUSR1));
```
franciscozdo, line 303: I think that you can't assume that parent will do all initialization needed before sending him the `SIGUSR1` signal. Please look at other tests in that file and check if you can synchronize using other signals (I think you can use similar mechanism to that in `signal_stop` test.)

bin/utest/signal.c
```diff
@@ -295,6 +296,93 @@ TEST_ADD(signal_sigsuspend) {
...
+  assert(__sigsetequal(&set, &current));
+
+  utest_child_exited(0);
+  return 0;
+}
+
+/* ======= signal_sigtimedwait_timeout ======= */
+static int sigusr2_handled = 0;
+
+void sigtimedwait_timeout_sigusr2_handler(int signo) {
```
franciscozdo, line 333: Why don't you name it `sigusr2_handler`?

bin/utest/signal.c
```diff
@@ -295,6 +296,93 @@ TEST_ADD(signal_sigsuspend) {
...
+  timeout.tv_sec = 0;
+  syscall_fail(sigtimedwait(&waitset, &info, &timeout), EINTR);
+
+  /* After the signal we should be interrupted only once, so next call to
+   * sigtimedwait will timeout. */
+  syscall_ok(kill(cpid, SIGUSR2));
+
+  /* This may be woken up by a timeout or by an interrupt. */
+  nanosleep(&timeout, NULL);
+  syscall_fail(sigtimedwait(&waitset, &info, &timeout), EAGAIN);
```
franciscozdo, line 380: I suggest extracting it into separate test case. You won't have to ensure that child stops sending us signals. We will be 100% sure that nothing strange happens.

sys/kern/signal.c
franciscozdo: Okay, I think that NetBSD has a little bit better described semantics here (https://man.netbsd.org/sigtimedwait.2). The say that if `timeout == NULL` then it behaves like `sigwaitinfo()` (waits for signal without the timeout). In that case we don't have any problem with order of operations:
1. Check if there are pending signals.
2. Check the timeout value (and return error or wait specified amount of time)

korbiniak: Again, I think that the NetBSD manual also do not describe what happens if the timeout value is invalid. They very explicitly say `if timeout is zero [...]` and `if timeout is NULL [...]`.

I'd expect a sentence like this: 'if the timeout value is invalid (i.e. it's false that `0 <= tv_nsec <= 1e9)` or tv_sec is negative, first check current pending signals, if there's no signal that we wait for then return an error.'.

korbiniak: I'm sorry to bring confusion. I've seen this behaviour in NetBSD, not OpenBSD.

korbiniak: As you can see [here](https://mimiker.ii.uni.wroc.pl/source/xref/NetBSD/sys/kern/sys_sig.c?r=53d22107#752), NetBSD returns immediately if timeout is invalid or tv_sec < 0. They do not check the pending signal mask in this case.

bin/utest/signal.c
korbiniak: This way I know that some particular test will be the only user of this function. 
Suppose someone in the future would like to use the sigusr2 for testing, then they'd need to create a new function, how would they name it then? They could reuse this function, but I advise against that. If one would like to change the behavior of the handler for some particular test they'd need to create a new function and give it some other name anyway.

bin/utest/signal.c
korbiniak: Good point.

sys/kern/signal.c
franciscozdo: Okay, It seems that manual may be outdated (I mean the commend about `EINVAL` in [error section](https://man.netbsd.org/sigtimedwait.2#:~:text=This%20error%20is%20only%20checked%20if%20no%20signal%20from%20set%20is%20pending%20and%20it%20would%0A%20%20%20%20%20be%20necessary%20to%20wait.).).

Maybe it is better to filter wrong arguments before anything is checked. It forces user to always pass correct arguments. I think we can leave it like that.

sys/kern/syscalls.c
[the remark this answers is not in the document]
franciscozdo: Okay, I understand why tsp is needed. You can put the call to `do_sigtimedwait` into if condition.

bin/utest/signal.c
franciscozdo: I suggested it because I see that in other signal tests we name functions that looks like that always in the same way. These functions are used in multiple tests and I think there is nothing wrong in that (they are used identically so why we would like to duplicate code?).

If you would like to add some functionality to such function then it is a reason to make function with unique name (and leave the old one unchanged because we expect some specific behavior from it).

@cahirwpz @pwit81 I think that we can think about making such function a standard. They are used pretty often and maybe we don't have to invent new functions for just handling one signal for every written test?

sys/kern/sleepq.c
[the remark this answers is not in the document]
franciscozdo: Okay, I agree with that, but I also think that @cahirwpz should take a look at that.

bin/utest/signal.c
```diff
@@ -295,6 +296,114 @@ TEST_ADD(signal_sigsuspend) {
...
+  syscall_ok(kill(ppid, SIGUSR1));
+  return 0;
+}
+
+TEST_ADD(signal_sigtimedwait) {
+  syscall_ok(signal(SIGCONT, sigcont_handler));
+  sigset_t set, current, waitset;
+  __sigemptyset(&set);
+  __sigaddset(&set, SIGUSR1);
+  __sigaddset(&set, SIGCONT);
```
pj1031999, line 311: s/__sig...set/sig...set/

see /lib/libc/gen/sig...set.c

bin/utest/signal.c
```diff
@@ -295,6 +296,114 @@ TEST_ADD(signal_sigsuspend) {
...
+  syscall_ok(sigprocmask(SIG_SETMASK, &set, NULL));
+
+  utest_spawn(sigtimedwait_child, NULL);
+
+  siginfo_t info;
+  __sigemptyset(&waitset);
+  __sigaddset(&waitset, SIGUSR1);
+  assert(sigtimedwait(&waitset, &info, NULL) == SIGUSR1);
```
pj1031999, line 319: NetBSD man:
> The assert() macro may be removed at compile time with the [cc(1)](https://man.netbsd.org/cc.1) option `-DNDEBUG.`

I think that you don't want to call anything inside assert.

bin/utest/signal.c
```diff
@@ -295,6 +296,114 @@ TEST_ADD(signal_sigsuspend) {
   return 0;
 }
 
+/* ======= signal_sigtimedwait ======= */
+int sigtimedwait_child(void *arg) {
```
pj1031999, line 300: Why it's not static?

bin/utest/signal.c
```diff
@@ -295,6 +296,114 @@ TEST_ADD(signal_sigsuspend) {
...
+
+  syscall_ok(sigprocmask(SIG_BLOCK, NULL, &current));
+  assert(__sigsetequal(&set, &current));
+
+  utest_child_exited(0);
+  return 0;
+}
+
+/* ======= signal_sigtimedwait_timeout ======= */
+int sigtimedwait_timeout_child(void *arg) {
```
pj1031999, line 330: static?

bin/utest/signal.c
pj1031999, line 300: __unused

bin/utest/signal.c
pj1031999, line 330: __unused?

sys/kern/syscalls.c
```diff
@@ -1330,5 +1330,32 @@ static int sys_kevent(proc_t *p, kevent_args_t *args, register_t *res) {
...
+
+  error = do_sigtimedwait(p, set, &ksi, u_timeout ? &timeout : NULL);
+  if (error)
+    return error;
+
+  if (u_info) {
+    error = copyout_s(ksi.ksi_info, u_info);
+  }
```
pj1031999, line 1357: You are not consistent with style. { }

sys/kern/sleepq.c
```diff
@@ -324,7 +324,8 @@ int sleepq_wait_timed(void *wchan, const void *waitpt, mtx_t *mtx,
...
+  td->td_flags |= (timeout > 0) ? TDF_SLPTIMED : 0;
```
pj1031999, line 328: `if (timeout > 0) td->td_flags |= TDF_SLPTIMED` is more readable and or with 0 is nop.

sys/kern/signal.c
```diff
@@ -201,6 +204,93 @@ int do_sigsuspend(proc_t *p, const sigset_t *mask) {
...
+   * Calculate timeout based on tsp.
+   *
+   * NULL pointer means an infinite timeout.
+   * {.tv_sec = 0, .tv_nsec = 0} means to not block.
+   * Timeout is valid only if 0 <= tv_nsec <= 1e9. If this is not true,
+   * immediately return EINVAL. If tv_sec < 0 immediately return EAGAIN.
+   *
+   * We set the timeout value such that:
+   *   - timeout == -1: do not block
+   *   - timeout == 0: wait indefinitely
```
pj1031999, line 224: Is it possible? if `timeout` is set to 0 then you change it *always* to -1 or 1.

sys/kern/signal.c
```diff
@@ -201,6 +204,93 @@ int do_sigsuspend(proc_t *p, const sigset_t *mask) {
...
+    if (error == ETIMEDOUT) {
+      error = EAGAIN;
+      goto out;
+    }
+
+    pending = td->td_sigpend.sp_set;
+    __sigandset(&waitset, &pending);
+    if ((sig = __sigfindset(&pending))) {
+      sigpend_get(&td->td_sigpend, sig, ksi);
+      error = 0;
```
pj1031999, line 283: error value is lost here. If `sleepq_wait_timed` will return any error other than `ETIMEDOUT` you ignore that.

pj1031999 (commented):
What is a purpose of /etc/rc.init /etc/rc.shutdown? How it is related to `sigtimedwait` or `/dev/tty`?

sys/kern/signal.c
korbiniak: Yes, `timeout == 0` iff `tsp == NULL`

include/sys/signal.h
```diff
@@ -202,6 +202,8 @@ void sig_onexec(proc_t *p);
 int do_sigaction(signo_t sig, const sigaction_t *act, sigaction_t *oldact);
 int do_sigprocmask(int how, const sigset_t *set, sigset_t *oset);
 int do_sigsuspend(proc_t *p, const sigset_t *mask);
+int do_sigtimedwait(proc_t *p, sigset_t waitset, ksiginfo_t *kinfo,
```
korbiniak, line 205: Change argument name: `kinfo` -> `ksi`

bin/utest/signal.c
korbiniak: Thanks, didn't know that.

sys/kern/signal.c
korbiniak: I think `sleepq_wait_timed` can only return `ETIMEDOUT` or `EINTR`, so here we work under assumption that it returned `EINTR`. This checks if we were interrupted by a signal from the waitset or not.

sys/kern/sleepq.c
korbiniak: Right, dunno why I made it this way.

sys/kern/syscalls.c
korbiniak: Right. Which one is preferable in our project?

sys/kern/signal.c
pj1031999: Implementation can change in the future, it's usually a good practice to catch all possible errors.

Something like:
```
error = foo(...);
switch (error) {
  case 0: break;
  case ERROR1: { /* handle ERROR1 */ }
  case ERROR2: { /* handle ERROR2 */ }
  default: { /* unknown error - panic? */ }
}
```

sys/kern/signal.c
pj1031999: ack

bin/utest/signal.c
franciscozdo: @pj1031999 Notice, that in utests we use our own implementation of asserts that comes from `utest.h`.

bin/utest/signal.c
cahirwpz: Indeed `__sig*` is not a part of POSIX interface, please have a look at [sigsetops](https://man.netbsd.org/sigsetops.3). The issue here is that, nonetheless, those functions are exposed to non-kernel code, instead of being hidden as in FreeBSD: https://mimiker.ii.uni.wroc.pl/source/xref/FreeBSD/lib/libc/gen/sigsetops.c?r=7a56009c#40

bin/utest/signal.c
cahirwpz: Right. There're still some confusing deficiencies of our userspace test framework. This is one of them. Personally I'd:
 * disallow using `assert` from by adding `-DNDEBUG` to `utest` Makefile,
 * rename our `assert` to `check`,
 * provide a set of functions with `x` prefix that guarantee to return with success as FreeBSD's [xmalloc](https://www.freebsd.org/cgi/man.cgi?query=xmalloc)

sys/kern/signal.c
cahirwpz: What @pj1031999 proposes seems to be a nice and uniform way of error handling. However this style was not employed by *BSD kernels and we didn't inherit that. Introducing it at the moment would cause inconsistencies I'd prefer to avoid in the long run.

sys/kern/syscalls.c
cahirwpz: For single line if-clause we don't use `{ }`. If in doubt look at surrounding code and see what's more prevalent.

bin/utest/signal.c
cahirwpz: @franciscozdo @korbiniak It's definitely a good idea to factor it out.

I'd do it by extending our current `utest.h` interface by adding:
 * `void signal_count(int signo)` - registers a generic handler for `signo` signal,
 * `int signal_counter(int signo)` - returns a number of times signal `signo` was delivered.

If you have a better idea how to name those functions then feel free to do so.

sys/kern/sleepq.c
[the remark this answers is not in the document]
cahirwpz: I think it has to do with possible deficiencies in `sleepq` API. ATM we have two `wait` functions - with and without timeout. IMO there should be `wait` and `wait_intr` with optional timeout. That should let us express all use cases for putting a thread to sleep.

Please note that non-interruptible `wait` should probably be only used by the code that does not interface user-space directly - i.e. the code that does not understand the concept of a signal or need not be burdened with complexity of extra error handling (`ETIMEDOUT`). Examples would be waking up a thread from interrupt or "fast" device drivers code. On contrary the interruptible `wait` should be used in "slow" device drivers code (where `read` or `write` may be interrupted) and all syscalls that can tolerate being interrupted.

Perhaps what I wrote above should be redacted and put somewhere into `sleepq.h`.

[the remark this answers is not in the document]
cahirwpz: To answer the question - the change is ok, since function is documented as follows:

```c
/*! \brief Performs interruptible sleep with timeout.
 *
 * Timed sleep is also interruptible. [...]
 */
```

cahirwpz:
@korbiniak I'd like to take over the PR and finish the work. WDYT?

PR #1323, opened 2022-08-03
[mem] Kmalloc slab allocator.

MichalBlk:
Kernel malloc implementation based on slabs.

This PR contains changes introduced in #1336.

PR #1324, opened 2022-08-07
[mem] Add a procedure to obtain the size of a pointed vmem segment.

MichalBlk:
The new implementation of the general purpose kernel memory allocator (#1323) needs a way to fetch the size of a selected vmem segment.


cahirwpz (approved):
LGTM

PR #1325, opened 2022-08-08
[mem] Remove the size argument from vmem_free.

MichalBlk:
It turns out that our current implementation of `vmem_free` assumes that only entire segments will be freed. This assumption makes the `size` argument useless as the size is contained in the corresponding boundary tag.


cahirwpz (approved):
LGTM

PR #1328, opened 2022-08-21
[evdev] Fix evtest build.

MichalBlk:
If we call make recursively from `evtest` during the download phase, the build fails while obtaining dependencies for `evtest.c` as the sources aren't patched yet.


cahirwpz (approved):
LGTM

PR #1329, opened 2022-08-24
[dev] BAR address should be size-aligned.

MichalBlk:
Unfortunately, along with removing the resource manager (#1322) we have violated the alignment of PCI BARs. If BAR alignment is too small, the lower bits are ignored and we end up with a virtual address accessing some other parts of the configuration space.


cahirwpz (approved):
LGTM

PR #1332, opened 2022-09-04
[dev] Basic USB HUB support.

MichalBlk:
We need basic support for USB hubs as on some boards the number of root ports is constrained and we must rely on external hubs to provide additional attachment points.


cahirwpz (approved):
LGTM

PR #1336, opened 2022-09-23
[lockdep][RISCV] Fix pmap lock acquiring order.

MichalBlk:
There is a cycle between `pmap::mtx` and `kernel_pmap::mtx`. The conflicting scenarios (RISC-V only):
1. We run out of kernel virtual address space and `pmap_growkernel` must be called. This results in `pmap_growkernel_md` being called which first acquires `kernel_pmap::mtx` and then `pmap::mtx`.
2. When entering a new mapping to the user's virtual address space, we must allocate a pv entry from the appropriate pool. However, if the pool happens to be empty, `kmem_alloc` gets called which results in a call to `pmap_kenter`. `pmap_enter` first locks `pmap::mtx`, and then `pmap_kenter` locks `kernel_pmap::mtx`.

The solution is simply changing the acquiring order in RISC-V's `pmap_growkernel_md` implementation.

### DIFF
--- a/sys/riscv/pmap.c
+++ b/sys/riscv/pmap.c
@@ -146,15 +146,19 @@ void pmap_md_growkernel(vaddr_t maxkvaddr) {
   pmap_t *kmap = pmap_kernel();
   pmap_t *umap = pmap_user();
 
+  if (umap)
+    mtx_lock(&umap->mtx);
+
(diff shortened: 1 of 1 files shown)


cahirwpz (approved):
Though I sense here a bad smell I'll accept it as a temporary workaround.

PR #1342, opened 2022-10-09
[exec] Fix user stack setup.

MichalBlk:
The current code assumes that malloc's `ALIGNMENT` is divisible by `STACK_ALIGN`. Although it turns out to be always valid for the present implementation, this assumption should be dropped.

### DIFF
--- a/sys/kern/ustack.c
+++ b/sys/kern/ustack.c
@@ -85,6 +85,8 @@ void ustack_relocate_ptr(ustack_t *us, vaddr_t *ptr_p) {
 
 void ustack_finalize(ustack_t *us) {
   assert(!finalized_p(us));
+  /* We have to align the stack size as that much memory will be copied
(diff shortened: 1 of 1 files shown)


sys/kern/ustack.c
```diff
@@ -85,8 +85,8 @@ void ustack_relocate_ptr(ustack_t *us, vaddr_t *ptr_p) {
 
 void ustack_finalize(ustack_t *us) {
   assert(!finalized_p(us));
-  /* Cannot fail because initially stack is aligned to STACK_ALIGN. */
-  ustack_align(us, STACK_ALIGN);
+  ptrdiff_t size = us->us_bottom - us->us_top;
```
cahirwpz, line 88: Please provide a comment on what's the expected effect of this line.

cahirwpz (approved):
LGTM

PR #1344, opened 2022-10-17
[usb] Fix UHCI USB host controller driver.

MichalBlk:
The current code attempts to acquire blocking locks while holding a spin lock. The solution is to move IRQ handling into an ithread context.


cahirwpz (approved):
LGTM

PR #1345, opened 2022-10-22
[execve] Set protection bits during ELF load

franciscozdo:
This PR aims to introduce setting protection bits when binary is loaded during execve. Currently we don't set any protection bits (and all memory has `rwx` rights).


sys/tests/utest.c
```diff
@@ -60,6 +59,13 @@ static int utest_generic(const char *name, int status_success) {
 #define UTEST_ADD_SIGNAL(name, sig)                                            \
   UTEST_ADD(name, MAKE_STATUS_SIG_TERM(sig), 0)
 
+UTEST_ADD_SIGNAL(vmmap_text_w, SIGSEGV);
+UTEST_ADD_SIGNAL(vmmap_rodata_w, SIGSEGV);
+#ifndef __mips__
+UTEST_ADD_SIGNAL(vmmap_data_x, SIGSEGV);
+UTEST_ADD_SIGNAL(vmmap_rodata_x, SIGSEGV);
+#endif
```
franciscozdo, line 67: I have no idea why these tests still don't work on mips.

sys/kern/vm_map.c
```diff
@@ -196,8 +196,22 @@ void vm_map_delete(vm_map_t *map) {
...
+  klog("vm_map_protect: 0x%x - 0x%x %c%c%c", start, end,
+       (prot & VM_PROT_READ) ? 'r' : '-', (prot & VM_PROT_WRITE) ? 'w' : '-',
+       (prot & VM_PROT_EXEC) ? 'x' : '-');
+
+  ent = vm_map_find_entry(map, start);
+  assert(ent);
+  assert(ent->end == end);
+
+  ent->prot = prot;
+  pmap_protect(map->pmap, start, end, prot);
```
pj1031999, line 214: IIUC it find's only first `vm_map_entry` - what with others? What is a semantic of this function?

sys/aarch64/pmap.c
```diff
@@ -76,7 +76,9 @@ pte_t pte_make(paddr_t pa, vm_prot_t prot, unsigned flags) {
 }
 
 pte_t pte_protect(pte_t pte, vm_prot_t prot) {
-  return vm_prot_map[prot] | (pte & (~ATTR_AP_MASK & ~ATTR_XN));
+  return vm_prot_map[prot] |
+         (pte & (~ATTR_AP_MASK & ~ATTR_XN & ~ATTR_SW_FLAGS)) |
+         (pte & ATTR_AP_USER); /* This flag shouldn't be cleaned */
```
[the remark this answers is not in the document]
franciscozdo: What do you mean by 'follow pattern from prevoius line'? This flag can't be cleaned so if it is set then we must be sure that it is set in result pte (it may be cleared with `ATTR_AP_MASK`).

sys/kern/vm_map.c
franciscozdo: Okay, I have updated it. Function now can be used only with whole entries which are inside [start, end] region. I don't want to support splitting now because it will be used when `mprotect` is implemented. (Then I will also add detailed tests for that splitting.)

include/sys/klog.h
```diff
@@ -29,6 +29,7 @@ typedef enum {
   KL_FILE,    /* filedesc & file operations */
   KL_FILESYS, /* filesystems */
   KL_TTY,     /* terminal subsystem */
+  KL_VMMAP,   /* TODO: remove */
```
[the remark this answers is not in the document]
franciscozdo: Removed

sys/tests/utest.c
cahirwpz: Actually only `vmmap_rodata_x` does not work ;-)

cahirwpz: OK. I was wrong. It just fails in another place.

We're going to need `siginfo_t` filled from the kernel to make sure our tests do what they really should check.

cahirwpz:
@franciscozdo I'll take over the change in order to fix problems on MIPS. I'll try to record all the steps and explain to you how I approached the problem. We'll discuss that during our weekly meeting.

cahirwpz (approved):
LGTM

PR #1346, opened 2022-11-03
Create a hook to verify code style

franciscozdo:
This PR:
  - creates a pre-commit hook
  - adds script to install that hook
  - updates implementation of `verify-pycodestyle.py`

**NOTE**: The script that installs the git hooks is the only way that came to my mind to make sure every one has hooks installed. Although it still has to be run manually.  That means we cannot get rid of verify code style stage in CI.

### DIFF
--- a/scripts/pre-commit
+++ b/scripts/pre-commit
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+err=false
+
+# Verify C code style
+
+echo "Verifying C code style..."
+CFILES=$(git diff --cached --name-only | grep -E ".*\.[hc]")
+
+if ! [ "$CFILES" == "" ]; then
+  for file in $CFILES; do
+    clang-format-14 -n --Werror --style=file -i $file || err=true
+  done
+fi
+
+# Verify Python code style
+
+echo "Verifying Python code style..."
+PYFILES=$(git diff --cached --name-only | grep -E ".*\.py")
+PYEXTRA=$(git diff --cached --name-only | grep -E "launch")
+
+if ! [ "$PYFILES" == "" ] || ! [ "$PYEXTRA" == "" ]; then
+  pycodestyle ${PYEXTRA} ${PYFILES} || err=true
+fi
+
+echo $err
+if $err ; then
+  exit 1
+fi
+
+echo "Code style OK"
+exit 0
--- a/git-setup.sh
+++ b/git-setup.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# Check if pre-commit hook exists
+[ -f .git/hooks/pre-commit ] \
+  && echo "You already have a pre-commit hook installed. Please remove it if you want to use our hook." \
+  && exit 1
+
+# Install the pre-commit hook
+ln -s ./scripts/pre-commit .git/hooks/pre-commit \
+  || echo "Failed to install pre-commit hook" \
+  && exit 1
+
+echo Hooks successfully installed.
+exit 0
--- a/verify-pycodestyle.sh
+++ b/verify-pycodestyle.sh
@@ -1,8 +1,15 @@
 #!/bin/sh
 
-PYFILES=$(find . \( -name 'toolchain' -prune \)\
-              -o \( -name '.*.py' -prune \) \
-              -o \( -name '*.py' -printf '%P\n' \))
+PYFILES=$(find . \( -name '*.py' \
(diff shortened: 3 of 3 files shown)


franciscozdo:
@cahirwpz Hey, could you also look at this PR? It depends on the #1381.

franciscozdo:
@cahirwpz I have spent some time on testing the new `make format` and git hook, but I'd be grateful if you will play with them to check if everything works as expected.

franciscozdo:
@cahirwpz I think I have finished it. The only file that was formatted, but now is excluded is `sys/libkern/copystr.c` (it is taken from NetBSD so it shouldn't be formatted anyway).

build/tools.mk
```diff
@@ -83,3 +83,34 @@ PATCH   = patch
 GENASSYM = $(TOPDIR)/sys/script/genassym.sh
 YACC	= byacc
 GZIP	= gzip -9
+
+dump-tools:
+	echo "CC       ${CC}"
```
cahirwpz, line 88: Wouldn't it be easier if you just dumped everything in following format: `CC=${CC}` ?
Then you could use `source` command from bash to include all variables.

scripts/pre-commit
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+err=0
+
+echo "Verifying C code style..."
+
+FORMAT_TOOL=$(make dump-tools | grep FORMAT | cut -d' ' -f2-)
```
cahirwpz, line 7: `export $(make dump-tools)` ?

scripts/pre-commit
```diff
@@ -0,0 +1,39 @@
...
+CFILES=$(git diff --cached --name-only | grep -E ".*\.[hc]" | grep -vf .format-exclude)
+
+err=0
+for file in $CFILES; do
+  # check stashed version only
+  git show :$file | $FORMAT_TOOL -n --Werror || err=1
```
cahirwpz, line 13: Why not immediately `... || exit 1` here?

scripts/pre-commit
```diff
@@ -0,0 +1,39 @@
...
+
+echo "Verifying Python code style..."
+
+PYFILES=$(git diff --cached --name-only | grep -E ".*\.py")
+PYEXTRA=$(git diff --cached --name-only | grep -E "launch")
+
+err=0
+for file in $PYFILES $PYEXTRA; do
+  # check stashed version only
+  git show :$file | pycodestyle - || err=1
```
cahirwpz, line 30: Ditto

verify-format.sh
```diff
@@ -1,17 +1,24 @@
-#!/bin/sh
+#!/bin/bash
 
-PAGER=cat
+MODIFY=-n
+if [ "$1" == "--fix" ]; then
+  MODIFY=
+fi
+
+FORMAT_TOOL=$(make dump-tools | grep FORMAT | cut -d' ' -f2-)
```
cahirwpz, line 8: `export $(make dump-tools)` ?

Makefile
```diff
@@ -7,11 +7,14 @@ SUBDIR = sys lib bin usr.bin etc include contrib
 
 all: install
 
-format: setup
 build: setup
 install: build
 distclean: clean
 
+format:
+	echo "'make format' is deprecated. Use 'verify-format.sh --fix' instead."
```
cahirwpz, line 15: Why not just call the command here?

cahirwpz (commented):
A few small improvements here and it'll be all right.

scripts/pre-commit
franciscozdo: I want to print all formatting errors here. If we exit after first error it may not give enough information. 

Let's say we are adding a file that shouldn't be formatted, but we forgot to add it to `.format-exclude` . Then we try to commit this file, but we also add some other file which also has bad formatting. Then we may got information about first error in file which we expect to be formatted. But we won't know that we forgot to add a file to `.format-exclude` because it isn't reported.

build/tools.mk
franciscozdo: Right, it is better.

scripts/pre-commit
franciscozdo: Actually `source <(make dump-tools)` works.

Makefile
franciscozdo: Okay, now we're calling `verify-format.sh` here.

cahirwpz (approved):
LGTM

PR #1347, opened 2022-11-04
Experiment with porting suckless sbase to mimiker

cahirwpz:
Main goal of this PR is to port as much of sbase as we can. We need to fix warnings in their code. Add missing function and possibly syscalls. But firstly let's identify which binaries we want to compile.


cahirwpz:
I think we're not going to merge this code with mimiker but use external reference to their repository and apply some patches with [quilt](https://en.wikipedia.org/wiki/Quilt_(software)) to manage our local changes.

cahirwpz:
Superseded by #1361

PR #1348, opened 2022-11-08
[RISC-V] Smarter update kernel part of user top PD

cahirwpz:
This is an issue on RISC-V where we must update upper part of each user pmap if top page directory of kernel pmap was modified.

### DIFF
--- a/sys/riscv/pmap.c
+++ b/sys/riscv/pmap.c
@@ -52,15 +52,8 @@ static const pte_t vm_prot_map[] = {
  * Page directory.
  */
 
-void pde_write(pde_t *pdep, paddr_t pa, int lvl, vaddr_t va) {
-  *pdep = PA_TO_PTE(pa) | PTE_V;
-
-  /* Check if we need to propagate changes to other pmaps. */
-  if (kern_addr_p(va) && lvl == 0) {
-    pmap_t *kmap = pmap_kernel();
-    kmap->md.generation++;
-    assert(kmap->md.generation);
-  }
+pde_t pde_make(int lvl, paddr_t pa) {
+  return PA_TO_PTE(pa) | PTE_V;
 }
 
 /*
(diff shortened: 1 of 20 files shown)


sys/riscv/pmap.c
```diff
@@ -84,27 +84,71 @@ inline pte_t pte_protect(pte_t pte, vm_prot_t prot) {
...
   __set_satp(umap->md.satp);
   __sfence_vma();
 }
 
+void pmap_md_update(pmap_t *umap) {
+  if (!umap)
+    return;
+
+  update_kernel_pd(umap);
+  __sfence_vma();
```
MichalBlk, line 149: If the generation is up to date, then the `__sfence_vma` is needless. You could move this statement after the `memcpy` invocation.

sys/riscv/pmap.c
```diff
@@ -84,27 +84,71 @@ inline pte_t pte_protect(pte_t pte, vm_prot_t prot) {
...
+      *pdep = pde_make(0, kmap_l0_pages[L0_INDEX(va)]);
+    }
+
+    kmap->md.generation++;
+    assert(kmap->md.generation);
+  }
+}
+
+void pmap_md_activate(pmap_t *umap) {
+  update_kernel_pd(umap);
```
MichalBlk, line 138: We should consider the case when `umap` is `NULL` (i.e. switching to a kernel thread). Currently, this isn't resulting in a kernel oops, as we don't update `curpmap` correctly (see #1353). You can keep it as it is, and I'll update #1353 after this PR is merged. Or merge that PR first and update this procedure.

MichalBlk (approved):
LGTM. Please add to the description an explanation regarding the lock cycle (similar to description in #1336).

sys/riscv/pmap.c
cahirwpz: Done!

sys/riscv/pmap.c
cahirwpz: Done!

cahirwpz:
@MichalBlk I added a comment on why `kmap_l0_lock` is needed.

PR #1353, opened 2022-11-11
[vm] Fix VM map switching.

MichalBlk:
If we switch to a kernel thread, the current user pmap should and userspace virtual memory map should be set to `NULL`.


cahirwpz:
Subsumed by #1348

PR #1361, opened 2022-11-28
Add suckless sbase tools

panantoni01:
The goal of this PR is to port [sbase](https://git.suckless.org/sbase/files.html) - a collection of suckless unix tools. We would like to have a single binary containing all the required tools - just like [busybox](https://busybox.net).


contrib/sbase/patches/creat2open.patch
```diff
@@ -0,0 +1,51 @@
+Index: sbase/sbase/libutil/cp.c
+===================================================================
+--- sbase.orig/sbase/libutil/cp.c	2022-12-06 19:10:45.123093136 +0100
++++ sbase/sbase/libutil/cp.c	2022-12-07 22:19:23.039798679 +0100
+@@ -111,14 +111,14 @@
+ 			cp_status = 1;
+ 			return 0;
+ 		}
+-		if ((f2 = creat(s2, st.st_mode)) < 0 && cp_fflag) {
++		if ((f2 = open(s2, O_CREAT|O_WRONLY|O_TRUNC, st.st_mode)) < 0 && cp_fflag) {
```
cahirwpz, line 10: Please add `creat` wrapper for `open` in exactly the same way as NetBSD did that in:
https://mimiker.ii.uni.wroc.pl/source/xref/NetBSD/lib/libc/compat-43/creat.c?r=53d22107#46

bin/sbase/basename.c
```diff
@@ -0,0 +1,37 @@
+/* See LICENSE file for copyright and license details. */
```
cahirwpz, line 1: You can already remove `bin/sbase` directory from the PR.

contrib/sbase/patches/cols.patch
```diff
@@ -0,0 +1,11 @@
+Index: sbase/sbase/cols.c
+===================================================================
+--- sbase.orig/sbase/cols.c	2022-12-08 22:33:35.296210736 +0100
++++ sbase/sbase/cols.c	2022-12-08 22:49:28.236179265 +0100
+@@ -1,5 +1,6 @@
+ /* See LICENSE file for copyright and license details. */
+ #include <sys/ioctl.h>
++#include <sys/ttycom.h>
```
cahirwpz, line 8: This manifests a problem with our header files. Please note that `sys/ioctl.h` in NetBSD includes `sys/ttycom.h` by default: https://mimiker.ii.uni.wroc.pl/source/xref/NetBSD/sys/sys/ioctl.h?r=53d22107#42

contrib/sbase/Makefile
```diff
@@ -0,0 +1,312 @@
...
+$(BUILDDIR)sbase/$(LIBUTIL): $(OBJ_LIBUTIL)
+
+$(BUILDDIR)sbase/$(LIBUTF): $(OBJ_LIBUTF) 
+
+download-here: sbase/.git
+
+build-before: quilt-patch
+
+install-here:
+	@echo "[INSTALL] /bin/sbase/sbase-box"
```
cahirwpz, line 307: Please use `$(BINDIR)` here as well.

contrib/sbase/Makefile
```diff
@@ -0,0 +1,312 @@
...
+	logger
+
+BIN_SURELY_DISABLED =\
+	chroot\
+	flock\
+	hostname\
+	mknod\
+	tftp
+
+CFLAGS.sbase/basename.c += -Dmain=basename_main
```
cahirwpz, line 189: In theory you could use `eval` function to set `CFLAGS.*.c` variables, but I suspect it's not that easy and may be even impossible. If you feel like trying to do it please refer to: https://www.gnu.org/software/make/manual/html_node/Eval-Function.html

contrib/sbase/Makefile
```diff
@@ -0,0 +1,312 @@
...
+	paste\
+	pathchk\
+	printf\
+	sed\
+	sort\
+	tail\
+	tar\
+	tee
+
+BIN_DONT_WORK =\
```
cahirwpz, line 169: Please explain what are the show stoppers for each of binaries below. We can always provide userspace stubs for non-crucial system calls. Here's the list of currently missing system call implementations: https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/lib/libc/sys/missing.S

cahirwpz: I recognize `date`, `touch` and `xargs` as frequently used. I'd really like to have them in working condition.

contrib/sbase/Makefile
```diff
@@ -0,0 +1,312 @@
+TOPDIR = $(realpath ../..)
+
+.SUFFIXES:
```
cahirwpz, line 3: Why do you need `.SUFFIXES` rule?

contrib/sbase/Makefile
```diff
@@ -0,0 +1,312 @@
...
+SOURCES_LIBUTIL = $(addprefix sbase/,$(LIBUTILSRC))
+SOURCES_LIBUTF = $(addprefix sbase/,$(LIBUTFSRC))
+OBJ_LIBUTIL = $(SOURCES_LIBUTIL:%.c=$(BUILDDIR)%.o)
+OBJ_LIBUTF = $(SOURCES_LIBUTF:%.c=$(BUILDDIR)%.o)
+LIB_SBASE = $(addprefix sbase/,$(LIBUTF)) $(addprefix sbase/,$(LIBUTIL))
+
+PROGRAM = sbase-box
+BINDIR = bin/sbase
+BUILDDIR = build/
+SOURCES = $(SOURCES_SBASE)
```
cahirwpz, line 282: Why do you treat `LIBUTF` and `LIBUTIL` differently than other sources? Wouldn't adding them to `SOURCES` variable simplify this `Makefile`?

lib/libc/gen/readdir.c
```diff
@@ -56,10 +56,10 @@ struct dirent *_readdir_unlocked(DIR *dirp, int skipdeleted) {
       dirp->dd_loc = 0;
     }
     if (dirp->dd_loc == 0 && !(dirp->dd_flags & __DTF_READALL)) {
-      dirp->dd_seek = lseek(dirp->dd_fd, (off_t)0, SEEK_CUR);
```
cahirwpz, line 59: So far `ls` command worked without an issue... why this change is needed?

contrib/sbase/box_src_generator.py
```diff
@@ -0,0 +1,52 @@
...
+            print('#include <libgen.h>')
+            print('#include <stdio.h>')
+            print('#include <stdlib.h>')
+            print('#include <string.h>')
+            print('#include "util.h"')
+
+            for prefix in main_prefixes:
+                print(f'int {prefix}_main(int, char**);')
+
+            print('int main(int argc, char *argv[]) { \
```
cahirwpz, line 22: Please take into account that the output file should still be readable by human being. It would be better if you could make the output formatted and not using weird constructs .

contrib/sbase/box_src_generator.py
```diff
@@ -0,0 +1,52 @@
...
+                char *s = basename(argv[0]);')
+            print(f'if(!strcmp(s,\"{program}\")) {{ \
```
cahirwpz, line 24: Double `{` and `if(0) ;` - that looks nasty...

contrib/sbase/box_src_generator.py
```diff
@@ -0,0 +1,52 @@
...
+                argc--; argv++; s = basename(argv[0]); }} if(0) ;')
+
+            if "test" in tools:
+                print('else if (!strcmp(s, \"[\")) \
```
cahirwpz, line 28: You don't need `else` if each clause returns.

cahirwpz (changes requested):

contrib/sbase/Makefile
panantoni01: Missing syscalls/functions for BIN_DONT_WORK tools are:

- `date` - missing `clock_settime` syscall
- `getconf` - `pathconf` syscall
- `time` - `getrusage` syscall and `_SC_CLK_TCK` argument for `sysconf`
- `touch` - `strptime` libc function. This also requires `tzalloc`, `tzfree`, `tzgetgmtoff`, `tzset` functions and `timezone`, `tzname` variables definitions
- xargs - requires `ARG_MAX` to be >=4096, however increasing the value to 65536 (default value from NetBSD) results in kernel panic (e.g while executing `# ls | xargs` in `/`).

contrib/sbase/Makefile
panantoni01: I think it's just a leftover after some initial work... I'll get rid of it

lib/libc/gen/readdir.c
panantoni01: `ls` still works fine, but there is an issue with `find` - without this change, the last `readdir` call in this loop: https://git.suckless.org/sbase/file/find.c.html#l981 returns 0, but also sets errno to `EINVAL` and we get a lot of confusing `readdir: Invalid argument` messages after each recursive call to `find` function. The errno value in `readdir` is set by the last `lseek` call - after all the directory entries have been read by `getdents`, then the current file read offset is at the end of directory, so `lseek` sets  `EINVAL`.

cahirwpz: This requires further investigation. I think `find` was implemented under assumptions of `readdir` returns errors as Linux `glibc` would do. Obviously there may be some differences between `glibc` and `libc` from NetBSD.

contrib/sbase/Makefile
panantoni01: 4. I managed to add the functions, but adding `tzname`, `timezone` variables is more complex - there are declarations in lib/libc/time/localtime.c and lib/libc/time/private.h, but they are disabled by some ifdefs (e.g https://github.com/cahirwpz/mimiker/blob/master/lib/libc/time/private.h#L529). Thus I need to figure out how to safely enable these declarations

contrib/sbase/Makefile
panantoni01: This seems interesting, I'll try it out

contrib/sbase/Makefile
panantoni01: This was a leftover from the original sbase version of Makefile. I've added libutil and libutf sources to`SOURCES` and simplified the Makefile, so that we don't build `libutil.a` and `libutf.a`.

cahirwpz (changes requested):
Please disable `bin/` and `usr.bin/` tools (by removing them from respective `Makefile`-s ) which are replaced by `sbase` and install them directly into `/bin` directory.

contrib/sbase/patches/format-specifiers.patch
```diff
@@ -0,0 +1,140 @@
+Index: sbase/sbase/cal.c
+===================================================================
+--- sbase.orig/sbase/cal.c	2022-12-07 22:20:25.006463307 +0100
++++ sbase/sbase/cal.c	2022-12-08 00:50:50.422832055 +0100
+@@ -111,7 +111,7 @@
+ 				cur_month -= 12;
+ 				cur_year += 1;
+ 			}
+-			snprintf(month_year, sizeof(month_year), "%s %zu", smon[cur_month], cur_year);
++			snprintf(month_year, sizeof(month_year), "%s %lu", smon[cur_month], cur_year);
```
cahirwpz, line 10: `%zu` is a perfectly correct format specifier. Perhaps we have a problem with our `size_t` definition? After all in NetBSD it's defined as `__SIZE_TYPE__` (through `_BSD_SIZE_T_` macro) and not as `unsigned long` (https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/include/stddef.h#5)

lib/libc/gen/readdir.c
panantoni01: `readdir` in glibc does not return errno if the end of directory is reached unlinke NetBSD's `readdir`, thus suckless's `find` was probably implemented under such assumption. It seems, that we can either add a patch for `find`, so that no confusing warnings are printed, or we can adjust our `readdir` function, so that it does not return `errno==EINVAL` on end of directory.

contrib/sbase/Makefile
panantoni01: I managed to set `CFLAGS.*.c` variables using custom Makefile function and `eval`: https://github.com/cahirwpz/mimiker/pull/1361/commits/b9ceb4a9b2f247ea6ea916a9c97ce23e6fed664a

contrib/sbase/patches/format-specifiers.patch
panantoni01: `size_t` and `ssize_t` issues have been fixed and `format-specifiers.patch` has been removed

contrib/sbase/Makefile
cahirwpz: 4. [strptime](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strptime.html) is a part of POSIX standard. We need to add it sooner or later.

lib/libc/gen/readdir.c
cahirwpz: 1. [readdir](https://pubs.opengroup.org/onlinepubs/9699919799/functions/readdir.html) as defined by POSIX standard does not allow the function to return `EINVAL`
2. Linux [readdir](https://man7.org/linux/man-pages/man3/readdir.3.html) only returns `EBADF`
3. `readdir` from *BSD says (in the manual page) that the function may also fail and set errno for any of the errors specified for the routine [getdents(2)](https://www.freebsd.org/cgi/man.cgi?query=getdents&sektion=2&apropos=0&manpath=FreeBSD+13.1-RELEASE+and+Ports)

IMO we should fix our `readdir` not to propagate unwanted errors from `getdents`. A user should never need to reach directly for `getdents` system call to perform directory reading. Please filter out `errno` on return path from `readdir` function. It must not return any other values than mentioned in POSIX standard.

usr.bin/Makefile
```diff
@@ -2,7 +2,7 @@
 
 TOPDIR = $(realpath ..)
 
-SUBDIR = env id login stat wc script su
+SUBDIR = id login stat script su
```
cahirwpz, line 5: FTR `id`, `login`, `stat` and `su` are part of `ubase`

contrib/sbase/Makefile
panantoni01: 4. I managed to add `tzname` and `timezone` variables - now `strptime` and thus `touch` work fine. However this required some considerable changes in `lib/libc/time/localtime.c` and `lib/libc/time/strptime.c` - https://github.com/cahirwpz/mimiker/pull/1361/commits/1bb5e4eabb2e761615fcc9af6f3fb04077cae1b7.

contrib/sbase/Makefile
```diff
@@ -0,0 +1,215 @@
...
+	sed\
+	sort\
+	tail\
+	tar\
+	time\
+	tee\
+	touch\
+	xargs
+
+BIN_DONT_WORK =\
```
cahirwpz, line 168: Please merge `BIN_DONT_WORK`, `BIN_QUESTIONABLE` and `BIN_SURELY_DISABLED` into single `BIN_DISABLED` variable.

contrib/sbase/Makefile
```diff
@@ -0,0 +1,215 @@
+TOPDIR = $(realpath ../..)
+
+HDR =\
```
cahirwpz, line 3: In all `Makefile` files I remember we put extra space before backslash (purely cosmetic remark).

lib/libc/gen/readdir.c
cahirwpz: Please put `lseek` into original position, but clear `errno` afterwards.

panantoni01: `lseek` is back in the original position and we don't return `errno` on end of directory: https://github.com/cahirwpz/mimiker/pull/1361/commits/0a28ff65ea3bd519f835bbbeb1a279b98113bc49

lib/libc/time/localtime.c
```diff
@@ -317,13 +299,9 @@ static void scrub_abbrs(struct state *sp) {
 
 static void update_tzname_etc(const struct state *sp,
                               const struct ttinfo *ttisp) {
-#if HAVE_TZNAME
```
cahirwpz, line 320: Wouldn't it be better (less changes) if you forced [HAVE_TZNAME](https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/lib/libc/time/private.h?r=10da8877#208) and [USG_COMPAT](https://mimiker.ii.uni.wroc.pl/source/xref/mimiker/lib/libc/time/private.h?r=10da8877#200) to be defined?

panantoni01: I've done it in such a way in https://github.com/cahirwpz/mimiker/pull/1361/commits/1b55b6fab21ebe4e6a9c112583dc0cd1014b0499.

cahirwpz (approved):
LGTM

PR #1363, opened 2022-11-30
Save additional information to signals

franciscozdo:
In this PR I pass additional information to use in signal handlers.

* make `sig_trap` common (all architectures had the same implementation)
* Import some signal codes from other implementations (to fill in then `si_code` field in siginfo structure)


cahirwpz (approved):
Looks mostly ok to me.

Some remarks:
* the code translating errno to signo/sigcode is repeating between `trap.c` files
* sys/signal.h is not formatted with clang-format
* sys/gen/signal.c could be moved to sys/signal.c since it looks mostly machine independant

cahirwpz:
@pwit81 Please have a look at this PR. You're going to need functionality it has introduced.

PR #1364, opened 2022-12-01
Use sigaction in all tests that need to handle SIGSEGV

franciscozdo:
I want to use code that I have prepared in previous PR to make better checks for SIGSEGV in all user tests.


bin/utest/mmap.c
```diff
@@ -94,15 +94,26 @@ static void munmap_good(void) {
...
+  if (sigsetjmp(return_to, 1) == 0) {
+    /* Try to access freed memory. It should raise SIGSEGV */
+    int data = *((volatile int *)(addr + 0x2000));
+    (void)data;
+    assert(0);
+  }
+
+  /* Check if SIGSEGV was handled correctly */
+  assert(sigsegv_address == (addr + 0x2000));
+  assert(sigsegv_code = SEGV_MAPERR);
```
cahirwpz, line 113: `=` vs `==` bug here ;-)

bin/utest/mmap.c
```diff
@@ -94,15 +94,26 @@ static void munmap_good(void) {
   assert(result == 0);
 }
 
-/* Don't call this function in this module */
 int test_munmap_sigsegv(void) {
```
cahirwpz, line 97: I think that with clever code shuffling you could get nice abstraction for signal handling and rewrite code to something like this:

```c
int test_munmap_sigsegv(void) {
  void *addr = mmap_anon_prw(NULL, 0x4000);
  munmap(addr, 0x4000);

  siginfo_t si;
  if (expect_signal(SIGSEGV, &si)) {
    /* Try to access freed memory. It should raise SIGSEGV */
    int data = *((volatile int *)(addr + 0x2000));
    (void)data;
  }

  /* Check if SIGSEGV was handled correctly */
  assert(got_signal(SIGSEGV));
  assert(si.si_addr == (addr + 0x2000));
  assert(si.si_code == SEGV_MAPERR);

  return 0;
}
```

Beware it can require some macrology around `expect_signal` since it should call sigsetjmp directly and not as subfunction. Please refer to https://gcc.gnu.org/onlinedocs/gcc-12.2.0/gcc/Statement-Exprs.html#Statement-Exprs

cahirwpz (commented):
I think you could spend a little bit more time to provide good abstraction for signal testing.

bin/utest/mmap.c
franciscozdo: Okay. I have prepared sth like that.

bin/utest/util.c
```diff
@@ -1,10 +1,45 @@
...
+  memset(siginfo_ptr, 0, sizeof(siginfo_t));
+  sa.sa_sigaction = signal_handler;
+  sa.sa_flags = SA_SIGINFO;
+  int err = sigaction(_handled_signo, &sa, &_old_sa);
+  assert(err == 0);
+  _save_info = siginfo_ptr;
+}
+
+void check_signal(int signo, void *addr, int code) {
+  assert(_save_info->si_signo == signo);
```
cahirwpz, line 33: Since `assert` is defined as macro and uses `__FILE__` and `__LINE__` to report when the assertion failed, all messages originating from `check_signal` will point to `util.c` lines 33-35. That would makes those diagnostic messages quite useless.

bin/utest/vm_map_prot.c
```diff
@@ -83,68 +54,40 @@ int test_vmmap_data_x(void) {
...
+  /* Check execute */
   func_int_t fun = (func_int_t)ro_data_str;
-
   printf("func: 0x%p\n", fun);
-
-  if (sigsetjmp(return_to, 1) == 0) {
+  TEST_EXPECT_SIGNAL(SIGSEGV, &si, fun, SEGV_ACCERR) {
     /* Executing from rodata segment. */
     fun(1);
     assert(0);
```
cahirwpz, line 89: Why `assert(0)` here (and in many other places)? Is it possible to move it to `TEST_EXPECT_SIGNAL` macro, so the loop will perform `exit(1)` if the program has not received the signal?

bin/utest/signal.c
```diff
@@ -67,11 +68,14 @@ int test_signal_abort(void) {
...
 
 /* ======= signal_segfault ======= */
-/* This test shall be considered success if the process gets terminated with
-   SIGABRT */
 int test_signal_segfault(void) {
   volatile struct { int x; } *ptr = 0x0;
-  ptr->x = 42;
+
+  siginfo_t si;
+  TEST_EXPECT_SIGNAL(SIGSEGV, &si, ptr, SEGV_MAPERR) {
```
cahirwpz, line 75: That looks much better but... since you pass extra arguments to `TEST_EXPECT_SIGNAL` it cannot be used to test the reception of other signals – think of @pwit81 who will continue work on `SIGILL`, `SIGTRAP`, `SIGBUS` and so on.

After the loop you could use another macro `GOT_SIGSEGV(&si, ptr, SEGV_MAPERR)`.

Or perhaps you could build `EXPECT_SIGSEGV` on top of `EXPECT_SIGNAL` ?

bin/utest/util.h
```diff
@@ -14,3 +16,20 @@ void wait_for_signal(int signo);
...
  * slave side. */
 void open_pty(int *master_fd, int *slave_fd);
+
+extern jmp_buf signal_return;
+
+void setup_handler(int signo, siginfo_t *siginfo_ptr);
+void check_signal(int signo, void *addr, int code);
+void restore_handler(void);
+
+#define TEST_EXPECT_SIGNAL(signo, siginfo_ptr, addr, code)                     \
```
cahirwpz, line 26: I think you can shorten it to just `EXPECT_SIGNAL` and it'll be ok.

cahirwpz (commented):
Looks much better but with little effort you could make `TEST_EXPECT_SIGNAL` a little bit more useful.

cahirwpz (approved):
LGTM

PR #1368, opened 2022-12-19
New functions and changes needed for suckless sbase

panantoni01:
This PR adds functions, defines and fixes needed for porting suckless sbase:
- POSIX ARG_MAX, NAME_MAX, PATH_MAX values
- fmemopen()
- llabs()
- add missing include in include/sys/ioctl.h
- fix for aarch64/bcopy.S (memmove function)
- creat() - wrapper for open syscall

### DIFF
--- a/lib/libc/string/aarch64/bcopy.S
+++ b/lib/libc/string/aarch64/bcopy.S
@@ -278,7 +278,7 @@ backward_tiny:
 	ret
 9:
 	/* length is small(<32), and src or dst may be unaligned */
-	eor	TMP_X, SRC0, DST0
+	eor	TMP_X, SRC0, DST
 	ands	TMP_X, TMP_X, #7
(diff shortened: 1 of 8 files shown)


lib/libc/string/aarch64/bcopy.S
```diff
@@ -278,7 +278,7 @@ backward_tiny:
 	ret
 9:
 	/* length is small(<32), and src or dst may be unaligned */
-	eor	TMP_X, SRC0, DST0
+	eor	TMP_X, SRC0, DST
```
cahirwpz, line 281: Since this attempts to fix a bug that is possibly present in NetBSD please provide short explanation before this line, i.e. what are the symptoms, when the bug is triggered and why this change is supposed to fix it.

cahirwpz (approved):
LGTM

PR #1369, opened 2022-12-27
Tests with signals upgrade

pwit81:
This PR removes all remaining usages of `UTEST_ADD_SIGNAL` macro and removes the macro itself. Affected tests are rewritten to use `EXPECT_SIGNAL`/`CLEANUP_SIGNAL` macro pair.


franciscozdo (changes requested):
I think you could also add detailed checks of fields returned in `siginfo` structure. I think that macros similar to `CHECK_SIGSEGV` should be enough. You can check for description of the expected values for each signal on the [siginfo manual page](https://www.freebsd.org/cgi/man.cgi?query=siginfo&apropos=0&sektion=0&manpath=FreeBSD+13.1-RELEASE+and+Ports&arch=default&format=html).

cahirwpz:
> I think you could also add detailed checks of fields returned in `siginfo` structure. I think that macros similar to `CHECK_SIGSEGV` should be enough. You can check for description of the expected values for each signal on the [siginfo manual page](https://www.freebsd.org/cgi/man.cgi?query=siginfo&apropos=0&sektion=0&manpath=FreeBSD+13.1-RELEASE+and+Ports&arch=default&format=html).

@pwit81 I agree with @franciscozdo. Alternatively this could be done in follow-up PR.

cahirwpz:
@franciscozdo After a discussion with @pwit81 we decided to move further work on testing cpu exceptions & traps to separate PR.

PR #1371, opened 2022-12-28
Add tools from suckless ubase

panantoni01:
This PR adds tools from suckless [ubase](https://git.suckless.org/ubase/) in a similar way to sbase (https://github.com/cahirwpz/mimiker/pull/1361)


cahirwpz (changes requested):
Please move `libc` and kernel related changes to two separate PRs.

sys/kern/main.c
```diff
@@ -40,6 +40,19 @@ static void mount_fs(void) {
   do_fchmodat(p, AT_FDCWD, "/tmp", ACCESSPERMS | S_ISTXT, 0);
 }
 
+static void open_fds(void) {
```
cahirwpz, line 43: Since that code was moved to #1390 we can safely remove it from this PR.

panantoni01: Reverted all the changes in `sys/kern.main.c` in https://github.com/cahirwpz/mimiker/pull/1371/commits/ac3245cf093932672daac3f36f3689a38e062b11

cahirwpz (commented):
First batch of comments.

contrib/ubase/patches/watch.patch
```diff
@@ -0,0 +1,31 @@
+Index: ubase/ubase/watch.c
+===================================================================
+--- ubase.orig/ubase/watch.c	2023-01-01 22:32:22.325582802 +0100
++++ ubase/ubase/watch.c	2023-01-01 22:33:42.278913458 +0100
+@@ -17,7 +17,7 @@
+ {
+ 	char cmd[BUFSIZ];
+ 	char *end;
+-	useconds_t interval = 2 * 1E6;
```
cahirwpz, line 9: [useconds_t](http://bxr.su/xref/NetBSD/sys/sys/types.h#324) should be defined in `sys/types.h` as `unsigned int`.

contrib/ubase/patches/watch.patch
```diff
@@ -0,0 +1,31 @@
...
++		interval = period;
+ 		break;
+ 	default:
+ 		usage();
+@@ -52,7 +52,7 @@
+ 		printf("\x1b[2J\x1b[H"); /* clear */
+ 		fflush(NULL);
+ 		system(cmd);
+-		usleep(interval);
++		sleep(interval);
```
cahirwpz, line 28: `usleep` is defined in `unistd.h`

contrib/ubase/patches/su.patch
```diff
@@ -0,0 +1,42 @@
...
+ static int pflag = 0;
+@@ -21,9 +22,9 @@
+ static int
+ dologin(struct passwd *pw)
+ {
+-	char *shell = pw->pw_shell[0] == '\0' ? "/bin/sh" : pw->pw_shell;
++	const char *shell = pw->pw_shell[0] == '\0' ? "/bin/sh" : pw->pw_shell;
+ 	char *term = getenv("TERM");
+-	clearenv();
++	environ = envinit;
```
cahirwpz, line 21: What about defining `clearenv` in `util.h` and adding it to `libutil/`?

contrib/ubase/patches/respawn.patch
```diff
@@ -0,0 +1,88 @@
...
+-		pollset->fd = open(fifo, O_RDONLY | O_NONBLOCK);
+-		if (pollset->fd < 0)
+-			eprintf("open %s:", fifo);
+-		pollset->events = POLLIN;
+-	}
+-
+ 	while (1) {
+-		if (fifo) {
+-			pollset->revents = 0;
+-			polln = poll(pollset, 1, -1);
```
cahirwpz, line 57: As a challenge please consider implementing `poll` emulation in similar way `select` was added here https://github.com/cahirwpz/mimiker/blob/master/lib/libc/sys/select.c

contrib/ubase/patches/libutil-passwd.patch
```diff
@@ -0,0 +1,46 @@
...
+-		errno = 0;
+-		spw = getspnam(pw->pw_name);
+-		if (!spw) {
+-			if (errno)
+-				weprintf("getspnam: %s:", pw->pw_name);
+-			else
+-				weprintf("who are you?\n");
+-			return -1;
+-		}
+-		p = spw->sp_pwdp;
```
cahirwpz, line 37: From [getpwent(3)](https://man.netbsd.org/NetBSD-9.0/getpwent.3):

> These functions have been written to ``shadow'' the password file, e.g. allow only certain programs to have access to the encrypted password.  If the process which calls them has an effective uid of 0, the encrypted password will be returned, otherwise, the password field of the returned structure will point to the string `*'.

You can fetch `p` here from `pw->pw_passwd`.

contrib/ubase/patches/getty.patch
```diff
@@ -0,0 +1,103 @@
+Index: ubase/ubase/getty.c
+===================================================================
+--- ubase.orig/ubase/getty.c
++++ ubase/ubase/getty.c
+@@ -10,13 +10,12 @@
+ #include <stdlib.h>
+ #include <string.h>
+ #include <unistd.h>
+-#include <utmp.h>
```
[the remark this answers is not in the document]
panantoni01: Added utmp.c in https://github.com/cahirwpz/mimiker/pull/1371/commits/7fe649d47f666e3fb06240dd2bd319f926aef47c

contrib/ubase/patches/getty.patch
```diff
@@ -0,0 +1,103 @@
...
+-	if (fchmod(fd, 0600) < 0)
+-		weprintf("fchmod %s:", tty);
+ 	if (fd > 2)
+ 		close(fd);
+ 
+@@ -88,37 +74,14 @@ main(int argc, char *argv[])
+ 	sigemptyset(&sa.sa_mask);
+ 	sigaction(SIGHUP, &sa, NULL);
+ 
+-	/* Clear all utmp entries for this tty */
```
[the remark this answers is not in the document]
panantoni01: Restored in https://github.com/cahirwpz/mimiker/pull/1371/commits/7ed288d7fe76e46da1c165f5407f858b89173766

contrib/ubase/patches/getty.patch
```diff
@@ -0,0 +1,103 @@
...
+-		} while (1);
+-		if (ferror(fp))
+-			weprintf("%s: I/O error:", UTMP_PATH);
+-		fclose(fp);
+-	}
+-
+ 	if (argc > 2)
+ 		return execvp(argv[2], argv + 2);
+ 
+-	if (gethostname(hostname, sizeof(hostname)) == 0)
```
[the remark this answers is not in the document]
panantoni01: Added `gethostname` implementation in https://github.com/cahirwpz/mimiker/pull/1371/commits/93ad9666867a430a378aa52faba93dcf61617393 and restored the original `getty` source code in https://github.com/cahirwpz/mimiker/pull/1371/commits/c36b315e3a8f6fc559c50e9efc21a204b8c575b7. Also, `HOST_NAME_MAX` definition was needed - https://github.com/cahirwpz/mimiker/pull/1371/commits/c10e63a542576f54fc79f87642fb8b3dbe09a100.

contrib/ubase/patches/watch.patch
panantoni01: Using `usleep()` to sleep for more than 1 second is considered an error in Mimiker and NetBSD (https://github.com/cahirwpz/mimiker/blob/master/lib/libc/gen/usleep.c#L46). I assumed that a user may want to use `watch` program with an interval longer than 1s, so I decided to use `sleep()` here, which does not have this limitation.

contrib/ubase/patches/su.patch
panantoni01: I'm not sure in which file in `ubase/libutil/` should I implement this function. 
I tried to create a new separate `ubase/libutil/clearenv.c` file for it by adding a new `quilt` patch and add it to `LIBUTILSRC` in https://github.com/panantoni01/mimiker/blob/ubase/contrib/ubase/Makefile#L22. However with this approach the new source file cannot compiled during the build process, because quilt patches for ubase are applied **after** generating `*.d` files for the ubase object files.

contrib/ubase/patches/getty.patch
```diff
@@ -0,0 +1,103 @@
...
+ #include <string.h>
+ #include <unistd.h>
+-#include <utmp.h>
+ 
+ #include "config.h"
+ #include "util.h"
+ 
+-static char *tty = "/dev/tty1";
+-static char *defaultterm = "linux";
++static char *tty = "/dev/uart";
```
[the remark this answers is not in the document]
panantoni01: It seems to me, that `_PATH_TTY` should be left as it is, because this should be the definition of a path representing controlling terminal for each process (in NetBSD we have the same value: http://bxr.su/NetBSD/include/paths.h#98) and we shouldn't change it to a device file name.

contrib/ubase/patches/libutil-passwd.patch
panantoni01: I'm not sure I understand what is to do here. I deleted this part of `pw_check` because it handles a case of a shadowed password. But since we don't have password shadowing (e.g `getspnam` function) implemented in Mimiker, I decided to remove this source code fragment.

cahirwpz: BSDs do have password shadowing but done in a more elegant way than in Linux. `master.passwd` is equivalent to `shadow`. However all the data including encrypted passwords is available through single call to `getpwent`. The implementation chooses between `/etc/passwd` and `/etc/master.passwd` based on `euid`.

Please note that our `getpwent` function is incomplete and lacks support for `/etc/passwd`.

contrib/ubase/patches/su.patch
cahirwpz: Ouch... maybe implement it as `static inline` in `util.h` then? Since the functionality is used more than once I'd like to keep it in a single place.

contrib/ubase/Makefile
```diff
@@ -0,0 +1,135 @@
...
+	libutil/proc.c           \
+	libutil/putword.c        \
+	libutil/recurse.c        \
+	libutil/strlcat.c        \
+	libutil/strlcpy.c        \
+	libutil/strtonum.c       \
+	libutil/tty.c
+
+LIBUTILSRC_DONT_WORK = \
+	libutil/agetcwd.c \
```
cahirwpz, line 40: Is there any reason why you skipped those?

contrib/ubase/patches/getty.patch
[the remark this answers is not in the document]
cahirwpz: Hmm... I'm afraid we don't have any support for `/dev/tty`. Would you be interested in implementing it? My best guess is that it should not be more complex than our `dev_procstat.c`. PTAL at [tty_tty.c](http://bxr.su/FreeBSD/sys/kern/tty_tty.c) from FreeBSD and [tty_tty.c](http://bxr.su/NetBSD/sys/kern/tty_tty.c) from NetBSD.

If you decide to implement it then we should meet to figure out what is the right strategy to follow.

contrib/ubase/patches/watch.patch
panantoni01: It seems to me, that it's better to have `interval` here as `unsigned int`, since we've decided to use `sleep()` (which takes `unsigned int` as an argument) instead of `usleep()` (which takes `useconds_t`). If we left `useconds_t` here, it would indicate that `interval` is a value in microseconds, which could be misleading.

cahirwpz: I agree! This makes sense.

contrib/ubase/Makefile
panantoni01: `agetcwd()` function uses `apathmax()`, which needs `pathconf()`. This is a syscall (https://man.netbsd.org/pathconf.2), that we don't have implemented.

contrib/ubase/patches/su.patch
panantoni01: Added `clearenv` as `static inline` in a separate patch for `util.h` in https://github.com/cahirwpz/mimiker/pull/1371/commits/122906fd65f14419f8638a3346bd2dc9ca12f5d6

contrib/ubase/Makefile
cahirwpz: The system call should be relatively easy to add. Have a look at [sys_pathconf](http://bxr.su/NetBSD/sys/kern/vfs_syscalls.c#sys_pathconf) in NetBSD. Through redirection layers it reaches filesystem specific `VOP_PATHCONF` for instance [ufs_pathconf](http://bxr.su/NetBSD/sys/ufs/ufs/ufs_vnops.c#ufs_pathconf). Let me know if you'd like to implement it.

contrib/ubase/patches/getty.patch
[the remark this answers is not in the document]
panantoni01: I think that actually we do have some support for `/dev/tty`- https://github.com/cahirwpz/mimiker/blob/master/sys/kern/tty.c#L1177. We even use this device file while getting a password from a user - https://github.com/cahirwpz/mimiker/blob/master/lib/libc/gen/getpass.c#L96.

contrib/ubase/patches/login.patch
```diff
@@ -0,0 +1,84 @@
+Index: ubase/ubase/login.c
+===================================================================
+--- ubase.orig/ubase/login.c	2023-05-30 17:51:24.466728782 +0200
++++ ubase/ubase/login.c	2023-05-30 17:54:52.876734143 +0200
+@@ -10,45 +10,21 @@
+ #include <string.h>
+ #include <time.h>
+ #include <unistd.h>
+-#include <utmp.h>
```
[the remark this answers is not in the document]
panantoni01: In `login` we have this `writeutmp` function which is called after a successful login ([ubase/login.c](https://git.suckless.org/ubase/file/login.c.html#l120)). However in Mimiker we don't have `ttyname` implemented ([ttyname.c](https://github.com/cahirwpz/mimiker/blob/master/lib/libc/gen/ttyname.c)) which is necessary for `writeutmp` to work correctly, so I needed to remove the `writeutmp` call and its definition (otherwise we get ` error: unused function` during compilation).

contrib/ubase/patches/getty.patch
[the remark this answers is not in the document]
cahirwpz: Ok. I think I needed that kind of refresher. Seems like my memory is good but short ;-)

Since `tty` variable is a default value for terminal device let's leave it as is.

contrib/ubase/Makefile
cahirwpz: Sounds great! Please create a GitHub issue for `pathconf` syscall and describe there how you'd approach the implementation.

contrib/ubase/patches/respawn.patch
cahirwpz: Is it something you'd like to commit to?

contrib/ubase/patches/login.patch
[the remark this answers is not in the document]
cahirwpz: What about fixing output from `ttyname` as follows:
```c
if (!tty)
    tty = "unknown";
```

contrib/ubase/patches/respawn.patch
cahirwpz: This has potentially a big impact since `poll` functionality is used in wide variety of programs.

panantoni01: > Is it something you'd like to commit to?

For now, I'd rather start dealing with `pathconf` and `TIOCNOTTY` ioctl (https://github.com/cahirwpz/mimiker/pull/1390#issuecomment-1596098488) and then try to implement `poll`.

contrib/ubase/patches/login.patch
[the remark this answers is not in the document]
panantoni01: I added `tty = "unknown";` and ported `writeutmp` in https://github.com/cahirwpz/mimiker/pull/1371/commits/369a272be9c5051134d5dbcc9445fb20185fddf4.

contrib/ubase/Makefile
panantoni01: I've added the issue here: https://github.com/cahirwpz/mimiker/issues/1397

contrib/ubase/patches/respawn.patch
cahirwpz: Since `pathconf` and `TIOCNOTTY` PRs have already been merged, we're left only with `poll` I guess.

panantoni01: Since `poll` has been recently implemented, I restored almost entire `respawn.c` original source code (https://github.com/cahirwpz/mimiker/pull/1371/commits/e05e66377e665e827ebd198fcdd7b4fb95c44860).

contrib/ubase/patches/libutil-passwd.patch
panantoni01: It seems to me that the idea here was to simply use `getpwnam()` instead of `getspnam()` - https://github.com/cahirwpz/mimiker/pull/1371/commits/5cc5f83eba2987771a43998e704d3d24e7e1e729.

contrib/ubase/Makefile
```diff
@@ -0,0 +1,133 @@
...
+	dmesg             \
+	eject             \
+	fallocate         \
+	free              \
+	freeramdisk       \
+	fsfreeze          \
+	halt              \
+	hwclock           \
+	insmod            \
+	killall5          \
```
cahirwpz, line 69: What stops us from porting `killall5` tool?

contrib/ubase/Makefile
```diff
@@ -0,0 +1,133 @@
...
+	last              \
+	lastlog           \
+	lsmod             \
+	lsusb             \
+	mesg              \
```
cahirwpz, line 74: What stops us from porting `mesg` tool?

contrib/ubase/Makefile
```diff
@@ -0,0 +1,133 @@
...
+	libutil/putword.c        \
+	libutil/recurse.c        \
+	libutil/strlcat.c        \
+	libutil/strlcpy.c        \
+	libutil/strtonum.c       \
+	libutil/tty.c
+
+BIN = \
+	clear             \
+	getty							\
```
cahirwpz, line 43: Formatting issues - stray backslash sign.

cahirwpz (commented):
We're very close for this PR to be ready for a merge.

contrib/ubase/patches/respawn.patch
```diff
@@ -0,0 +1,13 @@
+Index: ubase/ubase/respawn.c
+===================================================================
+--- ubase.orig/ubase/respawn.c	2023-08-31 00:14:55.569483078 +0200
++++ ubase/ubase/respawn.c	2023-08-31 00:22:33.282805231 +0200
+@@ -6,7 +6,7 @@
+ 
+ #include <errno.h>
+ #include <fcntl.h>
+-#include <poll.h>
++#include <sys/poll.h>
```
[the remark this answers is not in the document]
panantoni01: Added this file in https://github.com/cahirwpz/mimiker/pull/1371/commits/419049bd8efdb4122622121c00554c58d54c943d and removed `respawn.patch` in https://github.com/cahirwpz/mimiker/pull/1371/commits/1c345cd6a499c56ee0f0dbd06b6a100b349fd1c2

contrib/ubase/Makefile
panantoni01: Fixed in https://github.com/cahirwpz/mimiker/pull/1371/commits/9d96f1da00f1d03baa86b9a9d4001e679e4e6f00

contrib/ubase/Makefile
panantoni01: I added this tool (https://github.com/cahirwpz/mimiker/pull/1371/commits/4c69c2dd6880688b0b34c2eb726850afc236c9ff); it required a small fix in the source code (https://github.com/cahirwpz/mimiker/pull/1371/commits/c200ca70abccd5599d291ef08b3c0ad2e18e3ae8) due to `-Wsometimes-uninitialized` error.

etc/master.passwd
```diff
@@ -1,4 +1,4 @@
-root:root:0:0::0:0:root:/root:/bin/ksh
+root:root:0:0::0:0:root:/:/bin/ksh
```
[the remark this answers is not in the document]
panantoni01: I added `/root` directory and mounted `tmpfs` there (https://github.com/cahirwpz/mimiker/pull/1371/commits/4d66700759b0ff4677373b4656928e842e291b96) - this is now the default home directory for `root` user.
By the way I'm not sure that adding empty `bin/utest/extra/root/.keep` file was a correct way to create a `/root` directory in `initrd.cpio`, but I saw no solution that could be more elegant.

contrib/ubase/Makefile
panantoni01: `killall5` implementation in `ubase` is dependent on `proc` pseudo filesystem (https://git.suckless.org/ubase/file/killall5.c.html#l89)

cahirwpz (approved):
LGTM

cahirwpz:
@pj1031999 Could you review the PR?

pj1031999 (approved):

PR #1372, opened 2023-01-02
Add sinit submodule in contrib/sinit

panantoni01:
The goal of this PR is to add sinit as a submodule in contrib/sinit and build it according to the new method of porting external software. Changes made:

- add sinit submodule
- etc - add rc.init rc.shutdown
- config.patch - change rc.* files path
- sigwait function and sigtimedwait syscall stub


etc/rc.init
```diff
@@ -0,0 +1,15 @@
+
```
cahirwpz, line 1: I think shebang won't be recognized if there are preceding white spaces in a file.

cahirwpz (commented):
`rc.init` needs a small fix, otherwise looks good to me.

cahirwpz (approved):
LGTM

PR #1373, opened 2023-01-04
Fix kernel memory allocation on large blocks

panantoni01:
ARG_MAX value is changed from 3072 to 65536, `kva_map` and `kfree` are fixed


cahirwpz (approved):
LGTM

PR #1374, opened 2023-01-06
Userspace test framework overhaul (part 1)

cahirwpz:
* add TEST_ADD macro to define new tests
* add set of macros in utest.h to perform checks
* shorten duration of some tests that used sleep/nanosleep
* add ualarm implementation from NetBSD
* allow klog to be dumped to file
* each user test logs start time
* launch: correctly report test timeout
* launch: print output from console on test run failure or timeout

### DIFF
--- a/bin/utest/cred.c
+++ b/bin/utest/cred.c
@@ -7,40 +7,34 @@
 
 TEST_ADD(get_set_uid) {
   uid_t ruid, euid, suid;
-  int error;
 
   /* check if nothing fail if we put wrong addresses */
-  error = getresuid((void *)1, (void *)1, (void *)1);
-  assert(error < 0 && errno == EFAULT);
+  syscall_fail(getresuid((void *)1, (void *)1, (void *)1), EFAULT);
 
-  error = getresuid(&ruid, &euid, &suid);
-  assert(error == 0);
+  syscall_ok(getresuid(&ruid, &euid, &suid));
 
   /* assume we are running tests as root */
   assert(ruid == 0 && euid == 0 && suid == 0);
 
   /* as root we chan change id to ony other */
   ruid = 1, euid = 2, suid = 3;
-  error = setresuid(ruid, euid, suid);
-  assert(error == 0);
+  syscall_ok(setresuid(ruid, euid, suid));
 
-  getresuid(&ruid, &euid, &suid);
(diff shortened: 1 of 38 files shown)


bin/utest/cred.c
```diff
@@ -1,155 +1,130 @@
...
-int test_get_set_groups(void) {
+TEST_ADD(get_set_groups) {
   gid_t rgrp[NGROUPS_MAX], gidset[NGROUPS_MAX] = {0, 1, 2, 3, 4, 5};
-  int r, ngroups = 6;
+  const int ngroups = 6;
   uid_t euid;
 
   /* check if we are a root at start */
   getresuid(NULL, &euid, NULL);
+  // TODO(fzdob): this fails with EFAULT
```
cahirwpz, line 88: @franciscozdo Why this function fails with EFAULT?

bin/utest/mmap.c
```diff
@@ -56,53 +53,39 @@ static void mmap_with_hint(void) {
...
-  result = munmap(addr, 0x1000);
-  assert_ok(result);
+  syscall_ok(munmap(addr, 0x1000));
 
   /* munmapping again fails */
-  result = munmap(addr, 0x1000);
-  printf("result: %d\n", result);
-  assert(errno == EINVAL);
-  // assert_fail(result, EINVAL);
+  syscall_ok(munmap(addr, 0x1000));
```
cahirwpz, line 72: @franciscozdo Should it return `EINVAL`?

bin/utest/cred.c
franciscozdo: I can make an PR that fixes that return value for `getresuid` (and for `getresgid` where is the same problem).

bin/utest/mmap.c
franciscozdo: No, that was mistake, and I forgot to remove that line when I was cleaning previous changes.

bin/utest/cred.c
franciscozdo: The manual entry I had referred to: https://www.freebsd.org/cgi/man.cgi?query=getresuid&sektion=2&n=1

cahirwpz: Please fix it when this PR is merged. I'll leave TODO comment to be removed when the bug is fixed.

PR #1376, opened 2023-01-07
Fix error returned from getresuid and getresgid 

franciscozdo:
`getresuid` and `getresgid` return EFAULT when some address passed as argument is invalid. NULL value is also treated as invalid addres. This PR allows to pass NULL value as an argument for these syscalls (to retrieve only uids or gids for which non-NULL address is provided).

In manual entry ([getresuid(2)](https://www.freebsd.org/cgi/man.cgi?query=getresuid&sektion=2&n=1)) there is not specified what should happen if we pass NULL addres to these functions. Maybe that is the reason why we had choose to return error in such cases in the past.

### DIFF
--- a/bin/utest/cred.c
+++ b/bin/utest/cred.c
@@ -25,7 +25,8 @@ int test_get_set_uid(void) {
   error = setresuid(ruid, euid, suid);
   assert(error == 0);
 
-  getresuid(&ruid, &euid, &suid);
+  error = getresuid(&ruid, &euid, &suid);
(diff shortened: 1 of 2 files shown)


cahirwpz:
Please resolve conflicts. Take into account recent changes to user test infrastructure introduced by #1374

franciscozdo:
@cahirwpz I have resolved conflicts and I think it's ready.

cahirwpz (approved):
LGTM

PR #1378, opened 2023-01-07
[sleepq] Add a mutex to the sleepq wait interface.

MichalBlk:
As the current sleepq interface lacks any mechanism to lock the destination chain before calling the appropriate wait function, we at least need a way to release a provided lock before falling asleep.
The `sleepq_lock` mechanism will be introduced as part of merging #1367.

Required by `sigtimedwait` (see #1320).


cahirwpz (approved):
LGTM

PR #1379, opened 2023-02-23
[vm] Simplify virtual memory subsystem

franciscozdo:
This PR introduces amaps and removes vm_objects from virtual memory subsystem.

Current implementation is complicated and we don't need that now. If we replace vm_objects with amaps implementation will be simpler. It will also be easier to implement more complicated operations on virtual memory.


sys/kern/vm_amap.c
```diff
@@ -0,0 +1,164 @@
...
+
+  return ENOMEM;
+}
+
+vm_amap_t *vm_amap_clone(vm_aref_t aref) {
+  vm_amap_t *amap = aref.amap;
+  if (!amap)
+    return NULL;
+
+  vm_amap_t *new = vm_amap_alloc(amap->slots - aref.offset + 16);
```
pj1031999, line 92: null pointer dereference. vm_amap_alloc can fail

sys/kern/vm_map.c
```diff
@@ -416,19 +436,31 @@ vm_map_t *vm_map_clone(vm_map_t *map) {
...
+      if (it->aref.amap) {
+        if (it->flags & VM_ENT_SHARED) {
+          vm_amap_hold(it->aref.amap);
+          ent->aref = it->aref;
+        } else {
+          ent->aref.offset = 0;
+          ent->aref.amap = vm_amap_clone(it->aref);
+          if (!ent->aref.amap) {
+            klog("Unable to clone amap!");
+          }
```
pj1031999, line 460: from this point - is ent a valid entry or not? because you ignore this  and always put ent into new map entries

include/sys/vm_amap.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <machine/vm_param.h>
+
+typedef struct vm_amap vm_amap_t;
+typedef struct vm_aref vm_aref_t;
+
+struct vm_aref {
+  int offset;      /* offset in slots */
+  vm_amap_t *amap; /* underlying amap */
+};
```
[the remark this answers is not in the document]
franciscozdo: I have to expose it because I use it inside `vm_map_entry` structure. I don't want to store it there as pointer.

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,164 @@
...
+    pool_free(P_VM_AMAP_STRUCT, amap);
+    return NULL;
+  }
+
+  amap->ref_cnt = 1;
+  amap->slots = slots;
+  return amap;
+}
+
+int vm_amap_extend(vm_aref_t *aref, int slots) {
```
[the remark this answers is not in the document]
franciscozdo: I have removed it. For now I don't want to resize amaps. When we resize entry it is possible to do it if underlying amap is big enough (that's the reason why I add some constant to amap size during the allocation).

I am also not sure if resizing amap is easy or even possible. There might be a problem when amap is used in shared entry. I don't want to deal with it in this PR.

include/sys/vm_physmem.h
```diff
@@ -16,6 +16,9 @@ void _vm_physseg_plug(paddr_t start, paddr_t end, bool used);
 /* Allocates contiguous big page that consists of n machine pages. */
 vm_page_t *vm_page_alloc(size_t n);
 
+/* Allocates pages as vm_page_alloc and zreos them. */
+vm_page_t *vm_page_alloc_zero(int npages);
```
cahirwpz, line 20: `size_t n` as in function above?

include/sys/vm_amap.h
```diff
@@ -0,0 +1,43 @@
+#ifndef _SYS_VM_AMAP_H_
+#define _SYS_VM_AMAP_H_
+
+#include <machine/vm_param.h>
+
+typedef struct vm_amap vm_amap_t;
+typedef struct vm_aref vm_aref_t;
+
+struct vm_aref {
+  int offset;      /* offset in slots */
```
cahirwpz, line 10: Why not `uintptr_t` here?

include/sys/vm_amap.h
```diff
@@ -0,0 +1,43 @@
...
+  vm_amap_t *amap; /* underlying amap */
+};
+
+#define VM_AREF_EMPTY ((vm_aref_t){.offset = 0, .amap = NULL})
+
+/* Allocate new amap with specified number of slots. */
+vm_amap_t *vm_amap_alloc(int slots);
```
cahirwpz, line 17: Why not `size_t` if slots cannot be negative?

include/sys/vm_amap.h
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Returns new amap with ref_cnt equal to 1.
+ */
+vm_amap_t *vm_amap_clone(vm_aref_t aref, int slots);
+
+/* Operations on amap's ref_cnt. */
+void vm_amap_hold(vm_amap_t *amap);
+void vm_amap_drop(vm_amap_t *amap);
+
+vm_page_t *vm_amap_find_page(vm_aref_t aref, int offset);
```
cahirwpz, line 31: Same thing with `int` arguments here.

include/sys/vm_amap.h
```diff
@@ -0,0 +1,43 @@
...
+int vm_amap_add_page(vm_aref_t aref, vm_page_t *frame, int offset);
+void vm_amap_remove_pages(vm_aref_t aref, int offset, int n_slots);
+
+static inline int vaddr_to_slot(vaddr_t addr) {
+  return addr / PAGESIZE;
+}
+
+/* Handy functions. */
```
cahirwpz, line 39: Well... the comment is not very useful. Please describe the functions properly.

cahirwpz (commented):
Just a quick preliminary CR...

Could you add a comment in `vm_amap.c` what is the source of "inspiration" for this solution and what are its' limitation compared to original work?

include/sys/vm_physmem.h
franciscozdo: I have changed all uses of `int` to `size_t`.

include/sys/vm_amap.h
franciscozdo: done.

include/sys/vm_physmem.h
cahirwpz: Have you actually missed this one?

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,173 @@
...
+  SCOPED_MTX_LOCK(&amap->mtx);
+  if (bit_test(amap->pg_bitmap, offset)) {
+    return EINVAL;
+  }
+  amap->pg_list[offset] = frame;
+  bit_set(amap->pg_bitmap, offset);
+  return 0;
+}
+
+static void _vm_amap_remove_pages(vm_amap_t *amap, size_t start,
```
cahirwpz, line 151: Please rename it to `vm_amap_find_page_unlocked` and require the caller to acquire mutex if needed.

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,173 @@
...
+}
+
+void vm_amap_remove_pages(vm_aref_t aref, size_t start, size_t nslots) {
+  if (!aref.amap)
+    return;
+  _vm_amap_remove_pages(aref.amap, aref.offset + start, nslots);
+}
+
+static void vm_amap_free(vm_amap_t *amap) {
+  _vm_amap_remove_pages(amap, 0, amap->slots);
```
cahirwpz, line 169: Should this function acquire `amap->mtx`?

sys/kern/vm_map.c
```diff
@@ -87,13 +86,13 @@ vm_map_t *vm_map_new(void) {
   return map;
 }
 
-vm_map_entry_t *vm_map_entry_alloc(vm_object_t *obj, vaddr_t start, vaddr_t end,
+vm_map_entry_t *vm_map_entry_alloc(vm_aref_t aref, vaddr_t start, vaddr_t end,
```
cahirwpz, line 89: Since in majority cases you pass `VM_AREF_EMPTY` to `vm_map_entry_alloc` I'd suggest removing this argument and implement `vm_map_entry_copy` separately.

sys/kern/vm_map.c
```diff
@@ -425,39 +437,74 @@ void vm_map_dump(vm_map_t *map) {
...
   }
 }
 
+static vm_map_entry_t *entry_clone_shared(vm_map_t *map, vm_map_entry_t *ent) {
+  if (!ent->aref.amap) {
+    /* We need to create amap, because we won't be able to share it if it
+     * is not created now. */
+    int slots = vaddr_to_slot(ent->end - ent->start);
+    ent->aref.amap = vm_amap_alloc(slots);
+    if (!ent->aref.amap) {
```
cahirwpz, line 449: Unnecessary curly braces again.

sys/kern/vm_map.c
```diff
@@ -425,39 +437,74 @@ void vm_map_dump(vm_map_t *map) {
          (it->prot & VM_PROT_READ) ? 'r' : '-',
          (it->prot & VM_PROT_WRITE) ? 'w' : '-',
          (it->prot & VM_PROT_EXEC) ? 'x' : '-');
-#if 0
-    vm_object_dump(it->object);
-#endif
   }
 }
 
+static vm_map_entry_t *entry_clone_shared(vm_map_t *map, vm_map_entry_t *ent) {
```
cahirwpz, line 443: First argument is unused? Why not full name `vm_map_entry_clone_shared`?

sys/kern/vm_map.c
```diff
@@ -425,39 +437,74 @@ void vm_map_dump(vm_map_t *map) {
...
+    }
+  }
+
+  vm_map_entry_t *new = vm_map_entry_copy(ent);
+  vm_amap_hold(ent->aref.amap);
+  new->aref = ent->aref;
+  return new;
+}
+
+static vm_map_entry_t *entry_clone_copy(vm_map_t *map, vm_map_entry_t *ent) {
```
cahirwpz, line 460: Similar remarks as for function above.

sys/mips/pmap.c
```diff
@@ -52,7 +52,7 @@ pte_t pte_make(paddr_t pa, vm_prot_t prot, unsigned flags) {
 }
 
 pte_t pte_protect(pte_t pte, vm_prot_t prot) {
-  return (pte & ~PTE_PROT_MASK) | vm_prot_map[prot];
+  return (pte & ~PTE_PROT_MASK & ~PTE_RI & ~PTE_XI) | vm_prot_map[prot];
```
cahirwpz, line 55: I think `PTR_RI` and `PTE_XI` are already a part of `PTE_PROT_MASK`

cahirwpz (changes requested):

sys/kern/vm_amap.c
franciscozdo: I think not. It is only used when `ref_cnt` of amap has dropped to 0 so no one else is using this amap and it shold be safe.

include/sys/vm_physmem.h
franciscozdo: Right. Fixed.

sys/kern/fork.c
```diff
@@ -66,6 +66,8 @@ int do_fork(void (*start)(void *), void *arg, pid_t *cldpidp) {
 
   /* Clone the entire process memory space. */
   child->p_uspace = vm_map_clone(parent->p_uspace);
+  if (child->p_uspace == NULL)
+    return ENOMEM;
```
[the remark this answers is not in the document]
franciscozdo: I have created functions to free thread and process structures.

I have also considered a different approach. To create new `vm_map` before thread and process and then just write it to process structure when it is created. I don't know which is better so I've done it this way to preserve order of operations here.

sys/kern/vm_amap.c
franciscozdo: done.

sys/kern/vm_map.c
franciscozdo: done.

sys/kern/vm_map.c
franciscozdo: done.

sys/mips/pmap.c
franciscozdo: Yes it is. Removed from here.

sys/kern/fork.c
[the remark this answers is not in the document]
cahirwpz: I like the proposal since newly introduced `proc_error_free` and `thread_error_free` look weird.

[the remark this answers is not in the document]
franciscozdo: done

include/sys/vm_amap.h
```diff
@@ -0,0 +1,49 @@
...
+
+typedef struct vm_amap vm_amap_t;
+typedef struct vm_aref vm_aref_t;
+
+struct vm_aref {
+  size_t offset;   /* offset in slots */
+  vm_amap_t *amap; /* underlying amap */
+};
+
+#define VM_AREF_EMPTY ((vm_aref_t){.offset = 0, .amap = NULL})
```
cahirwpz, line 16: Since this is used internally only in one place... do we need to have it defined in public interface?

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,172 @@
...
+  return amap->ref_cnt;
+}
+
+size_t vm_amap_slots(vm_amap_t *amap) {
+  return amap->slots;
+}
+
+vm_amap_t *vm_amap_alloc(size_t slots) {
+  slots += EXTRA_AMAP_SLOTS;
+  vm_amap_t *amap = pool_alloc(P_VM_AMAP_STRUCT, M_WAITOK);
```
cahirwpz, line 57: Oh. I have an interesting observation here. It appears that `pool_alloc` always has `M_WAITOK` semantics, which states it never returns NULL - so `nullptr` checks are unnecessary.

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,172 @@
...
+  if (!amap)
+    return NULL;
+
+  amap->pg_list =
+    kmalloc(M_AMAP, slots * sizeof(vm_page_t *), M_ZERO | M_WAITOK);
+
+  if (!amap->pg_list)
```
cahirwpz, line 64: ditto

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,172 @@
...
+    goto list_fail;
+
+  amap->pg_bitmap =
+    kmalloc(M_AMAP, slots * sizeof(bitstr_t), M_ZERO | M_WAITOK);
+  if (!amap->pg_bitmap)
```
cahirwpz, line 69: ditto

include/sys/vm_amap.h
```diff
@@ -0,0 +1,49 @@
...
+
+/*
+ * Allocate new amap with specified number of slots.
+ *
+ * Returns new amap with ref_cnt equal to 1.
+ */
+vm_amap_t *vm_amap_alloc(size_t slots);
```
cahirwpz, line 23: I think it always returns non-NULL value. Please verify and remove NULL checks.

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,172 @@
...
+}
+
+void vm_amap_remove_pages(vm_aref_t aref, size_t start, size_t nslots) {
+  if (!aref.amap)
+    return;
+  SCOPED_MTX_LOCK(&aref.amap->mtx);
+  vm_amap_remove_pages_unlocked(aref.amap, aref.offset + start, nslots);
+}
+
+static void vm_amap_free(vm_amap_t *amap) {
```
cahirwpz, line 167: If you use that function only once why bother and not inline it?

sys/kern/vm_map.c
```diff
@@ -425,39 +439,75 @@ void vm_map_dump(vm_map_t *map) {
...
+  new->aref = ent->aref;
+  return new;
+}
+
+static vm_map_entry_t *vm_map_entry_clone_copy(vm_map_t *map,
+                                               vm_map_entry_t *ent) {
+  vm_map_entry_t *new = vm_map_entry_copy(ent);
+
+  /* If there was an amap we need to clone it. */
+  if (ent->aref.amap) {
```
cahirwpz, line 467: Please use early exit in here as the if-clause constitutes majority of the work performed by the function.

cahirwpz (commented):
I think it's the last slew of my rumblings.

sys/kern/vm_amap.c
franciscozdo: done

sys/kern/vm_amap.c
franciscozdo: Okay, I have removed all the NULL checks here.

include/sys/vm_amap.h
franciscozdo: Yes. According to previous comments we will always get requested memory and return with valid amap.

cahirwpz (approved):
LGTM

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,152 @@
...
+  vm_amap_t *new = vm_amap_alloc(slots);
+
+  SCOPED_MTX_LOCK(&amap->mtx);
+  for (size_t slot = 0; slot < slots; slot++) {
+    size_t old_slot = aref.offset + slot;
+
+    if (!bit_test(amap->pg_bitmap, old_slot))
+      continue;
+
+    vm_page_t *new_pg = vm_page_alloc_zero(1);
```
cahirwpz, line 85: Why zero page if you perform a copy in line below?

cahirwpz: Why not bulk allocate all pages with a single call to [vm_pagelist_alloc](https://mimiker.ii.uni.wroc.pl/source/s?refs=vm_pagelist_alloc&project=mimiker)?

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,152 @@
...
+  bit_set(amap->pg_bitmap, offset);
+  return 0;
+}
+
+static void vm_amap_remove_pages_unlocked(vm_amap_t *amap, size_t start,
+                                          size_t nslots) {
+  for (size_t i = start; i < start + nslots; i++) {
+    if (!bit_test(amap->pg_bitmap, i))
+      continue;
+    vm_page_free(amap->pg_list[i]);
```
cahirwpz, line 129: Maybe peform bulk free operation with [vm_pagelist_free](https://mimiker.ii.uni.wroc.pl/source/s?refs=vm_pagelist_free&project=mimiker)?

sys/kern/vm_map.c
```diff
@@ -493,21 +531,29 @@ int vm_page_fault(vm_map_t *map, vaddr_t fault_addr, vm_prot_t fault_type) {
...
-  vm_page_t *frame = vm_object_find_page(ent->object, offset);
+  bool insert = false;
+  int offset = vaddr_to_slot(fault_page - ent->start);
+  vm_page_t *frame = vm_amap_find_page(ent->aref, offset);
 
-  if (frame == NULL)
-    frame = obj->vo_pager->pgr_fault(obj, offset);
+  if (frame == NULL) {
+    insert = true;
+    frame = vm_page_alloc_zero(1);
```
cahirwpz, line 548: If that's happen to be the only reference to `vm_page_alloc_zero` then please remove the function and zero out the page here.

cahirwpz (commented):
Performance related hints / remarks.

cahirwpz (commented):
Performance related hints / remarks.

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,156 @@
...
+
+vm_amap_t *vm_amap_alloc(size_t slots) {
+  slots += EXTRA_AMAP_SLOTS;
+  vm_amap_t *amap = pool_alloc(P_VM_AMAP_STRUCT, M_WAITOK);
+
+  amap->pg_list =
+    kmalloc(M_AMAP, slots * sizeof(vm_page_t *), M_ZERO | M_WAITOK);
+
+  amap->pg_bitmap =
+    kmalloc(M_AMAP, slots * sizeof(bitstr_t), M_ZERO | M_WAITOK);
```
pj1031999, line 61: bitstr_size or am I missing something?

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,156 @@
...
+
+  SCOPED_MTX_LOCK(&amap->mtx);
+  for (size_t slot = 0; slot < slots; slot++) {
+    size_t old_slot = aref.offset + slot;
+
+    if (!bit_test(amap->pg_bitmap, old_slot))
+      continue;
+
+    vm_page_t *new_pg = vm_page_alloc(1);
+    pmap_copy_page(amap->pg_list[old_slot], new_pg);
```
pj1031999, line 86: NULL check?

sys/kern/vm_map.c
```diff
@@ -160,10 +161,18 @@ static vm_map_entry_t *vm_map_entry_split(vm_map_t *map, vm_map_entry_t *ent,
 
   vm_map_entry_t *new_ent = vm_map_entry_copy(ent);
 
+  /* Split amap if it is present. */
+  vm_aref_t old = ent->aref;
+  if (old.amap) {
+    vm_amap_hold(old.amap);
+    int offset = vaddr_to_slot(splitat - ent->start);
```
pj1031999, line 168: int / size_t / vaddr_t - please don't mix every int size in one line ;)

sys/kern/vm_amap.c
```diff
@@ -0,0 +1,156 @@
...
+ *  - Amaps are not resizable. We allocate more memory for each amap
+ *    (adjustable with EXTRA_AMAP_SLOTS) to allow resizing of small amounts.
+ *  - Amap is managing a simple array of referenced pages so it may not be the
+ *    most effective implementation.
+ */
+
+/* Amap size will be increased by this number of slots to easier resizing. */
+#define EXTRA_AMAP_SLOTS 16
+
+struct vm_amap {
```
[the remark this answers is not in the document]
franciscozdo: Right. That will be more clear too.

sys/kern/vm_amap.c
franciscozdo: True, I missed that such macro exists.

sys/kern/vm_amap.c
franciscozdo: Right. I should have missed it or remove it with other NULL check that appeared to be unnecessary. I have restored it and updated other places that may got NULL value when the allocation here fails.

PR #1380, opened 2023-03-02
libc changes for suckless sbase

panantoni01:
This PR adds some changes/features needed for suckless sbase tools:
- `clock_settime` syscall stub
- `strptime` libc function
- fixed `readdir` function - it doesn't return errno on end of directory
- `UTIME_NOW ` and `UTIME_OMIT` defines


cahirwpz (approved):
LGTM

PR #1381, opened 2023-04-29
Update python files

franciscozdo:
~I had a problem with launch script due to an update of API made in `libtmux v0.17.0`. (Accessing pane id is done in a different way now.) I have updated the launch script to determine which version is used and to access pane id using the right way.~

1. Ignore python files in current python venv
2. Pin versions of packages in `requirements.txt` (proposed versions work for me. Tested on Python 3.9 and 3.10.)

### DIFF
--- a/launch
+++ b/launch
@@ -261,7 +261,7 @@ class Launchable():
         cmd = ' '.join([self.cmd] + list(map(shlex.quote, self.options)))
         self.window = session.new_window(
             attach=False, window_name=self.name, window_shell=cmd)
-        self.pid = int(self.window.attached_pane.pane_index)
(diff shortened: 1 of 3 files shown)


cahirwpz (approved):

cahirwpz (commented):
One more thing - why are we limited to exact versions of the libraries? Cannot we use `>=` operator?

franciscozdo:
E.g. in `libtmux` in 0.17.0 there is a change of API and it causes problems in our `launch` script (and something more has changed because if we update `launch` after updating `libtmux` there are some problems with klog printing). I fixed them to don't spend time on debuging errors in python code after installing packages on new system (where they might be newer).

PR #1382, opened 2023-04-29
git: ignore changes in submodules

franciscozdo:
After building Mimiker git status shows changes in submodules. I they shouldn't be visible.


cahirwpz (approved):
LGTM

PR #1383, opened 2023-05-05
vm: Fix prot check in vm_page_fault

franciscozdo:
I have found, that checks of protection bits in `vm_page_fault` are done wrong. On aarch64 the `fault_type` has always set `VM_PROT_READ`. Then if access was read, then it didn't trigger any of these checks (because both read and write prot was set).

I have also added the check for exec bit.


cahirwpz (approved):
LGTM

PR #1385, opened 2023-05-16
proc: Free `p_args` from proper kmalloc pool

franciscozdo:
`p_args` was freed different poll than it was allocated on.


cahirwpz (approved):
LGTM

PR #1388, opened 2023-05-30
Kernel changes for ubase tools

panantoni01:
This PR introduces some kernel changes needed for porting tools from suckless ubase (https://github.com/cahirwpz/mimiker/pull/1371)


sys/kern/main.c
```diff
@@ -57,24 +70,19 @@ static __noreturn void start_init(__unused void *arg) {
   error = session_enter(p);
```
cahirwpz, line 70: Is it mandatory to enter a session before we enter the user space for the first time?

sys/kern/main.c
```diff
@@ -57,24 +70,19 @@ static __noreturn void start_init(__unused void *arg) {
...
+  if (test) {
+    open_fds();
     ktest_main(test);
+  }
 
   char *init = kenv_get("init");
   if (init == NULL)
     init = "/sbin/init";
 
+  if (strcmp("/sbin/init", init))
```
cahirwpz, line 83: Please describe here that it is a workaround and what it is for.

cahirwpz (commented):
Please move `open_fds` hack to a separate PR since this is still controversial to me and I have not made up my mind about it.

sys/kern/main.c
panantoni01: Added a comment in a19a4a7f2d2056822227eb9c628abe015471b88c

sys/kern/main.c
panantoni01: Actually it is not mandatory, provided that the session is created by the `init` process. For example in NetBSD it's on the side of `init` process to create a session - http://bxr.su/NetBSD/sbin/init/init.c#249.c. Therefore I changed the way the session is created in the case where we run `/sbin/init` - https://github.com/cahirwpz/mimiker/pull/1390/commits/567449066ac930d0572665fe6d6c1887c284b439

cahirwpz:
Please limit the changes in this PR to TIOCFLUSH handling.

cahirwpz (approved):

PR #1390, opened 2023-06-04
Workaround for /sbin/init

panantoni01:
This PR changes the way `/sbin/init` is started from kernel space:
* kernel does not open 0, 1, 2 file descriptors for `/sbin/init`, which means that`/sbin/init` doesn't get a controlling terminal
* session is created in `/sbin/init` process

### DIFF
--- a/contrib/sinit/patches/series
+++ b/contrib/sinit/patches/series
@@ -1,1 +1,2 @@
+sinit.patch
 config.patch
--- a/contrib/sinit/patches/sinit.patch
+++ b/contrib/sinit/patches/sinit.patch
@@ -1,14 +1,12 @@
 Index: sinit/sinit/sinit.c
 ===================================================================
---- sinit.orig/sinit/sinit.c	2023-06-04 17:28:33.209715149 +0200
-+++ sinit/sinit/sinit.c	2023-06-04 17:29:49.403045971 +0200
-@@ -37,7 +37,8 @@
+--- sinit.orig/sinit/sinit.c	2023-06-04 17:29:56.303045745 +0200
++++ sinit/sinit/sinit.c	2023-06-04 21:49:07.799201063 +0200
+@@ -37,6 +37,7 @@
  
  	if (getpid() != 1)
  		return 1;
--	chdir("/");
 +	setsid();
-+  	chdir("/");
+ 	chdir("/");
  	sigfillset(&set);
  	sigprocmask(SIG_BLOCK, &set, NULL);
- 	spawn(rcinitcmd);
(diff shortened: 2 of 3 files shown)


cahirwpz:
What if you revert to original source code but pass `O_NOCTTY` to `do_open`?

panantoni01:
> What if you revert to original source code but pass `O_NOCTTY` to `do_open`?

In the scenario where we run `/sbin/init` as our `init` process this solution works fine, because `init` does not get ctty and therefore it can be later acquired by `ksh`.

However if the user wants to run some other program as `init` - e.g. `./launch init=/bin/ksh` - then this process won't have ctty, which results e.g. in lack of job control in `ksh`.

cahirwpz:
Yeah. You're right. That's no solution.

Basically I think that we're trying to workaround a genuine bug / missing functionality in the kernel. I'm trying to figure out what we could have missed when implementing terminal subsystem. 

Since our terminal subsystem is derived from FreeBSD implementation we could add `TIOCNOTTY`:
```
TIOCNOTTY void
		 In the	past, when a process that did not have a controlling
		 terminal (see The Controlling Terminal	in `termios(4)`) first
		 opened	a terminal device, it acquired that terminal as	its
		 controlling terminal.	For some programs this was a hazard as
		 they did not want a controlling terminal in the first place,
		 and this provides a mechanism to disassociate the controlling
		 terminal from the calling process.  It	must be	called by
		 opening the file /dev/tty and calling TIOCNOTTY on that file
		 descriptor.

		 The current system does not allocate a	controlling terminal
		 to a process on an open() call: there is a specific ioctl
		 called	TIOCSCTTY to make a terminal the controlling terminal.
		 In addition, a	program	can fork() and call the	setsid() sys-
		 tem call which	will place the process into its	own session -
		 which has the effect of disassociating	it from	the control-
		 ling terminal.	 This is the new and preferred method for pro-
		 grams to lose their controlling terminal.

		 However, environmental	restrictions may prohibit the process
		 from being able to fork() and call the	setsid() system	call
		 to disassociate it from the controlling terminal.  In this
		 case, it must use TIOCNOTTY.
```
... and use it in `sinit.c` - @panantoni01 what do you think?

panantoni01:
This looks like a good idea. Once we have this ioctl implemented, we can simply call it in `sinit.c` before spawning `rc.init`. This would make it possible to acquire a controlling terminal in `getty.c`.

cahirwpz:
Is it something you'd like to implement?

panantoni01:
> Is it something you'd like to implement?

Yes, this looks like a nice challenge.

cahirwpz:
@panantoni01 My current understanding is that this PR is obsoleted by #1399. Is that correct?

panantoni01:
> @panantoni01 My current understanding is that this PR is obsoleted by #1399. Is that correct?

Yes, that's right. I guess that once #1399 is merged, we can close this PR.

cahirwpz:
#1399 has been merged rendering this PR obsolete, hence closing it.

PR #1391, opened 2023-06-05
Kernel function trace

franciscozdo:
In this PR I try to create a tool that will allow us to analyze how kernel is performing.

The main idea is to save all events (enter and exit from a function) with some data (thread and timestamp). Gathered data may be analyzed later using python utilities (`kft/*.py`).


franciscozdo:
Merged in  #1393 and #1418. Closing.

PR #1396, opened 2023-06-20
GDB: Improve tools to debug vm structures

franciscozdo:
In this PR I move all commands that operate on VM structures to be subcommands of `vm` command and create new ones

###### Commands added in this PR
- `vm map` -- (replace `kdump vm_map`) and interpret optional argument - pid of process we want to see the vm map
-  `vm pmap_kernel` -- (replace `kdump pmap`)
-  `vm pmap_user` -- dump pmap of specified proc (for current if not specified)
- `vm segment` -- see info about specified segment in current proc `vm_map`
- `vm segment_proc` -- see info about specified segment in `vm_map` of proc specified by pid
- `vm freepages` -- (replace `kdump freepages`)
- `vm physseg` -- (replace `kdump physseg`)

### DIFF
--- a/sys/debug/cpu.py
+++ b/sys/debug/cpu.py
@@ -135,10 +135,10 @@ class PageTableMips():
         print(table)
 
 
-class PageTableAarch64():
+class PageTableAArch64():
     def __init__(self, pmap):
         self._pmap = pmap
-        print("Page table not implemented for Aarch64")
+        print("Page table not implemented for AArch64")
 
     def print(self):
         pass
@@ -156,7 +156,7 @@ class PageTableRiscv():
 if get_arch() == 'mips':
     PageTable = PageTableMips
 elif get_arch() == 'aarch64':
-    PageTable = PageTableAarch64
+    PageTable = PageTableAArch64
 elif get_arch() == 'riscv':
     PageTable = PageTableRiscv
 else:


sys/debug/virtmem.py
```diff
@@ -1,43 +1,207 @@
...
+                                UserPmap(),
+                                VmMapDump(),
+                                CurrentSegmentInfo(),
+                                ProcSegmentInfo(),
+                                VmPhysSeg(),
+                                VmFreePages(),
+                                ])
+
+
+def _print_mips_pmap(pmap):
```
cahirwpz, line 25: Please move all machine dependant procedures to `cpu.py`. There provide a separate class for each CPU.

After it's done export this set of classes that are valid for the current architecture. Then provide class / function aliases at the top level like so:

```python
class PageTableAArch64:
   ...
   
def PrintPageTableAarch64(...):
   ...

if get_arch() == 'aarch64':
   PageTable = PageTableAArch64
   PrintPageTable = PrintPageTableAArch64
elif get_arch() == 'mips':
   PageTable = PageTableMIPS
   PrintPageTable = PrintPageTableMIPS
else:
   ...
```

sys/debug/virtmem.py
```diff
@@ -1,43 +1,207 @@
...
+        print('Amap refs:   {}'.format(amap.ref_cnt))
+
+        amap.print_pages(seg.amap_offset, seg.pages)
+    else:
+        print('Amap: NULL')
+
+
+def _segment_info(proc, segment):
+    entries = proc.p_uspace.get_entries()
+    if segment < 4096:
```
cahirwpz, line 147: Why magic constant 4096?

sys/debug/vm_map.py
```diff
@@ -0,0 +1,82 @@
...
+        return 'vm_map_entry[{:#08x}-{:#08x}]'.format(self.start, self.end)
+
+    @property
+    def amap_ptr(self):
+        return self.aref['amap']
+
+    @property
+    def pages(self):
+        size = self.end - self.start
+        return int(size / 4096)
```
cahirwpz, line 32: `size // 4096` to perform integer division ?

sys/debug/vm_map.py
```diff
@@ -0,0 +1,82 @@
...
+
+    @property
+    def amap(self):
+        if int(self.aref['amap']) == 0:
+            return None
+        else:
```
cahirwpz, line 38: `else` not needed here

sys/debug/vm_map.py
```diff
@@ -0,0 +1,82 @@
...
+    __ctype__ = 'struct vm_amap'
+    __cast__ = {'slots': int,
+                'ref_cnt': int}
+
+    def __repr__(self):
+        return 'vm_amap[slots={}, refs={}]'.format(self.slots, self.ref_cnt)
+
+    def _get_bitmap(self):
+        bitmap = []
+        bitssize = ((self.slots) + 7) >> 3
```
cahirwpz, line 59: Why extra parens around `self.slots`?

sys/debug/virtmem.py
```diff
@@ -1,43 +1,207 @@
...
+        table = TextTable(types='ittttt', align='rrrrrr')
+        table.header(['segment', 'start', 'end', 'prot', 'flags', 'amap'])
+        for idx, seg in enumerate(entries):
+            table.add_row([idx, hex(seg.start), hex(seg.end), seg.prot,
+                           seg.flags, seg.aref])
         print(table)
 
 
+def _print_segment(seg, pid, id):
+    print('Segment {} in proc {}'.format(id, pid))
```
cahirwpz, line 127: Please use f-strings if possible.

sys/debug/vm_map.py
```diff
@@ -0,0 +1,82 @@
...
+        return self.aref['amap']
+
+    @property
+    def pages(self):
+        size = self.end - self.start
+        return int(size / 4096)
+
+    @property
+    def amap(self):
+        if int(self.aref['amap']) == 0:
```
cahirwpz, line 36: `if not self.aref['amap']` ?

sys/debug/utils.py
```diff
@@ -21,6 +25,29 @@ def relpath(path):
...
 
+def get_arch():
+    try:
+        _ = gdb.parse_and_eval('aarch64_init')
+        return 'aarch64'
+    except gdb.error:
+        pass
+
+    try:
+        _ = gdb.parse_and_eval('riscv_init')
```
cahirwpz, line 36: Consider using output of `gdb.architecture_names` function.

sys/debug/virtmem.py
franciscozdo: I have done something like that.

franciscozdo: For now only Mips is supported, but I have provided stup classes for other architectures.

cahirwpz (commented):
Some remarks left. What do you think about employing `argparse` to parse arguments in `__call__` method. It'd be nice if `help user-defined` showed all subcommands, so the user does not have to consult the source code to know how to use a command.

Issue #1397, opened 2023-06-21
pathconf syscall


panantoni01:
The goal of this issue is to implement [pathconf](https://man.netbsd.org/pathconf.2) syscall. This is going to be done in the following way:
* add `sys_pathconf` in `syscalls.c`
* add `VOP_PATHCONF` in `vnode.h`
* implement `do_pathconf` in `vfs_syscalls.c`
* implement `tmpfs_vop_pathconf` and add it to [tmpfs_vnodeops](https://github.com/cahirwpz/mimiker/blob/master/sys/kern/tmpfs.c#L617)
* implement `initrd_vnode_pathconf` and add it to [initrd_vops](https://github.com/cahirwpz/mimiker/blob/master/sys/kern/initrd.c#L340)

cahirwpz:
Did you mean `initrd_vnode_pathconf`?

panantoni01:
> Did you mean `initrd_vnode_pathconf`?

Yes, I've just corrected it.

panantoni01:
PR https://github.com/cahirwpz/mimiker/pull/1398 has been merged, closing.

PR #1398, opened 2023-06-27
pathconf syscall

panantoni01:
This PR adds support for `pathconf` syscall according to https://github.com/cahirwpz/mimiker/issues/1397.


cahirwpz (approved):
LGTM

PR #1399, opened 2023-07-05
tty: add TIOCNOTTY ioctl

panantoni01:
This PR adds `TIOCNOTTY` ioctl, which is used to disassociate the controlling terminal from a process. This ioctl should be used in `sinit.c`, so that a controlling terminal can be disassociated in the init process and later acquired in `getty` (related PR: https://github.com/cahirwpz/mimiker/pull/1390).

### DIFF
--- a/contrib/sinit/patches/sinit.patch
+++ b/contrib/sinit/patches/sinit.patch
@@ -1,0 +1,24 @@
+Index: sinit/sinit/sinit.c
+===================================================================
+--- sinit.orig/sinit/sinit.c
++++ sinit/sinit/sinit.c
+@@ -6,6 +6,8 @@
+ #include <stdio.h>
+ #include <stdlib.h>
+ #include <unistd.h>
++#include <fcntl.h>
++#include <sys/ioctl.h>
+ 
+ #define LEN(x)	(sizeof (x) / sizeof *(x))
+ #define TIMEO	30
+@@ -40,6 +42,10 @@ main(void)
+ 	chdir("/");
+ 	sigfillset(&set);
+ 	sigprocmask(SIG_BLOCK, &set, NULL);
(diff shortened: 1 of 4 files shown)


contrib/sinit/patches/sinit.patch
```diff
@@ -0,0 +1,31 @@
...
+-	int sig;
++	int sig, fd;
+ 	size_t i;
+ 
+ 	if (getpid() != 1)
+@@ -40,6 +42,8 @@
+ 	chdir("/");
+ 	sigfillset(&set);
+ 	sigprocmask(SIG_BLOCK, &set, NULL);
++	fd = open("/dev/tty", O_RDWR);
```
cahirwpz, line 27: AFAIU descriptors 0, 1, 2 are already opened and attached to `/dev/tty`, aren't they?

cahirwpz (approved):
Only a single remark to the userspace code - otherwise looks good.

contrib/sinit/patches/sinit.patch
panantoni01: Yes, that's right. I just added this `open` here, so that it is clear that we call `TIOCNOTTY` ioctl on the controlling terminal. But I guess, that we could simply call the ioctl on descriptor 0,1 or 2 (and maybe add an explaining comment).

cahirwpz: Please add a comment then.

panantoni01: Added a comment in https://github.com/cahirwpz/mimiker/pull/1399/commits/506acc63e667c751409c6ffe20d37eeade9336cc

cahirwpz (approved):
LGTM

PR #1400, opened 2023-07-06
sbase: add getconf and hostname

panantoni01:
This PR adds the following missing tools from suckless sbase:
* `hostname` with `gethostname` and `sethostname` function stubs
* `getconf`


cahirwpz (approved):
LGTM

PR #1401, opened 2023-07-07
tests: add mmap_private test

franciscozdo:
New test that checks if requested memory is truly private. I have checked that all tests pass if we intentionally mark all mapped memory as shared. So we need that test (actually it appeared during work on CoW implementation).


cahirwpz (approved):
LGTM

PR #1414, opened 2023-08-02
[clock] synchronize now

pj1031999:
Use explicitly atomics for now.

Pathological scenario reported by sanitizer:

thread1:                        thread2:
    - gettime                           - clock_cb
        - return now           <race>           - now = bt2st(bin)

Found-by: KCSAN


cahirwpz (approved):
LGTM

PR #1415, opened 2023-08-05
Add poll syscall as a wrapper of kevent

panantoni01:
This PR provides `poll` syscall implemented as a wrapper of `kevent` syscall.

### DIFF
--- a/lib/libc/sys/poll.c
+++ b/lib/libc/sys/poll.c
@@ -1,6 +1,7 @@
 #include <sys/poll.h>
 #include <sys/time.h>
 #include <sys/event.h>
+#include <fcntl.h>
 #include <stdlib.h>
 #include <unistd.h>
 #include <errno.h>
@@ -27,7 +28,7 @@ int poll(struct pollfd *fds, nfds_t nfds, int timeout) {
     return -1;
   }
 
-  kq = kqueue();
+  kq = kqueue1(O_CLOEXEC);
   if (kq < 0)
     return -1;
 
--- a/lib/libc/sys/poll.c
+++ b/lib/libc/sys/poll.c
@@ -1,73 +1,76 @@
-#include<sys/poll.h>
-#include<sys/time.h>
-#include<sys/event.h>
-#include<stdlib.h>
-#include<unistd.h>
-#include<errno.h>
+#include <sys/poll.h>
+#include <sys/time.h>
+#include <sys/event.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <errno.h>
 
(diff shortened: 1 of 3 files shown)


cahirwpz (commented):
Please use `clang-format-14` on `lib/libc/sys/poll.c`. Looks good at first glance.

lib/libc/sys/poll.c
```diff
@@ -0,0 +1,74 @@
...
+    timeout_ts.tv_sec = timeout / 1000;
+    timeout_ts.tv_nsec = (timeout % 1000) * 1000000;
+  }
+  /* if timeout == -1, then poll should wait indefinitely */
+  else if (timeout < -1) {
+    errno = EINVAL;
+    return -1;
+  }
+
+  kq = kqueue();
```
pj1031999, line 30: It's not an issue now but if we will add proper threading support you can leak fd here.

Let's consider the following scenario:
```
td1:                             td2:
  - poll
    - kqueue
                      <leak>        - fork
    - close
```

*BSDs introduced `kqueue1`  as fix for that issue (no atomic close-on-exec). It allows to pass O_CLOEXEC as argument with standard semantic of this flag.

It's probably not a problem now but it will beat us in the future ;)

pj1031999: Yes. We already have only `kqueue1` and `kqueue` just calls it with `0`.

I think that they should be inherited by a child - but it's a different discussion that is out of scope.

panantoni01: Added O_CLOEXEC in https://github.com/cahirwpz/mimiker/pull/1415/commits/9eb99046af9a71baf4e6a6c1b0221bb93f063b6d

lib/libc/sys/poll.c
```diff
@@ -0,0 +1,75 @@
...
+
+    if (events[i].flags & EV_ERROR) {
+      errno = events[i].data;
+      if (errno == EBADF)
+        pfd->revents |= POLLNVAL;
+      else
+        pfd->revents |= POLLERR;
+    } else if (events[i].filter == EVFILT_READ)
+      pfd->revents |= POLLIN;
+    else if (events[i].filter == EVFILT_WRITE)
```
cahirwpz, line 66: small nit: please be consistent and use braces here if you used them above.

cahirwpz (approved):
Please note that implementing POLLHUP is tricky as it is not standardized (sic!).
One can read more here: https://www.jookia.org/wiki/POLLHUP_polling#NetBSD

lib/libc/sys/poll.c
panantoni01: Added braces here

pj1031999 (approved):

PR #1417, opened 2023-10-30
amap: remove bitmap

franciscozdo:
Remove bitmaps from amaps. Actually bitmaps are not needed because we can check if anon is present based by checking if its slot is Null.


cahirwpz (approved):
LGTM

PR #1419, opened 2023-11-29
Add pselect syscall as a wrapper of kevent

panantoni01:
This PR provides `pselect` syscall implementation. Since the operation of `pselect` and `select` is almost identical, I reorganized the `lib/libc/sys/select.c` file, so that:
* `pselect` contains all the core implementation
* `select` converts its `struct timeval` argument to `struct timespec` and calls `pselect`

### DIFF
--- a/lib/libc/sys/select.c
+++ b/lib/libc/sys/select.c
@@ -21,9 +21,14 @@ int pselect(int nfds, fd_set *restrict readfds, fd_set *restrict writefds,
         return -1;
     }
 
-    kq = kqueue1(O_CLOEXEC);
-    if (kq < 0)
+    if (sigmask && sigprocmask(SIG_SETMASK, sigmask, &sigs))
         return -1;
+
+    kq = kqueue1(O_CLOEXEC);
+    if (kq < 0) {
+        ret = -1;
+        goto restore_sigs;
+    }
 
     events = malloc(2 * nfds * sizeof(struct kevent));
     if (!events) {
@@ -44,9 +49,6 @@ int pselect(int nfds, fd_set *restrict readfds, fd_set *restrict writefds,
     if (writefds != NULL)
         FD_ZERO(writefds);
 
-    if (sigmask && sigprocmask(SIG_SETMASK, sigmask, &sigs))
-        return -1;
-
     ret = kevent(kq, events, nevents, events, nevents, timeout);
     if (ret == -1)
         goto end;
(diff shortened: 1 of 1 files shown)


lib/libc/sys/select.c
```diff
@@ -50,8 +43,11 @@ int select(int nfds, fd_set *restrict readfds, fd_set *restrict writefds,
         FD_ZERO(readfds);
     if (writefds != NULL)
         FD_ZERO(writefds);
-    
-    ret = kevent(kq, events, nevents, events, nevents, timeout == NULL ? NULL : &timeout_ts);
+
+    if (sigmask && sigprocmask(SIG_SETMASK, sigmask, &sigs))
+        return -1;
```
pj1031999, line 48: `kq` is leaking here. The same for `events`.

lib/libc/sys/select.c
```diff
@@ -5,29 +5,22 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <errno.h>
+#include <signal.h>
 
-int select(int nfds, fd_set *restrict readfds, fd_set *restrict writefds,
-           fd_set *restrict exceptfds, struct timeval *restrict timeout) {
+int pselect(int nfds, fd_set *readfds, fd_set *writefds,
+            fd_set *exceptfds, const struct timespec *timeout,
+            const sigset_t *sigmask) {
```
pj1031999, line 12: nit: non consistent `const` and `restrict` modifiers for pointers - see `select`s arguments.

lib/libc/sys/select.c
pj1031999, line 48: >     If sigmask is a non-null pointer, then the pselect() function shall
>     replace the signal mask of the caller by the set of signals pointed to by
>     sigmask before examining the descriptors, and shall restore the signal
>     mask of the calling thread before returning.

the question is how do we want to interpret _before_ - I'm thinking about moving it before `kqueue`.

panantoni01: Moved the `sigprocmask` call before `kqueue` - https://github.com/cahirwpz/mimiker/pull/1419/commits/2dc7f82d58a0e638af3ca13518660648daae6eab.

lib/libc/sys/select.c
panantoni01: Added missing `restrict` modifiers for `pselect` pointer arguments - https://github.com/cahirwpz/mimiker/pull/1419/commits/8e37006c10e6c623b5b485d018ee0f6208e83cb4

lib/libc/sys/select.c
panantoni01: Since the `sigprocmask` has been moved before `kqueue` (https://github.com/cahirwpz/mimiker/pull/1419/commits/2dc7f82d58a0e638af3ca13518660648daae6eab), no more leaks occur.

pj1031999 (approved):
lgtm

cahirwpz (approved):
LGTM

